PR #2, opened 2015-12-24
Split into two specs

justmoon:
Splitting the crypto-conditional transactions spec into two.

The Cryptographic Conditions spec defines cryptographic conditions which can be used in distributed systems to allow components to securely react to events happening in other components. Since cryptographic fulfillments can be passed on by anyone using any protocol, transport and method they provide very loose coupling between components unlike most secure channels (e.g. TLS).

The Cryptographic Distributed Transactions spec defines an implementation of two-phase commit (2PC) using cryptographic conditions.

Work in progress, not ready to merge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,10 @@ This repository contains a collection of various specifications and documentatio
 
 ## Specs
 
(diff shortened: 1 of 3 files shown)


justmoon:
Outdated PR, closing.

PR #4, opened 2016-05-23
OWSP -> SPSP, and text edits

emschwartz:
This moves the renamed Simple Payment Setup Protocol out of the architecture doc and into its own RFC (just a placeholder for now)


justmoon:
LGTM

PR #5, opened 2016-05-24
Packet and plugin specs

justmoon:
Add packet and plugin specs and various cleanup and updates.


emschwartz:
More work to be done but LGTM

PR #35, opened 2016-07-07
docs(0002): markdown version

mDuo13:
Downsides:
- GitHub doesn't render the markdown nicely (neither does Dactyl right now)
- Some newer ITU references aren't in the xml2rfc database but the converter assumes they will be, causing errors. Renamed to "itu" as a workaround.

Upsides:
- Still pretty easy to write by hand
- Generated HTML has very few non-whitespace differences from the original generated HTML. 
- Fixed a couple typos while I was at it.


justmoon:
LGTM

PR #38, opened 2016-07-07
Generate first diagram from DOT

justmoon:
This is the first diagram being generated from a dot file. This makes it easier to edit and track changes to the diagrams we use in Interledger specs.

Note that we can add an SVG stylesheet to make it look nicer if we want.


emschwartz:
Do SVGs work if we want to convert specs into the RFC format?

justmoon:
No, but the only spec that makes sense for IETF to ratify is crypto-conditions. So it's fine to use SVG in the other specs.

emschwartz:
I would propose sticking with ASCII art, at least for now. @justmoon what do you think?

bensharaf:
https://www.rfc-editor.org/rse/format-faq/#9 looks like RFCs will be supporting SVG

justmoon:
Going to withdraw this PR for now. I think it is worth going for SVG in the future, but we have more important things to worry about right now.

PR #40, opened 2016-07-08
Updated RFC3 to be used for IETF ID

adrianhopebailie:


emschwartz:
I think we actually don't want to change the format to the RFC one, that was just for submission to the IETF

PR #60, opened 2016-07-25
Docs: clarify LedgerPlugin interface

bensharaf:
From conversations with @justmoon . Clarifies when errors should be thrown, which methods require connection prior to their invocation, adds a `"ledger"` field to transfers so that the connector doesn't have to add it, and fixes conflicting information regarding `getInfo()`.

### DIFF
--- a/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
+++ b/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
@@ -104,7 +104,7 @@ Query whether the plugin is currently connected.
 #### getInfo
 <code>ledgerPlugin.getInfo() ⇒ Promise.&lt;[LedgerInfo](#class-ledgerinfo)></code>
 
-Retrieve some metadata about the ledger.
+Retrieve some metadata about the ledger. Plugin must be connected, otherwise the promise should reject.
 
 ###### Example Return Value
 ```json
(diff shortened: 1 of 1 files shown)


0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
```diff
@@ -104,7 +104,7 @@ Query whether the plugin is currently connected.
 #### getInfo
 <code>ledgerPlugin.getInfo() ⇒ Promise.&lt;[LedgerInfo](#class-ledgerinfo)></code>
 
-Retrieve some metadata about the ledger.
+Retrieve some metadata about the ledger. Plugin must be connected.
```
justmoon: What happens if it isn't?

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
```diff
@@ -187,12 +192,14 @@ For a detailed description of these properties, please see [`OutgoingTransfer`](
 #### fulfillCondition
 <code>ledgerPlugin.fulfillCondition( **transferId**:String, **fulfillment**:Buffer ) ⇒ Promise.&lt;null></code>
 
-Submit a fulfillment to a ledger. The ledger plugin or the ledger MUST automatically detect whether the fulfillment is an execution or cancellation condition fulfillment.
+Submit a fulfillment to a ledger. The ledger plugin or the ledger MUST
+automatically detect whether the fulfillment is an execution or cancellation
```
justmoon: > The ledger plugin or the ledger MUST automatically detect whether the fulfillment is an execution or cancellation condition fulfillment.

You can take that sentence out I think since we're no longer exposing cancellation conditions on the ILP layer.

justmoon:
LGTM

Issue #31, opened 2016-07-01
Proposal: ILP Address Mapping


justmoon:
So far we haven't explained how ILP addresses would actually map to local ledger addresses. We need a scheme similar to the Address Resolution Protocol (ARP) in the IP stack.
## Flow
1. When an Interledger module wants to send a payment, it must first determine whether the payment is local for any of its ledger plugins. Local means that the address starts with the ledger plugin's address prefix.
2. If yes, send through that plugin.
3. If no, resolve using ILP routing table to a connector's ILP address and send to that address. (All next hop addresses in the routing table MUST be local.) Typically, clients' routing tables would contain a single entry with a catch-all (gateway).
### Examples

We are `us.fed.wf.alice`.

_Note:_ The `send()` API is simplified in this example.

```
core.isLocalAddress("us.fed.wf.bob") -> true
plugin.send("us.fed.wf.bob", "30.00")
  - Is local? Yes (otherwise throw exception)
  - My prefix is "us.fed.wf", so local identifier is "bob"

core.isLocalAddress("uk.gov.bofe.rbs.john") -> false
(not local, => look up in routing table - returns gateway ilp address)
plugin.send("us.fed.wf.connie", "30.00")
  - Is local? Yes (otherwise throw exception)
  - My prefix is "us.fed.wf", so local identifier is "connie"

core.isLocalAddress("us.fed.wf.nerd.frednoob") -> true
plugin.send("us.fed.wf.nerd.frednoob", "30.00")
  - Is local? Yes (otherwise throw exception)
  - My prefix is "us.fed.wf", so local identifier is "nerd"
```
## Address Format Change

Before:

```
us.fed.wf/bob
```

After:

```
us.fed.wf.bob
```

Previously, we thought that being able to differentiate an account address from a ledger address was a useful feature. But as we investigated the details of the address resolution process, we discovered that this distinction makes it impossible for `us.fed.wf/bob` to be a ledger with subaccounts.

We were able to find the scheme outlined in this issue which properly allows for local address resolution without distinguishing between ledger addresses and account addresses.
## Ledger Plugin API Changes

We want to add a new method to the ledger plugin:

```
getPrefix() -> Promise.<String>
```

`getPrefix` returns the ledger plugin's ILP address `prefix` (which may be configured, automatically detected, hard-coded, etc.) 

If an ILP address begins with a ledger plugin's prefix, the address is considered local with respect to that plugin.

[...]

PR #62, opened 2016-08-01
RFC-4: ILP addresses, remove cancellationCondition, add rejectIncomingTransfer

emschwartz:
Updates the ledger plugin interface spec with various changes that have been discussed in the issues.

### DIFF
--- a/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
+++ b/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
@@ -23,7 +23,6 @@ This spec depends on the [ILP spec](../0003-interledger-protocol/).
 | | [**getPrefix**](#getprefix) ( ) `⇒ Promise<String>` |
 | | [**getAddress**](#getaddress) ( ) `⇒ Promise<String>` |
 | | [**getBalance**](#getbalance) ( ) <code>⇒ Promise.&lt;String></code> |
-| | [**getConnectors**](#getconnectors) ( ) <code>⇒ Promise.&lt;Array.&lt;String>></code> |
 | | [**send**](#send) ( transfer ) <code>⇒ Promise.&lt;null></code> |
 | | [**fulfillCondition**](#fulfillcondition) ( transferId, fulfillment ) <code>⇒ Promise.&lt;null></code> |
 | | [**replyToTransfer**](#replytotransfer) ( transferId, replyMessage ) <code>⇒ Promise.&lt;null></code> |
@@ -149,11 +148,6 @@ The mapping from the ILP address to the local ledger address is dependent on the
 <code>ledgerPlugin.getBalance() ⇒ Promise.&lt;String></code>
 
 Return a decimal string representing the current balance. Plugin must be connected, otherwise the promise should reject.
-
-#### getConnectors
-<code>ledgerPlugin.getConnectors() ⇒ Promise.&lt;Array.&lt;String>></code>
-
-Return an array of opaque local destination identifiers representing neighboring connectors. Plugin must be connected, otherwise the promise should reject.
 
 #### Event: `connect`
 <code>ledgerPlugin.on('connect', () ⇒ )</code>
--- a/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
+++ b/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
@@ -26,7 +26,7 @@ This spec depends on the [ILP spec](../0003-interledger-protocol/).
 | | [**send**](#send) ( transfer ) <code>⇒ Promise.&lt;null></code> |
 | | [**fulfillCondition**](#fulfillcondition) ( transferId, fulfillment ) <code>⇒ Promise.&lt;null></code> |
 | | [**replyToTransfer**](#replytotransfer) ( transferId, replyMessage ) <code>⇒ Promise.&lt;null></code> |
-| | [**rejectIncomingTransfer**](#rejectincomingtransfer) ( transferId ) <code>⇒ Promise.&lt;null></code> |
+| | [**rejectIncomingTransfer**](#rejectincomingtransfer) ( transferId, rejectMessage ) <code>⇒ Promise.&lt;null></code> |
 
 ###### Events
 | Name | Handler |
@@ -215,9 +215,9 @@ Submit a fulfillment to a ledger. Plugin must be connected, otherwise the promis
 **TODO**: Define what the message format is.  Plugin must be connected, otherwise the promise should reject.
 
 #### rejectIncomingTransfer
-<code>ledgerPlugin.rejectIncomingTransfer( **transferId**:String ) ⇒ Promise.&lt;null></code>
+<code>ledgerPlugin.rejectIncomingTransfer( **transferId**:String, **rejectMessage**:Buffer ) ⇒ Promise.&lt;null></code>
 
-Reject an incoming transfer that is held pending the fulfillment of its `executionCondition` before the `expiresAt` time.
+Reject an incoming transfer that is held pending the fulfillment of its `executionCondition` before the `expiresAt` time. `rejectMessage` MAY be supplied to provide details on why the transfer was rejected.
 
 This MAY be used by receivers or connectors to reject incoming funds if they will not fulfill the condition or are unable to forward the payment. Previous hops in an Interledger transfer would have their money returned before the expiry and the sender or previous connectors MAY retry and reroute the transfer through an alternate path.
 
--- a/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
+++ b/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
@@ -305,14 +305,15 @@ means that a transfer you created has timed out.
 <code>class Transfer</code>
 
 The `Transfer` class is used to describe local ledger transfers. Only
-[id](#id), [address](#address), [ledger](#ledger), and [amount](#amount) are required; the other
+[id](#id), [account](#account), [ledger](#ledger), and [amount](#amount) are required; the other
 fields can be left undefined (but not any other false-y value) if unused.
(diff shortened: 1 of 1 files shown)


0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
```diff
@@ -20,11 +20,13 @@ This spec depends on the [ILP spec](../0003-interledger-protocol/).
 | | [**disconnect**](#disconnect) ( ) `⇒ Promise.<null>` |
 | | [**isConnected**](#isconnected) ( ) `⇒ Boolean` |
 | | [**getInfo**](#getinfo) ( ) <code>⇒ Promise.&lt;[LedgerInfo](#class-ledgerinfo)></code> |
+| | [**getPrefix**](#getprefix) ( ) `⇒ Promise<String>` |
+| | [**getAddress**](#getaddress) ( ) `⇒ Promise<String>` |
```
adrianhopebailie: What's the thinking behind `getPrefix`? Who would call this and what's the use case? Is it discussed in an existing issue?

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
```diff
@@ -199,12 +208,19 @@ Submit a fulfillment to a ledger. Plugin must be connected, otherwise the promis
 
 **TODO**: Define what the message format is.  Plugin must be connected, otherwise the promise should reject.
 
+#### rejectIncomingTransfer
+<code>ledgerPlugin.rejectIncomingTransfer( **transferId**:String ) ⇒ Promise.&lt;null></code>
+
+Reject an incoming transfer that is held pending the fulfillment of its `executionCondition` before the `expiresAt` time.
+
```
adrianhopebailie: Does this mean the protocol doesn't allow ledgers to reject a transfer that is unconditional?

adrianhopebailie: What should a connector do with an optimistic payment that it can't forward?

sentientwaffle:
https://github.com/interledger/five-bells-connector/issues/198 is going to remove `getPrefix` and just include it in the transfer objects as `ledger`. The only other place it is called is https://github.com/interledger/js-ilp-core/blob/master/src/lib/client.js#L105, which could just use `getAddress()` instead.

emschwartz:
@sentientwaffle https://github.com/interledger/rfcs/issues/31 says `core` should use the `getPrefix` method to determine whether an address is "local" to a ledger plugin

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
emschwartz: > Does this mean the protocol doesn't allow ledgers to reject a transfer that is unconditional?

The protocol doesn't allow or disallow any functionality ledgers (and corresponding plugins) choose to implement. This interface specifies the functionality necessary for clients and connectors to use plugins interchangeably for sending and receiving ILP payments.

Right now the "authorized credits" feature is not part of this interface. It could be added, but we need to consider whether this is part of the minimal functionality we expect from all ledgers and plugins.

Note that ledger plugin implementations MAY emit more events and provide more functionality than specified here. However, we should define the interface such that the vast majority will implement all of the minimum functionality described here.

> What should a connector do with an optimistic payment that it can't forward?

Send it back or keep it.

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
emschwartz: See `Ledger Plugin API Changes` proposed here: https://github.com/interledger/rfcs/issues/31#issue-163459486

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
adrianhopebailie: > The protocol doesn't allow or disallow any functionality ledgers (and corresponding plugins) choose to implement. 

I disagree. A protocol, by definition, MUST define what implementations of the protocol are allowed and not allowed to do. That's what a protocol is, a set of rules that everyone must follow to be interoperable.

> This interface specifies the functionality necessary for clients and connectors to use plugins interchangeably for sending and receiving ILP payments.

I think this is the crux of the misunderstandings around this plugin interface and the protocol itself. The protocol defines the required behaviour of ledgers and connectors however the ledger plugin is a representation of the ledger interface running in the connector's processing space.

The design of the ledger interface **for the purpose of defining the requirements of a ledger to be conformant to ILP** should be a ledger API (even an abstract one) that is assumed to be running in the ledger's processing space.

Once you have this interface defined you can define a way this can be implemented as a plugin that runs in the connector's processing space instead.

By doing it as a plug-in first we are making design decisions or inferring things about the protocol that are not correct.

For example, you have said that _the "authorized credits" feature is not part of this interface_ but that's an implementation decision. The protocol needs to define how a ledger and connector interact when a transfer is received by a connector that has no condition and the connector wishes to reject it. 

The reasons for rejecting the transfer (eg: it's not authorized) are irrelevant, those are business rules which are implementation specific. The rules of the protocol either need to prevent this from happening (all transfers must have a condition) or define the behavior of the participants when it does (receivers can reject an incoming transfer even if it has no condition).

> Note that ledger plugin implementations MAY emit more events and provide more functionality than specified here. However, we should define the interface such that the vast majority will implement all of the minimum functionality described here.

No, a protocol needs to define the behavior for all cases not just the majority. If this interface is supposed to be an implicit definition of ILP then it must define the necessary events and methods used by all participants for all cases.

If the purpose of this interface is not to cover all cases for an ILP transfer then we need to be explicit about that and have somewhere else that this behaviour is documented.

> Send it back or keep it.

How? How does a connector that can't route a payment send it back and notify the sender that it is linked to the original payment? The protocol needs to define this behavior, if not in this interface then we need to do it somewhere else.

emschwartz: Continuing discussion on https://github.com/interledger/rfcs/issues/64

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
```diff
@@ -176,19 +198,12 @@ implement zero-amount transfers differently than other transfers.
 ```js
 p.send({
   id: 'd86b0299-e2fa-4713-833a-96a6a75271b8',
-  account: 'https://ledger.example/accounts/connector',
+  address: 'example.ledger.connector',
```
justmoon: I think `account` is better than `address` because address is the type/format of the value, whereas account is the thing described. Which address is it? It's the address of the account used by this ledger plugin.

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
```diff
@@ -288,16 +303,15 @@ means that a transfer you created has timed out.
...
 fields can be left undefined (but not any other false-y value) if unused.
 
 ###### Fields
 | Type | Name | Description |
 |:--|:--|:--|
 | `String` | [id](#id) | UUID used as an external identifier |
-| `String` | [account](#account) | Local source or destination account ID |
+| `String` | [address](#address) | ILP Address of the source or destination account |
 | `String` | [amount](#amount) | Decimal transfer amount |
-| `String` | [ledger](#ledger) | Ledger address |
```
justmoon: I think `ledger` would be there and be the prefix, e.g. `example.ledger.` - it's convenient to have so you can easily tell which ledger plugin this came from / should go to.

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
justmoon: `getAddress` should be `getAccount` imo, see https://github.com/interledger/rfcs/pull/62/commits/6e6851e26897b6d23d64eee723b9a334a4d3d28a#r73805565

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
```diff
@@ -213,6 +214,13 @@ Submit a fulfillment to a ledger. Plugin must be connected, otherwise the promis
 
 **TODO**: Define what the message format is.  Plugin must be connected, otherwise the promise should reject.
 
+#### rejectIncomingTransfer
+<code>ledgerPlugin.rejectIncomingTransfer( **transferId**:String ) ⇒ Promise.&lt;null></code>
```
justmoon: I think this should include a return message (of type `Buffer`), so we can give more context why we rejected the transfer.

emschwartz: Do you mean a return method or another parameter?

justmoon: Another parameter.

Though this may require some more systematic discussion. We have `fulfill` (success), `rejectIncomingTransfer` (failure) and `replyToTransfer` (neutral). Right now, we can send messages only with the latter.

* What use cases are there for `replyToTransfer`?
* Is there a use case for returning a message with success (`fulfill`)?

emschwartz: Added the `rejectMessage`.

> What use cases are there for replyToTransfer?

Originally it was intended for sending error messages back. It might be useful in the case of Optimistic payments to notify the sender whether the transfer got through. I can't think of anything else right now though.

> Is there a use case for returning a message with success (fulfill)?

The only thing I could think of would be something like "I did this one, but you (sender) should do something different next time". Not a very important use case. One way to look at it would be "it's finished, we don't need to talk about it anymore". If we want to add something like this it should be done in a new PR.

justmoon:
Just waiting for [this comment](https://github.com/interledger/rfcs/pull/62#discussion_r73805593) to be addressed, otherwise LGTM.

Issue #55, opened 2016-07-14
RFC-4: Flatten ledger plugin PluginOptions


justmoon:
Mentioned in #31 under "LedgerPlugin API Changes". The current PluginOptions object which is passed to the LedgerPlugin constructor has an `auth` field and a `store` field. However, the `store` may or may not be there depending on the plugin, and the contents of `auth` are plugin specific. The `auth` object can be flattened into the PluginOptions object, and special fields like store which cannot be represented in JSON are identified with a preceding underscore. The format would change from:

```js
{
  auth: { /* plugin specific fields*/ }
  store: {get, put, del}
}
```

To:

```js
{
  /* plugin specific fields */,
  _store: {get, put, del}
}
```

justmoon:
Just to document the rationale for posterity: It turned out that plugins had custom options that didn't really fall under the term `auth`. In fact most of the time the plugin options are quite different. As a result, we decided to make the options object custom to each plugin and just reserve properties starting with underscore as special (i.e. they should not be used by the individual plugins.)

emschwartz:
@bensharaf this is done, right?

bensharaf:
Yeah.

PR #67, opened 2016-08-04
update PluginOptions

bensharaf:
Addresses issue https://github.com/interledger/rfcs/issues/55 . Also removes `canConnectToLedger`, because it uses the outdated `auth` field.


emschwartz:
LGTM

Issue #65, opened 2016-08-03
RFC-9: Questions about design + purpose of SPSP


emschwartz:
I am looking over SPSP and have a number of questions about its design and purpose.

It seems like many of the options it was originally intended to bundle have since ceased to be options and are now fixed in other layers (e.g. sending "Universal" mode payments is now just sending ILP payments, and much of the request/response payment flow has been implemented in [ITP](https://github.com/interledger/rfcs/blob/master/0011-interactive-transport-protocol/0011-interactive-transport-protocol.md)).

As far as I can tell, this is what SPSP provides:

## SPSP Features
- Optional email-style identifiers for human readability
- Handling end-to-end communication (which is explicitly left out of ITP) using HTTP
- Standard API for peer-to-peer and invoice payments, two common use cases

## Use Cases

Steps that are actually SPSP (versus a lower level protocol) are in **bold** (it's not that much)

### Sending Peer-to-peer Payment, Fixed Destination Amount
**0. (Optional) Discover - resolve human-readable identifier into URL**
**1. Setup - post payment params (amount, maybe sender id), get ILP Packet**
2. Quote - check how much it would cost to delivery the packet
3. Execute

### Paying Invoice
**0. (Optional) Discover - resolve human-readable identifier into URL**
**1. Setup - get ILP Packet (no params)**
2. Quote - check how much it would cost to delivery the packet
3. Execute

### Sending Peer-to-peer Payment, Fixed Source Amount
**0. (Optional) Discover - resolve human-readable identifier into URL**
**1. Query - get ILP address from receiver endpoint**
2. Quote - find out how much receiver will receive
**3. Setup - post payment params, get packet**
4. Execute (create transfer based on original source amount, NOT based on packet)

## Questions

- Does the fixed destination amount use case make sense to include? How would the sender know how many units it would want the destination to receive if it doesn’t even know what ledger the destination is on?
- If the only reason we need the Query step is to return the ILP address for fixed source amount payments so you can quote a payment to it, can we combine that endpoint with the Setup one? (https://github.com/interledger/rfcs/issues/49 proposes combining Query and Quote, but Query goes to the receiver endpoint, whereas Quote goes through core ILP/ILQP to the connector)
[...]

Issue #49, opened 2016-07-13
RFC-9: Combine Query and Quoting phases


justmoon:
The query phase should just include all information needed to locally calculate the behavior/cost for different amounts, to avoid a separate quoting call.

adrianhopebailie:
Do we need a quoting phase at all?

How a node (sending client or connector) decides what it thinks the cost will be to send a payment to an address is not tightly coupled to the actual sending of the payment.

The node makes some decision about how much to transfer on one of it's connected ledgers when it passes that ledger a new ILP payment, how it comes to that decision isn't actually part of the protocol.

I'd propose that we mention quoting as an orthogonal process that helps the node decide how much it must transfer but not make it an explicit phase in the protocol

justmoon:
> Do we need a quoting phase at all?

No, this issue proposes removing it.

emschwartz:
As mentioned in https://github.com/interledger/rfcs/issues/65 I'm not sure whether we should combine Query + Quote or Query + Setup.

The Query step goes to the receiver while the Quote goes to the connector. How could the receiver know the full path from the sender?

adrianhopebailie:
I don't think quoting should be specified as part of any of the other protocols at all. Quoting is orthogonal to everything else in ITP and SPSP. It's a step in the overall process but not and end-to-end protocol.

The fact that SPSP tries to handle both fixed sender and fixed receiver pricing complicates things because it means we have to fit quoting in the middle sometimes but can do it after the SPSP steps other times.

Maybe we should break out and define the steps more clearly so we can be clear what steps a particular protocol handles?

This would also allow us to define SPSP in terms of these steps and how the order changes if the sender or receiver price is fixed.

1. __Account Discovery__ : What we call "query" now should be "address discovery" (i.e. discovering the ILP address of the receiver) because calling it query assumes that info will always be fetched from the receiver via a query (true for SPSP but not for all setup protocols).

2. __Destination Amount Discovery__: This can either be done via quoting or it is known up front. It could also be provided by the receiver in response to a query where the identifier is for an invoice not an account.

3. __Condition and Timeout Discovery__: Constructing the condition and determining the timeout. In ITP this is done by the receiver and sent back to the sender to use but it could also be provided in an invoice (declarative) or provided by some other system.

4. __Verification__: The process of preparing the payment may involve getting other information such as receiver names, public keys, images etc to verify the receiver.

[...]

Issue #37, opened 2016-07-07
Currency on simple-payment-setup-protocol Invoice 


rozzamozza:
There is am amount in the invoice protocol but no currency or currency code.
What's the policy/agreement re currencies. Shouldn't amounts always be accompanied by a currency?

emschwartz:
In the Interledger model we say that a ledger is a system that tracks assets of one and only one type. Therefore, the currency is defined by the ledger. In some application layer protocols, the currency code or symbol may be included to be able to display it for the user, but it is not necessary for the core protocol.

PR #75, opened 2016-08-09
SPSP overhaul

emschwartz:
depends on https://github.com/interledger/rfcs/pull/76
resolves https://github.com/interledger/rfcs/issues/65
resolves https://github.com/interledger/rfcs/issues/57
resolves https://github.com/interledger/rfcs/issues/37
SPSP now consists of a single receiver endpoint that you can query using HTTP GET or set up a payment with using HTTP POST
discovery phase is now optional and in an appendix
use ILP addresses
remove quoting phase (resolves https://github.com/interledger/rfcs/issues/49)
add details about data formats


0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -1,47 +1,104 @@
...
+8. The sender's SPSP client uses its ILP module to create a transfer **using the chosen source amount, NOT by quoting the ILP packet**, and attach the ILP packet to the transfer.
+9. The receiver's ILP module registers the incoming transfer held pending the fulfillment of the Crypto Condition. It validates that the transfer matches the packet and regenerates the condition fulfillment using the [Interactive Transport Protocol (ITP)](../0011-interactive-transport-protocol). The ILP module submits the fulfillment to execute the transfer and claim the funds.
+10. The sender's SPSP client receives a notification from its ILP module that the transfer has been executed, including the condition fulfillment from the recipient, and notifies the sender that the payment is completed.
+
+#### Invoice
+
+1. The sender's SPSP client queries the receiver endpoint.
+2. The receiver endpoint responds with the receiver info, including the invoice status, amount, and currency code.
+3. The sender's SPSP client submits the sender's info, including the sender's ILP address to the receiver endpoint.
```
emschwartz: @justmoon what do you think the client should submit to the receiver endpoint for an invoice? An empty `POST` or some information about themselves? I put the sender's ILP address in case the receiver might have the sender send to a different ILP address depending on where they're sending from.

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -137,11 +187,12 @@ HTTP/1.1 201 Created
 Content-Type: application/json
 
 {
-  "receipt_condition": "cc:1:1:47DEQpj8HBSa-_TImW-5JCeuQeRkm5NMpJWZG3hSuFU:1"
+  "packet": "asdglkjlsdfoiuaosiduf...",
+  "condition": "cc:0:3:47DEQpj8HBSa-_TImW-5JCeuQeRkm5NMpJWZG3hSuFU:0"
 }
```
emschwartz: @justmoon If the `condition` is communicated separately from the `packet` should this be some binary format to prevent having to encode and decode base64?

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -1,47 +1,104 @@
...
+## Specification
+
+The receiver endpoint will respond to HTTP `GET` and `POST` requests in the following manner:
+
+### Query (`GET <receiver>`)
+
+The sender queries the receiver endpoint to get information about the type of payment that can be made to this receiver:
 
 ``` http
 GET /api/receivers/bob HTTP/1.1
```
adrianhopebailie: 1. Is ```GET /api/receivers/?identifer=bob@red.ilpdemo.org``` valid?
2. Is ```GET /api/receivers/bob?currency_code=ZAR``` valid?
3. Is ```GET /api/receivers/bob?type=payee``` valid?

emschwartz: My initial answers:

1. Don't think so. 
2. Good question. I _think_ it's on the receiver to decide that.
3. Unless we have types aside from `payee` and `invoice`, no. If there are only these two I think it makes sense for an endpoint to be one or the other. If someone gives you an invoice receiver URL, then you'll use it as such. If you have a payee endpoint I can't think of a use case where you'd need to go from that to an invoice (because you'd get the invoice URL from some other context)

Another way to look at this though would be that the receiver can give you any valid URL that you can treat as opaque and it will work. I think I like that better.

adrianhopebailie: My thinking here is ito the restrictions we may implicitly be making on discovery.

Assume a sender has some identifier for the receiver and uses some discovery mechanism to resolve that to the SPSP receiver endpoint URL. All they have then is the URL and the identifier and they may need to still perform functions like enquiring what accounts that receiver has or who the receiver is.

Related to recent comments in #65 

If we say that 1. above is not allowed then we are saying that discovery mechanisms MUST always have a unique URL per identifier (as opposed to allowing a URL from a single service provider/ledger). Is that a restriction we are happy with?

Similar with 2, if the discovery mechanism simply maps a receiver identifier to the SPSP endpoint then we need to assume that that receiver may have multiple accounts right?

emschwartz: I think there's an argument to be made for returning all the data we think (or say, since we're defining the scope of this protocol) senders will need in the Query response. Who the receiver is is probably included in that. Are what accounts the receiver has relevant? Arguably, they get to decide which account you're dealing with.

There are three ways I could see us going regarding query parameters:

1. The sender must treat the receiver endpoint as completely opaque. The URL MAY include query parameters but those are decided by the recipient and cannot be modified by the sender.
2. The recipient can include some query parameters in the URL but we also define some standard ones that receivers SHOULD understand.
3. The base receiver URL cannot have query parameters in it and we standardize some query parameters that the receiver SHOULD understand.

justmoon: > Another way to look at this though would be that the receiver can give you any valid URL that you can treat as opaque and it will work. I think I like that better.

:+1: 

> Are what accounts the receiver has relevant?

That's a question about the use cases. SPSP could allow the sender to select from multiple choices of destination. I don't think that's a very common use case, since most receivers today deal only in their native currency. And it complicates the sender implementation. I'd leave it out of this version of SPSP.

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -1,62 +1,127 @@
 # Simple Payment Setup Protocol (SPSP)
 
-The Simple Payment Setup Protocol (SPSP) is a basic application-layer Interledger protocol for wallet-to-wallet or wallet-to-merchant payments.
+## Preface
 
-## Stages
+This document describes the Simple Payment Setup Protocol (SPSP), a basic application-layer Interledger protocol for exchanging payment information between senders and recipients.
```
adrianhopebailie: Suggest:

This document describes the Simple Payment Setup Protocol (SPSP), a basic protocol for exchanging payment information between senders and recipients to setup an Interledger payment.

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -1,62 +1,127 @@
...
+
+#### Receiver
+
+The specific HTTP endpoint on the SPSP used for setting up a payment.
+
+## Overview
+
+### Relation to Other Protocols
+
+SPSP is used for exchanging payment information before an ILP payment is initiated. It uses the [Interactive Transport Protocol (ITP)](../0011-interactive-transport-protocol) for generating conditions.
```
adrianhopebailie: Are we still using ITP? I thought we had a better name for that process?

emschwartz: See https://github.com/interledger/rfcs/pull/76
I'll update this PR so it depends on that other one

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -1,62 +1,127 @@
...
 
 {
   "type": "payee",
-  "ledger": "https://red.ilpdemo.org/ledger",
-  "account": "https://red.ilpdemo.org/ledger/accounts/bob",
-  "payments": "https://red.ilpdemo.org/api/receiver/bob/payments"
+  "account": "ilpdemo.red.bob",
+  "currency_code": "USD",
+  "currency_symbol": "$",
+  "first_name": "Bob",
```
adrianhopebailie: What about sending to a company?
Could this just be `name` or `receiver_name`?

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -1,62 +1,127 @@
...
 HTTP/1.1 200 OK
 Content-Type: application/json
 
 {
   "type": "payee",
-  "ledger": "https://red.ilpdemo.org/ledger",
-  "account": "https://red.ilpdemo.org/ledger/accounts/bob",
-  "payments": "https://red.ilpdemo.org/api/receiver/bob/payments"
+  "account": "ilpdemo.red.bob",
+  "currency_code": "USD",
```
adrianhopebailie: This was a HUGE debate in the wpwg. What about alt currencies?

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -96,52 +174,102 @@ Content-Type: application/json
...
+  "additional_routing_info": {
+    "something": "that helps the sender's client deliver the packet to the recipient"
+  }
 }
 ```
+|---|---|---|
+| `packet` | Object | JSON representation of the ILP Packet used by the sender's client to construct the binary ILP Packet |
+| `condition` | Crypto Condition | Execution condition the sender must use for the payment to the recipient |
+| `additional_routing_info` | Object | Additional info used by the sender's client to route the payment to the recipient |
```
adrianhopebailie: Not sure what this additional_routing_info would be? Shouldn't we just leave it out until we have a clearer definition?

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
emschwartz: This is for display purposes so we could say that both `currency_code` and `currency_symbol` are just strings and it is recommended that currencies that do have ISO strings use those for the code. Sender UIs will need a way to render strings they don't have hard-coded in.

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
emschwartz: When paying companies you'd probably use an invoice, no?

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
emschwartz: We already know we're going to need the receiver to specify the rate curve from a well known ledger to their ledger, but I'm not entirely sure about the data format. It's probably a good idea to specify it more precisely here, since clients would need to understand that.

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
justmoon: Makes sense!

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
justmoon: I'm with @adrianhopebailie, we shouldn't assume anything about the format of the name. Even for individuals, `first_name` + `last_name` doesn't work in many cases. And organizations could definitely be payees.

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
justmoon: We may not have `additional_routing_info` after all. We're still working on the details of the routing, but so far it's looking like just routing all the way to the receiver interactively is actually fine (and much simpler.) This isn't the right place to get into the details, but I'm with Adrian, let's remove it until we have more clarity.

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
justmoon: Eventually, we'll probably want to overhaul the asset system. I'm thinking that there would be a very generic, catch-all way of describing an asset uniquely and then a bunch of templates that allow you to provide richer definitions for certain asset types, e.g. currencies, bonds, etc.

I specifically **don't** want us to get distracted by that right now, since it's not part of the ILP layer and it's quite a large task to do properly.

justmoon: So I'm in favor of sticking with `currency_code` and `currency_symbol` for now.

justmoon:
LGTM

adrianhopebailie:
LGTM 👍

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
vhpoet: @emschwartz did you forget to update this part?

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -1,62 +1,126 @@
...
+2. The receiver endpoint responds with the receiver info, including the recipient's ILP address.
+3. The sender choses an amount of their currency or assets to send.
+4. The sender's SPSP client uses its ILP module to quote how much of the recipient's currency or asset type will be delivered to the recipient's ILP address for the given source amount.
+5. The sender accepts the quote.
+6. The sender's SPSP client submits the payment information, including the destination amount, to the receiver endpoint.
+7. The receiver endpoint responds with an ILP Packet and Crypto Condition corresponding to the destination amount.
+8. The sender's SPSP client uses its ILP module to create a transfer **using the chosen source amount, NOT by quoting the ILP packet**, and attach the ILP packet to the transfer.
+9. The receiver's ILP module registers the incoming transfer held pending the fulfillment of the Crypto Condition. It validates that the transfer matches the packet and regenerates the condition fulfillment using the [Interactive Transport Protocol (ITP)](../0011-interactive-transport-protocol). The ILP module submits the fulfillment to execute the transfer and claim the funds.
```
vhpoet: `0011-interactive-transport-protocol` is removed.

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -1,62 +1,126 @@
...
 
-### Query
+1. The sender's SPSP client queries the receiver endpoint.
+2. The receiver endpoint responds with the receiver info, including the invoice status, amount, and currency code.
+3. The sender's SPSP client submits the sender's info, including the sender's ILP address to the receiver endpoint.
+4. The receiver endpoint responds with an ILP Packet and Crypto Condition corresponding to the destination amount.
+5. The sender's SPSP client uses its ILP module to get a quote in their currency or asset type for the ILP transfer.
+6. The sender accepts the quote.
+7. The sender's SPSP client uses its ILP module to initiate the ILP transfer.
+8. The receiver's ILP module registers the incoming transfer held pending the fulfillment of the Crypto Condition. It validates that the transfer matches the packet and regenerates the condition fulfillment using the [Interactive Transport Protocol (ITP)](../0011-interactive-transport-protocol). The ILP module submits the fulfillment to execute the transfer and claim the funds.
```
vhpoet: same here

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -1,62 +1,126 @@
...
+9. The sender's SPSP client receives a notification from its ILP module that the transfer has been executed, including the condition fulfillment from the recipient, and notifies the sender that the payment is completed.
 
-Any SPSP recipient will create a receiving endpoint called the *receiver*. The sender queries this endpoint to get information about the type of payment that can be made to this receiver:
+## Specification
 
+The receiver endpoint is a URI used by the sender to query information about the sender and set up payments. The receiver URI MAY contain query string parameters. The sender SHOULD treat the URI as opaque.
+
+The receiver endpoint MUST respond to HTTP `GET` and `POST` requests in the following manner:
```
vhpoet: should this be HTTPS?

Issue #56, opened 2016-07-17
Define normalization algorithm for header data to be used in conditions


adrianhopebailie:
From: https://github.com/interledger/rfcs/issues/54#issuecomment-232892378

To use header data in the condition it must be normalized before it is signed/hashed. The algorithm for this needs to be standardized so it is repeatable and deterministic.

emschwartz:
To be more precise, the ASN.1 `SEQUENCE` types already define the ordering of the fields within a given header. I think what you're saying is that we should have a recommended ordering for the different headers attached to a packet (see [IPv6](https://tools.ietf.org/html/rfc2460#section-4.1))

adrianhopebailie:
Yes. If the data that is being signed includes multiple headers (ILP + ITP for example) then we need to define how to normalize the combined data

Issue #68, opened 2016-08-04
ILP Addresses per transaction?


adrianhopebailie:
In ITP we specify that the receiver can set a UUID for the transaction however it seems equally viable to simply append that to the receiver's address allowing the receiver to:

1. Include the transaction identifier in the condition
2. Easily reconcile the transaction later

emschwartz:
You can add transaction identifiers to your ILP address.

One difference that may come into play later is that ILP addresses will be visible to everyone in the path, whereas parts of the ILP packet may be encrypted. This isn't done now so at the moment it may not make a difference.

> 1. Include the transaction identifier in the condition
> 2. Easily reconcile the transaction later

Both of these are covered either way

adrianhopebailie:
> Both of these are covered either way

Right, but by putting them in the address you can have only the ILP header in the condition and not need any other data. That's quite efficient and less complex.

It means that the receiver only has to give the sender the ILP header.

This was also more of a general question about whether we should encourage this and design for it explicitly. I think it's a nice feature of how we do addresses that we don't really exploit

emschwartz:
Great point. I think I'll actually be able to remove the use of the `data` field in the `js-ilp` implementation of ITP as soon as [this](https://github.com/interledger/js-ilp-core/issues/32) is done.

Issue #66, opened 2016-08-03
RFC-11: Rename ITP to Payment Request Protocol (PRP) 


emschwartz:
Our analogy with IP Transport protocols is imprecise and calling ITP and other such protocols is confusing because they aren't actually "transporting" anything. Unless we can come up with a better name for this layer I would propose changing the name to something more descriptive of the actual functionality, like Payment Request Protocol.

emschwartz:
cc @justmoon @adrianhopebailie

adrianhopebailie:
+1 on not using "transport"

I think in this instance the protocol (ITP) is dependent on the receiver providing all the data so it is in effect generating a payment request. On that basis I think Payment Request Protocol sounds good.

justmoon:
Agree with not using transport. Not a big fan of "payment request protocol" because it doesn't describe how a payment is requested, which is more SPSP's domain. I like "Interactive", because the requirement to talk to the receiver is an important characteristic of this protocol. (As opposed to non-interactive protocols that would use asymmetric crypto.)

Maybe "Interactive Condition Protocol (ICP)"? (And the people using it could be called [juggalos](https://en.wikipedia.org/wiki/Insane_Clown_Posse)... :wink:)

emschwartz:
I think your point about "interactive" is a good one.

However, I don't think "Interactive Condition Protocol" is right because this protocol isn't just about conditions. The receiver sets all of the parameters of the packet -- address, amount, additional routing info, and the condition. That was why I thought of those packets as payment requests. You're right though that that creates confusion with requests as HTTP requests.

"Interactive Packet Protocol" :/

"Interactive Packet Setup"... Is this a full "protocol"?

adrianhopebailie:
> "Interactive Packet Setup"... Is this a full "protocol"?

Yes and no. 

Short of discovery of the receiver's service endpoint this protocol handles everything required to setup the payment right?

What RFC 11 doesn't define is the transport specific for the protocol (e.g. How to do the interaction via HTTP vs JSON RPC vs ?) so it would be useful to understand where this will be defined if not as part of RFC 11 before naming the protocol in RFC 11.

emschwartz:
This intentionally leaves out discovery and the communication protocol
because having something like this is useful when you want to integrate
with different communication protocols.

SPSP (built on top of this) specifies using HTTP.

I'm also working on another that would use a different communication method.

adrianhopebailie:
[...]

PR #76, opened 2016-08-16
Replace ITP with the Interledger Payment Request

emschwartz:
- based on the discussion in https://github.com/interledger/rfcs/issues/66 and https://github.com/interledger/rfcs/issues/68
- puts the focus on the data format, rather than the protocol (though a brief overview of the flow is still in there for clarity)
- uses request-specific address instead of a dedicated request ID field (as proposed by @adrianhopebailie in https://github.com/interledger/rfcs/issues/68)
- includes a JS snippet to show how a preimage sha256 condition can be generated from a JSON payment request


emschwartz:
cc @justmoon @adrianhopebailie

emschwartz:
Resolves https://github.com/interledger/rfcs/issues/56

0011-interledger-payment-request/0011-interledger-payment-request.md
```diff
@@ -0,0 +1,135 @@
...
+| Field | Type | Short Description |
+|:--|:--|:--|
+| `address` | ILP Address | Request-specific address |
+| `amount` | Decimal String | Amount requested by the recipient |
+| `condition` | Crypto Condition | Execution condition for the payment |
+| `expires_at` | [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) Timestamp | Expiry of the request |
+| `additional_routing_info` | Object | Details used by the sender's client to route the payment |
+| `additional_headers` | Base64 String | Additional headers for the ILP packet |
+
+#### address
```
emschwartz: @justmoon Note I called it `address` rather than `account` because it is now request-specific

emschwartz:
One thing I realized while writing this is that if we use request-specific addresses as proposed in https://github.com/interledger/rfcs/issues/68, receivers probably need to accept a bit of overpayment.

I like the request-specific address idea, but it means that a connector cannot know if it is the last connector, or whether something after them is going to take some fees. This means they must apply whatever their normal rate is and pass on the resulting amount. If they happen to be the last connector this may mean that they are still a small amount over the precise amount requested by the recipient.

This could be seen as an argument against using request-specific addresses. However, I think it that using the address to route a payment to a specific application listening on one account is a perfectly valid use of the address format. Also, connectors really shouldn't make assumptions about who will take fees and should always pass on payments according to their normal rate.

0011-interledger-payment-request/0011-interledger-payment-request.md
justmoon: Can you clarify this?

You mean you called it address because you don't know whether it's an account or something else?

emschwartz: Correct. If we're using request-specific addresses it's not really an "account" address anymore. It's... just an address for that request

adrianhopebailie:
> it means that a connector cannot know if it is the last connector, or whether something after them is going to take some fees

Not always. The entity hosting the SPSP receiver may be the same entity operating the last connector in the chain so they know how much they need to send. This is probably a highly likely scenario.

justmoon:
> receivers probably need to accept a bit of overpayment.

I'm not a fan of that because that's something that would be visible to the end user. As a receiver, I want to receive the exact amount I expect.

I think the difficulty here boils down to me opening Pandora's box with the subledgering feature. So let's see how we can close it again... (ideally while keeping the subledgering feature and also keeping request-specific addresses)

> it means that a connector cannot know if it is the last connector

Connectors can currently [cleanly differentiate](https://github.com/interledger/five-bells-connector/blob/master/src/lib/route-builder.js#L127) between forwarding to another connector and local delivery.

So I would say for now they should deliver the exact amount when delivering locally. That does put some constraints on subledgers and request-specific addresses, namely, they can't charge fees and can't use different currencies.

For request-specific addresses that seems totally fine.

For subledgers, we do have the option of peering with a connector instead of using the subledgering option if we need exchange rate or fee flexibility. For the kinds of use cases I was thinking of with the subledgering (personal ledgers managing apps, accounts or corporate ledgers which are essentially accounting systems) this would be fine.

It's likely that we'll decide at some point that we want untrusted subledgers to also be able to charge fees and set exchange rates. The way I would do that is with a header that gives the last "regular" connector the information that it needs to deliver the correct amount. But I'd like to remove the time pressure by using the solution with the above stated limitations before thinking about how to design a more feature-rich version.

0011-interledger-payment-request/0011-interledger-payment-request.md
```diff
@@ -78,6 +82,10 @@ The Crypto Condition to be used as the execution condition for the payment. See
 
 The timestamp when the request expires. Recipients SHOULD NOT fulfill the conditions of incoming transfers that arrive after the expiry has passed. Senders MAY use the request expiry to determine the expiry of the transfer they put on hold for the first connector.
 
+#### data
+
+An arbitrary JSON object that should be included in the ILP Packet's user data header.
```
emschwartz: @justmoon What's the current thinking on data headers in the ILP Packet? Would it make sense to have this as JSON and have the sender serialize it into one of the headers or would that be much more complicated than having this encoded in base64 in the payment request?

emschwartz:
I agree that receiving too much is not a good user experience, but I'm not sure I totally understand how connectors know if they are delivering locally.

How does a connector that has an account on `us.fed.bigbank.` differentiate between what it needs to do with a payment for `us.fed.bigbank.someaccount.somerequestid` and `us.fed.bigbank.littlebank.accountholder`? Is the idea that if a connector doesn't have a routing table entry for the `littlebank` connector it will assume it is the last hop and try using `ilp-core` to deliver the exact amount specified in the packet to the `littlebank` account on `us.fed.bigbank.`?

It seems like the request-specific addresses will break how `ilp-core` [determines what ledger an address is on](https://github.com/interledger/js-ilp-core/blob/master/src/lib/core.js#L178-L184) and maybe how it determines whether an address is "local".

On a different topic, @justmoon What's the latest status on having data header(s) in the ILP Packet? Should the `data` field in the request be JSON that the sender serializes into binary to include in a data header in the packet? Or should the request data be communicated as a base64 string?

justmoon:
> Is the idea that if a connector doesn't have a routing table entry for the littlebank connector it will assume it is the last hop and try using ilp-core to deliver the exact amount specified in the packet to the littlebank account on us.fed.bigbank.?

Yes, except I don't like the word *assume*. We are *defining* the behavior as:

* If a connector **forwards** the payment to another connector, it will send enough such that it believes enough money will arrive at the receiver.

* If a connector **delivers** a payment locally, it will send the exact amount specified in the ILP header.

When a connector is delivering a payment to you to a local ledger, it should not assume that the payment will be forwarded further than that. If we open up that possibility, we'll basically end up reinventing every feature that we need in normal ILP forwarding again for that case. If the "receiver" wants to forward the payment fine, but the connector before shouldn't have to know or care.

In the future we may add more features to give more power and flexibility to the receiver. The "additional routing info" header would be pretty much what you'd need. I think the minimum would be some table where it tells you the amount at different prefixes, e.g.

| Prefix | Amount |
|---|---|
| `ilpdemo.red.` | 221 |
| `ilpdemo.red.bob.` |  201 |
| `ilpdemo.red.bob.sub.` | 200 |
| `ilpdemo.red.bob.sub.doubledollars.` | 100 |

emschwartz:
Ok, I like the idea of giving those two behaviors different names.

The one case that still irks me a little is it seems that a connector could mistakenly try to deliver funds when they should be forwarding them if they do not have a certain connector in their routing table.

The case would be something like you have a small bank that opens up and connects to and addresses themselves in terms of a bigger bank. It peers with some of the connectors but not all of them on the big bank ledger. Now you have a payment that comes in through a connector that doesn't know about the small bank. The small bank has correctly named their account on the bigger bank ledger so this ignorant connector will send the packet to the right place. _However_, because the connector doesn't realize it's sending to another connector it will take too much of the money and try to deliver less than the required amount to the connector. Because of this, the payment will be rejected. Would the ignorant connector retry the transfer applying its normal rate (figuring, maybe there is a connector there so it wants the extra money)?

adrianhopebailie:
> but I'm not sure I totally understand how connectors know if they are delivering locally

I am in the same boat. How does a connector know if the address it is delivering to is local to the ledger it is making the transfer on or not?

Assume the destination address is `ilpdemo.red.bob.sub` and the ledger prefix for the ledger it is doing the transfer on is `ilpdemo.red.`. 

How does the connector know if the transfer it makes is a forward or local delivery?

adrianhopebailie:
> However, because the connector doesn't realize it's sending to another connector it will take too much of the money and try to deliver less than the required amount to the connector. Because of this, the payment will be rejected. Would the ignorant connector retry the transfer applying its normal rate (figuring, maybe there is a connector there so it wants the extra money)?

I think the connector would get back a "transfer rejected" message from the receiving connector and then could update it's routing tables to reflect that there is another connector on the other side of that transfer and then get a quote from that connector before retrying.

justmoon:
> The case would be something like you have a small bank that opens up and connects to and addresses themselves in terms of a bigger bank. It peers with some of the connectors but not all of them on the big bank ledger.

This scenario does a great job of highlighting the limitations of subledgering. I would recommend one of two things to the bank in question:

1. Either choose a prefix which has a long common prefix with the larger bank (so that you're close-by in people's routing tables) or
2. Include a dedicated header that clarifies the ambiguity.

Per @emschwartz's request, I've created a separate ticket (#77), so let's move the discussion there, since it really is only tangentially related to this PR.

---

Back on topic: LGTM

PR #79, opened 2016-08-18
Create proposal-address-mapping.md

adrianhopebailie:
Proposal for mapping existing accounts to ILP addresses

### DIFF
--- a/proposal-address-mapping.md
+++ b/proposal-address-mapping.md
@@ -1,0 +1,20 @@
+# Interledger Address Mapping
+
+The Interldger Protocol defines a universal addressing scheme for any digital asset account. In order to transfer assets into existing accounts it is necessary to address these accounts.
+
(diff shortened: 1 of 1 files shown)


adrianhopebailie:
@justmoon , @emschwartz any thoughts on this?

emschwartz:
If you want this in the repo I think it should go in as a numbered RFC

adrianhopebailie:
@emschwartz - it's just a proposal for now, prob could have submitted as an issue first.

It was a lot more verbose in v1 (more explanations etc) but I lost v1 before saving (the perils of editing in the browser) so this was a quick lean v2 to get the idea out there

emschwartz:
Yeah if it's more of a proposal I'd say put it as an issue instead of in the repo. Seems like a fine start to me

PR #80, opened 2016-08-18
Add getFulfillment to Ledger Plugin Interface

emschwartz:
this is important for senders and connectors in case they miss the notification of a fulfillment and need to retrieve it later


justmoon:
LGTM

PR #82, opened 2016-08-19
Fix old mentions of ITP in SPSP

emschwartz:
based on comments from @vhpoet on https://github.com/interledger/rfcs/pull/75#discussion-diff-75408084


vhpoet:
LGTM

PR #84, opened 2016-08-19
RFC-4: LedgerPlugin#getFulfillment can return null

sentientwaffle:
It returns null if there is no fulfillment (this includes the case where the transfer doesn't exist).

cc: @emschwartz

### DIFF
--- a/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
+++ b/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
@@ -22,7 +22,7 @@ This spec depends on the [ILP spec](../0003-interledger-protocol/).
 | | [**getPrefix**](#getprefix) ( ) `⇒ Promise<String>` |
 | | [**getAccount**](#getaccount) ( ) `⇒ Promise<String>` |
 | | [**getBalance**](#getbalance) ( ) <code>⇒ Promise.&lt;String></code> |
-| | [**getFulfillment**](#getfulfillment) ( transferId ) <code>⇒ Promise.&lt;String></code> |
+| | [**getFulfillment**](#getfulfillment) ( transferId ) <code>⇒ Promise.&lt;String|null></code> |
 | | [**send**](#send) ( transfer ) <code>⇒ Promise.&lt;null></code> |
 | | [**fulfillCondition**](#fulfillcondition) ( transferId, fulfillment ) <code>⇒ Promise.&lt;null></code> |
 | | [**replyToTransfer**](#replytotransfer) ( transferId, replyMessage ) <code>⇒ Promise.&lt;null></code> |
@@ -144,9 +144,11 @@ The mapping from the ILP address to the local ledger address is dependent on the
 Return a decimal string representing the current balance. Plugin must be connected, otherwise the promise should reject.
 
 #### getFulfillment
-<code>ledgerPlugin.getFulfillment( transferId ) ⇒ Promise.&lt;String></code>
+<code>ledgerPlugin.getFulfillment( transferId ) ⇒ Promise.&lt;String|null></code>
 
 Return the fulfillment of a transfer if it has already been executed.
+
+Return `null` if there is no fulfillment.
 
 #### Event: `connect`
 <code>ledgerPlugin.on('connect', () ⇒ )</code>


bensharaf:
should it throw an error/reject instead?

sentientwaffle:
I don't think so, since not having a fulfillment isn't an error per se.

emschwartz:
@justmoon what do you think of this question about returning null vs throwing an error when a transfer hasn't been fulfilled yet? and what about if you try to get a fulfillment for a transfer that doesn't exist at all?

adrianhopebailie:
> I don't think so, since not having a fulfillment isn't an error per se

+1, but you need to differentiate between this and the case of a bad transfer ID or missing transfer. The latter should throw an exception. 

i.e. You are requesting a resource of type A using the key of a linked resource of type B (which you assume exists).

Saying A doesn't exist is not "exceptional", it implies, "try again later".
Saying B doesn't exist is an error condition in this case.

It's worth thinking about the DoS attack vectors we may introduce though if this is a public interface and we are revealing more info than necessary. Prob not applicable here.

emschwartz:
@adrianhopebailie that was my initial thought too. @sentientwaffle pointed out that the `five-bells-ledger` will give you a 404 error in either case, which makes in the REST context.

To me the argument for throwing an error if the transfer doesn't exist would be stronger if we also had a `getTransfer` method in this interface. We don't right now because it isn't needed for making ILP payments (`getFulfillment` is needed for senders and connectors who may miss a notification about a fulfillment). I think there may be many ledger implementations where it's more complicated to differentiate between these two cases, and I'm not sure you would handle them any differently anyway.

adrianhopebailie:
> To me the argument for throwing an error if the transfer doesn't exist would be stronger if we also had a `getTransfer` method in this interface.

Why do you say that? As a caller I want to know why my call failed so I know if I should retry or if it is likely to keep failing.

emschwartz:
We don't have a way to get a transfer from the history right now, so (insofar as this is an implicit definition of the functionality we expect from a ledger -- I know you hate that @adrianhopebailie) this means that ledger (plugin) implementations don't necessarily need a full history of all transfers that have happened. They do, however, need to store conditions and fulfillments that they've seen, both so that they enforce the rule that you cannot submit a new transfer with a condition that has already been fulfilled and so that we can implement this fulfillment lookup feature. It seems that many implementations may store the fulfillments separately from the transfers (like the `five-bells-ledger` does), so it would be simplest to have the `getFulfillment` method just tell you whether it has or hasn't seen that fulfillment.

This does make me wonder though whether you should `getFulfillment`s by their `transferId` OR by the `condition`...

adrianhopebailie:
> This does make me wonder though whether you should getFulfillments by their transferId OR by the condition

:boom: I think that's the answer! They should definitely be indexed by condition. Binary indexes are nice and efficient and the best part is they are a fixed size.

emschwartz:
This would also suggest that `fulfillCondition` should not expect the `transferId`, which may be fine but we would also need to change `five-bells-ledger` accordingly

adrianhopebailie:
I feel like I recall this conversation from a few months ago...

How does a ledger behave if it has multiple transfers on hold with the same condition? Is there a reason it shouldn't release them all? If not then it makes sense for ledgers to just always index both conditions and fulfillments by the condition fingerprint.

justmoon:
> How does a ledger behave if it has multiple transfers on hold with the same condition? Is there a reason it shouldn't release them all?

It conflicts with the idea that only the receiver can post the fulfillment (which is @bensharaf's elegant way of solving the "no unwanted incoming payments" problem without extra round trips.)

This idea only works with universal mode (not optimistic, not atomic), but it's my favorite solution right now, because we do think of ILP as primarily universal mode I think.

So I'm against "fulfillTransfer without transfer id".

As for getFulfillment - it seems more consistent with the rest of the API to use the transfer ID.

Some ledgers may not support `getFulfillment`, which means if you miss the notification, you're SOL.

> They do, however, need to store conditions and fulfillments that they've seen, both so that they enforce the rule that you cannot submit a new transfer with a condition that has already been fulfilled

This is an advanced feature, so I would let that feature be influenced by our API choices, but not vice versa.

> and so that we can implement this fulfillment lookup feature. 

That's begging the question. You can implement the lookup feature by transfer ID as well, no need to store an index by condition.

As for the main question in this thread - I prefer a specific error. We can return something in the error JSON from Five Bells Ledger that disambiguates the two cases.

adrianhopebailie:
> It conflicts with the idea that only the receiver can post the fulfillment (which is @bensharaf's elegant way of solving the "no unwanted incoming payments" problem without extra round trips.)

This is an implementation feature. If you guys think it's one that we should have on our implementation then I have no objection. It doesn't really impose anything at the protocol level.

Instead of binding the condition to the transfer why not bind it to the receiver explicitly? i.e. Only a caller that has permissions on the receiving account can submit a fulfillment that impacts that account. Would probably have more complexity in implementation...

> Some ledgers may not support getFulfillment, which means if you miss the notification, you're SOL.

Agreed, again this is implementation specific stuff that doesn't really make any impact on the design of the core protocol as far as I can tell.

> As for the main question in this thread - I prefer a specific error. We can return something in the error JSON from Five Bells Ledger that disambiguates the two cases.

👍  - as with the events we should provide granularity if we can and let consumers choose to handle both cases the same if they want to.

justmoon:
Sounds like we have a decision.

* Add more detail to the 404 error in five-bells-ledger (by having two new error classes for either case which both return 404, but with different JSON)
* `getFulfillment` is defined to throw with different exceptions `TransferNotFoundError` and `FulfillmentNotFoundError` (names up for debate) upon failure

cc/ @sentientwaffle

PR #90, opened 2016-08-24
initial draft of ledger API

diminator:
I started and initial draft for guidelineing the API as discussed in last call.

@adrianhopebailie: Is this the kind of format you were looking for?

cc @justmoon @emschwartz


0004-ledger-plugin-interface/0004-ledger-interface.md
```diff
@@ -0,0 +1,203 @@
...
+```
+
+
+### Transfers
+
+#### Retrieve transfer item
+
+##### Request
+```http
+GET /transfers/<ID> HTTP/1.1
```
justmoon: The format for `<ID>` should be specified. Recommend a lowercase UUID.

0004-ledger-plugin-interface/0004-ledger-interface.md
```diff
@@ -0,0 +1,203 @@
...
+    'name': <string: optional>,
+    'connector': <uri: optional>
+}
+```
+
+#### Update account
+
+##### Request
+```http
+PUT /accounts/<ID> HTTP/1.1
```
justmoon: The format for `<ID>` should be specified. Recommend an arbitrary string with characters `A-Z`, `a-z`, `0-9`, `-`, case-insensitive.

0004-ledger-plugin-interface/0004-ledger-interface.md
```diff
@@ -0,0 +1,203 @@
...
+
+### Accounts
+
+#### Retrieve account
+
+##### Request
+```http
+GET /accounts/<ID> HTTP/1.1
+Host: ledger.my
+Authorization: <token>/<cert>/<password>/...
```
justmoon: Should add a section to flesh out how authorization works. I'd be fine with having a placeholder for now.

justmoon:
@emschwartz @adrianhopebailie 

Each RFC number should be only one document, and we should never redefine/reassign. (Better to have a hole/deprecated RFC than a number that meant different things at different times.)

IL-RFC 0004 is the JavaScript ledger plugin interface. So we should assign a new RFC number for this imo.

0012-ledger-web-api/0012-ledger-web-api.md
```diff
@@ -0,0 +1,193 @@
...
+- Easy adoption of existing interledger plugins
+
+<p align="center">
+  <img width="70%" height="70%" src ="./static/ilp_interface.png" />
+</p>
+
+## Requirements
+
+In order for ledgers to connect seamlessly with any of the ILP plugins `ilp-plugin-<ledgerID>`, the ledger needs to provide:
+- a RESTful HTTP API with endpoints to `transfer` (or `transaction`, `asset`) objects
```
adrianhopebailie: a RESTful HTTP API with endpoints to initiate `transfers` and submit `fulfillments`.

0012-ledger-web-api/0012-ledger-web-api.md
```diff
@@ -0,0 +1,193 @@
...
+- A standardized interface for inter-ledger communication
+- Easy adoption of existing interledger plugins
+
+<p align="center">
+  <img width="70%" height="70%" src ="./static/ilp_interface.png" />
+</p>
+
+## Requirements
+
+In order for ledgers to connect seamlessly with any of the ILP plugins `ilp-plugin-<ledgerID>`, the ledger needs to provide:
```
adrianhopebailie: "In order for ledgers to integrate seamlessly with connectors the ledger needs to provide:"

0012-ledger-web-api/0012-ledger-web-api.md
```diff
@@ -0,0 +1,193 @@
+# Ledger Web API
+
+This spec defines some guidelines for **RESTful ledger APIs** to easily integrate with the Interledger Protocol suite.
+
+The advantages of following this spec when designing a ledger are:
+- A standardized interface for inter-ledger communication
+- Easy adoption of existing interledger plugins
```
adrianhopebailie: "Easy adoption by connectors that implement the standard Web client (Example: ilp-plugin-web)"

We should definitely look at creating this `lp-plugin-web` project soon. It would be a connector plugin that expects to connect to a ledger running this API.

0012-ledger-web-api/0012-ledger-web-api.md
```diff
@@ -0,0 +1,193 @@
...
+
+<p align="center">
+  <img width="70%" height="70%" src ="./static/ilp_interface.png" />
+</p>
+
+## Requirements
+
+In order for ledgers to connect seamlessly with any of the ILP plugins `ilp-plugin-<ledgerID>`, the ledger needs to provide:
+- a RESTful HTTP API with endpoints to `transfer` (or `transaction`, `asset`) objects
+- PUSH notifications to register changes in the ledger objects, albeit using websockets, long-polling (EPOLL), or server-sent events.
```
adrianhopebailie: PUSH notifications to notify connectors of changes in the ledger objects and incoming transfers or fulfillments. This could be through websockets, long-polling (EPOLL), or server-sent events.

0012-ledger-web-api/0012-ledger-web-api.md
```diff
@@ -0,0 +1,193 @@
...
+
+##### Request
+```http
+GET /accounts/<uuid: ID> HTTP/1.1
+Host: ledger.example
+Authorization: <token>/<cert>/<password>/...
+Content-Type: application/json
+```
+
+##### Response
```
adrianhopebailie: Would be useful to explain what the fields are in the responses. Maybe just a placeholder for now?

0012-ledger-web-api/0012-ledger-web-api.md
```diff
@@ -0,0 +1,193 @@
...
+}
+```
+
+### Transfers
+
+#### Retrieve transfer item
+
+##### Request
+```http
+GET /transfers/<uuid|base58: ID> HTTP/1.1
```
adrianhopebailie: What's the motivation for the base58 (think we should standardize on UUID for this API. If the underlying ledger uses something else then it would probably need to maintain a mapping of one to the other.

0012-ledger-web-api/0012-ledger-web-api.md
```diff
@@ -0,0 +1,193 @@
...
+Host: ledger.example
+Authorization: <token>/<cert>/<password>/...
+Content-Type: application/json
+```
+
+##### Response
+```http
+HTTP/1.1 200 OK
+ledgerTransfer
```
adrianhopebailie: Does this include the fulfillment if it exists? If so I'd say we don't need:
`GET /transfers/<uuid: ID>/fulfillment`

diminator:
> IL-RFC 0004 is the JavaScript ledger plugin interface. So we should assign a new RFC number for this imo.

assigned it to RFC-0012

0012-ledger-web-api/0012-ledger-web-api.md
diminator: You mean the object structure and field types?
Or something more general?

0012-ledger-web-api/0012-ledger-web-api.md
diminator: Yeah, this is a bit of a PITA. Many blockchains use `base58` encodings of a hash (ie bitcoin, ethereum, multichain, bigchaindb, others). I don't think it will be possible to enforce this at the server side for most...

one idea was to have the clients/connectors or the plugins keep track of the mappings.

0012-ledger-web-api/0012-ledger-web-api.md
diminator: Could be, I was following the pattern in five-bells.Here you can GET/PUT fulfillments

In many cases though it would be part of the transfer object.
Depends on what our target ledger type is: sql-based or append only?

0012-ledger-web-api/0012-ledger-web-api.md
```diff
@@ -0,0 +1,193 @@
...
+    ilpFulfillment
+```
+
+##### Response
+```http
+HTTP/1.1 200 OK
+ilpFulfillment
+```
+
+## PUSH notifications (websockets)
```
adrianhopebailie: We will need to define what these messages would look like? Just JSON objects?

0012-ledger-web-api/0012-ledger-web-api.md
adrianhopebailie: > You mean the object structure and field types?

Yeah

0012-ledger-web-api/0012-ledger-web-api.md
adrianhopebailie: If a ledger is exposing this interface then it was either built with this interface in mind or this interface is being hosted by an adaptor on top of the underlying ledger so in the case of ledgers like Bitcoin I'd say only the latter is possible. In this case I'd expect the adaptor to maintain the mappings.

The adaptor could probably do some clever stuff like write the UUID into the transfer memo so that it can find it from the underlying ledger later and not need to persist the mappings for long.

0012-ledger-web-api/0012-ledger-web-api.md
adrianhopebailie: > Depends on what our target ledger type is: sql-based or append only?

I think this consideration should be independent of the API design. From the perspective of the consumer of the API they have a transfer resource they are getting so that they can check if it has an attached fulfillment yet.

The right way to submit a fulfillment is probably a PUT on the transfer resource endpoint?

emschwartz:
Shall we close this in favor of https://github.com/interledger/rfcs/pull/125?

PR #91, opened 2016-08-26
RFC-9: Request a quote before the payment setup

vhpoet:
This PR would make the `five-bells-wallet` payment flow more convenient. It also makes the fixed source and destination flows the same.

If this flow gets accepted, we should merge the flows in the docs.

See also https://github.com/interledger/five-bells-wallet/pull/25

### DIFF
--- a/0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
+++ b/0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
@@ -55,10 +55,10 @@ We assume that the sender knows the receiver endpoint (see [Appendix A: (Optiona
 1. The sender's SPSP client queries the receiver endpoint.
 2. The receiver endpoint responds with the receiver info, including the receiver's currency code.
 3. The sender chooses an amount of the receiver's asset to deliver.
-4. The sender's SPSP client submits the payment information, including the destination amount, to the receiver endpoint.
-5. The receiver endpoint responds with an [Interledger Payment Request](../0011-interledger-payment-request), which includes the execution condition.
-6. The sender's SPSP client uses its ILP module to get a quote in their currency or asset type for the ILP transfer.
-7. The sender accepts the quote.
+4. The sender's SPSP client uses its ILP module to get a quote in their currency or asset type for the ILP transfer.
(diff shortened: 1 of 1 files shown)


0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -55,10 +55,10 @@ We assume that the sender knows the receiver endpoint (see [Appendix A: (Optiona
...
 2. The receiver endpoint responds with the receiver info, including the receiver's currency code.
 3. The sender chooses an amount of the receiver's asset to deliver.
-4. The sender's SPSP client submits the payment information, including the destination amount, to the receiver endpoint.
-5. The receiver endpoint responds with an [Interledger Payment Request](../0011-interledger-payment-request), which includes the execution condition.
-6. The sender's SPSP client uses its ILP module to get a quote in their currency or asset type for the ILP transfer.
-7. The sender accepts the quote.
+4. The sender's SPSP client uses its ILP module to get a quote in their currency or asset type for the ILP transfer.
+5. The sender accepts the quote.
+6. The sender's SPSP client submits the payment information, including the destination amount, to the receiver endpoint.
+7. The receiver endpoint responds with an [Interledger Payment Request](../0011-interledger-payment-request), which includes the execution condition.
```
emschwartz: It's actually important at this step that the sender quote the payment request to make sure it matches what they thought it would be. It's kind of annoying to have two quotes in the flow but I'm not sure we can avoid it.

vhpoet:
doc talks about two endpoints: query and setup. Shouldn't we also add query?

emschwartz:
What do you mean? The doc specifies [Query](https://github.com/interledger/rfcs/blob/master/0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md#query-get-receiver) and [Setup](https://github.com/interledger/rfcs/blob/master/0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md#setup-post-receiver)

vhpoet:
@emschwartz sorry I meant a quote

vhpoet:
nevermind

Issue #93, opened 2016-08-30
RFC-4: PROPOSAL: Add Error Types


bensharaf:
# Reasoning

Currently, most functions in the LPI return promises which will be rejected if something goes wrong. The interface assumes that all errors emitted have plugin-specific types and messages. This poses a problem when it's important to distinguish different failure types (eg. https://github.com/interledger/rfcs/pull/84)

Adding some standard error types would make the error values of functions unambiguous. For instance, a malformed transfer will appear different from a repeat transfer. In the case of plugin-specific errors, the generic `Error` could still be used.

[Ilp-plugin-tests](https://github.com/interledger/js-ilp-plugin-tests) currently only sees whether functions failed or succeeded. A function that is supposed to fail could be failing for the wrong reason, which would not be picked up by the tests. These mistakes lead to bugs, but are not possible to prevent while keeping the tests cross-plugin. Error types would eliminate this problem.

This change would also let tools such as [ilp-connector](https://github.com/interledger/js-ilp-connector) give more descriptive errors.

# Possible Error Types

| Function | Error Types Returned |
|----------|------------|
| `connect()` | `InvalidFieldsError`, `UnreachableError` |
| `getFulfillment()` | `TransferNotFoundError`, `MissingFulfillmentError` |
| `send()` | `InvalidFieldsError`, `NotAcceptedError`, `RepeatError` |
| `fulfillCondition()` | `InvalidFieldsError`, `TransferNotFoundError`, `NotAcceptedError`, `RepeatError` |
| `replyToTransfer` | `TransferNotFoundError` |
| `rejectIncomingTransfer` | `TransferNotFoundError`, `NotAcceptedError`, `RepeatError` |

- `InvalidFieldsError` - The arguments or configuration were invalidated client-side.
- `UnreachableError` - Could not connect due to network issues.
- `TransferNotFoundError` - A Transfer with the given ID doesn't exist.
- `MissingFulfillmentError` - Transfer has not yet been fulfilled.
- `RepeatError` - That transfer has already been sent/fulfilled/rejected.
- `NotAcceptedError` - The ledger rejected the operation (eg. If balance is too low, or fulfillment doesn't match condition).

**cc** @justmoon @emschwartz @adrianhopebailie

emschwartz:
👍 . Another class of errors has to do with transfers being expired (for `getFulfillment`, `send`, `fulfillCondition`).

[...]

PR #94, opened 2016-09-01
Feat: Add error types

bensharaf:
Resolves https://github.com/interledger/rfcs/issues/93 . Documents the different error types and their purposes, as well as which functions return them.


emschwartz:
LGTM

Issue #95, opened 2016-09-07
RFC-4: Define autoconfig mechanism


justmoon:
Ledgers should be able to provide client configuration automatically. (This would be similar to DHCP in the IP world.) Currently the reference implementation provides two features to that effect:
- `getPrefix` allows clients to automatically detect the ledger's prefix
- `getConnectors` allows clients to automatically get a list of connectors to peer with

We should consider consolidating these sorts of features as a single "autoconfig" object. It could even be part of `getInfo`. 

See also interledger/five-bells-ledger#321

emschwartz:
Connectors are now included in [`LedgerInfo`](https://github.com/interledger/rfcs/blob/master/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md#class-ledgerinfo) and plugins can get the prefix from the ledger if they want

PR #97, opened 2016-09-08
RFC-4: LedgerPlugin#getInfo includes "connectors"

sentientwaffle:
Include a list of recommended connectors on `LedgerInfo`.

See: https://github.com/interledger/rfcs/issues/95 and https://github.com/interledger/js-ilp-plugin-bells/pull/40


emschwartz:
LGTM

PR #98, opened 2016-09-09
docs(004): add AlreadyRolledBackError, AlreadyFulfilledError, DuplicateIdError

bensharaf:
`fulfillCondition` should no longer throw when fulfilling a fulfilled transfer, in order to remain idempotent. `AlreadyRolledBackError` should be emitted when someone tries to fulfill a cancelled transfer and also when `getFulfillment` is used to fetch a cancelled transfer. `AlreadyFulfilledError` is emitted when trying to reject a fulfilled transfer. `DuplicateIdError` is emitted when sending a transfer with a duplicate ID that does not otherwise match the existing transfer.


emschwartz:
LGTM

PR #101, opened 2016-09-15
[Draft] ILQP HTTP+JSON Format

mDuo13:


0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md
```diff
@@ -1,7 +1,104 @@
...
+
+| Field                         | Type        | Description                    |
+|:------------------------------|:------------|:-------------------------------|
+| `source_address`              | ILP Address | Address of the source's account in the source ledger. |
+| `destination_address`         | ILP Address | Address of the destination's account in the destination ledger. |
+| `source_amount`               | Number      | Fixed amount to debit from the source's account. (Required unless `destination_amount` specified.) |
+| `destination_amount`          | Number      | Fixed amount to credit to the destination's account. (Required unless `source_amount` specified.) |
+| `destination_expiry_duration` | Number      | (Optional) Number of milliseconds before the transfer in the destination ledger expires. |
+| `source_expiry_duration`      | Number      | (Optional) Number of milliseconds before the transfer in the source ledger expires. |
+| `slippage`                    | Number      | (Optional) Custom slippage ***amount? percentage?*** to assume for the exchange between ledgers. Lower slippage means the payment is cheaper but more likely to fail due to market movement. |
```
emschwartz: It's a percentage:

> CONNECTOR_SLIPPAGE (default: 0.001 = 0.1%) The ratio for overestimating exchange rates to prevent payment failure if the rate changes.

From https://github.com/interledger/js-ilp-connector/blob/master/README.md#L106

0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md
```diff
@@ -1,7 +1,104 @@
...
+
+### Response Format
+
+The response uses the HTTP status code 200 OK and contains a JSON object with the following fields:
+
+| Field                         | Type        | Description                    |
+|:------------------------------|:------------|:-------------------------------|
+| `source_connector_account`    | ILP Address | The address of the connector's account in the source ledger where it should receive the first transfer. |
+| `source_ledger`               | ILP Address | The address of the ILP-enabled ledger where the source account is. |
+| `source_amount`               | String      | Decimal number amount of currency the ***TODO: connector's account should receive(?)*** in the source ledger. |
```
emschwartz: 👍 We assume ledger fees are paid out of band, or they would be included in that amount.

emschwartz:
Added 2 minor comments but otherwise LGTM

0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md
mDuo13: I'm concerned that the assumption that ledger fees are paid out of bound doesn't generalize. At the very least, it's definitely not true of the Ripple Consensus Ledger. (Even for XRP transfers, the sender ends up getting debited the transaction cost on top of the amount sent.) So I'd really much rather pick one (the amount sent or the amount received in the source ledger) and stick with it rather than leave it vague.

justmoon: > I'm concerned that the assumption that ledger fees are paid out of bound doesn't generalize.

Paying ledger fees out of band is a choice, not an assumption. We're saying that it is the responsibility of the sender of a local transfer to pay the fee. Each connector will recoup that cost from the person asking them to do something.

If the fees are high enough that the connector can't risk getting stuck with the cost of failed payments, it can simply collect some fees upfront when the user registers. Note that the connector and its customer/peer share a ledger by definition, so they can pay each other with a simple local transfer.

0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md
```diff
@@ -1,7 +1,104 @@
...
+
+* The **source ledger** is where the sender and the connector both have accounts.
+* The **destination ledger** is where the receiver and the connector both have accounts.
+
+The sending and receiving ledger are different ledgers. Otherwise, there should be no reason to use ILQP.
+
+**Parties:**
+
+* The **sender** is the party being debited at the start of the overall chain.
+* The **source** is the party being debited in a single link of the chain. In a simple payment, the source is also the sender.
```
justmoon: What's a "simple payment"?

PR #107, opened 2016-10-13
docs(004,008): add messaging; s/send/sendTransfer

### LINKED ISSUE #96: RFC-4: Split messaging from transfers
Currently the `send` method handles both messaging (sending a zero-value transfer) and transfers.

However, in practice the delivery guarantees we want for messaging are probably weaker than for a zero transfer, so it makes sense to split the two calls.

Proposal:
``` typescript
sendTransfer = (transfer:OutgoingTransfer) => {}
sendMessage = (message:OutgoingMessage) => {}
```

Where `OutgoingMessage` contains only the `account`, `ledger`, `data` and `custom` fields of the `OutgoingTransfer` object.

sentientwaffle:
* Add `plugin.sendMessage(message)` and `plugin.on("incoming_message")`.
* Update ILQP to use messaging instead of HTTP.
* Rename `plugin.send` to `plugin.sendTransfer`.

Fixes #96

### DIFF
--- a/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
+++ b/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
@@ -243,13 +243,15 @@ For a detailed description of these properties, please see [`OutgoingTransfer`](
 Plugin must be connected, otherwise the promise should reject. Send a ledger-local message.
 If there is a problem with the structure or validity of the message, then `sendMessage` should throw an error in the form of a rejected promise.
 
+Messaging is used by connectors for [quoting](../0008-interledger-quoting-protocol/) and broadcasting routes.
+
 ###### Parameters
 | Name | Type | Description |
 |:--|:--|:--|
 | message | <code>[OutgoingMessage](#outgoingmessage)</code> | Properties of the message to be created |
 
 ###### Returns
-**`Promise.<null>`** A promise which resolves when the message has been submitted (but not necessarily delivered.)
+**`Promise.<null>`** A promise which resolves when the message has been submitted (but not necessarily delivered). To ensure delivery, you must build a mechanism for that on top.
 
 Throws `InvalidFieldsError` if required fields are missing from the message or malformed.
 Throws `NotAcceptedError` if the message is rejected by the ledger.
--- a/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
+++ b/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
@@ -23,9 +23,9 @@ This spec depends on the [ILP spec](../0003-interledger-protocol/).
 | | [**getAccount**](#getaccount) ( ) `⇒ Promise<String>` |
 | | [**getBalance**](#getbalance) ( ) <code>⇒ Promise.&lt;String></code> |
 | | [**getFulfillment**](#getfulfillment) ( transferId ) <code>⇒ Promise.&lt;String></code> |
-| | [**send**](#send) ( transfer ) <code>⇒ Promise.&lt;null></code> |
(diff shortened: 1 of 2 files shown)


0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
```diff
@@ -236,6 +237,35 @@ p.send({
 
 For a detailed description of these properties, please see [`OutgoingTransfer`](#outgoingtransfer).
 
+#### sendMessage
+<code>ledgerPlugin.sendMessage( **message**:[OutgoingMessage](#outgoingmessage) ) ⇒ Promise.&lt;null></code>
```
emschwartz: It would be good to include an explanation of what this feature is for. Also what should ledger plugin implementor do if the underlying ledger doesn't support messaging?

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
```diff
@@ -236,6 +237,35 @@ p.send({
...
+Plugin must be connected, otherwise the promise should reject. Send a ledger-local message.
+If there is a problem with the structure or validity of the message, then `sendMessage` should throw an error in the form of a rejected promise.
+
+###### Parameters
+| Name | Type | Description |
+|:--|:--|:--|
+| message | <code>[OutgoingMessage](#outgoingmessage)</code> | Properties of the message to be created |
+
+###### Returns
+**`Promise.<null>`** A promise which resolves when the message has been submitted (but not necessarily delivered.)
```
emschwartz: We need to figure out at what level retries should happen. Should you assume that the message will get through if the promise resolves? Or is there a case when you'd need to retry it?

0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md
```diff
@@ -55,22 +57,34 @@ The request uses the following query parameters:
...
+  "method": "quote_request",
+  "id": "721e4126-98a1-4974-b35a-8a8f4655f934",
+  "data": {
+    "source_amount": "100.25",
+    "source_address": "example.eur-ledger.alice",
+    "destination_address": "example.usd-ledger.bob",
+    "source_expiry_duration": "6000",
+    "destination_expiry_duration": "5"
+  }
+}
```
emschwartz: Is this the `data` in the `OutgoingMessage`?

0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md
```diff
@@ -79,19 +93,41 @@ The response uses the HTTP status code 200 OK and contains a JSON object with th
...
+```js
 {
-  "source_connector_account": "mark",
-  "source_ledger": "example.eur-ledger.",
-  "source_amount": "100.25",
-  "source_expiry_duration": "6000",
-  "destination_ledger": "example.usd-ledger.",
-  "destination_amount": "105.71",
-  "destination_expiry_duration": "5000"
+  "method": "quote_response",
```
emschwartz: Should be `error`

0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md
sentientwaffle: Yes

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
sentientwaffle: I'll add an explanation of the feature's purpose.
If the underlying ledger doesn't directly support messaging it can use zero-amount transfers, right?

PR #121, opened 2016-10-31
docs(0004): Add from/to to messaging

sentientwaffle:
Related to https://github.com/interledgerjs/five-bells-ledger/issues/346


sentientwaffle:
This isn't needed because of how ledger plugin factories work.

PR #124, opened 2016-11-11
Updates to crypto-conditions

adrianhopebailie:
Moved string encoding to appendix
Dropped featurebitmask in favour of subtypes
Dropped weights from threshold type
Changed some field sizes
Re-write abstract, intro and others


0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -114,312 +114,288 @@ This specification is a part of the [Interledger Protocol](https://interledger.o
...
 
-The SHA-256 feature suite provides the SHA-256 hash function. SHA-256 is a cryptographic hash function published by the US National Institute of Standards and Technology that produces 256 bits of output. This feature suite is assigned the feature bit 2^0 = 0x01.
+The field is encoded as a variable length octet string and specifies the set of types an implementation must support in order to be able to successfully validate the fulfillment of this condition. This is the set of types and subtypes of the top-level condition and all sub-crypto-conditions, recursively. 
 
-## PREIMAGE {#preimage-feature-suite}
-The PREIMAGE feature suite provides conditions that use a preimage as a one-time signature. This feature suite is assigned the feature bit 2^1 = 0x02.
+Each bit represents a type. The list of known types is the IANA-maintained [Crypto-Condition Type Registry](#crypto-conditions-type-registry) .
```
nbougalis: So, a number of issues.

First of all, I don't see where we define the types for the existing 5 cryptoconditions. We can't use their types, since those are an `ENUM` and can't be used as a bitfield, at least as defined since some types have more than 1 bit set.

If we keep `type` as an `ENUM` and make the enum members have values usable as bitfields we're limited to 8 condition types (or 16, 24, 32, etc). Obviously not the best idea.

If we keep type as is, but want `subtypes` to be a bitfield we need to "duplicate" each type in the `ENUM` with a "subtype" whose value is a power of 2; that just seems error-prone and unnecessarily complex.

Also, if we are going to treat these values as a bitfield, then we should probably encode them as such: using an OER "bit string" without a size constraint. Not as `INTEGER`.

I think that the existing `type` + `feature` setup works well. It also allows us to easily "reuse" a type in a safe way. For example, want a `PREFIX-SHA-512`? Don't define a new type; just use the existing `PREFIX` type with `SHA-512` as the feature suite.

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -114,312 +114,288 @@ This specification is a part of the [Interledger Protocol](https://interledger.o
...
+#### Examples
 
-A prefix condition prepends the message to be validated with a constant string before passing it on to the subcondition's validation function. <!-- TODO: better explanation of why -->
+The following bitmask indicates the presence of the subtypes PREIMAGE-SHA-256 (type number 0) and RSA-SHA-256 (type number 3)
 
-## THRESHOLD {#threshold-feature-suite}
-The THRESHOLD feature suite provides conditions that have several weighted subconditions and a threshold number. This feature suite is assigned the feature bit 2^3 = 0x08.
+    Hex
+    1    2    3    4    
+    0x09 0x00 0x00 0x00
```
nbougalis: Not sold; what's the `0x09` for? Also, does this show the OER length prefix? If so, where is it?

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -114,312 +114,288 @@ This specification is a part of the [Interledger Protocol](https://interledger.o
...
+    Binary
+    1         2         3         4          
+    0000 1001 0000 0000 0000 0000 0000 0000  
+    
+The following bitmask indicates the presence of the subtype ED25519 (type number 4)
 
-Threshold conditions provide flexible multi-signing, such as requiring "M-of-N" subconditions be fulfilled. Subconditions can also be weighted so that one subcondition can count multiple times towards meeting the threshold.
+    Hex
+    1    2    3    4    
+    0x10 0x00 0x00 0x00
```
nbougalis: Again, not sold... what's the `0x10`? Where's the length prefix?

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -114,312 +114,288 @@ This specification is a part of the [Interledger Protocol](https://interledger.o
...
+    1         2         3         4           
+    0001 0000 0000 0000 0000 0000 0000 0000  
+    
+The following bitmask indicates the presence of an unknown subtype with type number 10.
 
-## RSA-PSS {#rsa-pss-feature-suite}
-The RSA-PSS feature suite provides the RSS-PSA signature algorithm. RSA-PSS is a signature algorithm based on the RSA cryptosystem, which relates to the problem of factoring the product of two large prime numbers. This feature suite is assigned the feature bit 2^4 = 0x10.
+    Hex
+    1    2    3    4    
+    0x00 0x04 0x00 0x00
```
nbougalis: More confusion...

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -114,312 +114,288 @@ This specification is a part of the [Interledger Protocol](https://interledger.o
...
 The following condition types are defined in this version of the specification. Future versions of this spec MAY introduce new feature suites and condition types, which SHALL be registered in the IANA maintained [Crypto-Condition Type Registry](#crypto-conditions-type-registry).
 
 ## PREIMAGE-SHA-256 {#preimage-sha-256-condition-type}
-PREIMAGE-SHA-256 is assigned the type ID 0. It relies on the SHA-256 and PREIMAGE feature suites which corresponds to a feature bitmask of 0x03.
+
+PREIMAGE-SHA-256 is assigned the type ID 0. It relies on the availability of the SHA-256 digest algorithm.
 
 This type of condition is also called a "hashlock". By creating a hash of a difficult-to-guess 256-bit random or pseudo-random integer it is possible to create a condition which the creator can trivially fulfill by publishing the random value. However, for anyone else, the condition is cryptographically hard to fulfill, because they would have to find a preimage for the given condition hash.
 
 Implementations MUST ignore any input message when validating a PREIMAGE-SHA-256 fulfillment.
```
nbougalis: A bit about the rationale is important here, I think.

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -114,312 +114,288 @@ This specification is a part of the [Interledger Protocol](https://interledger.o
...
 
-condition
-: is the subcondition which the subfulfillment must match.
+subcondition
+: is the sub-condition derived from the sub-fulfillment.
+
+
+### MaxFulfillmentLength {#prefix-sha-256-condition-type-maxfulfillmentlength}
 
+The maxFulfillmentLength is the size, in bytes, of the prefix plus the maxFulfillmentLength length of the sub-condition.
```
nbougalis: Just pointing out that, strictly speaking, this field isn't necessary in this instance, since we have the information to derive this directly in the condition fingerprint: the prefix, and the subcondition. Of course, we can't just skip it, so...

Food for thought: do we gain anything by storing the _SHA256_ digest, à la `preimage-sha-256` instead?

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -428,89 +404,57 @@ subfulfillment
...
+      subconditions SEQUENCE OF Condition,
     }
 
-    ThresholdSubcondition ::= SEQUENCE {
-      weight INTEGER (0..4294967295),
-      condition Condition
-    }
-
-The list of conditions is sorted first based on length, shortest first. Elements of the same length are sorted in lexicographic (big-endian) order, smallest first.  
+The list of sub-conditions is sorted first based on length, shortest first. Elements of the same length are sorted in lexicographic (big-endian) order, smallest first.
```
nbougalis: Can you elaborate? Does the sorting take place on the binary encoded version of the conditions?

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -428,89 +404,57 @@ subfulfillment
...
-The list of conditions is sorted first based on length, shortest first. Elements of the same length are sorted in lexicographic (big-endian) order, smallest first.  
+The list of sub-conditions is sorted first based on length, shortest first. Elements of the same length are sorted in lexicographic (big-endian) order, smallest first.  
 
 threshold
-: threshold MUST be an integer in the range 1 ... 2^32 - 1. In order to fulfill a threshold condition, the weights of the provided fulfillments MUST be greater than or equal to the threshold.
+: threshold MUST be an integer in the range 1 ... 255. In order to fulfill a threshold condition, the number of provided sub-fulfillments MUST be greater than or equal to the threshold.
 
 subconditions
-: is the set of subconditions, each provided as a tuple of weight and condition.
+: is the set of sub-conditions either provided in the fulfillment or derived from the sub-fulfillments provided.
```
nbougalis: This could be a little clearer: the condition _MUST_ include all subconditions, not a subset. Not sure how to word it in a more approachable way.

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -428,89 +404,57 @@ subfulfillment
...
-: is the numeric weight of this subcondition, i.e. how many times it counts against the threshold. It MUST be an integer in the range 1 ... 2^32 - 1.
+: is the set of sub-fulfillments.
 
-condition
-: is the subcondition if this subcondition is not fulfilled.
-
-fulfillment
-: is the subfulfillment if this subcondition is fulfilled.
+subconditions
+: is the set of sub-conditions provided in place of any unfufilled sub-fulfillment.
```
nbougalis: Either define the sorting order or explicitly allow any order.

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -428,89 +404,57 @@ subfulfillment
...
 threshold
-: is a number and MUST be an integer in the range 1 ... 2^32 - 1. In order to fulfill a threshold condition, the weights of the provided fulfillments MUST be greater than or equal to the threshold.
+: is a number and MUST be an integer in the range 1 ... 255. In order to fulfill a threshold condition, the sum of the provided fulfillments MUST be greater than or equal to the threshold.
 
 subfulfillments
-: is the set of subconditions and subfulfillments, each provided as a tuple of weight and condition or fulfillment.
-
-weight
-: is the numeric weight of this subcondition, i.e. how many times it counts against the threshold. It MUST be an integer in the range 1 ... 2^32 - 1.
+: is the set of sub-fulfillments.
```
nbougalis: Either define the sorting order or explicitly allow any order.

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -717,102 +638,151 @@ This section to be expanded in a later draft.  <!-- TODO --> For now, see the te
...
+BASE10
+: Variable-length integer encoded as a base-10 (decimal) number. Implementations MUST reject encoded values that are too large for them to parse. Implementations MUST be tested for overflows.
+
+BASE16
+: Variable-length integer encoded as a base-16 (hexadecimal) number. Implementations MUST reject encoded values that are too large for them to parse. Implementations MUST be tested for overflows. Encodings may have an odd number of characters as the encoding excludes leading zeros.
+
+BASE64URL
+: Base64-URL encoding. See [RFC4648](#RFC4648), Section 5.
+
+## Condition URI Format {#string-condition-format}
```
nbougalis: I know I'm going to lose this "fight", but I've got to try again: defining a custom string format gets us nothing and only adds complexity, especially when you consider that the string format itself requires OER parsing anyways.

Instead of this complicated encoding, with fields and separators and field-specific encodings, I'd_ recommend something a lot simpler; assume the existence of a function called `TO_BLOB` which converts a condition or fulfillment to the corresponding binary format. Then define the string version of a condition `C` or fulfillment `F` as follows:

```
"cc:" BASE64URL(TO_BLOB(C))
"cf:" BASE64URL(TO_BLOB(F))
```

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -114,312 +114,288 @@ This specification is a part of the [Interledger Protocol](https://interledger.o
...
-Fulfillments are binary encoded as:
+Conditions are encoded as follows:
 
-    Fulfillment ::= SEQUENCE {
+    Condition ::= SEQUENCE {
       type ConditionType,
-      payload OCTET STRING
+      fingerprint OCTET STRING,
+      maxFulfillmentLength INTEGER (0..MAX),
+      subtypes OCTET STRING
```
nbougalis: Not sold on this...

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -114,312 +114,288 @@ This specification is a part of the [Interledger Protocol](https://interledger.o
...
 
-## PREIMAGE {#preimage-feature-suite}
-The PREIMAGE feature suite provides conditions that use a preimage as a one-time signature. This feature suite is assigned the feature bit 2^1 = 0x02.
+Each bit represents a type. The list of known types is the IANA-maintained [Crypto-Condition Type Registry](#crypto-conditions-type-registry) .
 
-The fingerprint of a preimage condition is the hash of an arbitrary value. The payload of a preimage fulfillment is the hashed arbitrary value before hashing, also known as the preimage. Conditions that use this preimage MUST also rely on a cryptographically secure hashing algorithm. Since cryptographically secure hashing functions are preimage-resistant, only the original creator of a preimage condition can produce the preimage, as long as it contains a large amount of random entropy.
+The presence of one of more sub-crypto-conditions of a specific type is indicated by setting the numbered bit corresponding to the type number of that type. Bits are numbered from the least significant bit of the first byte to the most significant bit of the last byte.
```
nbougalis: See http://www.oss.com/asn1/resources/books-whitepapers-pubs/Overview%20of%20OER.pdf section 2.7

0002-crypto-conditions/0002-crypto-conditions.md
adrianhopebailie: It significantly simplifies the whole spec. Open to ideas about how it could be done better but if you don't need the field then you consume a single byte to indicate it's empty.

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -114,312 +114,288 @@ This specification is a part of the [Interledger Protocol](https://interledger.o
...
-| Field                | Value | Description |
-|----------------------|-------|----------------------------------------------|
-| preface              | `cf`  | Constant. Indicates this is a fulfillment. |
-| type                 | `0`   | Type 0 is [PREIMAGE-SHA-256][]. |
-| payload              | `VGhlIG...pbGRl` | The BASE64URL-encoded SHA-256 preimage of the condition, since this is a PREIMAGE-SHA-256 type fulfillment. In this case, it is an arbitrary string. |
+Implementations which index conditions MUST use the entire encoded condition as the key, not just the fingerprint - as different conditions of different types may have the same fingerprint. 
 
+For most condition types, the fingerprint is a cryptographically secure hash of the data which defines the condition, such as a public key. 
 
+This field is encoded as a fixed length (32 byte) octet string for all currently known condition types.
```
adrianhopebailie: Will fix

0002-crypto-conditions/0002-crypto-conditions.md
adrianhopebailie: We've intentionally not used the BIT STRING type but rather the OCTET STRING type because the BIT STRING orders bits from highest order down on each bit which seems odd.

If you think that's a mistake I'm happy to review.

0002-crypto-conditions/0002-crypto-conditions.md
adrianhopebailie: > First of all, I don't see where we define the types for the existing 5 cryptoconditions. We can't use their types, since those are an ENUM and can't be used as a bitfield, at least as defined since some types have more than 1 bit set.

Each type has a numeric type ID (starting from the PREIMAGE-SHA-256 type with ID 0). The idea is that each bit flag position corresponds to the type ID of a type that is either used in the compound condition or one of its subconditions (bit set) or not.

Will review text to clarify.

0002-crypto-conditions/0002-crypto-conditions.md
adrianhopebailie: Those examples don't include the encoding stuff, just the raw bytes that will be encoded. That said they need to be updated because I changed that field to variable length.

While we have defined OER as our canonical encoding format I was trying to avoid being too tightly bound to that and so avoiding putting things like the length indicator in the example.

If the consensus is that we should put it in I'm happy to do so.

adrianhopebailie: > what's the 0x10? 

It's the first byte of the bitmap. `0001 0000`

Bit 4 is set indicating that there is an ED25519 condition used within this compound condition.

0002-crypto-conditions/0002-crypto-conditions.md
adrianhopebailie: Second byte of the bitmap is 0x04. 

Bit 2 of the second byte is set indicating that there is a sub condition used within this compound condition that has the type ID 10. i.e. If you see this then reject it because nobody knows what type 10 is 😄

0002-crypto-conditions/0002-crypto-conditions.md
adrianhopebailie: > Implementations MUST ignore any input message when validating a PREIMAGE-SHA-256 fulfillment.

For this line specifically?

0002-crypto-conditions/0002-crypto-conditions.md
adrianhopebailie: Not sure I follow?

The purpose of this section in each type definition is to be explicit about HOW to derive this number. If you're saying that this is intuitive that's fine I'd still like to be explicit.

0002-crypto-conditions/0002-crypto-conditions.md
adrianhopebailie: Will fix

0002-crypto-conditions/0002-crypto-conditions.md
adrianhopebailie: > especially when you consider that the string format itself requires OER parsing anyways

Is that true? The idea here is to have a format that can be processed easily by systems that don't implement OER. They may not be able to validate a fulfillment but at a minimum they can determine if they should accept the condition (perhaps for validation by another system). To do that they need to see the type, subtypes and maxfulfillmentlength fields.

0002-crypto-conditions/0002-crypto-conditions.md
nbougalis: If the ID is the value in the ENUM, it won't work. Consider, for example a value of 5. Does it mean 4+1 (prefix + ed25519) or 3+2 (threshold+RSA)?

ORing values together and storing them in a bitfield works great but only if the values are all powers of 2.

If we change the `ENUM` values and type IDs to be powers of two, then we limit ourselves to 8 condition types under the current spec definition.

0002-crypto-conditions/0002-crypto-conditions.md
nbougalis: I don't see how this is simpler than the features we currently have; in fact I think it's marginally worse, but I don't really feel strongly enough about this to "demand" one over the other.

However (in addition to my other comments elsewhere in the PR) I think that if we go with subtypes, we need three changes:

First, since the field is, essentially, a bitstring, I'd suggest using the correct OER/ASN.1 type: BIT STRING.

Second, both the type and subtypes field should be BIT STRING. Rationale: they both contain the same kind of information: types; only difference is that one will only allow one flag to be set, while the other multiple flags.

Last, I'd put the two fields next to each other.

0002-crypto-conditions/0002-crypto-conditions.md
nbougalis: Well, that's ok I guess it's not like a `BIT STRING` is somehow more efficient. It just documents the intent of the field a little better.

As long as we explicitly document all the details (such as how bits are ordered, etc) then an `OCTET STRING` is fine. You may disregard previous comments about using `BIT STRING`.

0002-crypto-conditions/0002-crypto-conditions.md
nbougalis: Follow-up: I'm fine with `OCTET STRING` instead of `BIT STRING` buy my other comments stand.

adrianhopebailie: > I don't see how this is simpler than the features we currently have; in fact I think it's marginally worse, but I don't really feel strongly enough about this to "demand" one over the other.

What value do you get at as an implementer knowing that a compound condition you're processing has subconditions requiring you to support certain features. What you're trying to determine is whether you can evaluate the fulfillment of those subconditions later and that requires more than just the features it requires knowledge about those subcondition types too.

Consider a future where we add the SHA-512 feature suite and somebody implements support for RSA-SHA-256 and PREIMAGE-SHA-512 but not RSA-SHA-512 (bad example but not impossible).

They see a THRESHOLD-SHA-256 condition that says it has subconditions with the features SHA-512, SHA-256 and RSA. How do they know if that means they'll encounter an RSA-SHA-512 subfulfillment?

> First, since the field is, essentially, a bitstring, I'd suggest using the correct OER/ASN.1 type: BIT STRING.

I agree. @justmoon isn't crazy for them but given how things have evolved I think he may agree now too.

> Second, both the type and subtypes field should be BIT STRING. Rationale: they both contain the same kind of information: types; only difference is that one will only allow one flag to be set, while the other multiple flags.

This is an interesting idea and means we'll never run out of type IDs although it's obviously less efficient than a fixed size INTEGER. Again, I'm inclined to agree. We should use the right ASN.1 type and not think too much about the final encoded form. A few bytes here and there for length indicators is nothing relative to large prefixes and signatures.

> Last, I'd put the two fields next to each other.

Agreed. This is a hangover from trying to make it optional and make conditions constant size if you exclude the field. Not sure I see the value of that now.

0002-crypto-conditions/0002-crypto-conditions.md
nbougalis: I guess if all you want is to peel out the type, subtype and fulfillment length, then you don't need OER parsing since that only applies for the payload. 

That's a convincing argument. Thanks.

0002-crypto-conditions/0002-crypto-conditions.md
adrianhopebailie: We are not tracking features anymore at all, just types. There is no combination of features simply an ID (the index of the enum) for each type. Subtypes is just a bitmap that says these are the types of any subconditions.

0002-crypto-conditions/0002-crypto-conditions.md
nbougalis: Feel free to ignore the first part of my comment - it was part of my thought process leading up to the second part (hashing the prefix in the condition) which was the point I _really_ wanted to make.

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -110,387 +110,380 @@ informative:
...
-Weighted signatures allow more complex relationships than simple M-of-N signing. For example, a weighted condition can support an arrangement of subconditions such as, "Either Ron, Adi, and Leonard must approve; or Clifford must approve."
+Multi-level signatures allow more complex relationships than simple M-of-N signing. For example, a weighted condition can support an arrangement of subconditions such as, "Either Ron, Mac, and Ped must approve; or Smithers must approve."
 
 
+## Crypto-conditions as a signature scheme
 
-# Format
+Crypto-conditions is a signature scheme for compound signatures which has similar properties to most other signature schemes, such as:
+  1. Validation of the signature (the fulfillment) is done using a public key (the condition) and a message as input
+  2. The same public key can be used to validate multiple different signatures, each against a different message
```
bachase: How does reuse work with the requirement of unique fingerprints?

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -110,387 +110,380 @@ informative:
...
+  3. It is not possible to derive the signature from the public key
+  
+However, the scheme also has a number of features that make it unique such as:
+  1. It is possible to derive the same public key from any valid signature without the message
+  2. It is possible for the same public key and message to be used to validate multiple signatures. For example, the fulfillment of an m-of-n condition will be different for each combination of n signatures.
+  4. Composite signatures use one or more other signatures as components allowing for recursive signature validation logic to be defined.  
+  3. A valid signature can be produced using different combinations of private keys if the structure of the compound signature requires only specific combinations of internal signatures to be valid  (m of n signature scheme).
```
bachase: Typo numbers 3 and 4 are swapped.

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -110,387 +110,380 @@ informative:
...
 
+## Crypto-conditions as a signature scheme
 
-# Format
+Crypto-conditions is a signature scheme for compound signatures which has similar properties to most other signature schemes, such as:
+  1. Validation of the signature (the fulfillment) is done using a public key (the condition) and a message as input
+  2. The same public key can be used to validate multiple different signatures, each against a different message
+  3. It is not possible to derive the signature from the public key
+  
+However, the scheme also has a number of features that make it unique such as:
```
bachase: I have a hard time understanding points 1-4 below.  Probably just my lack of experience in the domain, but thought it worth pointing out.

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -110,387 +110,380 @@ informative:
...
 
-## String Types
+It is also important that all participants in such a distributed system are able to evaluate, prior to the trigger being fired, that they will be capable of verifying the trigger. Determinism is useless if validation of the trigger requires algorithms or resources that are not available to all participants.
 
-Crypto-conditions use the following types within string encoding:
+Therefor conditions may be used as **distributable event descriptions** in the form of a *fingerprint*, but also *event meta-data* that allows the event verification system to determine if they have the necessary capabilities (such as required crypto-algorithms) and resources (such as heap size or memory) to verify the event notification later.
 
-BASE10
-: Variable-length integer encoded as a base-10 (decimal) number. Implementations MUST reject encoded values that are too large for them to parse. Implementations MUST be tested for overflows.
+Fulfillments are therefor **cryptographically verifiable event notifications** can be used to verify that the event occurred but also matches the given description.
```
bachase: Missing word?
> ... verifiable event notifications** ??? can be

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -110,387 +110,380 @@ informative:
...
 
-BASE16
-: Variable-length integer encoded as a base-16 (hexadecimal) number. Implementations MUST reject encoded values that are too large for them to parse. Implementations MUST be tested for overflows. No leading zeros.
+When using crypto-conditions as a trigger it will often make sense for the message that is used for validation to be empty to match the signature of the trigger processing system's API. This makes crypto-conditions compatible with systems that use simple hash-locks as triggers.
```
bachase: This is saying empty messages are needed to conform to the surrounding system's API?

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -110,387 +110,380 @@ informative:
...
 
-BASE64URL
-: Base64-URL encoding. See [RFC4648](#RFC4648), Section 5.
+If a PKI signature scheme is being used for the triggers this would require a new key pair for each trigger which is impractical. Therefor the PREFIX compound type wraps a sub-crypto-condition with a message prefix that is applied to the message before signature validation. In this way a unique condition can be derived for each trigger even if the same key pair is re-used with an empty message.
```
bachase: What does PKI stand for?

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -110,387 +110,380 @@ informative:
...
+  2. The same public key can be used to validate multiple different signatures, each against a different message
+  3. It is not possible to derive the signature from the public key
+  
+However, the scheme also has a number of features that make it unique such as:
+  1. It is possible to derive the same public key from any valid signature without the message
+  2. It is possible for the same public key and message to be used to validate multiple signatures. For example, the fulfillment of an m-of-n condition will be different for each combination of n signatures.
+  4. Composite signatures use one or more other signatures as components allowing for recursive signature validation logic to be defined.  
+  3. A valid signature can be produced using different combinations of private keys if the structure of the compound signature requires only specific combinations of internal signatures to be valid  (m of n signature scheme).
+  
+## Crypto-conditions as a trigger in distributed systems
```
bachase: Caveat to my limited experience, this and the following feature section define the core problem crypto-conditions are solving.   IMHO, I'd consider dropping the others and focusing just on these.

Perhaps the other features can be a few sentences in the feature section intro?

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -110,387 +110,380 @@ informative:
...
 
-## Condition {#condition-format}
-Below are the string and binary encoding formats for a condition.
+The advantage of using crypto-conditions for such use cases as opposed to a turing complete contract scripting language is the fact that the outcome of a crypto-condition validation is deterministic across platforms as long as the underlying cryptographic primitives are correctly implemented.
 
-### String Format {#string-condition-format}
+# Validation of a fulfillment
 
-Conditions are ASCII encoded as:
+Validation of a fulfillment (F) against a condition (C) and a message (M), in the majority of cases, the following steps.
```
bachase: Typo?
> in the majority of cases, the following steps

should read

> in the majority of cases, follows these steps:

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -110,387 +110,380 @@ informative:
...
-      preimageSha256(0),
-      rsaSha256(1),
-      prefixSha256(2),
-      thresholdSha256(3),
-      ed25519(4)
-    } (0..65535)
+## Subfulfillments
 
-### Fields {#condition-format-fields}
+In validating a fulfillment for a compound crypto-condition it is necessary to validate one or more sub-fulfillments per step 3 above. In this instance the condition for one or more of these sub-fulfillments is often not available for comparison with the derived condition. Implementations MUST skip the first fulfillment validation step as defined above and only perform steps 2 and 3 of the validation.
```
bachase: This section is confusing without the having seen the crypto-condition type definitions.  I'd consider dropping and letting people read these details later on.

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -110,387 +110,380 @@ informative:
...
+Implementations of this spec MUST support encoding and decoding using Octet Encoding Rules (OER) as defined in [ITU.X696.2015](#itu.X696.2015). This is the canonical encoding format. Where alternate encoding rules are defined and used these should be applied through to all ASN.1 defined data structures.
 
-Fulfillments are ASCII encoded as:
+For example, implementations that choose to use DER encoding should also use DER encoding when assembling the content of a condition fingerprint before it is hashed.
 
-    "cf:" BASE16(type) ":" BASE64URL(payload)
+## Condition {#condition-format}
 
-### Binary Format {#binary-fulfillment-format}
+The binary encoding of conditions is dependant on their type. The overall size of conditions is not fixed as some types (compound conditions) contain additional fields (subtypes) and the algorithm for deriving the fingerprint differs by type therefor the fingerprint size also differs.
```
bachase: This reads like simple types do not have a subtypes field, but below it states simple types have subtypes equal to types field.

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -110,387 +110,380 @@ informative:
...
-| SHA-256    | 2^0 | 0x01 | The SHA-256 hashing algorithm. |
-| PREIMAGE   | 2^1 | 0x02 | The functionality of comparing a hash to a preimage. |
-| PREFIX     | 2^2 | 0x04 | The functionality of prefixing the fulfillment with a prefix before generating the condition. |
-| THRESHOLD  | 2^3 | 0x08 | The functionality of composing a condition out of several weighted subconditions. |
-| RSA-PSS    | 2^4 | 0x10 | The RSA-PSS signature algorithm. |
-| ED25519    | 2^5 | 0x20 | The ED25519 signature algorithm. |
+For most condition types, the fingerprint is a cryptographically secure hash of the data which defines the condition, such as a public key. 
 
-## SHA-256 {#sha-256-feature-suite}
+For types that use PKI signature schemes, the signature is intentionally not included in the content that is used to compose the fingerprint. This meansthe fingerprint can be calculated without needing to know the message or having access to the private key.
```
bachase: Typo, missing a space
> This meansthe fingerprint

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -110,387 +110,380 @@ informative:
...
+Future types may use different functions to produce the fingerprint which may have different lengths therefor the filed is encoded as a variable length string.
 
-## PREIMAGE {#preimage-feature-suite}
-The PREIMAGE feature suite provides conditions that use a preimage as a one-time signature. This feature suite is assigned the feature bit 2^1 = 0x02.
+### MaxFulfillmentLength
 
-The fingerprint of a preimage condition is the hash of an arbitrary value. The payload of a preimage fulfillment is the hashed arbitrary value before hashing, also known as the preimage. Conditions that use this preimage MUST also rely on a cryptographically secure hashing algorithm. Since cryptographically secure hashing functions are preimage-resistant, only the original creator of a preimage condition can produce the preimage, as long as it contains a large amount of random entropy.
+This is the maximum length of the essential fulfillment payload for a fulfillment that can fulfill this condition.
```
bachase: What does essential mean in this context?

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -110,387 +110,380 @@ informative:
...
-Crypto-conditions may also be useful in other contexts where a system needs to make a decision based on predefined criteria, such as smart contracts.
+The function returns a boolean indicating if the compound signature is valid or not. This property of crypto-conditions means they can be used in most scenarios as a replacement for existing signature schemes which also take as input, a public key (the condition), a signature (the fulfillment), and a message and return a boolean result.
 
-## Terminology
+# Terminology
 The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC2119][].
 
-Within this specification, the term "condition" refers to the hash of a description of a signed message. The hash function must be preimage-resistant.
+# Types
+Crypto-conditions are a standard format for expressing conditions and fulfillments. The format supports multiple algorithms, including different hash functions and cryptographic signing schemes. Crypto-conditions can be nested in multiple levels, with each level possibly having multiple signatures.
```
bachase: Even though they are defined in the introduction, I'd consider briefly redefining conditions and fulfillments again here.

0002-crypto-conditions/0002-crypto-conditions.md
nbougalis: Correct.

I'm proposing instead that we define a well-known cryptocondition `TRIGGER` type: it's simply an outer wrapper around a subfulfillment/subcondition; a `TRIGGER` requires the `message` parameter to be empty, and will fail to validate if a non-empty message is presented. This helps avoid accidentally created conditions that can't be fulfilled because they require a message to fulfill, but one isn't provided.

0002-crypto-conditions/0002-crypto-conditions.md
justmoon: @adrianhopebailie and I chatted in Seoul and he successfully convinced me about using BIT STRING for the bitmaps. Using the format intended for this is case makes sense and the encoding isn't as unreasonable as I thought.

As for the type, I still think it should not be a bitmap (with one bit), but an integer. It's more space efficient and you don't have to worry about validating canonicality (I.e. exactly one bit must be set.) Plus it's trivial to convert an integer n to a bitmap with the n-th bit set.

Within that, the type could be a fixed length integer or a CHOICE, either one is fine with me.

0002-crypto-conditions/0002-crypto-conditions.md
justmoon: > This helps avoid accidentally created conditions that can't be fulfilled because they require a message to fulfill, but one isn't provided.

Not following this - there is no such thing as a condition that requires a message to fulfill. A condition is just a public key, so given a `condition`, for any possible message `message`, there is a `fulfillment` that fulfills the condition with respect to that message.

A condition is like a public key and a public key can sign any message. The fact that our triggers are signing the empty message is a bit of an arbitrary decision. I can easily imagine someone else's implementation of triggers using crypto-conditions whether the message signed is some constant string defined by the protocol instead.

Sure you could have a public key type which says "this public key can only be used to sign the empty message", but that seems awfully specific and I'm not sure what the benefit of having this feature would be.

Triggers are just one of many ways that you can use a signature scheme and crypto-conditions aren't special compared to other signature schemes in that you can use them as triggers, so I don't see a need to add a special feature for triggers any more than Ed25519 or RSA or any other scheme has a special key format for triggers.

adrianhopebailie: Sorry, I confused myself. I forgot that conditions don't have the signature in them. 

@nbougalis how were you envisioning this happening, maybe I've missed something:

> accidentally created conditions that can't be fulfilled because they require a message to fulfill, but one isn't provided

0002-crypto-conditions/0002-crypto-conditions.md
justmoon: @nbougalis Your argument is spot on when it comes to the fulfillment. The fulfillment is a huge OER encoded blob. And the only things you'll want to know about it is its structure and whether it's valid. And to know either you need to be able to fully parse it anyway. There is very, very little value in knowing just the top level type of the fulfillment and nothing else. So having a top level string encoding makes no sense.

For the condition, it's kinda the opposite. Conditions have no structure past the top level. They consist only of a hash, a type, a max fulfillment length and maybe a subtypes bitmap. None of these things require OER to parse. I could easily see people wanting to build simple filters that can check whether a condition type is supported. It's also nice to be able to see the type just from looking at the condition, e.g. `cc:preimage-sha256:...`

The point is: I think it's very important to answer this question for the condition and the fulfillment separately because the design goals in each case are very different.

For the condition, I like the idea of sticking with something similar to the current format, like `cc:preimage-sha256:[hash]:1024` and `cc:threshold-sha256:[hash]:[subtypes bitmap]:15300`. For the fulfillment, I like the suggestion of switching to `"cf:" BASE64URL(TO_BLOB(F))`

There is still the issue that conditions also appear in threshold and prefix fulfillments as "subconditions". So we either need to have a binary encoding in addition to the string encoding (awkward) or we have to use the string encoding inside of the binary encoding (awkward).

We may be able to reduce the awkwardness of the latter option. Instead of having the fingerprint contents of a threshold condition include the condition as a string (e.g. `"cc:preimage-sha256:[hash]:1024"`), they could include a hash of the condition string instead (e.g. `0x13abd...`), then it would just be a Uint256 in the OER format, which saves a few bytes, is just as secure and solves the bad taste of putting string format inside of a binary format.

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -110,387 +110,380 @@ informative:
...
+This specification is a part of the [Interledger Protocol](https://interledger.org/) work. Feedback related to this specification should be sent to <ledger@ietf.org>.
 
 --- abstract
 
-Crypto-conditions provide a mechanism to describe a signed message such that multiple actors in a distributed system can all verify the same signed message and agree on whether it matches the description. This provides a useful primitive for event-based systems that are distributed on the Internet since we can describe events in a standard deterministic manner (represented by signed messages) and therefore define generic authenticated event handlers.
+Crypto-conditions defines a set of encoding formats and data structures for conditions and fulfillments.  A condition uniquely identifies a boolean circuit constructed from one or more logic gates, evaluated by either validating a cryptographic signature or verifying the preimage of a hash digest. A fulfillment is a data structure encoding one or more cryptographic signatures and hash digest preimages that can be used to evaluate the result of the circuit.
```
justmoon: Yeah, the abstract is a bit self-referential:

> Crypto-**conditions** defines a set of encoding formats and data structures for **conditions** and fulfillments.

Maybe we can consolidate these sections, they seem fairly redundant. Maybe move the first paragraph of the intro to the abstract and make the intro more of a section about why we are doing this, perhaps referencing Chris Allen's Smart Signatures etc.

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -110,387 +110,380 @@ informative:
...
-In the simplest case, the fulfillment can be a preimage that hashes to the condition, in which case the preimage is both the description and the message.
+## Simple and Compound Types
 
+Two categories of crypto-condition type exist. Simple crypto-conditions provide a standard encoding for existing primitives such as RSA and ED25519 signatures, or SHA256 hash digests where many of the scheme parameters have hardcoded values. As such multiple simple types maybe be defined to encode the same underlying scheme but where the parameters differ.
 
-## Features
-Crypto-conditions are a standard format for expressing conditions and fulfillments. The format supports multiple algorithms, including different hash functions and cryptographic signing schemes. Crypto-conditions can be nested in multiple levels, with each level possibly having multiple signatures.
+As an example, the types defined in this version of the specification all use the SHA-256 digest algorithm for derivation of the condition fingerprint. If a future version were to introduce SHA-512 as an alternative this would require that new types be defined for each existing type that must have its condition derived using SHA-512.
```
justmoon: A lot of repetition of the similar sounding words "derived" and "defined".

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -110,387 +110,380 @@ informative:
...
+
+A compound crypto-condition may have multiple branches and be defined to evaluate to TRUE even if only a subset of these evaluate to TRUE (as in an m-of-n signature scheme). As such the valid fulfillment of a compound condition could contain a combination of sub-fulfillments (that can be evaluated) but also sub-conditions that are simply used (in combination with the derived conditions from the sub-fulfillments) to derive the compound condition and compare this to the condition provided for validation.
+
+## Defining and supporting New types
+
+The crypto-conditions format has been designed so that it can be expanded. For example, you can add new cryptographic signature schemes or hash functions. This is important because advances in cryptography frequently render old algorithms insecure or invent newer, more effective algorithms.
+
+Implementations are not required to support all condition types therefor it is necessary to indicate which types an implementation must support in order to validate a fulfillment. For this reason, compound conditions are encoded with an additional field, subtypes, indicating the set of types and subtypes of all sub-crypto-conditions.
```
justmoon: I think we were considering (and have ultimately decided to?) change it so that only compound conditions contain a type field.

Notes for background:

* I'm assuming that conditions have only a string format, not a binary format. (See https://github.com/interledger/rfcs/pull/124/files#r90308714)
* I'm assuming that there may be conditions in the future which have other custom fields (other than subtypes)

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -110,387 +110,380 @@ informative:
...
+Each condition type uses one or more cryptographic primitives such as digest or signature algorithms. Compound types may contain sub-crypto-conditions of any type and indicate the set of underlying types in the subtypes field of the condition
 
-### Multi-Signature
+To verify that a given implementation can verify a fulfillment for a given condition, implementations MUST ensure they are able to validate fulfillments of all types indicated in the subtypes field of a compound condition. If an implementation encounters an unknown type it MUST reject the condition as it will almost certainly be unable to validate the fulfillment.
+
+## Multi-Signature
 Crypto-conditions can abstract away many of the details of multi-sign. When a party provides a condition, other parties can treat it opaquely and do not need to know about its internal structure. That allows parties to define arbitrary multi-signature setups without breaking compatibility.
```
justmoon: Good catch. I would say implementation can treat conditions opaquely with respect to their *data*, but not with respect to their *type* (and subtypes).

0002-crypto-conditions/0002-crypto-conditions.md
justmoon: Fingerprints are unique with respect to different public keys, not different messages. This sentence is talking about different messages.

0002-crypto-conditions/0002-crypto-conditions.md
justmoon: No, the trigger use case is not the core problem crypto-conditions are solving. Quite the opposite, it's the motivating use case, but that's mostly a historical accident. A much more typical use case for crypto-conditions would be multi-sig authentication.

For instance, when rippled added support for multi-sig, crypto-conditions would have been a very elegant way to express the keys. We just didn't have them back then, so we came up with a custom way of doing it. But that was a much more typical use case than SusPay.

0002-crypto-conditions/0002-crypto-conditions.md
justmoon: See my earlier comment. We probably want to change the section below to remove subtypes from simple types.

0002-crypto-conditions/0002-crypto-conditions.md
adrianhopebailie: > There is still the issue that conditions also appear in threshold and prefix fulfillments as "subconditions". So we either need to have a binary encoding in addition to the string encoding (awkward) or we have to use the string encoding inside of the binary encoding (awkward).
> 
> We may be able to reduce the awkwardness of the latter option. Instead of having the fingerprint contents of a threshold condition include the condition as a string (e.g. "cc:preimage-sha256:[hash]:1024"), they could include a hash of the condition string instead (e.g. 0x13abd...), then it would just be a Uint256 in the OER format, which saves a few bytes, is just as secure and solves the bad taste of putting string format inside of a binary format.

I don't agree that having two encodings is awkward. Having reviewed the spec with this change in mind I find mixed-encodings worse to be honest. If we only have a string encoding then we must define the rules about how to hash it for use as a sub-condition and implementations must use DER to generate the fingerprint then produce a string encoded condition and then hash that to get something binary that is then used as a sub-field in a DER encoding of the compound condition.

For a condition that goes a few levels deep you have to keep flipping between encoding rules instead of just using DER until you get to the top and then String encoding the top-level condition after you've generated the fingerprint.

I'd prefer to say:
1. All data structures are defined in ASN.1 (no change to what we have today)
2. DER is a our canonical encoding ruleset for encoding these data structures (DER preferred over OER by the crypto-community)
3. We define a second encoding ruleset for String/URI encoding that can be used to URI encode conditions. (Take our existing encoding rules and define how they'd be implemented off the same ASN.1 data structure base as the DER encoding rules)

The rules can be as simple as:

  * String encoding encodes data using ASCII characters according to the following rules:
    * All Conditions start with the prefix `cc:` followed by a type name as defined in the type registry and another `:` character
    * A SEQUENCE is encoded by encoding each field using the type specific rule below and separating each field with a `:` character.
    * INTEGER types are encoded as the string representation of the number (base 10)
    * OCTET STRING types are encoded as a base64url encoded string
    * BIT STRING types are encoded as a base 10 integer and represented as a string (need to think about this one but I think that will work)

0002-crypto-conditions/0002-crypto-conditions.md
adrianhopebailie: The abstract is supposed to be a short summary of the whole spec so there will be some redundancy (it's intentional).

In the interests of having the technical bits worked out for people to start implementing I'm going to leave this for now and maybe re-work later.

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -110,387 +110,380 @@ informative:
...
 
-The term "fulfillment" refers to a description of a signed message and a signed message that matches the description. We hash the description and compare that to the condition, and also compare the signed message to the description. If the message matches the description and the hash of the description matches the condition, we say that the fulfillment fulfills the condition.
+The different structures of different crypto-conditions and the algorithms they employ is defined by the type of the crypto-condition.
 
-In the simplest case, the fulfillment can be a preimage that hashes to the condition, in which case the preimage is both the description and the message.
+## Simple and Compound Types
 
+Two categories of crypto-condition type exist. Simple crypto-conditions provide a standard encoding for existing primitives such as RSA and ED25519 signatures, or SHA256 hash digests where many of the scheme parameters have hardcoded values. As such multiple simple types maybe be defined to encode the same underlying scheme but where the parameters differ.
```
adrianhopebailie: Fixed

adrianhopebailie: 👍

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -110,387 +110,380 @@ informative:
...
+As an example, the types defined in this version of the specification all use the SHA-256 digest algorithm for derivation of the condition fingerprint. If a future version were to introduce SHA-512 as an alternative this would require that new types be defined for each existing type that must have its condition derived using SHA-512.
+
+Compound crypto-conditions contain one or more sub-crypto-conditions. Compound crypto-conditions are used to construct the branches of the boolean circuit where a compound condition will evaluate to TRUE or FALSE based on the output of the evaluation of their sub-crypto-conditions. 
+
+A compound crypto-condition may have multiple branches and be defined to evaluate to TRUE even if only a subset of these evaluate to TRUE (as in an m-of-n signature scheme). As such the valid fulfillment of a compound condition could contain a combination of sub-fulfillments (that can be evaluated) but also sub-conditions that are simply used (in combination with the derived conditions from the sub-fulfillments) to derive the compound condition and compare this to the condition provided for validation.
```
adrianhopebailie: Have done some word-smithing to be clearer using the suggested text

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -110,387 +110,380 @@ informative:
...
+
+Implementations are not required to support all condition types therefor it is necessary to indicate which types an implementation must support in order to validate a fulfillment. For this reason, compound conditions are encoded with an additional field, subtypes, indicating the set of types and subtypes of all sub-crypto-conditions.
 
-This format has been designed so that it can be expanded. For example, you can add new cryptographic signature schemes or hash functions. This is important because advances in cryptography frequently render old algorithms insecure or invent newer, more effective algorithms.
+#Features
 
-The [Bitmask](#bitmask) of a crypto-condition indicates which algorithms it uses, so a compliant implementation can know whether it supports the functionality required to interpret the crypto-condition.
+Crypto-conditions offer many of the features required of a regular signature scheme but also others which make them useful in a variety of new use cases.
```
adrianhopebailie: :smile: I hate consecutive headings, personal style

0002-crypto-conditions/0002-crypto-conditions.md
adrianhopebailie: As with the intro and abstract I'm going to defer on trying to change this unless it is actually incorrect.

0002-crypto-conditions/0002-crypto-conditions.md
adrianhopebailie: Fixed

0002-crypto-conditions/0002-crypto-conditions.md
adrianhopebailie: Fixed

PR #125, opened 2016-11-15
Five Bells Ledger API specification

mDuo13:
(Formerly known as the "Common Ledger API")

This PR has been updated to a third iteration, following the work earlier in this PR which continued from PR #90.. In keeping with our current thinking, a common ledger API may not be needed in the long run, but a formalization of the Five Bells Ledger API currently in use is helpful.

Thus, this PR now more closely reflects the existing implementation of the [five-bells-ledger](https://github.com/interledgerjs/five-bells-ledger), including some of the deprecated and odd functionality such as using "actual" crypto-conditions instead of just hashlocks, supporting cancellation conditions, having multi-credit, multi-debit with a proposed stage, not separating the ID and client ID, lacking URL fields for some operations, etc.


0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1181 @@
...
+| `rejection_reason`     | String               | _(Server-provided)_ The reason the transfer was rejected. MUST appear if and only if `state` is `rejected`. |
+| `state`                | String               | _(Server-provided)_ The current state of the transfer. Valid states are `prepared`, `executed`, and `rejected`. |
+| `timeline`             | Object               | _(Server-provided)_ Timeline of the transfer's state transitions. |
+| `timeline.executed_at` | [Date-Time][]        | _(Server-provided)_ Time when the transfer was originally executed. MUST appear if and only if `state` is `executed`. This time MUST be equal to or later than the `prepared_at` time. |
+| `timeline.prepared_at` | [Date-Time][]        | _(Server-provided)_ Time when the transfer was originally prepared. MUST appear, even if the transaction is unconditional. |
+| `timeline.rejected_at` | [Date-Time][]        | _(Server-provided)_ Time when the transfer was originally rejected. MUST appear if and only if `state` is `rejected`. This time MUST be equal to or later than the `prepared_at` time. |
+
+[UUID]: http://en.wikipedia.org/wiki/Universally_unique_identifier
```
emschwartz: Why is this line just hanging out at the bottom here?

0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1181 @@
...
+
+
+### Message Resource
+[Message resource]: #message-resource
+
+Messages are sent through the ledger's [Send Message][] method and received in a [WebSocket][] subscription. All fields of the message are required:
+
+| Field    | Value   | Description                                             |
+|:---------|:--------|:--------------------------------------------------------|
+| `ledger` | [URL][] | The base [URL][] of this ledger. MUST be an HTTP(S) URL where the client can [Get Ledger Metadata][]. |
```
emschwartz: What's the point of having this field?

0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1181 @@
...
+
+
+### Asset Resource
+[Asset resource]: #asset-resource
+
+There are many kinds of resources that can be tracked in a ledger. The Asset Resource describes the resource tracked in this ledger, so that client applications can display appropriate information to users. An asset resource has the following fields:
+
+| Field            | Value     | Description                                   |
+|:-----------------|:----------|:----------------------------------------------|
+| `type`           | String    | The type of asset. Currently, the only supported type is `iso4217-currency`. |
```
emschwartz: What is this asset definition format based on (I remember there was some link posted but can't remember what it was)?

We really need to have a way to define custom asset types. I think it would be useful to make some of the other fields mandatory so clients can depend on them, no matter if you're defining something custom.

0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1181 @@
...
+
+
+### Account Resource
+[Account resource]: #account-resource
+
+An account resource contains the fields in the following table. Some fields are _Server-provided_, meaning they cannot be set directly by clients. Fields that are "Optional" or "Server-provided" in the following table may be omitted by clients submitting objects to the API, but those fields are not optional to implement. Fields of nested objects are indicated with a dot (`.`) character; no fields contain a dot literal.
+
+| Name                      | Type    | Description                            |
```
emschwartz: Should we add the concept of account owners here?

0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1181 @@
...
+#### Errors
+
+- [NotFoundError][]
+- [InvalidUriParameterError][]
+
+
+### Send Message
+[Send Message]: #send-message
+
+Try to send a notification to another account. The message is only delivered if the other account is subscribed to [account notifications](#subscribe-to-account) on a WebSocket connection. ILP Connectors use this method to share quote information.
```
emschwartz: And by clients to request quotes

0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1181 @@
...
+#### Example
+
+Request:
+
+```json
+POST /messages
+Content-Type: application/json
+
+{
+  "from": "https://blue.ilpdemo.ripple.com/ledger/accounts/bob",
```
emschwartz: Ledgers MUST validate the `from` value if it is set by the client. It is very important that receivers of messages can trust that the message came from the correct source.

0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1181 @@
...
+
+### Date-Time
+[Date-Time]: #date-time
+
+All dates and times should be expressed in [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) date-time format with precision to the second or millisecond. The time zone offset should always be `Z`, for no offset. In other words, the date-time should be a **string** matching the one of the following formats:
+
+| Precision   | Format                     |
+|:------------|:---------------------------|
+| Second      | `YYYY-MM-DDTHH:mm:ss.sssZ` |
+| Millisecond | `YYYY-MM-DDTHH:mm:ssZ`     |
```
emschwartz: In the vein of having fewer options and more standardization, I think we could just say that everyone needs to include milliseconds. Also, I think the values are reversed in this table

0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1181 @@
...
+
+A successful result uses the HTTP response code **200 OK** and contains a JSON object with the following fields. (A field name like `foo[].bar` indicates that the row describes the field `bar` in _each_ object contained in the array `foo`.)
+
+| Field               | Value              | Description                       |
+|:--------------------|:-------------------|:----------------------------------|
+| `asset_info`        | [Asset resource][] | Information on the currency or other asset this ledger tracks. |
+| `ilp_prefix`        | String             | The ILP Address prefix of the ledger. This must match the regular expression `^[a-zA-Z0-9._~-]+\.$` |
+| `connectors`        | Array of Objects   | Accounts belonging to recommended ILP Connectors. This list MAY be empty. |
+| `connectors[].id`   | [URL][]            | The unique ID of this connector's account. This is used when [messaging](#send-message) the connectors to get quotes. MUST be an HTTP(S) URL where a client can [Get Account][]. |
+| `connectors[].name` | [URL][]            | The unique `name` field of this connector's account. |
```
emschwartz: Do we need to provide the name? Will it be used by anything? I think we might be able to make `connectors` just an array of strings where each one is the connector ID (URL)

0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1181 @@
...
+```
+
+Response:
+
+```json
+HTTP/1.1 200 OK
+
+{
+  "currency_code": null,
+  "currency_symbol": null,
```
emschwartz: Would be better if in the example these weren't `null`

0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1181 @@
...
+  "debit_account": "http://usd-ledger.example/accounts/alice",
+  "credit_account": "http://usd-ledger.example/accounts/bob",
+  "amount": "50",
+  "execution_condition": "cc:0:3:8ZdpKBDUV-KX_OnFZTsCWB_5mlCFI3DynX5f5H2dN-Y:2",
+  "expires_at": "2016-10-25T08:41:59.795Z",
+  "state": "prepared",
+  "timeline": {
+    "prepared_at": "2015-06-16T00:00:01.000Z"
+  }
+}
```
emschwartz: This looks so much cleaner and simpler than before!

0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1181 @@
...
+```
+
+#### Errors
+
+- [InsufficientFundsError][] - The `debit_account` would go below its `minimum_allowed_balance` if this transfer were executed
+- [UnprocessableEntityError][] - The request is formatted properly but contains an otherwise-unspecified semantic problem.
+- [AlreadyExistsError][] - The `client_id` supplied is already used by another transfer.
+- [InvalidUriParameterError][] - One of the [URL][] or [UUID][] parameters was not formatted properly.
+- [InvalidBodyError][] - The request body was not properly-formatted JSON, or did not match the Content-Type provided.
```
emschwartz: Wouldn't this also happen if you left out a required field or added one that was not supported?

0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1181 @@
...
+
+**Authorization:** The owner of an account MUST be able to submit the fulfillment for a transfer that credits the account. A ledger MAY allow other account owners or unauthorized clients to submit a fulfillment. (If only the `credit_account` can submit a fulfillment, then crediting an account with a conditional transfer effectively requires permission from the account owner.)
+
+#### Request Format
+
+PUT to the `fulfillment` path of the transfer you want to execute, with the header `Content-Type: application/json` and a message body containing the fulfillment.
+
+**Metadata URL Name:** `transfer_fulfillment`. Suggested generic path: `/transfers/{client_id}/fulfillment`
+
+**Caution:** The ledger _MUST_ require the `Content-Type: application/json` header on this request. In the future, the API might also accept fulfillments in binary fulfillment at this same endpoint.
```
emschwartz: Why is this a **Caution**?

Also, what do you think about defining the binary method now (or making it only binary if we think that's better)? @justmoon

0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1181 @@
...
+- [UnprocessableEntityError][] - The request is formatted properly but contains an otherwise-unspecified semantic problem.
+- [InvalidBodyError][] - The request body was not properly-formatted JSON.
+
+
+### Reject Transfer
+[Reject Transfer]: #reject-transfer
+
+Reject a prepared transfer. A transfer can be rejected if and only if that transfer is in the `prepared` state. Doing so transitions the transfer to the `rejected` state.
+
+**Authorization:** The `credit_account` of a transfer MUST be able to reject the transfer. Unauthorized users and the `debit_account` MUST NOT be able to reject the transfer. Other account owners MUST NOT be able to reject the transfer unless those accounts are explicitly authorized to do so by the `credit_account` and those accounts are not the `debit_account`.
```
emschwartz: The second sentence is a bit confusing, especially since we haven't defined any type of delegation in this spec

0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1181 @@
...
+
+**Metadata URL Name:** `transfer_rejection`. Suggested generic path: `/transfers/{client_id}/rejection`
+
+##### Body Parameters
+
+The message body should be a JSON object with one top-level field:
+
+| Field              | Value  | Description                                    |
+|:-------------------|:-------|:-----------------------------------------------|
+| `rejection_reason` | String | The reason for rejecting the transfer. This is intended to be a machine-readable identifier. |
```
emschwartz: Should we try to define a number of `rejection_reason`s like we're defining error types?

0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1181 @@
...
+```json
+PUT /transfers/3a2a1d9e-8640-4d2d-b06c-84f2cd613204/rejection
+Content-Type: application/json
+
+{
+  "rejection_reason": "BlacklistedSender"
+}
+```
+
+Response:
```
emschwartz: This still seems odd to me that you send just the `rejection_reason` and get back the whole transfer. It's also inconsistent with how the fulfillments work because according to this logic you would want to see the whole transfer for that too.

I think it's probably better to just return the `rejection_reason` and rely on notifications to give you updates about the transfer state. In that vein, we should make a note about whether an `HTTP 200` indicates the transfer was definitely rejected

0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1181 @@
...
+  },
+  "note_to_self": {
+    "foo": "bar"
+  },
+  "id": "https://red.ilpdemo.org/ledger/transfers/63a61fa9-fca2-4779-9d50-49dc691b8fbf",
+  "fulfillment": "https://red.ilpdemo.org/ledger/transfers/63a61fa9-fca2-4779-9d50-49dc691b8fbf/fulfillment",
+  "state": "rejected",
+  "timeline": {
+    "prepared_at": "2016-10-21T22:45:01.000Z",
+    "cancelled_at": "2016-10-22T21:33:34.867Z"
```
emschwartz: `rejected_at`

0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1181 @@
...
```
emschwartz: `rejected_at`

0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1181 @@
...
+- [NotFoundError][]
+- [InvalidUriParameterError][]
+
+
+### Get Transfer Fulfillment
+[Get Transfer Fulfillment]: #get-transfer-fulfillment
+
+Retrieve the fulfillment for a transfer that has been executed or canceled. This is separate from the [Transfer resource][] because it can be very large.
+
+**Authorization:** The ledger MUST allow owners of the `credit_account` and `debit_account` to get the transfer's fulfillment. A ledger MAY allow other account owners or unauthorized users to get a transfer's fulfillment.
```
emschwartz: Might be helpful to include a note about why you would or would not allow other users to get a transfer's fulfillment. If you can't get another user's transfer you probably shouldn't be able to get their fulfillments (which would indicate that a transfer has happened). Not sure how this would work with the dependent transfers idea...

0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1181 @@
...
+- [InvalidUriParameterError][]
+- [UnprocessableEntityError][]
+
+
+### Get Auth Token
+[Get Auth Token]: #get-auth-token
+
+Get a token that can be used to authenticate future requests.
+
+**Note:** This method is REQUIRED for ledgers to authenticate [WebSocket][] connections. If the ledger does not authenticate WebSocket connections, this method is OPTIONAL. The ledger MAY allow clients to authenticate for any other methods of the API using the tokens returned by this method.
```
emschwartz: I think the ledger MUST authenticate WebSocket connections

0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1181 @@
...
+## WebSocket
+[WebSocket]: #websocket
+
+**Metadata URL Name:** `websocket`. Suggested generic path: `/websocket` (with `wss://` protocol)
+
+Clients can subscribe to live, read-only notifications of ledger activity by opening a [WebSocket](https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API) connection to this path and sending a subscription request.
+
+### WebSocket Authentication
+
+A ledger MAY require clients to authenticate themselves using a token in the `token` query parameter. If the ledger supports authentication, it MAY restrict the data that can be accessed, as long as it complies with the following rules:
```
emschwartz: I don't understand why you would allow unauthenticated users to subscribe to anything

0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1181 @@
...
+
+| Field     | Value            | Description                                   |
+|:----------|:-----------------|:----------------------------------------------|
+| `id`      | String or Number | The `id` value from the request. This helps distinguish responses from other messages. |
+| `jsonrpc` | String           | MUST have the value `"2.0"` to indicate this is a JSON-RPC 2.0 message. |
+| `result`  | Number           | Updated number of active account subscriptions on this WebSocket connection. In practice, this is usually the length of the `params` array from the request. |
+
+Later, the ledger responds with [notifications](#websocket-notifications) whenever any of the following occurs:
+
+- A transfer affecting the account is prepared (`event` type: `transfer.create`)
```
emschwartz: I think we should change the event names to match the transfer statuses

0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1181 @@
...
+#### Subscribe to Transfer
+
+The client sends a JSON object to the ledger with the following fields:
+
+| Field     | Value            | Description                                   |
+|:----------|:-----------------|:----------------------------------------------|
+| `id`      | String or Number | An arbitrary identifier for this request. MUST NOT be `null`. |
+| `jsonrpc` | String           | MUST have the value `"2.0"` to indicate this is a JSON-RPC 2.0 request. |
+| `method`  | String           | MUST be the value `"subscribe_transfer"`.      |
+| `params`  | Array            | Each member of this array must be the [URL][] of a transfer to subscribe to, as a string. This array replaces any existing transfer subscriptions on this WebSocket connection. If the array length is zero, the client is unsubscribed from all transfer notifications. |
```
emschwartz: Looks like `params` is sometimes an array and sometimes an object -- is that allowed?

0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1181 @@
...
+##### Event Types
+[Event Types]: #event-types
+
+A ledger MAY define custom `event` types. A ledger MUST support at least the following `event` values:
+
+| Value             | Resource              | Description                      |
+|:------------------|:----------------------|:---------------------------------|
+| `transfer.create` | [Transfer resource][] | Occurs when a new transfer is prepared. Sent to clients subscribed to the `debit_account` and/or the `credit_account`. If the transfer is unconditional, this notification indicates the state of the transaction after execution. The `related_resources` field is omitted. |
+| `transfer.update` | [Transfer resource][] | Occurs when a transfer changes state from `prepared` to `executed` or `rejected`. If the transfer was executed, the `execution_condition_fulfillment` field of `related_resources` MUST contain the fulfillment. If the transfer was rejected, the `related_resources` field is empty. |
+| `message.sent`    | [Message resource][]  | Occurs when someone else sends a message. |
```
emschwartz: Should be `message.send` to be consistent with the tense of the others

0012-common-ledger-api/0012-common-ledger-api.md
sentientwaffle: Right now ilp-core uses `name` and ignores `id` (see https://github.com/interledgerjs/ilp-core/blob/11a019344b254d33d8cddb7ddfef0642d47977f2/src/lib/client.js#L227-L237).

0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1181 @@
...
+`^[-+]?[0-9]*[.]?[0-9]+([eE][-+]?[0-9]+)?$`
+
+Client applications can decode numeric strings to whatever representation provides sufficient precision for their specific needs. The ledger should report information about the precision it uses in the [ledger metadata][Get Ledger Metadata] response.
+
+
+### Assets and Currency
+
+The Common Ledger API provides access to a ledger containing a single currency or asset. The [Get Ledger Metadata][] method reports the type of asset or currency used, as an [Asset resource][].
+
+If a provider supports multiple currencies, the provider can server Common Ledger APIs separately for each currency. For each method described in this document, there would be a version of it prefixed by a different currency code. For example, the [Prepare Transfer][] method could be available for different currencies at the following locations:
```
dncohen: Why restrict the prefix to just a currency code? Why not allow https://ledger.example.com/USD.bitstamp/transfers and https://ledger.example.com/USD.gatehub/transfers, for example?

mDuo13: Yeah, that's what's intended, it's just not clear from how it's written now. (I'll fix later.)

0012-common-ledger-api/0012-common-ledger-api.md
mDuo13: In JSONRPC 2.0, params is a "structured" value -- in other words, yes. The examples in their spec contain both.

That said, it's probably wise to make this an object so it's forward compatible and similar to the Subscribe to Account call.

0012-common-ledger-api/0012-common-ledger-api.md
mDuo13: I agree.

0012-common-ledger-api/0012-common-ledger-api.md
mDuo13: I can imagine a ledger whose _purpose_ is to share transfers publicly, in which case you wouldn't need to authenticate anyone for the WebSocket API.

0012-common-ledger-api/0012-common-ledger-api.md
mDuo13: The "Caution" callout here is to point out specific behavior that is innocuous now but would lead to spec incompatibility when we add a binary format.

0012-common-ledger-api/0012-common-ledger-api.md
mDuo13: I'm not sure. I thought that would be an UnprocessableEntityError, but maybe not.

0012-common-ledger-api/0012-common-ledger-api.md
mDuo13: I like making it an array of ID URLs.

0012-common-ledger-api/0012-common-ledger-api.md
mDuo13: I'd rather make milliseconds optional because it saves some implementers the trivial concern of whether their programming language's ISO-String function provides milliseconds by default.

0012-common-ledger-api/0012-common-ledger-api.md
mDuo13: This line defines the "[UUID] reference link that first appears in the table above it.

0012-common-ledger-api/0012-common-ledger-api.md
mDuo13: Added a clarification about this.

0012-common-ledger-api/0012-common-ledger-api.md
mDuo13: https://gist.github.com/Fluidbyte/2973986

0012-common-ledger-api/0012-common-ledger-api.md
mDuo13: I think it serves as a linkback so you can find URLs for other methods, find out what currency this resource uses, etc.

0012-common-ledger-api/0012-common-ledger-api.md
emschwartz: I think it works fine without the second sentence. If we want to add a section about delegation later we could and it should be fairly obvious that accounts with permissions delegated to them would not be able to do what the delegating accounts cannot.

0012-common-ledger-api/0012-common-ledger-api.md
emschwartz: expired, unwanted, insufficient_liquidity or insufficient_time (for the connector to pass it on), invalid_rate, unsupported_condition...

Now that I'm thinking about this though I'm wondering if it actually makes sense to have this field in the transfer at all.

We need a protocol for communicating errors down the path analogous to ICMP. That will be defined on the same level as the quoting protocol, and would probably use the message functionality for communication. Since we need that protocol anyway, I'm not sure it makes sense to either define `rejection_reason`s on this level or even include them on this level. Most of the reasons will be _inter_-ledger ones anyway, so the messages used should be more separate from this ledger layer.

0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1201 @@
...
+
+An account resource usually has an owner who can [authenticate](#authorization-and-authentication) to the ledger. A ledger MUST serve all the following fields to clients authenticated as the account owner. A ledger MAY serve a subset of fields, containing at least the `name` and `id` of the account, to clients not authenticated as the account owner.
+
+| Name                      | Type    | Description                            |
+|:--------------------------|:--------|:---------------------------------------|
+| `name`                    | String  | Name of the account. A ledger MAY require this to be unique. MUST match the regular expression `^[a-zA-Z0-9._~-]{1,256}$`. |
+| `fingerprint`             | String  | _(Optional)_ A fingerprint of the account owner's client certificate. This field MUST be available if and only if the ledger supports client certificate authentication. MUST match the regular expression `^[0-9A-Fa-f]{2}(:[0-9A-Fa-f]{2}){19,127}$`. |
```
clark800: We are currently discussing a proposal to replace fingerprints with hostnames, which makes certificate changes easier.

0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1201 @@
...
+  "client_id": "3a2a1d9e-8640-4d2d-b06c-84f2cd613204",
+  "ledger": "http://usd-ledger.example",
+  "debit_account": "http://usd-ledger.example/accounts/alice",
+  "credit_account": "http://usd-ledger.example/accounts/bob",
+  "amount": "50",
+  "execution_condition": "cc:0:3:8ZdpKBDUV-KX_OnFZTsCWB_5mlCFI3DynX5f5H2dN-Y:2",
+  "expires_at": "2016-10-25T08:41:59.795Z",
+  "state": "prepared",
+  "timeline": {
+    "prepared_at": "2015-06-16T00:00:01.000Z"
```
clark800: In our implementation we changed "prepared_at" to "created_at" so that the same name makes sense for transfers with and without execution conditions.

0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1201 @@
...
+```
+
+#### Errors
+
+- [Unauthorized][]
+
+
+
+
+## WebSocket
```
clark800: Not sure if this will cover all the desired use cases, but in our implementation we just have an optional `notification_url` stored in each account and a notification is POSTed to this URL when there is activity on that account. This makes it really simple because there are no subscriptions to manage. The downside vs websockets is that the recipient has to have a server, but you could also make a separate microservice that provides the websockets if you want to keep the ledger minimal.

0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1201 @@
...
+
+A transfer object contains the fields from the following table. Some fields are _Server-provided_, meaning they cannot be set directly by clients. Fields that are "Optional" or "Server-provided" in the following table may be omitted by clients submitting transfer objects to the API, but **those fields are not optional to implement**. All fields, including the `memo` fields, must be implemented for the Common Ledger API to work properly. Fields of nested objects are indicated with a dot (`.`) character; no field names contain a dot literal.
+
+| Name                   | Type                 | Description                  |
+|:-----------------------|:---------------------|:-----------------------------|
+| `client_id`            | [UUID][]             | Client-provided UUID for this resource. MUST be unique within the ledger. |
+| `ledger`               | [URL][]              | Resource identifier for the ledger where the transfer occurs. MUST be an HTTP(S) URL where you can [get the ledger metadata][Get Ledger Metadata]. |
```
clark800: Why is this needed?

mDuo13: [answered above](https://github.com/interledger/rfcs/pull/125#discussion_r88204803).

0012-common-ledger-api/0012-common-ledger-api.md
clark800: It don't think this logically belongs to a transfer, so it seems awkward in terms of design.

It may also cause problems. For example, if the ledger is generating this value, it means the ledger has to know it's external hostname, which can be complicated if it is going through proxies or is changed or has multiple external names etc.

0012-common-ledger-api/0012-common-ledger-api.md
mDuo13: Personally I think it would be awesome if the ledger API didn't have to include a WebSocket component at all.

It's possible to do transfer notifications by serving Get Transfer through an http response that doesn't finish until the transfer is in a final state, but I'm not sure how you implement that. (I've seen it done in web-based games before -- the GET just sits there pending until the server has something to respond with.)

But that still doesn't resolve the question of how clients and connectors would communicate and authenticate each other for quoting.

0012-common-ledger-api/0012-common-ledger-api.md
mDuo13: The ledger already has to solve those problems in order to serve the transfer ID and fulfillment URLs, so getting rid of this field alone doesn't resolve that complexity.

I think it does belong in a transfer in the same way that many data structures in an object-oriented program contain a "parent" reference. It makes it clear at a glance what ledger this transfer is from. Imagine passing such a resource around in a system that interacts with multiple ledgers: between this field and the Get Ledger Metadata method, the transfer resource stands alone with no outside context or state necessary whatsoever.

0012-common-ledger-api/0012-common-ledger-api.md
clark800: My comment above answers that, no?

0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1201 @@
...
+Request:
+
+```json
+POST /transfers
+Content-Type: application/json
+
+{
+  "client_id": "3a2a1d9e-8640-4d2d-b06c-84f2cd613204",
+  "ledger": "http://usd-ledger.example",
+  "debit_account": "http://usd-ledger.example/accounts/alice",
```
clark800: I think this should just be "alice". This is coupling to a lot of unnecessary things. Like if the server switches from http to https it looks like this will cause problems. Or if a new version of the code switches the endpoint name from "accounts" to "account".

0012-common-ledger-api/0012-common-ledger-api.md
emschwartz: In an earlier iteration of the `five-bells-ledger` we were using RESThooks instead of Websockets. Switching to websockets was the single biggest speed boost we saw in end-to-end interledger payments. I wouldn't object to having the `notification_url` as an option too (only downside being another feature for implementers to include), but I would strongly oppose replacing websockets with a slower notification method (not counting WebRTC, MQTT or another real-time messaging protocol).

clark800: The performance improvement was due to removing the database accesses that were required for robust notifications. What I'm suggesting is non-robust HTTP POST notifications, which is what our implementation uses and it is faster than the five-bells-ledger with websockets.

0012-common-ledger-api/0012-common-ledger-api.md
adrianhopebailie: > I don't think this is a desirable goal

@clark800 - why do you say that? I would argue that we should be making the API simple from the perspective of consumers over implementors which suggests that providing context to any data is desirable.

0012-common-ledger-api/0012-common-ledger-api.md
adrianhopebailie: The approach taken in the W3C Webpayments API is to namespace the asset identifier and default to ISO4217 currency.

So my proposal is to make this a URI and for it to default to: `urn:iso:std:iso:4217`

0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1201 @@
...
+
+### Asset Resource
+[Asset resource]: #asset-resource
+
+There are many kinds of resources that can be tracked in a ledger. The Asset Resource describes the resource tracked in this ledger, so that client applications can display appropriate information to users. An asset resource has the following fields:
+
+| Field            | Value     | Description                                   |
+|:-----------------|:----------|:----------------------------------------------|
+| `type`           | String    | The type of asset. Currently, the only supported type is `iso4217-currency`. |
+| `code`           | String    | The currency code to represent this asset. For `iso4217-currency` assets, this MUST be a three-letter uppercase [ISO 4217](http://www.xe.com/iso4217.php) currency code. |
```
adrianhopebailie: Suggest using the following URL as reference: http://www.iso.org/iso/home/standards/currency_codes.htm

0012-common-ledger-api/0012-common-ledger-api.md
adrianhopebailie: The type of this member should be  URI and a value relative to the API endpoint should be allowed.

0012-common-ledger-api/0012-common-ledger-api.md
adrianhopebailie: By this logic we should change the API method to "Create Transfer" for consistency

0012-common-ledger-api/0012-common-ledger-api.md
clark800: Even if this saves a couple lines of code in the common use case, that doesn't necessarily mean it is simpler for consumers. I think it's best to design by principle and avoid the temptation to violate the principle to achieve small conveniences. In this case the principle is that a data structure should have a single purpose and only require the minimal amount of information needed to achieve that purpose. There are two purposes here: (1) execute a transfer on a ledger (2) specify a transfer in a way that can be communicated globally. The principle would suggest that the second would have a data structure like `{ledger: ledgerURL, transfer: transfer}`. If we collapse these two data structures into one, we force a burden on code (either client or server) that only cares about the first purpose, and also potentially cause confusion about why the "ledger" field needs to be there. It's hard to list all the disadvantages of "convenience hacks" and weigh them against the convenience because most of the disadvantages are in the higher-level design that it imposes vs the more obvious and low-level convenience, but high-level design is much more important than low-level convenience, which is why it's generally an easy decision to stick with the principles.

0012-common-ledger-api/0012-common-ledger-api.md
clark800: No reason?

0012-common-ledger-api/0012-common-ledger-api.md
adrianhopebailie: > Even if this saves a couple lines of code in the common use case, that doesn't necessarily mean it is simpler for consumers

This is not about lines of code it's about simplicity in understanding. 

> There are two purposes here: (1) execute a transfer on a ledger (2) specify a transfer in a way that can be communicated globally.

Your comment has me confused about what you're referring to because this thread was originally referring to the `ledger` field of a `Message Resource` (nothing to do with transfers).

Are we all talking about the same thing here?

clark800: I originally commented on the transfer object, but was pointed to this existing thread for messages, which is on the same topic. Same argument should apply to both.

adrianhopebailie: I think it's fairly common in a REST API to provide a resource property that is the base URL for the resource, especially in the case of resources that may be sent over something like websockets where the transport should be decoupled from the message content.

In fact if we're being strict about principles then the hypermedia principle of REST says we MUST have that property. That said, REST API design always becomes a compromise between convention and configuration. If we can assert that for this API the ledger URL for an account is always `../../` relative to the account URL (`https://ledger.example/accounts/alice`) then perhaps that's enough.

Finally, it seems very flaky to me to connect to `ws://ledger.example` and then get notifications with `from: alice` and `to: bob` as properties that I must figure out, based on the URL I connected to originally, actually resolve to `from: https://ledger.example/accounts/alice` and `to: https://ledger.example/accounts/bob`.

Perhaps the answer is to make it optional for the REST API calls and always present in websocket messages? (Not sure what behaviour you'd want to mandate though if the ledger URL property doesn't correspond to the URL of the resource itself, i.e. share a common base)

0012-common-ledger-api/0012-common-ledger-api.md
MatthewPhinney: I'd also be interested to know why this member should be a URI.

0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1181 @@
...
+| `timeline.executed_at` | [Date-Time][]        | _(Server-provided)_ Time when the transfer was originally executed. MUST appear if and only if `state` is `executed`. This time MUST be equal to or later than the `prepared_at` time. |
+| `timeline.prepared_at` | [Date-Time][]        | _(Server-provided)_ Time when the transfer was originally prepared. MUST appear, even if the transaction is unconditional. |
+| `timeline.rejected_at` | [Date-Time][]        | _(Server-provided)_ Time when the transfer was originally rejected. MUST appear if and only if `state` is `rejected`. This time MUST be equal to or later than the `prepared_at` time. |
+
+[UUID]: http://en.wikipedia.org/wiki/Universally_unique_identifier
+
+
+#### Client IDs
+
+The transfer's `client_id` field is one of the main ways of identifying the transfer; it is unique across the ledger, so implementations MAY use it as a primary unique key in a database. The client applications, not the server, generate the `client_id`, so the server SHOULD require that the field match the canonical form for a [UUID][], which is 32 lowercase hexadecimal digits, separated by hyphens into groups of 8-4-4-4-12.
```
sappenin: Can you clarify the intent of `client_id`?  

For example, for a single ILP Payment with transfers on `Ledger1` and `Ledger2`, is it intended that the `client_id` will be the same in both ledgers?  

I'm trying to figure out if this value is meant to correlate ILP payments across modules (i.e., like an ILP transaction id), or if this field has a different purpose.

0012-common-ledger-api/0012-common-ledger-api.md
MatthewPhinney: For transfers, what's wrong with `{ledger: ledgerURL, transfer: transfer}`?

Also, @mDuo13 you said the ledger solves the problem of knowing its external hostname when accessed through proxies, or when there are multiple external hostnames. How does it do this?

mDuo13: I didn't say it solves them, I said it has to solve them. I don't think the current implementation handles that, nor do I know how you'd handle it (maybe using some sort of config and/or looking at some metadata from the HTTP request). But for a ledger to work through a proxy or whatever, that's going to be necessary to serve any of the URL fields, including the ID and transfer fulfillment.

0012-common-ledger-api/0012-common-ledger-api.md
emschwartz: Good question. This field is only to ensure uniqueness within a single ledger. It is not meant as a way to correlate transfers across ledgers and actually should not be used that way.

The reason to use an id like this within one ledger is to prevent people from accidentally submitting the same transfer twice. If the transfer with that ID has already been submitted, POSTing it again will cause an error.

The `client_id` should not be kept the same across ledgers because of the potential for id squatting. If you are a connector that always uses the same id for the outgoing transfer as the incoming one, I can prevent you from executing the destination transfer if I know the source transfer's id (for example if I am the sender).

0012-common-ledger-api/0012-common-ledger-api.md
adrianhopebailie: The ledger is a resource identified by a URI.
The account is a resource that is relative to the ledger.

This is exactly the use case that URIs are designed for.

There is very well defined and deterministic logic for resolving absolute URIs from relative URIs given a base (the ledger) to solve for exactly the issues @clark800 raises such as changing the protocol.

If we are saying that a ledger MUST have an account endpoint URL or an account URL Template (RFC6570) then it makes sense for the accounts to be identified either by the variable that is substituted into the template or by a relative URL that is resolved against the account endpoint URL.

On the other hand if the only normative requirement is for a ledger to have a base URL then the account MUST be identified by a URL relative to the ledger URL `/accounts/alice` or an absolute URL.

0012-common-ledger-api/0012-common-ledger-api.md
adrianhopebailie: > The reason to use an id like this within one ledger is to prevent people from accidentally submitting the same transfer twice. If the transfer with that ID has already been submitted, POSTing it again will cause an error.

> The client_id should not be kept the same across ledgers because of the potential for id squatting. If you are a connector that always uses the same id for the outgoing transfer as the incoming one, I can prevent you from executing the destination transfer if I know the source transfer's id (for example if I am the sender).

There are some normative requirements there that we should document in the spec.

0012-common-ledger-api/0012-common-ledger-api.md
sappenin: Should the API maintain both values, allowing for a clear distinction between `created_at` and a `prepared_at`?  

Per RFC1, the term "Prepared" demands an escrowed payment, which is not accurate in an optimistic-mode transfer.  However, if we use the term `created_at`, but in some instances (e.g., Universal Transfer) we really mean "when this was prepared", it it could lead to confusion.  Likewise, choosing the value `prepared_at` when  we really mean "when this was created" (e.g., Optimistic Transfer), seems equally imprecise.

Along those lines, RFC1 talks about a `Proposed` state.  Would it make sense to allow transfers to be created in the `Proposed` state initially, and then (potentially at some point later) moved into the `Prepared` state, but only if its a conditional transfer?  

That would allow optimistic transfers to have a timeline like the following:

```
"timeline": {
  "created_at": "2015-06-16T00:00:01.000Z",
  "executed_at": "2015-06-16T00:00:02.000Z"
}
```

whereas conditional transfers could have a timeline like this:

```
"timeline": {
  "created_at": "2015-06-16T00:00:01.000Z",
  "prepared_at": "2015-06-16T00:00:01.000Z",
  "executed_at": "2015-06-16T00:00:02.000Z"
}
```

0012-common-ledger-api/0012-common-ledger-api.md
sappenin: What do you guys think about having the ledger generate this identifier value instead of the client?  

There are quite a few requirements for this identifier that should probably be the burden of the ledger since the ledger is the one mandating them -- e.g., unique across the ledger; conforming to a particular pattern that the ledger can use as a PK in the ledger's database, etc.  Plus, the point about preventing id-squatting seems like something that should be the responsibility of the ledger as opposed to the client (who might mess it up or be unable to enforce the requirement).

For my proposal to work, though, the API would need to do something like the following to facilitate the creation of a transfer, which makes it a bit more chatty, but IMHO makes the API more robust: 

1. POST the transfer payload (minus client_id) to `/transfers`
1. Receive the created transfer as a response, with a server-supplied transfer `id`.
1. When ready, execute a PUT to `/transfers/{transfer_id}` (or possibly a PATCH) to update the state of the transfer to indicate to the ledger that the transfer is ready to be acted upon (e.g., update the state from "proposed" to "prepared", or something like that).

While the second call requiring clients to overtly trigger the processing of a transfer is somewhat of a negative, this two-phase creation has the added benefit of allowing clients to operate upon the server-side state of a transfer using possibly multiple remote calls before actually having the ledger do anything with it.

emschwartz: This seems unnecessarily complicated to me. A trivial way for clients to avoid id squatting is to either use a completely random UUID or one that is deterministically generated from some detail about the interledger payment (as is _supposed_ to happen in the connector [here](https://github.com/interledgerjs/ilp-connector/blob/master/src/lib/route-builder.js#L148-L158)). Full HTTP round trips are relatively slow and we should avoid adding them wherever possible.

Whether or not the ledger uses the `client_id` as the database primary key, it shouldn't be difficult for ledgers using any standard database to enforce uniqueness on that value. A previous version of this spec said ledgers could optionally enforce `client_id` uniqueness on a per-account basis, but it seemed like having more options would just make implementing this spec more complicated.

0012-common-ledger-api/0012-common-ledger-api.md
emschwartz: I think having accounts identified by a relative URL is better than expecting all ledgers to have an account endpoint. Some ledgers might want to have an account aliasing or internal sub-accounting feature, in which case I might want you to send money to `/aliases/ceb8a630-2d9d-4ce2-ad81-72a862659ae8` instead of my real account `/accounts/1234567890`.

0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1201 @@
...
+[Asset resource]: #asset-resource
+
+There are many kinds of resources that can be tracked in a ledger. The Asset Resource describes the resource tracked in this ledger, so that client applications can display appropriate information to users. An asset resource has the following fields:
+
+| Field            | Value     | Description                                   |
+|:-----------------|:----------|:----------------------------------------------|
+| `type`           | String    | The type of asset. Currently, the only supported type is `iso4217-currency`. |
+| `code`           | String    | The currency code to represent this asset. For `iso4217-currency` assets, this MUST be a three-letter uppercase [ISO 4217](http://www.xe.com/iso4217.php) currency code. |
+| `symbol`         | String    | Symbol to use in user interfaces with amounts of this asset. For example, "$". |
+| `decimal_digits` | Number    | _(Optional)_ Typical number of decimal places to display for amounts of this asset. MUST be a non-negative integer. This MAY be different from the precision and scale used internally by the ledger, which are reported in the [Get Ledger Metadata][] field. |
```
sappenin: What are some scenarios where the `decimal_digits` attribute of an Asset would diverge from the metadata's `precision` and `scale` attributes?

0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1201 @@
+# Common Ledger API
+
+The Common Ledger API is a RESTful API served by a ledger (or an adapter), which provides functionality necessary for ILP compatibility. The Common Ledger API provides a single standard API that a ledger can serve in order to ease integration with other Interledger Protocol components and applications, such as the reference ILP Client and ILP Connector. This is not the only way a ledger can become ILP-enabled, but it provides a template that minimizes the integration work necessary for compatibility with ILP software.
+
+## Overview
+
+The Common Ledger API defines the following data types and structures:
+
+- [Transfer resource][]
+- [Account resource][]
```
diminator: From a blockchain perspective, the account resource belongs to the wallet, not the ledger itself.
Blockchain or append-only ledgers only rely on pubkeys.
I'm not saying we should throw this out of the door, but it seems to me to get this API working for both legacy payment ledgers and blockchain ledgers such as BTC, ETH, BDB, etc there would be the need of a proxy API relaying between the chain and the wallet.

Any thoughts?

0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1201 @@
...
+The ledger MAY support additional methods, including but not limited to:
+
+- Create or Update Account
+- Health check status
+- Look up transfers by account or [Crypto-Condition][]
+
+
+## Conventions
+
+The API is designed to be [RESTful](https://en.wikipedia.org/wiki/Representational_state_transfer), extensible, and modular. By design, every resource type contains links to other methods, so you can "naturally" discover other API paths by starting from any object. Clients SHOULD NOT hardcode any [URL][] other than the path to the [Get Ledger Metadata][] method. A ledger MAY host different objects and API methods at different hostnames or base URLs, as long as all client applications can dereference the URLs in the usual way.
```
earizon: I'm wondering whether gPRC will be a better approach in the long term. There are many benefits out of the box. Transparent binary serialization, cross-language support, interface (auto)documentation in .proto files, better performance. No need for "parallel" websockets to transmit async events.

0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1201 @@
...
+
+The Common Ledger API does not define a way to add or remove money from the ledger; the methods defined here always maintain a zero sum across all account balances. However, you can effectively add money to a ledger by transferring it from an "issuer" account whose balance is allowed to go negative.
+
+A transfer object contains the fields from the following table. Some fields are _Server-provided_, meaning they cannot be set directly by clients. Fields that are "Optional" or "Server-provided" in the following table may be omitted by clients submitting transfer objects to the API, but **those fields are not optional to implement**. All fields, including the `memo` fields, must be implemented for the Common Ledger API to work properly. Fields of nested objects are indicated with a dot (`.`) character; no field names contain a dot literal.
```
earizon: I interpret that "client" means in this context the software connecting to the ledger while "server" is the ILP-compliant ledger. I think Using "client" and "server" could cause trouble. Explicitly speaking  about "wallets" (the local GUI connecting to the ledger), "ledger" and "connector" will be safer. For example in a  websocket stream  the difference between the client and server vanishes once the bidirectional message channel is created.

0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1201 @@
...
+| `memo`                 | Object               | _(Optional)_ Arbitrary data provided by the `debit_account` for the `credit_account`. The connector stores the ILP Header in this field. |
+| `note_to_self`         | Object               | _(Optional)_ Arbitrary data provided by the `debit_account` for itself. This field MUST be hidden when not authenticated as the `debit_account` or an admin. |
+| `id`                   | [URL][]              | _(Server-provided)_ Primary resource identifier for this transfer. MUST be an HTTP(S) URL where you can [get the transfer resource][Get Transfer]. |
+| `fulfillment`          | [URL][]              | (Server-provided) Path to the fulfillment for this transfer. MUST be an HTTP(S) URL where the client can [submit the fulfillment][Submit Fulfillment] or [get the fulfillment][Get Transfer Fulfillment]. MUST be provided if and only if this transfer has an `execution_condition`. |
+| `transfer_rejection`   | [URL][]              | (Server-provided) Path where a client can [reject this transfer][Reject Transfer]. MUST be an HTTP(S) URL. |
```
earizon: Wouldn't  rejection_webhook  be more 'standard' (https://en.wikipedia.org/wiki/Webhook) ?

I don't have a global picture about how transfers can be rejected using such URL. Is there any RFCs explaining it?

 As I understand "cancelling" and "rejecting" are two different terms in ILP:

- "cancelling" means creating a cancellation fulfilment matching a cancellation crypto-condition. 

- rejecting means that at some point the execution fulfilment didn't match the execution condition or there was a timeout. 

DOUBT:  How this 'transfer_rejection' match in the ILP protocol. Does it means the (sender originator?) ledger must fulfil the cancellation fulfilment ? In such case transfer_cancellation (or cancellation_webhook) will be more appropiate.
(Actually I didn't find anything related in the RFCs, only the code in the five-bells implementation)

0012-common-ledger-api/0012-common-ledger-api.md
justmoon: The description for this field could be improved. `decimal_digits` refers to the **conventionally displayed** number of decimal places for a given asset. It's purely a hint for display purposes and is ignored by everyone but human-facing amount formatters. (By contrast, precision and scale affect that actual available precision on a ledger.)

For example, a USD ledger might use a precision of 15 and a scale of 6. So, it can represent amounts like USD 10.123456. However, US dollars are typically shown with a precision of cents, so `decimal_digits` would be 2. How a client exactly renders this is up to the client, but it might show any precision beyond the `decimal_digits` in a deemphasized way, e.g. 10.12(3456)

0012-common-ledger-api/0012-common-ledger-api.md
mDuo13: Is there a definition for the `iso:` namespace of URNs somewhere? Why does `iso` appear twice in that string?

That said, I'm in support of making the currency identifier a URI.

0012-common-ledger-api/0012-common-ledger-api.md
mDuo13: As agreed below, there's a valid use case for unauthenticated WebSocket connections.

0012-common-ledger-api/0012-common-ledger-api.md
mDuo13: The current five-bells-ledger implementation has a `proposed` state that matches this description. However, one of the goals of the Common Ledger API (when we started on this project) was to simplify the `proposed` state out of existence.

I think I'm going to go ahead with changing the remaining state to "created" though.

0012-common-ledger-api/0012-common-ledger-api.md
```diff
@@ -0,0 +1,1201 @@
...
+
+```json
+POST /messages
+Content-Type: application/json
+
+{
+  "from": "https://blue.ilpdemo.ripple.com/ledger/accounts/bob",
+  "to": "https://blue.ilpdemo.ripple.com/ledger/accounts/alice",
+  "ledger": "https://blue.ilpdemo.ripple.com/ledger",
+  "data": {
```
mDuo13: This question is tricky. The size of the message data depends on the encoding. Any limit I could think of, upper or lower, would be totally arbitrary. Maybe we should decide on a minimum limit in bytes and/or characters of UTF-8, chosen based on being able to convey any reasonable quote / ILP header data we expect to need?

0012-common-ledger-api/0012-common-ledger-api.md
mDuo13: I agree on using "ledger" instead of "server."

However, I think the term "wallet" tends to be rather confusing and misleading. People use the term "wallet" to mean "client application", "key pair", or "web service" somewhat interchangeably, which is a problem. I'm going to stick with "client application" for now.

0012-common-ledger-api/0012-common-ledger-api.md
mDuo13: I agree that blockchains would need a proxy server to host a Common Ledger API, which would sit between the client application (wallet) and the blockchain server / p2p network. I'm not sure there's a way around that.

The "account information" returned can be quite minimal. In a blockchain case, "Get Account" could use the pubkey as the `name` and then just provide the known balance that can be spent by that key pair.

Maybe we need a special value for balances that are unknown in order to support this. I'm not really sure.

0012-common-ledger-api/0012-common-ledger-api.md
mDuo13: This API doesn't have any concept of "cancelling" in the ILP sense.

Rejecting transfers uses the "Reject Transfer" method also defined in this document. The receiver of a transfer can reject a conditional transfer if the execution condition hasn't been fulfilled yet.

0012-common-ledger-api/0012-common-ledger-api.md
diminator: Ok, we can split up the proxy requirements to somewhere else?
Maybe those will be more implementation specific - hence not RFC-ed?

mDuo13: I don't see how you'd split it off. The whole REST API would probably be served by some "proxy" server (not as in HTTP proxy).

0012-common-ledger-api/0012-common-ledger-api.md
mDuo13: I'd be willing to change it. The rationale behind the suggested behavior is that the rejection reason is generally represented as one field of the transfer object. The request here is the only case where the rejection reason appears outside of the transfer object. We can keep the number of returned resources low if we return the whole transfer object. This also has the added benefit of providing the full context and updated state of the transfer right away, which makes one usage pattern (which I expect to be common) more efficient -- the case where you get the updated transfer state immediately after rejecting a transfer. I think that's likely to happen a lot because apps may want to show an updated date/status thing for the transfer. If it can save the response to "Reject Transfer" then that saves us a full round trip.

Any further opinions one way or the other on this?

earizon:
In the five-bells-ledger and ilp-plugin-bells there is support for cancellation condition  and cancellation fulfillments (https://github.com/interledgerjs/ilp-plugin-bells/blob/master/src/lib/plugin.js, https://github.com/interledgerjs/five-bells-ledger/blob/master/src/models/transfers.js). I don't see any mention about them in this rfc or the 0004 either. 

Has cancellation been  removed, or is it missing on the rfcs ?

If it hasn't, the final state of a cancelled transfer is also "rejected" or must it be "cancelled"?

justmoon:
> Has cancellation been removed, or is it missing on the rfcs ?

Cancellation and a few other five-bells-ledger features, such as multi-debit/credit turned out not to be needed for ILP, so they aren't part of the ILP ledger adapter API (IL-RFC-4).

We expect that many ledgers will support more than the minimum feature set, but the Common Ledger API is intended to be pretty close to the minimum required functionality, so those optional features are also excluded here.

0012-common-ledger-api/0012-common-ledger-api.md
earizon: I was playing with an alternative solution to websocket based on gRPC. First thing I needed was to abstract all the  "websocket subscription and message stuff" as a bidirectional asynchronous stream. 

An early *proto defining the service is available at:
https://github.com/everis-innolab/ILP_gRPC_API/blob/master/ilp_ledger_subscriptions.proto

A comparative about a different alternative solutions for bidirectional async channels:

\  | websocket+JSON/RPC 2.0 | Client request + Client pooling | protobuf + gRPC | web-hooks
-- | ---------------------- | ------------------------------- | -------------   | ---------
Notes| - Current five-bells implementation - Once the ws connection is stablished client and ledger become peers.|- Simplest solution.  - Messages from client to ledger are sent in HTTP requests.  Messages from ledger to client are pulled during periodic pooling requests.|- bidirectional async stream are created with: streamN(stream msg) returns (stream msg) |both ledger and client must provide a callback URL for peers messages/events
web-browser support|YES|YES| "WORK IN PROGRESS" github.com/grpc/grpc-experiments/issues/159 github.com/grpc-ecosystem/grpc-gateway |NO Android support |YES|YES|YES|YES
iOS support|YES|YES|YES|YES
Standard|YES (IETF RFC 6455)|NO|"Open-Source / IETF Draft"|NO
Works on 3G, behind proxies/firewalls,...| "SOMETIMES" |YES|YES/NO/SOMETIMES?| NO (both peers require IP visibility to expose REST web-hooks)
automatic mapping object to/from network data"|PARTIAL and IN SOME LANGUAJES|NO|YES|NO
auto-documented |NO|NO| YES. Using *proto files ( *proto can also be used as ref doc.  for non-gRPC implementations ) |NO
cross-language support |NO|NO|YES|NO
speed|GOOD|SLOW |VERY GOOD|UNDEFINED
"deprecated" support (backward compatibility)|MANUAL|MANUAL|YES|MANUAL
PREFERED FOR| CONNECTORS & SERVER APPS | SORT-LIVE CONNECTIONS WHERE MOBILE CLIENTS AND RANDOM NETWORK SETUPS ARE EXPECTED. (""WALLETS"") | CONNECTORS & SERVER APPS | CONNECTORS & SERVER APPS IF WEBSOCKETS AND gRPC ARE NOT AVAILABLE. 


I'm also thinking whether it could be convenient to split this API in services for light clients ("wallets" that just connect time to time to the ledger posting new prepared Transfers and checking payments), server applications ("e-comerce" with permanent tcp/ip connections waiting for payments and fulfilling transfer) and connectors. Also many simultaneous "wallet" connections to ledger could be expected in some setups while only a very limited set of ilp-connectors and server apps ("e-commerce"). If I'm not wrong there is no distinction amongst them right now. The API treats them all equally.

sappenin: Great write-up of different channels @earizon!  I've been doing similar research as I implement a Java ledger conforming to this Common Ledger API, and have been wanting to experiment with gRPC as an alternative to Websockets.  I think once browsers fully support everything they need to ([here's](https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-WEB.md) one such proposal) then Websockets will probably make less sense in the browser (IMHO).

Your proposal for API separation is an interesting one.  In the short-term, it seems like the Websocket channel would be the minimum thing to support so that Browsers can connect, with gRPC maybe being an optional choice?  Or do you have a different idea?

earizon: @sappenin Yes I think Websocket can be considered required all (web/non-web) clients. But I'm not 100% sure that it will work in all network environments due to HTTP proxies. See for example [problems with whatsapp web using wss behind proxies @ google](https://www.google.com/search?channel=fs&q=whataspp+web+behind+proxy). A fall-back to client pooling could be useful too. It's clearly a suboptimal solution in terms of performance, but for clients that just want to connect "time to time" to make some random payment it could be useful. There are also [some other problems](https://deepstream.io/blog/load-balancing-websocket-connections/) when using permanent permanent tcp/ip connections and the number of clients start to grow.

adrianhopebailie: @sappenin and @earizon I think the core Java stuff is stabalizing to the point that we should definitely try a second adaptor implementation.

Could we try implementing a gRPC based interface between the client and your ledger?

You'd implement the org.interledger.ilp.ledger.LedgerAdaptor interface and then should be able to simply instantiate a Client with your adaptor instead of the REST adaptor.

earizon: @adrianhopebailie great news!

Can anyone confirm next claims (since it's not 100% clear to me from the spec). Using gRPC syntax I was thinking on:
```
    rpc setupLedger2ClientUnidirectionalAsyncTransferChannel(stream spec.ILPLocalTransfer) returns ( Empty );
    rpc setupBidirectionalAsyncMessagesChannel              (stream spec.ILPMessage)       returns ( stream spec.ILPMessage );
```
In common language:
- It's expected that the ledger will send transfer notifications to any ledger-client directly subscribed to a transfer or indirectly subscribed to any transfer affecting the account (as creditor or debitor).

- It's expected that any ledger or ilp-connector-ledger-client will send messages to its peer, to emit or receive events (those events are used to propagate routing information). ( Is there any use-case when an non-ilp-connector will need to listen for message events? )

mDuo13:
Rebased this onto the latest **master** and updated the intro to reflect the new intended goal (formalize the existing 5BL without making all the cosmetic / nice-to-have changes).

mDuo13:
Closed this in favor of #237. (Same contents, new thread.)

PR #129, opened 2016-12-07
docs(0004): plugin#isConnected checks for active websocket

sentientwaffle:
Related to https://github.com/interledgerjs/ilp-connector/issues/248

### DIFF
--- a/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
+++ b/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
@@ -121,7 +121,7 @@ Unsubscribe from ledger events.
 #### isConnected
 <code>ledgerPlugin.isConnected() ⇒ Boolean</code>
 
-Query whether the plugin is currently connected.
+Query whether the plugin is currently connected and has an active websocket.
 
 #### getInfo
 <code>ledgerPlugin.getInfo() ⇒ Promise.&lt;[LedgerInfo](#class-ledgerinfo)></code>


0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
```diff
@@ -123,6 +124,11 @@ Unsubscribe from ledger events.
 
 Query whether the plugin is currently connected.
 
+#### isOpen
+<code>ledgerPlugin.isOpen() ⇒ Boolean</code>
+
+Query whether the plugin is currently connected and has an active websocket.
```
emschwartz: Isn't this what `isConnected` is for? If plugin bells isn't currently giving useful information when that it called, the implementation of that method should be changed rather than adding a new one to the interface.

sentientwaffle: They are different. Example:

* Initially: `connected=false` and `open=false`.
* When `plugin.connect()` is called successfully, `connected=true` and `open=true`
* If some time later the websocket connection drops, `connection=true` but `open=false`.
* The plugin will try to reestablish the websocket, and when it succeeds, `connection=true` and `open=true`.
* When `disconnect()` if called, `connection=false` and `open=false`.

`isConnected()` is used to check that functions like `getPrefix()` and `getAccount()` can be called, because the first `connect()` call sets up the urls & metadata that the plugin needs.

`isOpen()` will be used by the connector for https://github.com/interledgerjs/ilp-connector/issues/248 -- i.e. to allow the connector to refuse to quote or settle payments without an _active_ websocket.

Perhaps the names should be changed to reduce confusion? `s/isConnected/isReady` and `s/isOpen/isConnected`, or similar. Though, that would be a breaking change.

emschwartz: Gotcha.

> Perhaps the names should be changed to reduce confusion? s/isConnected/isReady and s/isOpen/isConnected, or similar.

That would make most sense to me. It would be annoying to have to make that breaking change, and to have yet another method on this interface.

What if `isConnected` tells you whether it's connected right now and the methods like `getPrefix` reject if the plugin hasn't ever been connected? You wouldn't be able to check beforehand if the plugin has ever connected, but you would probably just call `plugin.connect().then(() => plugin.getPrefix())` anyway.

sentientwaffle: Make sense; updated.

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
```diff
@@ -121,7 +121,7 @@ Unsubscribe from ledger events.
 #### isConnected
 <code>ledgerPlugin.isConnected() ⇒ Boolean</code>
 
-Query whether the plugin is currently connected.
+Query whether the plugin is currently connected and has an active websocket.
```
emschwartz: Not all plugins will use websockets to connect

sentientwaffle:
Closing because the current wording works.

Issue #106, opened 2016-10-10
RFC-2: Fixed-size OER encoding of Condition


nbougalis:
Continuing from #105, having a `Condition` have a fixed-size binary encoding is a good idea. Assuming #105 is adopted, the `fingerprint` and `featureBitmask` fields use variable-length encoding.

We should consider imposing a limit on the number of "features" available. For example, we could allow up to 32 features.

Currently, we have five features: sha256, preimage, prefix, threshold, rsa_pss and ed25519. With a 4 byte unsigned int, we still have space for `(32-5) = 27` additional features. Alternatively, we could just make the field 64 bits, allowing us `(64-5) = 59` additional features. We can gain one extra bit by removing the `sha256` feature and defining that algorithm as always implicitly available.

Second, we should consider making the `fingerprint` field a fixed sized, instead of a variable-length octet string. The specification already includes SHA256, which, conveniently, outputs 32 bytes which is the current size of a fulfillment. So we could just define conditions to be exactly 32 bytes.

Potential problems with that is that SHA256 has shown some weaknesses. Our construct is fine for now, but we should be conservative. We could go for SHA512, doubling the number of bits available.

Doing so means that our hashlocks won't be directly compatible with Bitcoin's `OP_HASH256` and will increase the size of a condition, which are both things to consider; on the plus side any implementation which has SHA256 available should, also have SHA512 since they're in the same SHA-2 family.

adrianhopebailie:
As mentioned in #105 wouldn't adding a version allow us to be more rigid. i.e. Fix the bit-mask to 1 byte and the fingerprint to 32.

The cost is a single byte at the beginning that defaults to 0x00 and gives us 255 future versions.

justmoon:
In collaboration with @nbougalis and @adrianhopebailie, we ended up using BIT STRING encoding which is variable length. However, simple conditions (those that can't have subconditions) no longer contain a subtypes field at all and are therefore constant size.

Issue #92, opened 2016-08-30
RFC-2: Encoding of conditions and fulfillments


adrianhopebailie:
This proposal follows the discussion in #78 and on the community call on 24 August.

Crypto-conditions and fulfillments are data structures. They must be encoded and decoded between standardized forms such that they can be used in a deterministic fashion.

The crypto-condition and fulfillment data structures are defined in ASN.1 which means that the same structure can be encoded and decoded in different forms for different uses and translated between different forms and encodings.
## Requirements

**Canonical Form** 

In RFC-2 we should provide a canonical cross-platform encoding that can be used for transmission on the wire and machine to machine communications.

We have defined the OER binary encoding as the canonical form.

**Developer Friendly**

It is also useful to provide encodings that are human-readable for ease of use by developers. This is best served by JSON which is the de-facto standard for data encoding on the Web.

**Compact but Human Readable**

Much like the fingerprint of a public key it is valuable to have a concise text value that can be used to easily perform visual matching and comparison.

There is also some value in representing a condition as a resource and therefor having a URI encoding. The URI encoding of a fulfillment is less useful but is also provided for completeness.
## Conditions vs Fulfillments

Conditions have the benefit of being fixed size and are, in many respects, the equivalent of a public key fingerprint. Fulfillments on the other hand are far more complex and can be of almost any size.

Therefor it makes sense to define different encodings for conditions, especially in the URI form, but not to try and replicate these for fulfillments. That is not to say they couldn't be defined but they are unlikely to be used.
## Mixing Encodings

Since the encodings proposed here take the same shape it's possible to mix binary encoded fields into a string encoding.

In this proposal we use BASE16 (hex) when string encoding a raw binary field value and BASE64URL when the underlying data is OER encoded (i.e. including length indicators and sub-fields).

Since the data structure of fulfillments are infinitely recursive it's possible that, rather than fully expanding a payload into all of it's parts and string encoding each of these, users may wish to simply BASE64URL encode the OER encoded binary form of the payload.
## Types

Since the text encoding is intended to be easily human readable it makes sense for the type to also be easy to read. Therefor the following constants are defined for the condition types:
- preimage-sha-256
- prefix-sha-256
- threshold-sha-256
- rsa-sha-256
- ed25519
## Features

As with the types, it makes sense to list the features of a condition as human-readable strings. It's a debatable if these should be used in the URI encoding as they are quite verbose so in this proposal they are only used in the JSON encoding.
- sha-256
- preimage
- prefix
[...]

Issue #87, opened 2016-08-22
RFC-2: Replace weighted threshold conditions with regular (unweighted) thresholds


justmoon:
Having support for weighted thresholds is adding significant complexity. Most signficantly, it requires implementations to solve a version of the 0-1 knapsack problem everytime they calculate the `maxFulfillmentLength` of a threshold condition. To make matters worse, the algorithm has a complexity of _O(nW)_ where `n` is the number of items and `W` is the threshold. Most practical values may be fine, but it's still a possible DoS vector.

The use cases for weighted thresholds are pretty thin and most of them can likely be served by transforming the weighted threshold circuit into an unweighted one. Also #86 might be good enough for small weights, since we could just refer to the same condition multiple times.

adrianhopebailie:
Is the proposal to add a new type called `WEIGHTED-THRESHOLD-SHA-256` which is the same as the current `THRESHOLD-SHA-256` and then drop the weights from the current definition?

It's likely that few people will implement `WEIGHTED-THRESHOLD-SHA-256` if they are overly complex and unlikely to be used but that might just be for the ILP use case?

justmoon:
Yes, although I'd recommend that we do not define `WEIGHTED-THRESHOLD-SHA-256` at this stage, but rather wait for a need to arise. I.e. somebody to show up who has a use case for it. If that never happens, so be it.

justmoon:
Implemented in draft-02.

PR #131, opened 2016-12-13
Updated draft-02

### LINKED ISSUE #115: RFC-2: Are PREIMAGE and PREFIX FeatureSuites necessary?
There is an argument to be made that for an implementation to support these features they need to explicitly support the corresponding condition types: PREIMAGE-SHA-256 and PREFIX-SHA-256.

Neither requires support for any cryptographic functions beyond SHA-256 so we should either drop them OR consider merging the idea of features and types by:
1. Assigning condition types values that can be combined as flags  
   (i.e. 1,2,4,8... as opposed to 1,2,3,4...)
2. Changing the FeatureSuites field in the Condition to simply be the set of types of any subconditions.
3. Consider only using this field for "compound" types: PREFIX-SHA-256 and THRESHOLD-SHA-256

### LINKED ISSUE #120: RFC-2: Typos
Looking at https://tools.ietf.org/html/draft-thomas-crypto-conditions-01#section-2.4.2 and https://tools.ietf.org/html/draft-thomas-crypto-conditions-01#appendix-C I see:

```
ConditionType ::= INTEGER {
    preimageSha256(0),
    rsaSha256(1),
    prefixSha256(2),
    thresholdSha256(3),
    ed25519(4)
} (0..65535)
```

Those values are incorrect. Elsewhere, in the same document, we _correctly_ defines the type IDs. Specifically, in sections 4.2.1, 4.3.1 and 4.4.1 we define `prefixSha256` as type 1, `thresholdSha256` as type 2 and `rsaSha256` as type 3.  The reference implementation also uses the correct sequence.

adrianhopebailie:
Closed previous PR as there were too many comments to track properly.
All should have been addressed.
Some major changes based on feedback from SAAG at IETF Seoul

Sumary of changes:
* Changed to DER encoding from OER
* Changed string encoding to be simpler
* Dropped featurebitmask in favour of subtypes
* Dropped weights from threshold type
* Changed ASN.1 defs to:
  * use CHOICE for unions
  * use BIT STRING for subtypes bitmap
  * reference existing types like RSAPublicKey
* Re-wrote abstract, intro and others to clarify some concepts

Fixes #120
Makes #106 redundant and resolves #115 as we have dropped Features
Addresses #92 and #87


0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -110,508 +115,638 @@ informative:
...
+The binary encoding of conditions differs based on their type. All types define at least a fingerprint and maxFulfillmentLength sub-field. Some types, such as the compound condition types, define additional sub-fields that are required to convey essential properties of the crypto-condition (such as the sub-types used by sub-conditions in the case of the compound types).
 
-    cf:0:VGhlIG9ubHkgYmFzaXMgZm9yIGdvb2QgU29jaWV0eSBpcyB1bmxpbWl0ZWQgY3JlZGl0LuKAlE9zY2FyIFdpbGRl
+Each crypto-condition type has a type ID. The list of known types is the IANA-maintained [Crypto-Condition Type Registry](#crypto-conditions-type-registry).
 
+Conditions are encoded as follows:
 
-The example has the following attributes:
+    Condition ::= CHOICE {
+	  preimageSha256 SimpleCondition,
```
stevenroose: I thought the options in a CHOICE must have distinct tags? (See [Section 5.5 here](http://luca.ntop.org/Teaching/Appunti/asn1.html) f.e.). It doesn't matter too much though, this notation makes it clear what option to pick for what condition type.

stevenroose: It seems like tag identifiers for the choices are not specified though.. Not for `SimpleCondition` and `CompoundCondition` but also not for the fulfillments..

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -110,508 +115,638 @@ informative:
...
-Example fulfillment:
+### Example {#rsa-sha-256-example}
 
-    cf:3:gYCzDnqTh4O6v4NoUP9J4U-H4_ktXEbjP-yj5PCyI1hYCxF2WZX0uO6n-0cSwuHjFvf3dalT0jIhahadmmTdwAcSCkALN_KvwHe2L-ME3nTeahGexAdrUpxPYJawuq1PUz3wFzubgi_YXWX6S--pLY9ST2nLygE2vYDQlcFprsDglYGAjQM0-Z5B-953uQtJ5dXL1D5TWpM0s0eFF0Zty7J2Y3Nb0PqsR5I47a2wYlA7-106vjC8gHFdHVeSR6JksSrhj8YaMWfV0A6qhPz6hq-TqSKCXd4mf3eCpyyFYR_EyH5zXd56sJEU3snWlFbB_bKAW4si_qdfY9dT87YGUp_Grm0
+TODO
 
+## ED25519 {#ed25519-condition-type}
+ED25519 is assigned the type ID 4. It relies on the SHA-512 digest algorithm and the ED25519 signature scheme as the condition fingerprint is not a digest.
 
+The exact algorithm and encodings used for the public key and signature are defined in [I-D.irtf-cfrg-eddsa](#I-D.irtf-cfrg-eddsa) as Ed25519. SHA-512 is used as the hashing function for this signature scheme.
```
tonychain: Note that this draft should shortly be published as RFC 8032 (presently in IETF AUTH48 state, approved by one author but waiting for approval from another):

https://www.rfc-editor.org/auth48/rfc8032https://www.rfc-editor.org/auth48/rfc8032

0002-crypto-conditions/0002-crypto-conditions.md
adrianhopebailie: The tags are assumed to be the index if not provided.
I can add them but it won't affect the encoding.

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -110,508 +115,638 @@ informative:
...
+	  preimage PreimageFulfillment ,
+	  prefix PrefixFulfillment,
+      threshold ThresholdFulfillment,
+	  rsaSha256 RsaSha256Fulfillment,
+	  ed25519 Ed25519Fulfillment
+    }
 
-# Condition Types
-The following condition types are defined in this version of the specification. Future versions of this spec MAY introduce new feature suites and condition types, which SHALL be registered in the IANA maintained [Crypto-Condition Type Registry](#crypto-conditions-type-registry).
+    PreimageFulfillment ::= OCTET STRING
```
stevenroose: I would really like to ask to keep every fulfillment a SEQUENCE. It makes it a lot harder to encode this as just an octet string instead of a SEQUENCE with a single element.

stevenroose: F.e. Go, and I believe other languages to, allows to map native structs into DER sequences. F.e. this object here, would be automatically encoded into the SEQUENCE structure. But changing it to just an octet string would make things a lot harder.

https://github.com/stevenroose/cryptoconditions/blob/develop/ffpreimagesha256.go#L10

stevenroose:
Oh yeah, we spoke about SET's. See [Section 5.15](http://luca.ntop.org/Teaching/Appunti/asn1.html) of the Layman's Guide. SET means 1 or more, SET OF means zero or more. Also, and more importantly, DER encoding indeed specifies a sort order for elements in the set, namely lexicographic ordering just like the previous spec did.

adrianhopebailie:
SET OF is a problem because it can only contain elements of different types so we'll need to use SEQUENCE OF.

stevenroose:
```
ThresholdSha256FingerprintContents ::= SEQUENCE {
  threshold INTEGER (0..65535),
  subconditions SEQUENCE OF Condition
}
```

That section of the spec also mentions Condition sorting. If you use `SET OF` or `SET` (because it's one or more, I guess) there as well, it automatically sorts the conditions by encoding:

> DER encoding. Constructed. Contents octets are the same as for the BER encoding, except that there is an order, namely ascending lexicographic order of BER encoding. Lexicographic comparison of two different BER encodings is done as follows: Logically pad the shorter BER encoding after the last octet with dummy octets that are smaller in value than any normal octet. Scan the BER encodings from left to right until a difference is found. The smaller-valued BER encoding is the one with the smaller-valued octet at the point of difference.

adrianhopebailie:
This should be the final revision based on all feedback. Let me know if there are any critical issues.

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -110,662 +116,889 @@ informative:
...
+      subconditions        SET OF Condition
     }
 
-    ThresholdSubfulfillment ::= SEQUENCE {
-      weight INTEGER (0..4294967295) DEFAULT 1,
-      condition Condition OPTIONAL,
-      fulfillment Fulfillment OPTIONAL
+    -- Fulfillment 
+    ThresholdFulfillment ::= SEQUENCE {
+      subfulfillments      SEQUENCE OF Fulfillment,
```
justmoon: Let's change these to `SET OF`.

Set-of defines a canonical ordering which makes testing easier. It seemed like a good idea to have leniency here, but it actually just creates ambiguity. If we are going to use DER, we shouldn't use sequence-of when we actually don't care about the order.

See also: https://github.com/interledger/crypto-conditions/commit/7a6fce5853af22508c6f97ebe642f5cf04f5f887

stevenroose: @justmoon I'm not sure about if SET OF allows duplicates. It needs to be possible to add duplicates in order to give conditions a higher weight.

adrianhopebailie: @stevenroose I don't think SET OF explicitly disallows duplicates per ASN.1 I think the name "set" is misleading in this instance.

@justmoon Should we hold on this change until we can confirm this won't be a problem?

stevenroose: Yeah I couldn't find anything on it either, not an explicit mention of either. I guess I just assumed it, because that's how SETs are interpreted in a lot of programming languages. I agree that it's misleading.

justmoon: It may be misleadingly named, but there is no ambiguity. X.680 clearly says:

> **3.8.68 sequence-of types** Types defined by referencing a single component type; each value in the sequence-of type is an ordered list of zero, one or more values of the component type.
> [...]
> **3.8.73 set-of types** Types defined by referencing a single component type; each value in the set-of type is an unordered list of zero, one or more values of the component type.

A list is unambiguously something that can contain duplicate entries, so it's perfectly clear that this is allowed. Note also that the only difference in wording between `sequence-of` and `set-of` is the prefix *un*ordered, so if you're worried about sets not allowing duplicate entries, switching to a sequence wouldn't help.

X.690 says to encode (BER) a set-of exactly like a sequence-of, except that the ordering may or may not be preserved:

> **8.12 Encoding of a set-of value**
> **8.12.1** The encoding of a set-of value shall be constructed.
> **8.12.2** The text of 8.10.2 applies.
> **8.12.3** The order of data values need not be preserved by the encoding and subsequent decoding.

DER adds one more rule:

> **11.6 Set-of components**
> The encodings of the component values of a set-of value shall appear in ascending order, the encodings being compared as octet strings with the shorter components being padded at their trailing end with 0-octets.
> *NOTE – The padding octets are for comparison purposes only and do not appear in the encodings*

`SEQUENCE OF` is the **wrong** since we don't care about the order. I noticed the mistake while writing the test suite. The test suite was generating the exact same binary as the implementation-under-test in all cases except one: They would sometimes differ on the ordering of the subfulfillments and subconditions in an encoded threshold fulfillment. This is precisely why `SET OF` is in the standard. When you don't care about the ordering, `SEQUENCE OF` is a source of non-canonicality and `SET OF` fixes it by sorting the items.

justmoon: Looked at couple of implementations:

ffasn1dump (C): allows duplicates
asn1step (C++): allows duplicates
asn1.js (JS): allows duplicates
asn1crypto (Python): [allows duplicates](https://github.com/wbond/asn1crypto/blob/master/docs/universal_types.md#setof)
asn1 (Go): [allows duplicates](https://golang.org/pkg/encoding/asn1/)
BouncyCastle (Java): [allows duplicates](http://javadox.com/org.bouncycastle/bcprov-jdk15on/1.51/org/bouncycastle/asn1/ASN1Set.html)
PHPASN1 (PHP): [allows duplicates](https://github.com/fgrosse/PHPASN1/blob/master/lib/ASN1/Universal/Set.php)

0002-crypto-conditions/0002-crypto-conditions.md
```diff
@@ -110,662 +116,889 @@ informative:
...
+: An octet string representing the RSA signature on the message M.
 
-: Implementations MUST verify that the signature and modulus consist of the same number of octets and that the signature is numerically less than the modulus.
+: Implementations MUST verify that the signature is numerically less than the modulus.
 
-The message to be signed is provided separately. If no message is provided, the message is assumed to be an octet string of length zero.
+Note that the message that has been signed is provided separately. If no message is provided, the message is assumed to be an octet string of length zero.
 
-### Implementation {#rsa-sha-256-condition-type-implementation}
-The recommended modulus size as of 2016 is
```
libscott: I can't see where `ED25519-SHA-256` is depending on SHA-512. Am I missing something?

libscott: As far as I can see, this is the only hash involved in `ED25519-SHA-256`, so `Ed25519Sha512Fulfillment` doesn't make alot of sense to me.

libscott: I can't see where this condition type is using SHA-512, am i missing something?

libscott:
https://github.com/interledger/rfcs/pull/131/files#diff-d1eaed96e7b264dd2619e1cc1a71ab20R781

I can't see where SHA-512 is used in the Ed25519 condition type. It appears to use SHA-256(public_key) as the fingerprint, but no other hash function. Am I missing something?

libscott:
One more question, the preimage-sha256 cost is constant at `12`, but the ed25519-sha256 is all the way up to `131072`. I read that the cost is designed to "increase rapidly", but one number seems conspicuously out of proportion to the other, seeing as they are both constant size fulfillments with constant resource requirements.

tonychain:
> I can't see where SHA-512 is used in the Ed25519 condition type. It appears to use SHA-256(public_key) as the fingerprint, but no other hash function.

That confuses me as well. If this is for a prehash prior to Ed25519, I would suggest using the Ed25519ph IUF mode as specified in RFC-to-be 8032 (draft-irtf-cfrg-eddsa).

justmoon:
> I can't see where SHA-512 is used in the Ed25519 condition type. It appears to use SHA-256(public_key) as the fingerprint, but no other hash function. Am I missing something?

SHA-512 is used internally in Ed25519 to hash the message together with some other fields. It is also used to derive the secondary private values from the input private seed. (This is all as per the Ed25519 spec.) I agree that if this spec mentions SHA-512 at all, it should be clear about what it is referring to.

----

> Is there a reasoning for the ed25519-sha256 cost to be so high, or not returned as the byte count of the fulfillment like preimage-sha256?

First, some background: The cost feature is intended to limit the possibility of a denial-of-service attack against a system that authenticates requests using crypto-conditions. You want implementations to be able to set a cost limit such that legitimate use cases fall within the cost limit, while making it difficult to craft a fulfillment that falls within the limit but still creates an exorbitant amount of work.

Ideally, we would have a cost metric that perfectly matches the actual amount of work that an implementation has to do. However, that's impossible, if only because the exact resources required for verification will vary from implementation to implementation. Also the relative cost of memory vs CPU vs network bandwidth will vary and the cost should capture all of them. We are further constrained in that we want implementations to be simple - that's important for a variety of reasons, among them the fact that simpler implementations are easier to write with fewer bugs, which helps security. In summary, we have to devise some rough approximation that is both simple, yet still captures the true cost well enough to deter DoS attacks. 

Now to answer your question: The cost of RSA and Ed25519 conditions is so high because they require so much CPU work to verify. Based on our benchmarks of the different cryptographic operations (using `openssl speed` and [curvebench](https://github.com/justmoon/curvebench)), we calculated that an Ed25519 condition would take about the same amount of CPU time (plus/minus an order of magnitude) as a preimage condition with a ~150kB preimage (cost 150000). We rounded it to the nearest power of two (2^17). Ideally, if somebody else tried to derive the same constants by performing the same benchmarks and then rounding to the nearest power of two, they'd land on the same result. Obviously, if there was a huge advance in either SHA-256 or Ed25519 performance that would topple that balance. Also, I'm sure there are platforms where the balance between SHA-256 and Ed25519 is very different than it is on x86-64 (which is where we tested.) I don't see a good way to avoid those limitations and keep in mind that the cost only needs to be a rough approximation.

----

> That confuses me as well. If this is for a prehash prior to Ed25519, I would suggest using the Ed25519ph IUF mode as specified in RFC-to-be 8032 (draft-irtf-cfrg-eddsa).

It's not using Ed25519ph currently. Ed25519 (no ph) still uses SHA-512. The I-D advises against using the Ed25519ph variant:

> The Ed25519ph and Ed448ph variants are prehashed. This is mainly useful for inter-operation with legacy APIs [...] These variants SHOULD NOT be used.

It would give us a slight performance and DoS protection advantage to use Ed25519ph. But I agree with the I-D authors that it would be weaker overall.

Regarding the DoS issues - you could try to attack an implementation of the current spec by creating a `PREFIX([100000 byte string], THRESHOLD([100 ED25519 conditions]))`, the idea being that the hashing (Ed25519 actually hashes the message twice) will start to outweigh the cost of the signature verification. It would require some work to figure out if this attack is viable in practice. It's definitely a lot weaker than other attacks that are already being addressed with the latest version of the spec.

0002-crypto-conditions/0002-crypto-conditions.md
adrianhopebailie: Fixed in latest commit

adrianhopebailie:
Note that the CHOICE labels reflect the crypto-condition "type" which is an indicator of the algorithm/structure and the hashing function used to generate the fingerprint. These are also the same for both the condition and fulfillment.

The type names, on the other-hand, are more specific and differ between the conditions and fulfillments. E.g. Note the name of the ED25519 fulfillment is `Ed25519Sha512Fulfillment`.

```
Condition ::= CHOICE {
  preimageSha256   [0] SimpleSha256Condition,
  prefixSha256     [1] CompoundSha256Condition,
  thresholdSha256  [2] CompoundSha256Condition,
  rsaSha256        [3] SimpleSha256Condition,
  ed25519Sha256    [4] SimpleSha256Condition
}

Fulfillment ::= CHOICE {
  preimageSha256   [0] PreimageFulfillment ,
  prefixSha256     [1] PrefixFulfillment,
  thresholdSha256  [2] ThresholdFulfillment,
  rsaSha256        [3] RsaSha256Fulfillment,
  ed25519Sha256    [4] Ed25519Sha512Fulfillment
}
```

stevenroose:
Are the fingerprint contents supposed to be SEQUENCEs?

PR #140, opened 2017-01-03
Earizon seq diagram0004

earizon:
I have some trouble understanding the RFC 0004 for the ilp plugin. 

I created a sequence diagram (source in plantuml) to clarify ideas based on the RFC and the ilp-connector JS code.

Since it could be interesting for other people I leave it here as "pull request" for your consideration.

WARN: Probably there can be some (major?) errors in it.


adrianhopebailie:
@earizon I think this would be very helpful as a reference.

The plugin is also used by a stand-alone sender and receiver so it might be useful to show that too?

earizon:
@adrianhopebailie Yes, I think it will make sense. I'm continuing to detail/fix errors. Since the diagram started to get more and more complex while adding details I just split in subdiagrams. Still far from sure about all details. Specially in the  send transfer and fulfill case (0004-ledger-plugin-interface/sequence_diagrams/sendTransferAndFulfill_2Ledgers.pu). I'm writing it down while reading/learning the existing code.
For now the stand-alone sender and receiver is represented in sendTransferAndFulfill_2Ledgers as a local transfer (alt is local transfer?). This is the use-case that comes to my mind about stand-alone sender/receiver, but maybe I'm wrong or there are other use-cases I'm just not aware of.

earizon:
this PR is outdated

PR #142, opened 2017-01-05
crypto-conditions: Add SECP256K1-SHA-256 fulfillment

stevenroose:
Add a new fulfillment that implements an ECDSA signature verification over Bitcoin's secp256k1 curve.

The output is not provided since I didn't get the Grunt task to work.

### DIFF
--- a/0002-crypto-conditions/0002-crypto-conditions.md
+++ b/0002-crypto-conditions/0002-crypto-conditions.md
@@ -53,6 +53,7 @@ normative:
     RFC3280:
     RFC4055:
     RFC4648:
+    RFC5912:
     RFC6920:
     RFC8017:
     I-D.draft-irtf-cfrg-eddsa-08:
@@ -74,11 +75,23 @@ normative:
         seriesInfo:
             name: "ITU-T"
             value: "Recommendation ITU-T X.690, August 2015"
+    ANSI-X9.62-2005:
+        title: "Public Key Cryptography for the Financial Services Industry: The Elliptic Curve Digital Signature Algorithm (ECDSA)"
+        author:
+            organization: American National Standards Institute
+        date: 2005-11
+    SECG-SEC2:
+        title: "Standards for Efficient Cryptography"
+        target: http://www.secg.org/sec2-v2.pdf
+        author:
+            organization: Standards for Efficient Cryptography Group
+        date: 2000
 
 informative:
     RFC2119:
     RFC3110:
     RFC4871:
+    RFC6979:
     LARGE-RSA-EXPONENTS:
         title: Imperial Violet - Very large RSA public exponents (17 Mar 2012)
         target: https://www.imperialviolet.org/2012/03/17/rsados.html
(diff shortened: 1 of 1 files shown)


adrianhopebailie:
I will merge #131 as soon as I have validated that the ASN.1 changes don't break the new test harness at https://github.com/rfcs/crypto-conditions/

adrianhopebailie:
#131 is merged. Can you rebase and force push a new commit?

stevenroose:
@adrianhopebailie I just rebased.

The only thing missing is a cost value. I didn't fully follow the new cost calculation decisions. I don't thing ECDSA is operationally similar to either Ed25519 or RSA, so I don't know how to specify a sensible heuristic.

adrianhopebailie:
The cost functions are a very rough estimate of "the number of bytes that would need to be processed by the SHA-256 hash digest algorithm to produce the equivalent amount of work".

I'd recommend running some tests on a machine with a consistent resource configuration and comparing the results of performing a SHA-256 hash and computing RSA, ed25519 and ECDSA signatures on some fixed size messages.

This should give enough data to calculate a cost function that produces a relative cost value in line with the others we have already specified.

@justmoon may also have some suggestions in this regard as we did the initial tests on his machine and I think he still has some of the raw data.

stevenroose:
@justmoon, is it possible for you to come up with a value using your reference machine?

I think it's important to keep in mind that there are some very performant secp256k1 implementations out there that act only on that specific curve like libsecp126k1 or Go's btcec.

justmoon:
> @justmoon, is it possible for you to come up with a value using your reference machine?

Ok, I just ran a rough benchmark and I got 78.721us per signature verification for secp256k1.

Using the same reference number for SHA-256 (2200 bytes/us) that would give us: 78.7 * 2200 = 173140. The nearest power of two is the same as Ed25519, 131072.

That makes sense because Ed25519 and libsecp256k1 [should be a similar speed](http://justmoon.github.io/curvebench/benchmark.html).

stevenroose:
@justmoon so I can use the same heuristic as the Ed25519 one?

stevenroose:
The cost description of Ed25519 says 

> The public key and signature are a fixed size therefore the cost for an ED25519 crypto-condition is fixed at 131072.
> 
>     cost = 131072

However ECDSA signatures are not necessarily fixed-size. The sizes vary slightly.

stevenroose:
I pushed an update, adding a cost of 131072.

adrianhopebailie:
I want to tag the current draft before we push major changes like this so just waiting for some responses to #148

stevenroose:
Ok, sure. Makes sense.

adrianhopebailie:
@stevenroose can you close this PR and replicate it against: https://github.com/rfcs/crypto-conditions/

michielbdejong:
Can this be closed?

stevenroose:
@michielbdejong I should find some time to migrate it to the RFCs repository.  I might have overestimated the usage of this addition, though. Bitcoin cannot do on-chain signature verification of arbitrary messages. This support might land in the future with segwit script versioning, either for classic ECDSA or Schnorr (both over the secp256k1 curve). Until then, it might be useful for Ethereum (Ethereum can do arbitrary message signature verification using `ecrecover()`).

emschwartz:
Closing this PR now as it should move to the rfcs repo

Issue #141, opened 2017-01-05
RFC-4: getInfo should be synchronous and populated on connect


justmoon:
Currently `getInfo` is an asynchronous method and plugin users are responsible for caching the information returned from it. (See `InfoCache` in the connector.)

It's unclear when a plugin user should consider the information outdated. It would more sense if the plugin was in charge of caching this information because it has a better idea regarding when and how often the info is likely to change.

If that is the case, then the `getInfo` call should be a lightweight synchronous call.

Since the plugin may need to do asynchronous work in order to initially populate the information, the plugin should throw an error on `getInfo` until the first success call to `connect` has completed and the plugin has emitted the connect event.

PR #143, opened 2017-01-05
docs(0004): make plugin.get{Prefix,Account,Info} synchronous

sentientwaffle:
Fixes https://github.com/interledger/rfcs/issues/141

### DIFF
--- a/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
+++ b/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
@@ -135,7 +135,7 @@ Retrieve some metadata about the ledger. Plugin must be connected, otherwise the
   "scale": 4,
   "currencyCode": "USD",
   "currencySymbol": "$",
-  "connectors": [ "chloe" ]
+  "connectors": [ "us.fed.some-bank.chloe" ]
 }
 ```
 
@@ -535,7 +535,7 @@ Metadata describing the ledger. This data is returned by the [`getInfo`](#getinf
 | `Number` | [scale](#scale) | Digits allowed after decimal |
 | `String` | [currencyCode](#currencycode) | ISO three-letter currency code |
 | `String` | [currencySymbol](#currencysymbol) | UTF8 currency symbol |
-| `String[]` | [connectors](#connectors) | Names of recommended connectors |
+| `String[]` | [connectors](#connectors) | ILP addresses of recommended connectors |
 
 ### Fields
 
@@ -569,7 +569,7 @@ The currency symbol as one or more UTF8 characters.
 #### connectors
 <code>**connectors**:String[]</code>
 
-The names of recommended connectors.
+The ILP addresses of recommended connectors.
 
 ## Class: PluginOptions
 <code>class PluginOptions</code>
--- a/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
+++ b/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
@@ -18,7 +18,7 @@ This spec depends on the [ILP spec](../0003-interledger-protocol/).
 | | [**connect**](#connect) ( options ) `⇒ Promise.<null>` |
 | | [**disconnect**](#disconnect) ( ) `⇒ Promise.<null>` |
 | | [**isConnected**](#isconnected) ( ) `⇒ Boolean` |
-| | [**getInfo**](#getinfo) ( ) <code>⇒ Promise.&lt;[LedgerInfo](#class-ledgerinfo)></code> |
+| | [**getInfo**](#getinfo) ( ) <code>⇒ [LedgerInfo](#class-ledgerinfo)</code> |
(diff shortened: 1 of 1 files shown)


0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
```diff
@@ -18,7 +18,7 @@ This spec depends on the [ILP spec](../0003-interledger-protocol/).
 | | [**connect**](#connect) ( options ) `⇒ Promise.<null>` |
 | | [**disconnect**](#disconnect) ( ) `⇒ Promise.<null>` |
 | | [**isConnected**](#isconnected) ( ) `⇒ Boolean` |
-| | [**getInfo**](#getinfo) ( ) <code>⇒ Promise.&lt;[LedgerInfo](#class-ledgerinfo)></code> |
+| | [**getInfo**](#getinfo) ( ) <code>⇒ [LedgerInfo](#class-ledgerinfo)</code> |
 | | [**getPrefix**](#getprefix) ( ) `⇒ Promise<String>` |
 | | [**getAccount**](#getaccount) ( ) `⇒ Promise<String>` |
```
emschwartz: What about `getPrefix` and `getAccount`? I think those are in the same boat as `getInfo`

sentientwaffle: Updated

justmoon:
Should these just be fields of `LedgerInfo`?

Thinking out loud: The prefix definitely seems like a ledger property nowadays, so I think that could move to LedgerInfo. The account... less so - it's a property of the plugin, not really of the ledger. So maybe that one we keep as a separate method?

Also we may want to add an event if the ledger info changes, e.g. `info_change`. (If we go with the suggestion above: I don't think the account would ever change (?), unless the prefix changes, which would trigger the `info_change` event, so I don't think we need an event for that.)

sharafian:
Do we want to roll all of this functionality into `getInfo`? It seems like it's becoming a catch-all (a cache-all, even) for anything information the plugin is able to store.

It seems like what usually happens is a component like core will call `getInfo`, and then cache the metadata itself. If this were all synchronous, I think we could split `getInfo` into more methods, which could be called more freely, and wouldn't need to minimize calls by wrapping all this info into one object.

sharafian:
I also think that with these changes, we should bring back `getConnectors`, which could synchronously return a list of connector names (right now, everything that gets connectors runs `getInfo`, and then maps the connectors field to a list of connector names)

sentientwaffle:
@justmoon: I think adding `prefix` to `LedgerInfo` is a good idea (and removing `getPrefix`). Re: `info_change`: would that be an event emitted from the ledger to the plugin?

@sharafian: I'm not in favor of adding `getConnectors`; it adds to every plugin's interface without adding any new behavior. If we add `getConnectors`, why not `getCurrencyCode`, `getPrecision`, etc... having everything rolled into one function seems simpler to me.

emschwartz:
If the calls are going to be synchronous either way, I don't think it matters much whether we have one `getInfo` function that returns a bunch of data or individual functions for the various fields. If there's an update event that seems like an argument for rolling the different fields into one object.

Regarding `getConnectors`, I also don't care whether the data is in the info object or exposed through a separate method but I do like the suggestion of making it an array of connector account names -- unless we can think of some compelling use for including other information about the connectors.

sentientwaffle:
So, info would become:

```
{
  "prefix": "example.red."
  "precision": 10,
  "scale": 4,
  "currencyCode": "USD",
  "currencySymbol": "$",
  "connectors": [ "chloe" ]
}
```

I don't see a reason for `connectors` to be an Object anymore.

justmoon:
Why is `connectors` a list of names instead of a list of ILP addresses?

Seems like we should either add the ledger prefix everywhere we talk about accounts or strip it out everywhere.

sentientwaffle:
> If the routing table entry has a liquidity curve, the connector can respond immediately to the quote request. The connector calculates the amount by combining its rate (to the ledger a.b.x) with the curve it got from the next connector (a.b.x.z).

That's not exactly how it works now (see https://github.com/interledgerjs/ilp-core/blob/35d02b7a2c08da4add4c60beacd1dcacb63352b9/src/lib/core.js#L128-L131). It doesn't check for a liquidity curve, it checks if it is a local or a direct route. Even indirect routes have a curve so that we can recompute the source amount before relaying a quote (https://github.com/interledgerjs/ilp-core/blob/35d02b7a2c08da4add4c60beacd1dcacb63352b9/src/lib/core.js#L133-L146)

justmoon:
> Re: info_change: would that be an event emitted from the ledger to the plugin?

Depends on the ledger, could be push or could be polling. Some plugins may not implement this event at all. But having this event (and supporting it in the connector, as well as in the receiver) seems quite useful so that you avoid downtime if a ledger's prefix needs to be changed for any reason.

One more thing: Could you add something to the `connect` method description to the effect of "The plugin should ensure that the information returned by `getInfo` and `getAccount` is available and cached before emitting the connected event."

PR #147, opened 2017-01-16
Specify some normalization rules for URIs

adrianhopebailie:


0002-crypto-conditions/output/0002-crypto-conditions.html
```diff
@@ -457,7 +457,7 @@
 
   <meta name="dct.creator" content="Thomas, S., Reginelli, R., and A. Hope-Bailie" />
   <meta name="dct.identifier" content="urn:ietf:id:draft-thomas-crypto-conditions-02" />
```
adrianhopebailie: Correct! I actually think we should move this whole spec to it's own repo to allow it to be tagged before merging any more changes.

See #148

adrianhopebailie:
Migrated to rfcs/crypto-conditions

PR #149, opened 2017-01-16
Initial version of a spec providing ILP payments for HTTP calls

justmoon:
This spec was motivated by me needing a payment solution for [Unhash](https://github.com/justmoon/unhash), which is an HTTP-based protocol. I already [started the work](https://github.com/interledgerjs/ilp-kit/pull/179) to add webhooks as per this spec to ilp-kit.

This is based on @emschwartz's previous (unpublished) spec, which in turn is based on work we've done together with the rest of the [Codius](https://codius.org) team.

I'm posting this here for reference, **not** to start a discussion about what this standard should eventually look like. I liked the approach we took with the Five Bells Ledger API - build some initial thing first, then when it's up and running and we've learned a ton, let's have a more informed conversation about standardizing it and cleaning it up (Common Ledger API). So this is just the initial thing to get us up and running.


justmoon:
Oh and I skipped RFC number 13 to avoid bad luck. Bad enough that most of this spec was written on Friday 13th (1/13/2017), no need to pile it on.

adrianhopebailie:
@justmoon can I merge?

justmoon:
Would like to update this for the latest PSK/etc. changes before we merge. Stand by!

0014-http-ilp/0004-http-ilp.md
```diff
@@ -0,0 +1,245 @@
...
+In order to refill its balance, the client now creates an ILP payment with the following properties:
+
+* Destination: `us.nexus.ankita.~recv.filepay`
+* Amount: `100`
+* Condition: `SHA256(fulfillment)`
+* Memo: `pay_token`
+
+The `fulfillment` is generated as:
+
+* Fulfillment: `HMAC(SHA256, condition_seed, destination_account || destination_amount)`
```
michielbdejong: Would `destination_account || destination_amount` not always be equal to just `destination_account`? Are there situations where `destination_account` is undefined?

0014-http-ilp/0004-http-ilp.md
```diff
@@ -0,0 +1,245 @@
...
+```
+```
+[...]
+```
+
+Note that the client hasn't paid at this point and is only making the request to solicit a response from the server.
+
+### 2.2. Server rejects request, because token is new and unfunded
+
+The server returns an HTTP error of `402 Payment Required` and includes response headers showing the amount (how much the request would have cost), an ILP address and a condition seed.
```
michielbdejong: -> a base64-encoded condition seed. otherwise people will try to use that string directly for the HMAC operation below

0014-http-ilp/0004-http-ilp.md
michielbdejong: and I assume that here, in this example, `condition_seed` is 
`Buffer.from('SkTcFTZCBKgP6A6QOUVcwWCCgYIP4rJPHlIzreavHdU', 'base64')`
and not
`Buffer.from('SkTcFTZCBKgP6A6QOUVcwWCCgYIP4rJPHlIzreavHdU', 'utf8')`

michielbdejong: also, I would put a counter in the destination address, so that you get a different condition each time you pay to the same receiving party

michielbdejong:
I used the `X-Pay` header from this PR in https://github.com/interledger/tutorials/pull/2

0014-http-ilp/0004-http-ilp.md
```diff
@@ -0,0 +1,245 @@
...
+* Token: `HMAC(SHA256, client_secret, hostname)`
+
+### 2.1. Client attempts to call API using token
+
+The paid HTTP request is fired off:
+
+``` http
+POST /upload HTTP/1.1
+Host: myservice.example
+X-Pay-Token: 7y0SfeN7lCuq0GFF5UsMYZofIjJ7LrvPvsePVWSv450
```
emschwartz: Shouldn't it just be `Pay-Token`? See [RFC 6648](https://tools.ietf.org/html/rfc6648) on deprecating the X prefix

sharafian: Yeah, current unhash implementation uses `Pay` and `Pay-Token`

michielbdejong: oh, good to know :) I'll update that in my tutorial as well, then.

michielbdejong:
I created an updated version of this PR in https://github.com/interledger/rfcs/pull/306

adrianhopebailie:
@justmoon:
- Need to rename the file to match the directory number (0004 -> 0014)
- Need to add front-matter to the file to track draft number and provide title

michielbdejong:
merged in #306

Issue #132, opened 2016-12-14
How do we address small ledgers/currencies?


emschwartz:
Given our hierarchical address format and the assumption that the "Tier 1" connectors will need to limit the sizes of their routing tables, how are we going to address small ledgers or currencies?

If the small ledger uses the same currency as a larger one, this might be covered by the "subledger" concept (plus maybe some amount of source routing tags if there are fees collected in between the larger ledger and the small one).

What I don't see is how you make local currencies or ledgers discoverable using the current approach. If the Tier 1 connectors don't store routes to these smaller ledgers, it seems like that group could intentionally or unintentionally exclude smaller players.

This makes me wonder whether we should take some inspiration from IPFS' [`multiaddr`](https://github.com/multiformats/multiaddr) concept.

justmoon:
I used to think that the issue of custom currencies would be addressed by TRTs (#77).

One change I'd make to TRTs right now is to take out the curve. It's simpler and probably totally fine in practice just to recurse quoting all the way to the last connector.

---

That said, due to quote recursion, there is an even simpler way to accommodate custom currencies which doesn't even require us to add any extra features (like TRTs).

Key idea: The number of tier 1 systems doesn't necessarily have to be limited.

In the IP world, it's limited more because of the size of IP address space than because of the size of routing tables. Prefixes pretty much allow you to have an arbitrary number of tier 1s as long as they share some prefix.

Right now, we use ledger prefixes like `us.usd.nexus.`. But suppose we add one more layer of indirection, so my (USD) ledger prefix is actually `us.usd.nexus.main.`. Next, I tell my peers to route anything with prefix `us.usd.nexus.*` to me. Then I can turn around and peer with people who want to create custom currencies. I'll assign them prefixes starting with `us.usd.nexus.`, e.g. `us.usd.nexus.bobcustom`. Nobody's routing table (but mine) would have to grow, but those custom currencies would now be tier 1.

In this scenario, I wouldn't face a trust issue because I can set up my peering relationship with `us.usd.nexus.bobcustom.` such that I only accept routes for their ledger prefix (and below) from them.

This does lead to a bit more rigid routing structure, but I think that's fine. We can come up with more and more efficient routing in the future, securely detect shortcuts, etc. But for an MVP, we can live without that.

---

> This makes me wonder whether we should take some inspiration from IPFS' multiaddr concept.

[...]

Issue #43, opened 2016-07-10
Proposal: Our transport layers deal with condition generation, memo encryption, ...


justmoon:
Issue #28 proposes the removal of Optimistic, Universal, Atomic as transport protocols.

@emschwartz suggested we could bring back conditions as a condition layer, because they are end-to-end and need to be standardized.

In addition to that, this layer could also deal with encrypting the memo to the recipient (we already have the recipient's public key in some of the layer types mentioned above.) 
## User-defined Transport

There is no condition and no expiry. The memo is unencrypted.

``` js
sender.getCondition(condition) 
```
## Interactive Transport

Receiver generates the condition and provides a payment request (with a standard format) to the sender. 

``` js
receiver.generateCondition(ilpPacket)
sender.getCondition(multiPaymentRequest | previousFulfillment)
```

As an additional feature, we could have a condition whose fulfillment is another payment request (including a condition), such that multiple (pre-planned) payments can be expressed by one original payment request. This is useful for repeated payments or other predictable payments between the same participants.
## Passive Transport

Receiver has an Ed25519 public key. The sender generates the condition from this key and encrypts the memo with it. (Supercedes #30)

``` js
sender.getCondition(receiverPublicKey, ilpPacket:Buffer)
receiver.getFulfillment(receiverPrivateKey, finalTransfer)
// - Verify transfer has as much money as ilpPacket
// - Verify condition matches ilpPacket

// Alternative: sender-encrypted hashlock
sender.getCondition(receiverPublicKey, ilpPacket:Buffer)
receiver.getFulfillment(receiverPrivateKey, finalTransfer)
// - Verify transfer has as much money as ilpPacket claims
// - Verify condition matches ilpPacket
```

Open Questions:
- [ ] How to decide ILP header expiry for Passive case?
- [ ] How to decide source transfer expiry for Interactive & Passive cases?
- [ ] Do we have an equivalent to UDP/TCP ports (i.e. an application ID or something)
- [ ] Naming: Does it matter that we had a different meaning for UTP for a while? (Universal vs User-defined)

r-marques:
Note that ed25519 is a public key digital signature system and its not intended to be used to encrypt.
There may be a way to reuse the key pair for encryption with a different crypto system, maybe curve25519, but I am not sure.

emschwartz:
[...]

PR #154, opened 2017-02-03
ILP Addresses (RFC0015)

mDuo13:
This PR consolidates information from #132 and #31, plus summarizes up-to-date thoughts from the team on addresses in general.

Summary:

- Intro to what addresses are for and high-level view of how they're used for routing
- Rules for what's a valid ILP Address, including lists of currently-defined prefixes
- Descriptions of different segments and what they might be used for
- Examples and recommendations for segments

### DIFF
--- a/0015-ilp-addresses/0015-ilp-addresses.md
+++ b/0015-ilp-addresses/0015-ilp-addresses.md
@@ -28,7 +28,7 @@ ILP Addresses must meet the following requirements:
     - Hyphen (`-`)
 3. Each segment MUST be separated from other segments by a period character (`.`).
 4. Address prefixes MUST end in a period (`.`) character and MAY contain any number of segments after the allocation scheme prefix.
-6. Destination addresses MUST NOT end in a period (`.`) character, and MUST contain at least two segments after the allocation scheme prefix.
+5. Destination addresses MUST NOT end in a period (`.`) character, and MUST contain at least two segments after the allocation scheme prefix.
 
 The following ABNF specification defines the format for all ILP addresses and address prefixes:
 
--- a/0015-ilp-addresses/0015-ilp-addresses.md
+++ b/0015-ilp-addresses/0015-ilp-addresses.md
@@ -31,7 +31,7 @@ ILP Addresses must meet the following requirements:
 5. Destination addresses MUST NOT end in a period (`.`) character, and MUST contain at least two segments after the allocation scheme prefix.
 6. The total length of an ILP Address must be no more than **1023 characters** including the allocation scheme prefix, separators, and all segments.
 
-The following ABNF specification defines the format for all ILP addresses and address prefixes:
+The following ABNF specification defines the format for the contents of all ILP addresses and address prefixes. (You must also enforce the overal length requirement of 1023 characters or less.)
 
 ```abnf
 address     = scheme separator *prefix [segment]
@@ -51,11 +51,11 @@ You can also use the following regular expressions to verify the same requiremen
 
 | Address Type        | Regular Expression                                     |
 |:--------------------|:-------------------------------------------------------|
-| All addresses       | `^(g|private|example|peer|self|test[1-3])\.([a-zA-Z0-9_~-]+\.)*([a-zA-Z0-9_~-]+)?$` |
-| Address prefix      | `^(g|private|example|peer|self|test[1-3])\.([a-zA-Z0-9_~-]+\.)*$` |
-| Destination address | `^(g|private|example|peer|self|test[1-3])\.([a-zA-Z0-9_~-]+\.)+[a-zA-Z0-9_~-]+$` |
+| All addresses       | `(?=^.{1,1023}$)^(g|private|example|peer|self|test[1-3])[.]([a-zA-Z0-9_~-]+[.])*([a-zA-Z0-9_~-]+)?$` |
+| Address prefix      | `(?=^.{1,1023}$)^(g|private|example|peer|self|test[1-3])[.]([a-zA-Z0-9_~-]+[.])*$` |
+| Destination address | `(?=^.{1,1023}$)^(g|private|example|peer|self|test[1-3])[.]([a-zA-Z0-9_~-]+[.])+[a-zA-Z0-9_~-]+$` |
 
-(Note that neither the ABNF nor the regular expressions enforce the maximum length requirement. Remember to check that total length of the address is **1023 characters or less**.)
+(If your regular expression engine does not support lookahead, you must drop the first parenthesis and separately enforce the overall length requirement of 1023 characters or less.)
 
 ## Allocation Schemes
 
--- a/0015-ilp-addresses/0015-ilp-addresses.md
+++ b/0015-ilp-addresses/0015-ilp-addresses.md
@@ -1,9 +1,11 @@
 # ILP Addresses
 
-_ILP addresses_ provide a way route payments to their intended destination through a recursive series of hops, including any number of ILP Connectors. (This happens after the payment is set up on a higher level, such as [SPSP](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md).) Addresses can be subdivided into two categories:
+_ILP addresses_ provide a way route payments to their intended destination through a recursive series of hops, including any number of ILP Connectors. (This happens after the payment is set up on a higher level, such as [SPSP](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md).) Addresses are not meant to be user-facing, but allow several ASCII characters for easy debugging.
+
+Addresses can be subdivided into two categories:
 
 - **Destination Addresses** are complete addresses that can receive payments. A destination address always maps to one account in a ledger. (It can also provide more specific information, such as an invoice ID or a sub-account.) Destination addresses MUST NOT end in a period (`.`) character.
-- **Prefix Addresses** are incomplete addresses representing a group of destination addresses. Many depths of grouping are possible: groups of accounts or sub-accounts, an individual ledger or subledger, or entire neighborhoods of ledgers. Prefix addresses MUST end in a period (`.`) character. Payment setup protocols MUST reject payments to prefix addresses.
+- **Address Prefixes** are incomplete addresses representing a group of destination addresses. Many depths of grouping are possible: groups of accounts or sub-accounts, an individual ledger or subledger, or entire neighborhoods of ledgers. Address prefixes MUST end in a period (`.`) character. Payment setup protocols MUST reject payments to address prefixes.
 
 Both types of address usually contain one or more period characters as separators.
 
@@ -11,40 +13,39 @@ Both types of address usually contain one or more period characters as separator
 
 ILP Addresses must meet the following requirements:
 
-1. The address MUST begin with a prefix indicating the allocation scheme. The only supported allocation scheme for now is the [General Allocation Scheme](#general-allocation-scheme), which uses the prefix `g.`.
+1. The address MUST begin with a prefix indicating the allocation scheme. See [Allocation Schemes](#allocation-schemes) for more information.
 2. Each "segment" of the address MUST contain one or more of the following characters:
     - Alphanumeric characters, upper or lower case. (Addresses are **case-sensitive** so that they can contain data encoded in formats such as base64url.)
     - Underscore (`_`)
     - Tilde (`~`)
     - Hyphen (`-`)
 3. Each segment MUST be separated from other segments by a period character (`.`).
-4. Prefix addresses MUST end in a period (`.`) character and MAY contain any number of segments after the allocation scheme prefix.
+4. Address prefixes MUST end in a period (`.`) character and MAY contain any number of segments after the allocation scheme prefix.
 6. Destination addresses MUST NOT end in a period (`.`) character, and MUST contain at least two segments after the allocation scheme prefix.
+
+***TODO: ABNF definitions for address, destination address, and address prefix; update or remove regexes accordingly.***
 
 The following regular expressions summarize these requirements:
 
 | Address Type        | Regular Expression                         |
 |:--------------------|:-------------------------------------------|
 | All addresses       | `^g\.[a-zA-Z0-9._~-]*$`                    |
-| Prefix address      | `^g\.([a-zA-Z0-9_~-]+\.)*$`                |
+| Address prefix      | `^g\.([a-zA-Z0-9_~-]+\.)*$`                |
 | Destination address | `^g\.([a-zA-Z0-9_~-]+\.)+[a-zA-Z0-9_~-]+$` |
 
-## Example Addresses
+## Allocation Schemes
 
-`g.acme.bob` - a destination address to the account "bob" in the ledger "acme".
+The allocation scheme is the first part of an address, which indicates how the address is assigned. Here is a summary of the prefixes that are currently defined:
 
-`g.us.fed.ach.0.acmebank.swx0a0.acmecorp.sales.199.cdfa5e16-e759-4ba3-88f6-8b9dc83c1868.2` - destination address for a particular invoice, which can break down as follows:
-
-- Neighborhoods: `us`, `fed`, `ach`, `0`, `acmebank`
-- Ledger: `swx0a0`
-- Subledger: `acmecorp`
-- Account group: `sales`
-- Account: `199`
-- Interactions: `cdfa5e16-e759-4ba3-88f6-8b9dc83c1868`, `2`
-
-`g.` - the shortest possible prefix address. All entries that are in the general allocation scheme.
-
-`g.crypto.bitcoin.` - prefix address for the public Bitcoin blockchain
+| Prefix                       | Allocation Scheme                                       | Definition and Use Case |
+|:-----------------------------|:--------------------------------------------------------|:--|
+| `g.`                         | [General Allocation Scheme](#general-allocation-scheme) | Most ILP addresses used to send and receive real money. |
+| `private.`                   | Private allocation                                      | For ILP addresses that only have meaning in a private subnet or intranet. Analogous to the [192.168.0.0/16 range in IPv4](https://en.wikipedia.org/wiki/Private_network). |
+| `example.`                   | Examples                                                | For "non-real" addresses that are used as examples or in documentation. Analogous to ["555 phone numbers"](https://en.wikipedia.org/wiki/555_%28telephone_number%29) in the USA. |
+| `test1.`, `test2.`, `test3.` | Testing                                                 | For addresses used in tests, such as unit or integration tests of compatible software. |
+| `local.`                     | Ledger-local                                            | For addresses that are only valid in the context of a local ledger. Analogous to [link-local addresses](https://en.wikipedia.org/wiki/Link-local_address) in IP. |
+| `peer.`                      | Peering                                                 | Similar to ledger-local addresses, but specifically for use in a peering relationship. The [ilp-plugin-virtual](https://github.com/interledgerjs/ilp-plugin-virtual) is an example of an existing implementation that uses this. |
+| `self.`                      | Local loopback                                          | For addresses that are only valid on the local machine. |
 
 ## General Allocation Scheme
 
@@ -66,7 +67,7 @@ To make routing work well, we recommend including the following components as se
 
 Neighborhoods have no specific meaning, but serve as a quick shortcut for routing to the right area.
 
-- For regional currencies, use a continent code from the folowing table:
+- For regional currencies, use a continent code from the following table:
     | Segment | Continent     |
     |:--------|:--------------|
     | `af`    | Africa        |
@@ -111,7 +112,23 @@ Additional segments within an address. In most cases, any segments after the acc
 
 _**Rome's note:** I'm not sure if we should say ledger plugins can use these segments to decide which account to route to or not. If they can, I'm not sure how useful it is to draw a distinction between this and the account identifier._
 
+### Example General Allocation Scheme Addresses
 
+`g.acme.bob` - a destination address to the account "bob" in the ledger "acme".
+
+`g.us.fed.ach.0.acmebank.swx0a0.acmecorp.sales.199.cdfa5e16-e759-4ba3-88f6-8b9dc83c1868.2` - destination address for a particular invoice, which can break down as follows:
+
+- Neighborhoods: `us`, `fed`, `0`, `acmebank`
+- Payment rail: `ach`
+- Ledger: `swx0a0`
+- Subledger: `acmecorp`
+- Account group: `sales`
+- Account: `199`
+- Interactions: `cdfa5e16-e759-4ba3-88f6-8b9dc83c1868`, `2`
+
+`g.` - the shortest possible address prefix. Includes all entries that are in the general allocation scheme.
+
+`g.crypto.bitcoin.` - address prefix for the public Bitcoin blockchain
 
 # Routing
 
--- a/0015-ilp-addresses/0015-ilp-addresses.md
+++ b/0015-ilp-addresses/0015-ilp-addresses.md
@@ -1,0 +1,123 @@
+# ILP Addresses
+
+_ILP addresses_ provide a way route payments to their intended destination through a recursive series of hops, including any number of ILP Connectors. (This happens after the payment is set up on a higher level, such as [SPSP](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md).) Addresses can be subdivided into two categories:
+
+- **Destination Addresses** are complete addresses that can receive payments. A destination address always maps to one account in a ledger. (It can also provide more specific information, such as an invoice ID or a sub-account.) Destination addresses MUST NOT end in a period (`.`) character.
+- **Prefix Addresses** are incomplete addresses representing a group of destination addresses. Many depths of grouping are possible: groups of accounts or sub-accounts, an individual ledger or subledger, or entire neighborhoods of ledgers. Prefix addresses MUST end in a period (`.`) character. Payment setup protocols MUST reject payments to prefix addresses.
+
+Both types of address usually contain one or more period characters as separators.
+
+## Address Requirements
+
+ILP Addresses must meet the following requirements:
+
+1. The address MUST begin with a prefix indicating the allocation scheme. The only supported allocation scheme for now is the [General Allocation Scheme](#general-allocation-scheme), which uses the prefix `g.`.
+2. Each "segment" of the address MUST contain one or more of the following characters:
+    - Alphanumeric characters, upper or lower case. (Addresses are **case-sensitive** so that they can contain data encoded in formats such as base64url.)
+    - Underscore (`_`)
+    - Tilde (`~`)
+    - Hyphen (`-`)
+3. Each segment MUST be separated from other segments by a period character (`.`).
+4. Prefix addresses MUST end in a period (`.`) character and MAY contain any number of segments after the allocation scheme prefix.
+6. Destination addresses MUST NOT end in a period (`.`) character, and MUST contain at least two segments after the allocation scheme prefix.
+
+The following regular expressions summarize these requirements:
+
+| Address Type        | Regular Expression                         |
+|:--------------------|:-------------------------------------------|
+| All addresses       | `^g\.[a-zA-Z0-9._~-]*$`                    |
+| Prefix address      | `^g\.([a-zA-Z0-9_~-]+\.)*$`                |
+| Destination address | `^g\.([a-zA-Z0-9_~-]+\.)+[a-zA-Z0-9_~-]+$` |
+
+## Example Addresses
+
+`g.acme.bob` - a destination address to the account "bob" in the ledger "acme".
+
+`g.us.fed.ach.0.acmebank.swx0a0.acmecorp.sales.199.cdfa5e16-e759-4ba3-88f6-8b9dc83c1868.2` - destination address for a particular invoice, which can break down as follows:
+
+- Neighborhoods: `us`, `fed`, `ach`, `0`, `acmebank`
+- Ledger: `swx0a0`
+- Subledger: `acmecorp`
+- Account group: `sales`
+- Account: `199`
+- Interactions: `cdfa5e16-e759-4ba3-88f6-8b9dc83c1868`, `2`
+
+`g.` - the shortest possible prefix address. All entries that are in the general allocation scheme.
+
+`g.crypto.bitcoin.` - prefix address for the public Bitcoin blockchain
+
+## General Allocation Scheme
+
+The general allocation scheme for ILP Addresses is currently the only supported scheme. It uses the prefix `g.`.
+
+This scheme has no central issuing authority or mechanism, so more than one entity can use the same address. In this case, some connectors may prepare a route to a different account than intended. In this failure case, no money moves because the receiver does not send the fulfillment. Participants in the interledger can reduce the chances of encountering this failure case by choosing addresses carefully and by properly managing connectors' routing tables.
+
+To make routing work well, we recommend including the following components as segments in an address (in order):
+
+- [Neighborhoods](#neighborhoods)
+- [Payment rail](#payment-rail)
+- [Ledger prefix](#ledger-prefix)
+- [Subledger and account group](#subledger-and-account-group)
+- [Account identifier](#account-identifier)
+- [Interactions](#interactions)
+
+
+### Neighborhoods
+
+Neighborhoods have no specific meaning, but serve as a quick shortcut for routing to the right area.
+
+- For regional currencies, use a continent code from the folowing table:
+    | Segment | Continent     |
+    |:--------|:--------------|
+    | `af`    | Africa        |
+    | `as`    | Asia          |
+    | `eu`    | Europe        |
+    | `na`    | North America |
+    | `sa`    | South America |
+    | `oc`    | Oceania       |
+    | `an`    | Antarctica    |
+- For decentralized crypto-currency ledgers, use the `crypto` neighborhood.
+- For naturally-ocurring commodities or other globally-distributed currencies, use the `earth` neighborhood.
+- For currencies strongly associated with a particular international corporation, use the continent of that corporation's headquarters.
+
+It may also be useful to have 1-3 levels of "sub-neighborhoods" to group together related connectors or ledgers. We have no specific recommendations at this time.
+
+### Payment Rail
+
+The payment rail refers to a specific way some accounts or ledgers are connected outside of ILP, such as ACH in the U.S., SEPA in Europe, or other online payment systems such as PayPal or Alipay. We recommend using an all-lowercase abbreviation for the payment rail if one is applicable to the address. Examples:
+
+| Segment | Payment Rail                                                       |
+|:--------|:-------------------------------------------------------------------|
+| `ach`   | [United States Automated Clearing House](https://en.wikipedia.org/wiki/Automated_Clearing_House) |
+| `sepa`  | [Single Euro Payments Area](https://en.wikipedia.org/wiki/Single_Euro_Payments_Area) |
+
+***TODO: more rails?***
+
+### Ledger Prefix
+
+The ledger prefix is the unique name of the ledger. When doing a local delivery, the connector typically uses this segment to figure out which ledger plugin to use. As a result, it's difficult (but possible) to make a connector between two ledgers that use the same ledger prefix.
+
+### Subledger and Account Group
+
+Zero or more segments representing optional divisions within a ledger. If there are no fees involved in transferring between a ledger and a subledger, the subledger can be a "local delivery" from the connector's perspective. If fees are necessary for connecting to a subledger, it may require a connector. ***TODO: More clarifications and detailed recommendations needed.***
+
+### Account Identifier
+
+A unique identifier of the account within the ledger. The ledger plugin maps these to accounts within a ledger. For some ledgers, a simple conversion rule may suffice; other ledgers may require a lookup table. ***TODO: clarify exactly how five-bells-ledger-plugin or Common Ledger API plugin do this***
+
+### Interactions
+
+Additional segments within an address. In most cases, any segments after the account identifier are not expected to match anything in a routing table, so they won't usually affect the routing of a payment. They may be used by automated programs listening for payments to specific purposes or situations, such as invoices.
+
+_**Rome's note:** I'm not sure if we should say ledger plugins can use these segments to decide which account to route to or not. If they can, I'm not sure how useful it is to draw a distinction between this and the account identifier._
+
+
+
+# Routing
+
+_**Rome's note:** I might move this to RFC 0003 eventually._
+
+Connectors maintain a _routing table_ of ILP addresses. Routing is a recursive lookup through the routing tables of any number of connectors. When a connector receives a query, it finds the [longest prefix match](https://en.wikipedia.org/wiki/Longest_prefix_match) for the queried address. Then, it follows one of the following cases:
+
+- If the matching address is marked for local delivery, the connector prepares a transfer to that address in one of the ledgers connected to it. The connector's ledger plugin maps the ILP address to an account within the ledger. (This is the base case.)
+- If the matching address is marked as forwarded delivery, it has the address of another connector associated with it in the routing table. The connector makes a routing lookup on the connector associated with the address. (This is the recursive case.)
--- a/0015-ilp-addresses/0015-ilp-addresses.md
+++ b/0015-ilp-addresses/0015-ilp-addresses.md
@@ -13,7 +13,7 @@ Both types of address usually contain one or more period characters as separator
 
 Connectors maintain a _routing table_ of ILP addresses. Routing is a recursive lookup through the routing tables of any number of connectors. When a connector receives a query, it finds the [longest prefix match](https://en.wikipedia.org/wiki/Longest_prefix_match) for the queried address. Then, it follows one of the following cases:
 
-- If the matching address is marked for local delivery, the connector prepares a transfer to that address in one of the ledgers connected to it. The connector's ledger plugin maps the ILP address to an account within the ledger. (This is the base case.)
+- If the matching address is marked for local delivery, the connector prepares a transfer to that address in one of the ledgers connected to it. The connector maps the ILP address to an account within the ledger. (This is the base case.)
 - If the matching address is marked as forwarded delivery, it has the address of another connector associated with it in the routing table. The connector makes a routing lookup on the connector associated with the address. (This is the recursive case.)
 
 ## Address Requirements
@@ -29,6 +29,7 @@ ILP Addresses must meet the following requirements:
 3. Each segment MUST be separated from other segments by a period character (`.`).
 4. Address prefixes MUST end in a period (`.`) character and MAY contain any number of segments after the allocation scheme prefix.
 5. Destination addresses MUST NOT end in a period (`.`) character, and MUST contain at least two segments after the allocation scheme prefix.
+6. The total length of an ILP Address must be no more than **1023 characters** including the allocation scheme prefix, separators, and all segments.
 
 The following ABNF specification defines the format for all ILP addresses and address prefixes:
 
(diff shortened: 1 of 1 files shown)


0015-ilp-addresses/0015-ilp-addresses.md
```diff
@@ -0,0 +1,123 @@
...
+
+The following regular expressions summarize these requirements:
+
+| Address Type        | Regular Expression                         |
+|:--------------------|:-------------------------------------------|
+| All addresses       | `^g\.[a-zA-Z0-9._~-]*$`                    |
+| Prefix address      | `^g\.([a-zA-Z0-9_~-]+\.)*$`                |
+| Destination address | `^g\.([a-zA-Z0-9_~-]+\.)+[a-zA-Z0-9_~-]+$` |
+
+## Example Addresses
```
emschwartz: This seems like it should come after the explanation of the General Allocation Scheme, because it is an example of that scheme

0015-ilp-addresses/0015-ilp-addresses.md
```diff
@@ -0,0 +1,123 @@
...
+    | `eu`    | Europe        |
+    | `na`    | North America |
+    | `sa`    | South America |
+    | `oc`    | Oceania       |
+    | `an`    | Antarctica    |
+- For decentralized crypto-currency ledgers, use the `crypto` neighborhood.
+- For naturally-ocurring commodities or other globally-distributed currencies, use the `earth` neighborhood.
+- For currencies strongly associated with a particular international corporation, use the continent of that corporation's headquarters.
+
+It may also be useful to have 1-3 levels of "sub-neighborhoods" to group together related connectors or ledgers. We have no specific recommendations at this time.
```
sharafian: We should include how to address one ledger in terms of another, and also document the routing prefix vs. ledger prefix distinction. So `us.usd.cornelius.main.alice` might be an account on the ledger `us.usd.cornelius.main.`, but if you wanted to address in terms of the ledger you would use the address `us.usd.cornelius.myledger`.

0015-ilp-addresses/0015-ilp-addresses.md
```diff
@@ -0,0 +1,123 @@
+# ILP Addresses
+
+_ILP addresses_ provide a way route payments to their intended destination through a recursive series of hops, including any number of ILP Connectors. (This happens after the payment is set up on a higher level, such as [SPSP](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md).) Addresses can be subdivided into two categories:
+
+- **Destination Addresses** are complete addresses that can receive payments. A destination address always maps to one account in a ledger. (It can also provide more specific information, such as an invoice ID or a sub-account.) Destination addresses MUST NOT end in a period (`.`) character.
+- **Prefix Addresses** are incomplete addresses representing a group of destination addresses. Many depths of grouping are possible: groups of accounts or sub-accounts, an individual ledger or subledger, or entire neighborhoods of ledgers. Prefix addresses MUST end in a period (`.`) character. Payment setup protocols MUST reject payments to prefix addresses.
```
emschwartz: Address prefixes? It's not the address of a prefix..

0015-ilp-addresses/0015-ilp-addresses.md
```diff
@@ -0,0 +1,123 @@
...
+| `ach`   | [United States Automated Clearing House](https://en.wikipedia.org/wiki/Automated_Clearing_House) |
+| `sepa`  | [Single Euro Payments Area](https://en.wikipedia.org/wiki/Single_Euro_Payments_Area) |
+
+***TODO: more rails?***
+
+### Ledger Prefix
+
+The ledger prefix is the unique name of the ledger. When doing a local delivery, the connector typically uses this segment to figure out which ledger plugin to use. As a result, it's difficult (but possible) to make a connector between two ledgers that use the same ledger prefix.
+
+### Subledger and Account Group
```
emschwartz: I still don't see a difference between a subledger and an account group

0015-ilp-addresses/0015-ilp-addresses.md
```diff
@@ -0,0 +1,123 @@
...
+- [Subledger and account group](#subledger-and-account-group)
+- [Account identifier](#account-identifier)
+- [Interactions](#interactions)
+
+
+### Neighborhoods
+
+Neighborhoods have no specific meaning, but serve as a quick shortcut for routing to the right area.
+
+- For regional currencies, use a continent code from the folowing table:
```
sharafian: I don't think that region-based ILP addresses are helpful. The only thing that matters in the "proximity" of two ledgers is how many connectors operate between them. The ILP Kits being run by early adopters are going to be  easy to connect between, even though they're widely distributed across the globe.

0015-ilp-addresses/0015-ilp-addresses.md
```diff
@@ -0,0 +1,123 @@
...
+
+It may also be useful to have 1-3 levels of "sub-neighborhoods" to group together related connectors or ledgers. We have no specific recommendations at this time.
+
+### Payment Rail
+
+The payment rail refers to a specific way some accounts or ledgers are connected outside of ILP, such as ACH in the U.S., SEPA in Europe, or other online payment systems such as PayPal or Alipay. We recommend using an all-lowercase abbreviation for the payment rail if one is applicable to the address. Examples:
+
+| Segment | Payment Rail                                                       |
+|:--------|:-------------------------------------------------------------------|
+| `ach`   | [United States Automated Clearing House](https://en.wikipedia.org/wiki/Automated_Clearing_House) |
```
sharafian: In "example addresses," `ach` is listed as a neighborhood. Is this different?

0015-ilp-addresses/0015-ilp-addresses.md
```diff
@@ -0,0 +1,123 @@
...
+| `sepa`  | [Single Euro Payments Area](https://en.wikipedia.org/wiki/Single_Euro_Payments_Area) |
+
+***TODO: more rails?***
+
+### Ledger Prefix
+
+The ledger prefix is the unique name of the ledger. When doing a local delivery, the connector typically uses this segment to figure out which ledger plugin to use. As a result, it's difficult (but possible) to make a connector between two ledgers that use the same ledger prefix.
```
sharafian: We should say who is expected to choose the prefix of a ledger (the ledger can set it in their metadata, the connector can set it in their route broadcasts, etc.)

0015-ilp-addresses/0015-ilp-addresses.md
```diff
@@ -0,0 +1,123 @@
...
+
+### Account Identifier
+
+A unique identifier of the account within the ledger. The ledger plugin maps these to accounts within a ledger. For some ledgers, a simple conversion rule may suffice; other ledgers may require a lookup table. ***TODO: clarify exactly how five-bells-ledger-plugin or Common Ledger API plugin do this***
+
+### Interactions
+
+Additional segments within an address. In most cases, any segments after the account identifier are not expected to match anything in a routing table, so they won't usually affect the routing of a payment. They may be used by automated programs listening for payments to specific purposes or situations, such as invoices.
+
+_**Rome's note:** I'm not sure if we should say ledger plugins can use these segments to decide which account to route to or not. If they can, I'm not sure how useful it is to draw a distinction between this and the account identifier._
```
emschwartz: I think they would route the money to the same account, but they might use those to send out different notifications. For example, an IPR receiver might want to listen on the `alice` account but only for payments that have the receiver identifier right after the `alice` part.

0015-ilp-addresses/0015-ilp-addresses.md
```diff
@@ -0,0 +1,123 @@
...
+- Account: `199`
+- Interactions: `cdfa5e16-e759-4ba3-88f6-8b9dc83c1868`, `2`
+
+`g.` - the shortest possible prefix address. All entries that are in the general allocation scheme.
+
+`g.crypto.bitcoin.` - prefix address for the public Bitcoin blockchain
+
+## General Allocation Scheme
+
+The general allocation scheme for ILP Addresses is currently the only supported scheme. It uses the prefix `g.`.
```
sharafian: Are the allocation schemes part of ILP, or is this just recommended behavior?

0015-ilp-addresses/0015-ilp-addresses.md
```diff
@@ -0,0 +1,123 @@
...
+
+The ledger prefix is the unique name of the ledger. When doing a local delivery, the connector typically uses this segment to figure out which ledger plugin to use. As a result, it's difficult (but possible) to make a connector between two ledgers that use the same ledger prefix.
+
+### Subledger and Account Group
+
+Zero or more segments representing optional divisions within a ledger. If there are no fees involved in transferring between a ledger and a subledger, the subledger can be a "local delivery" from the connector's perspective. If fees are necessary for connecting to a subledger, it may require a connector. ***TODO: More clarifications and detailed recommendations needed.***
+
+### Account Identifier
+
+A unique identifier of the account within the ledger. The ledger plugin maps these to accounts within a ledger. For some ledgers, a simple conversion rule may suffice; other ledgers may require a lookup table. ***TODO: clarify exactly how five-bells-ledger-plugin or Common Ledger API plugin do this***
```
sharafian: is an account identifier always one segment?

0015-ilp-addresses/0015-ilp-addresses.md
```diff
@@ -0,0 +1,123 @@
...
+
+Additional segments within an address. In most cases, any segments after the account identifier are not expected to match anything in a routing table, so they won't usually affect the routing of a payment. They may be used by automated programs listening for payments to specific purposes or situations, such as invoices.
+
+_**Rome's note:** I'm not sure if we should say ledger plugins can use these segments to decide which account to route to or not. If they can, I'm not sure how useful it is to draw a distinction between this and the account identifier._
+
+
+
+# Routing
+
+_**Rome's note:** I might move this to RFC 0003 eventually._
```
sharafian: Could routing table behavior be a part of the Connector-Connector Protocol, or is that only for their interactions between one another?

0015-ilp-addresses/0015-ilp-addresses.md
emschwartz: I think "ILP" as such should only mandate a certain character set

0015-ilp-addresses/0015-ilp-addresses.md
emschwartz: I think so. Otherwise how would ledger plugins know what to split on?

sharafian: what if a ledger existed that had 2 segments to determine an account (eg `bob.smith` and `bob.jones` were two separate accounts). A ledger plugin could easily take off 2 segments after the ledger prefix instead of 1, but would this violate ILP addressing?

0015-ilp-addresses/0015-ilp-addresses.md
mDuo13: Same, to be honest.

0015-ilp-addresses/0015-ilp-addresses.md
mDuo13: I'm sorry, you lost me. What's the difference?

0015-ilp-addresses/0015-ilp-addresses.md
mDuo13: Good catch. The "example address" breakdown was copied from Evan's post in #132 and I neglected it to update with the payment rail thing.

0015-ilp-addresses/0015-ilp-addresses.md
```diff
@@ -0,0 +1,123 @@
...
+- **Destination Addresses** are complete addresses that can receive payments. A destination address always maps to one account in a ledger. (It can also provide more specific information, such as an invoice ID or a sub-account.) Destination addresses MUST NOT end in a period (`.`) character.
+- **Prefix Addresses** are incomplete addresses representing a group of destination addresses. Many depths of grouping are possible: groups of accounts or sub-accounts, an individual ledger or subledger, or entire neighborhoods of ledgers. Prefix addresses MUST end in a period (`.`) character. Payment setup protocols MUST reject payments to prefix addresses.
+
+Both types of address usually contain one or more period characters as separators.
+
+## Address Requirements
+
+ILP Addresses must meet the following requirements:
+
+1. The address MUST begin with a prefix indicating the allocation scheme. The only supported allocation scheme for now is the [General Allocation Scheme](#general-allocation-scheme), which uses the prefix `g.`.
```
michielbdejong: +1 to leaving this open,  I can think of one or two other schemes we might invent in the future.

sharafian: does it make sense to say that all addresses in the same allocation scheme are going to be close to one another? this would make it impossible to address one ledger in terms of another if they don't share allocation schemes.

sappenin: Is creating a custom scheme preferable to extending a URI, with restrictions, sort of like [RFC-7565](https://tools.ietf.org/html/rfc7565#section-7) did?

For example, a URI scheme could conform to the restrictions of section 2 below, but would instead look like: `ilp://earth.us.bank.account.bob`

0015-ilp-addresses/0015-ilp-addresses.md
```diff
@@ -0,0 +1,123 @@
...
+2. Each "segment" of the address MUST contain one or more of the following characters:
+    - Alphanumeric characters, upper or lower case. (Addresses are **case-sensitive** so that they can contain data encoded in formats such as base64url.)
```
sappenin: Why the restriction on character set?  This was briefly discussed on the call a few weeks back, but I don't recall the exact rationale.

0015-ilp-addresses/0015-ilp-addresses.md
mDuo13: Do you have any suggestions for what kind of recommendations to make for the topmost-level prefix?

0015-ilp-addresses/0015-ilp-addresses.md
mDuo13: I don't think there's any reason a ledger could use more than one segment. I'm not sure if it can use a variable number of segment, though.

0015-ilp-addresses/0015-ilp-addresses.md
```diff
@@ -0,0 +1,123 @@
+# ILP Addresses
+
+_ILP addresses_ provide a way route payments to their intended destination through a recursive series of hops, including any number of ILP Connectors. (This happens after the payment is set up on a higher level, such as [SPSP](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md).) Addresses can be subdivided into two categories:
```
sappenin: It might be worth adding some verbiage somewhere in here that ILP addresses are not meant to be user-facing (assuming that's still the case).  For example, here was some feedback from the mailing list from a few weeks back:

> Our experience so far is that the most confusing thing about ILP Addresses is that they are ASCII characters, whereas IP Addresses are just numbers. Many people have interpreted that to mean that ILP Addresses are user-facing. They are not! They are similarly ephemeral and low-level as IP Addresses, and a user should never see one. They are used for routing and nothing else.

0015-ilp-addresses/0015-ilp-addresses.md
sharafian: I do think `crypto` is a good top-level prefix, because there is actually a community around it. For large banks, it could make sense to use a currency/country code, or maybe to have some consortium of banks be the top-level prefix. Of course, this is less important for really big ledgers, because almost all connectors will have them in their routing tables.

Any ledger that isn't drawing a ton of traffic on its own will want to be addressed in terms of other ledgers. The current network of ILP early adopters might have its own top-level prefix, given that there's nothing to route in terms of yet.

0015-ilp-addresses/0015-ilp-addresses.md
sharafian: Evan sums it up here pretty well https://github.com/interledger/rfcs/issues/132#issuecomment-270390166

0015-ilp-addresses/0015-ilp-addresses.md
mDuo13: I think we need to clarify what it means to "leave this open" -

- If you're saying that implementation should validate addresses that don't start with `g.`, I can see some rationale for that. However, any such implementation is not going to be able to know whether the addresses are following all the rules of the other allocation schemes.
- If you're saying that people should be able to claim addresses that don't start with `g.` but are otherwise the same as the generally-allocated addresses, I think that's unwise. That defeats the purpose of using a general allocation prefix.

When we devise other allocation schemes, it's easy to update this spec with more prefixes _and_ the rules of the other allocation schemes. That protects the future allocation schemes from having existing addresses occupying their namespace, and it protects implementations from trying to understand addresses that are allocated through a 

Imagine a "pubkey" allocation scheme, where the address is a public key that can be looked upon a keyserver - one of the rules of the allocation scheme is that the pubkey exists and isn't revoked on a consensus of key servers. We may or may not want implementations to route things to addresses in the `pubkey.` space if those implementations don't know the difference.

So I'm for leaving this as-is until we have additional allocation schemes, and officially defining other allocation schemes as they come into use.

michielbdejong:
I like the fact that this only describes how to pick a name within the `g.` prefix, and leaves room for naming schemes we will probably invent in the future. I also like the fact that it's a recommendation by example, not a strict list (because that would make this document into a naming authority).

But I don't think ledgers can a priori be grouped into prefixes in terms of routability. For the internet, routing depends mainly on big physical cables. So having a prefix for e.g. Africa makes sense. For ILP, routability depends on the current state of trust lines. A trust line between A and B could be exhausted in one direction if A just sent a lot of money to B. That means routing tables need to update. Next, if a few payments back from B to A occurs, the same trustline might be exhausted in the other direction! And it would simultaneously open up again in the direction A to B. So it's both asymmetric and very fluctuating.

On the other hand, prefixes can be useful for connectors that specialize in certain groups of ledgers. It would make sense to route all traffic for ledgers that announce themselves as geographically Brazilian to a connector that is specialized in the Brazilian market. That connector is then also likely to handle a lot of traffic within the Brazilian prefix, so that way "regions" can exist. This would, however, probably give a bias where inter-region transactions are more expensive.

For decentralized ledgers, different software applications might assign different ledger names, but since there are < 1000 such ledgers, that's probably not a big problem, it would just require two or three entries instead of one then. If that situation changes, we can invent something outside the `g.` namespace. :)

Maybe the only improvement I would make to this naming scheme would be to introduce a clear separation between ledger and account. There can be millions of accounts on one ledger, but we know beforehand that it will be wasteful (as well as harmful to peer performance stats collection, actually) to try to route two accounts differently if they are on one and the same ledger.

I think what could help to understand more of this is a simulation. Maybe it doesn't even have to be with real ilp-kit nodes, we can just invent a routing algorithm, and see how that would fare with a given bulk of traffic in a simulated network.

0015-ilp-addresses/0015-ilp-addresses.md
emschwartz: @justmoon can probably give the best summary of the logic behind restricting the character set. I know that @winthan and Per Lind had some concerns about it though

0015-ilp-addresses/0015-ilp-addresses.md
mDuo13: I agree, but I don't know the "right" answer. Any experts wanna take this one?

0015-ilp-addresses/0015-ilp-addresses.md
mDuo13: I think the routing table behavior should not be part of Connector-Connector Protocol; the routing table behavior is part of routing, and thus part of the transport layer protocol. Several parts of ILP, including Connector-Connector Protocol, depend on how the routing works, but not vice versa.

0015-ilp-addresses/0015-ilp-addresses.md
mDuo13: What does it mean for addresses to be "close" to each other in this context?

emschwartz: Connectors in the same "neighborhoods" are close to one another in the sense that they are likely to be able to route to one another. Connectors can explicitly say they can route to addresses with a different allocation scheme (by broadcasting routes that don't start with `g.`) but I don't think it makes sense to mix schemes.

0015-ilp-addresses/0015-ilp-addresses.md
emschwartz: Correction: routing is not part of the transport layer protocol. I think Transport like it's used in the internet stack is somewhat confusing, because it's actually the layer on top of the addressing layer (IP) that deals with reliability, streams, retries, etc. In ILP, the [Transport layer deals with conditions, memo encryption, retries, etc.](https://github.com/interledger/rfcs/issues/43)

Nevertheless, I agree that the routing table behavior -- or at least some loose recommendation on how connectors should interpret addresses -- should probably be part of the address scheme specification.

sharafian: We want to keep the ILP address as solid as possible though, and routing is one of the areas that we're going to be changing the most during the early stages of ILP.

emschwartz: The route broadcast protocol and how connectors actually figure out their routes will change a lot. However, the address scheme should describe how to interpret the address. Maybe it shouldn't go into so much detail that it's talking about routing tables, but it's probably be better to give too much information in this type of spec than too little.

justmoon:
@mDuo13 Did you see my gist about ILP top level prefixes? https://gist.github.com/justmoon/c1f24b595502251e6a1922a0e2069dd5

I'm not sure if that should be part of this RFC or not. Probably yes.

0015-ilp-addresses/0015-ilp-addresses.md
justmoon: I added a suggestion to get rid of the slightly hacky `main`. https://github.com/interledger/rfcs/issues/132#issuecomment-278185368

0015-ilp-addresses/0015-ilp-addresses.md
justmoon: Very interesting question. Address translation should definitely be up to the ledger plugin imo. Whoever creates the ledger plugin will have a much better idea what would make sense for them than we do.

This is also related to the distinction I made between subledger and account group.

A subledger is a different system (think separate physical servers) whereas an account group is just a subdivision or grouping of a accounts within one ledger (think single database but accounts are grouped together).

To enable account groups I think the ledger would indeed strip off multiple segments and I could see that being useful.

Imagine an accounting ledger which uses `assets` and `liabilities` as account groups. So you could have an address like `us.example.assets.1012` and an address like `us.example.liabilities.7209` and they'd be translated into the local identifiers `assets:1012` and `liabilities:7209`, respectively. (Notice that I used a colon, which might just happen to be the format that this hypothetical accounting ledger uses for its local identifiers.) Nothing seems particularly wrong to me in this picture.

0015-ilp-addresses/0015-ilp-addresses.md
mDuo13: I looked for an ISO standard for continent codes, but there doesn't seem to be one. I used the next closest thing, which is [the list that PHP uses](http://php.net/manual/en/function.geoip-continent-code-by-name.php), which can also be found on [Wikipedia](https://en.wikipedia.org/wiki/List_of_sovereign_states_and_dependent_territories_by_continent_(data_file)) among other sites.

0015-ilp-addresses/0015-ilp-addresses.md
mDuo13: This should be fixed in the latest edit.

0015-ilp-addresses/0015-ilp-addresses.md
michielbdejong: >When we devise other allocation schemes, it's easy to update this spec with more prefixes and the rules of the other allocation schemes. 

Yes, I think that's what I was trying to say: For now, anything outside `g.` is incorrect, until we update this spec to add more schemes. So for now it's good we leave ourselves some unallocated space to do such things in the future.

mDuo13:
OK, so at this point I think we're down to just a few remaining open questions:

1. Where does the routing section belong? In this doc? CCP doc?

2. Should addresses allow zero-length segments? @justmoon suggested that some ledger might allow crazy things like the empty string as an account identifier, but that could screw with routing since `g.crazyledger.` could be the prefix of crazyledger but it also the address of an account on crazyledger. On the other hand, there might be some encoding for invoice IDs or whatnot that would benefit from the ability to have multiple dots in a row.

3. Should we add more payment rails to the spec? Right now I have `ach` and `sepa` only.

4. What kinds of top-level neighborhoods should we recommend? It sounds like people feel like country codes and continent codes may not be very useful. Should we go with [usenet's 'Big 8'](https://en.wikipedia.org/wiki/List_of_newsgroups#The_Big_8_hierarchies) instead?

5. Who decides what the ledger prefix of a ledger is? The connector? The ledger? Someone else?

0015-ilp-addresses/0015-ilp-addresses.md
emschwartz: I see 3 main options:

- Ledger chooses - the operator of a ledger, such as an instance of the Five Bells Ledger, can choose a ledger prefix and have the ledger report that prefix, for example through the API
- Plugin chooses - in the case of a cryptocurrency, the cryptocurrency itself wouldn't "tell you" what it's prefix is. But it would be in the interest of anyone writing a ledger plugin for that ledger to use the same value and hardcode it in their implementation. So when a client or connector asks the plugin for the prefix, it will return that value.
- Connectors choose - in case neither of the previous options are possible, connectors to a given ledger may choose to settle upon a common identifier for that ledger

0015-ilp-addresses/0015-ilp-addresses.md
```diff
@@ -0,0 +1,158 @@
...
+
+***TODO: more rails?***
+
+### Ledger Prefix
+
+The ledger prefix is the unique name of the ledger. When doing a local delivery, the connector typically uses this segment to figure out which ledger plugin to use. As a result, it's difficult (but possible) to make a connector between two ledgers that use the same ledger prefix.
+
+### Subledger
+
+Zero or more segments representing optional divisions within a ledger; for example, if a ledger is composed of partitions that run on different hardware, the partitions could be named here. If there are no fees involved in transferring between a ledger and a subledger, this can still be a "local delivery" from the connector's perspective.
```
emschwartz: This local delivery concept isn't totally clear, so it would be worth adding an explanation at the top or linking to somewhere else.

It's also worth mentioning that subledgers can be used for accounting purposes, for example denoting different divisions and subdivisions within a company. They don't necessarily need to be run on different machines.

0015-ilp-addresses/0015-ilp-addresses.md
```diff
@@ -0,0 +1,158 @@
...
+
+If fees are necessary for connecting to a subledger, payments to that subledger must be routed through a connector.
+
+### Account Identifier
+
+A unique identifier of the account within the ledger. The ledger plugin maps these to accounts within a ledger. For some ledgers, a simple conversion rule may suffice; other ledgers may require a lookup table. The five-bells-ledger-plugin reference implementation translates uses this full segment exactly as the account identifier.
```
emschwartz: Might be worth linking to the plugin

0015-ilp-addresses/0015-ilp-addresses.md
```diff
@@ -0,0 +1,158 @@
...
+
+If a ledger has different groupings of accounts, it might represent them as additional segments that are effectively part of the account identifier.
+
+### Interactions
+
+Additional segments within an address. In most cases, any segments after the account identifier are not expected to match anything in a routing table, so they won't usually affect the routing of a payment. Ledgers and ledger plugins may use the interactions segment of an address when generating notifications, so programs listening for payments can respond differently based on this portion of the address.
```
emschwartz: Arguably that is a kind of routing

0015-ilp-addresses/0015-ilp-addresses.md
```diff
@@ -0,0 +1,158 @@
...
+
+`g.acme.bob` - a destination address to the account "bob" in the ledger "acme".
+
+`g.us.fed.ach.0.acmebank.swx0a0.acmecorp.sales.199.cdfa5e16-e759-4ba3-88f6-8b9dc83c1868.2` - destination address for a particular invoice, which can break down as follows:
+
+- Neighborhoods: `us`, `fed`, `0`
+- Payment rail: `ach`
+- Ledger: `acmebank`
+- Subledger components: `swx0a0`, `acmecorp`
+- Account components: `sales`, `199`
```
emschwartz: It's a bit confusing to me why `sales` is considered an account component instead of a subledger.

0015-ilp-addresses/0015-ilp-addresses.md
```diff
@@ -0,0 +1,158 @@
...
+- Interactions: `cdfa5e16-e759-4ba3-88f6-8b9dc83c1868`, `2`
+
+`g.` - the shortest possible address prefix. Includes all entries that are in the general allocation scheme.
+
+`g.crypto.bitcoin.` - address prefix for the public Bitcoin blockchain
+
+# Routing
+
+_**Rome's note:** I might move this to RFC 0003 eventually._
```
emschwartz: Let's take this out before merging

0015-ilp-addresses/0015-ilp-addresses.md
```diff
@@ -0,0 +1,158 @@
...
+
+Connectors maintain a _routing table_ of ILP addresses. Routing is a recursive lookup through the routing tables of any number of connectors. When a connector receives a query, it finds the [longest prefix match](https://en.wikipedia.org/wiki/Longest_prefix_match) for the queried address. Then, it follows one of the following cases:
```
emschwartz: I think it would be helpful to put this section at the top of the doc

0015-ilp-addresses/0015-ilp-addresses.md
```diff
@@ -0,0 +1,158 @@
...
+## Allocation Schemes
+
+The allocation scheme is the first part of an address, which indicates how the address is assigned. Here is a summary of the prefixes that are currently defined:
+
+| Prefix                       | Allocation Scheme             | Definition and Use Case |
+|:-----------------------------|:------------------------------|:--------------|
+| `g.`                         | [General Allocation Scheme][] | Most ILP addresses used to send and receive real money. |
+| `private.`                   | Private allocation            | For ILP addresses that only have meaning in a private subnet or intranet. Analogous to the [192.168.0.0/16 range in IPv4](https://en.wikipedia.org/wiki/Private_network). |
+| `example.`                   | Examples                      | For "non-real" addresses that are used as examples or in documentation. Analogous to ["555 phone numbers"](https://en.wikipedia.org/wiki/555_%28telephone_number%29) in the USA. |
+| `test1.`, `test2.`, `test3.` | Testing                       | For addresses used in tests, such as unit or integration tests of compatible software. |
```
mDuo13: personally I think that makes sense, but maybe @justmoon can weigh in on whether there's a good reason to have different allocation schemes for testing. Maybe so that you can test rules around allocation schemes themselves?

justmoon: Well, I picked `test1`, `test2`, `test3` for three reasons:

1. As a nod to [RFC 5737](https://tools.ietf.org/html/rfc5737).
2. Because our [integration tests]() currently use `test1` and `test2`.
3. Because people like when things come in threes.

Argument #2 is probably the strongest - you should always try to write your spec such that it describes what you already do, unless you have a strong reason to change it.

0015-ilp-addresses/0015-ilp-addresses.md
```diff
@@ -0,0 +1,158 @@
...
+    | `sa`    | South America |
+    | `oc`    | Oceania       |
+    | `an`    | Antarctica    |
+- For decentralized crypto-currency ledgers, use the `crypto` neighborhood.
+- For naturally-ocurring commodities or other globally-distributed currencies, use the `earth` neighborhood.
+- For currencies strongly associated with a particular international corporation, use the continent of that corporation's headquarters.
+
+It may also be useful to have 1-3 levels of "sub-neighborhoods" to group together related connectors or ledgers. We have no specific recommendations at this time.
+
+### Payment Rail
```
justmoon: Agree, this is just a neighborhood.

There should probably eventually be some standard that defines the top-level neighborhoods under `g.`, but there is no way we can predict what they will end up being, there has to be some amount of organic allocation before we can figure out how to better cluster things together.

0015-ilp-addresses/0015-ilp-addresses.md
justmoon: Subledgers are run on different machines by definition, otherwise they are account groups.

0015-ilp-addresses/0015-ilp-addresses.md
justmoon: Because `sales` and its siblings are physically on the same (atomic) ledger, whereas `acmecorp` and its siblings are physically on different ledgers.

0015-ilp-addresses/0015-ilp-addresses.md
justmoon: @emschwartz pointed out that the current dictionary definition of "subledger" is basically what I was thinking of as an account group: https://en.wikipedia.org/wiki/Subledger

Based on that I agree that we should just combine account groups and subledgers into the same concept.

0015-ilp-addresses/0015-ilp-addresses.md
```diff
@@ -0,0 +1,158 @@
...
+
+1. The address MUST begin with a prefix indicating the allocation scheme. See [Allocation Schemes](#allocation-schemes) for more information.
+2. Each "segment" of the address MUST contain one or more of the following characters:
+    - Alphanumeric characters, upper or lower case. (Addresses are **case-sensitive** so that they can contain data encoded in formats such as base64url.)
+    - Underscore (`_`)
+    - Tilde (`~`)
+    - Hyphen (`-`)
+3. Each segment MUST be separated from other segments by a period character (`.`).
+4. Address prefixes MUST end in a period (`.`) character and MAY contain any number of segments after the allocation scheme prefix.
+6. Destination addresses MUST NOT end in a period (`.`) character, and MUST contain at least two segments after the allocation scheme prefix.
```
sentientwaffle: typo: `5.`, not `6.`

0015-ilp-addresses/0015-ilp-addresses.md
mDuo13: So the idea here is that it's not one or the other, but rather there are fallbacks here:

1. Preferably, the ledger itself defines a prefix to use
2. If the ledger doesn't, connectors using that ledger should come to a consensus regarding what prefix to use
3. If nobody else is using the ledger yet, and the ledger won't choose a prefix, the first connector arbitrarily chooses an appropriate prefix to use

mDuo13:
OK, so here's my to-do list on this one:
1. Explain small ledger addressing (#132) and how you can't tell just from looking at an address which parts are which
2. Finalize a list of "example neighborhoods" to include in the doc. So far there seems to be some agreement on these:
    - `g.crypto.`
    - `g.sepa.`
    - `g.dev.`
3. Move the routing overview to the beginning of the doc
4. "General Allocation" → "Global Allocation" & explain implications (the global allocation scheme is for people who intentionally want to be accessible to everyone else in the global scheme if liquidity and routing permit it)
5. Update wordings on subledger/account group (no distinction between them)
6. Add a "Transport Protocol" section (part of "interactions" really) w/ links to IPR and PSK
7. Update to explain who chooses the ledger prefix
8. Fix typos & remove draft notes

mDuo13:
I've finished my to-do list. This is now, as far as I can see, ready for merging, but please do take another look at it @justmoon @emschwartz et al., in particular the new sections from https://github.com/interledger/rfcs/pull/154/commits/f8bd4b067614c4de074ce678af70ae6c0b995eaa

0015-ilp-addresses/0015-ilp-addresses.md
```diff
@@ -54,105 +61,102 @@ The allocation scheme is the first part of an address, which indicates how the a
...
-    | `as`    | Asia          |
-    | `eu`    | Europe        |
-    | `na`    | North America |
-    | `sa`    | South America |
-    | `oc`    | Oceania       |
-    | `an`    | Antarctica    |
-- For decentralized crypto-currency ledgers, use the `crypto` neighborhood.
-- For naturally-ocurring commodities or other globally-distributed currencies, use the `earth` neighborhood.
-- For currencies strongly associated with a particular international corporation, use the continent of that corporation's headquarters.
+Large neighborhoods can be further subdivided into nested sub-neighborhoods; for example, `dev.luxembourg.` for development ledgers hosted in Luxembourg.
```
emschwartz: I would also mention that the point of neighborhoods is to group parties such that everyone in the same neighborhood is likely connected. If everyone in `dev` knows about everyone else in `dev`, then it's a useful neighborhood. If not, but everyone in `dev.luxembourg` is connected, then that's a more useful neighborhood.

0015-ilp-addresses/0015-ilp-addresses.md
```diff
@@ -54,105 +61,102 @@ The allocation scheme is the first part of an address, which indicates how the a
...
 
-It may also be useful to have 1-3 levels of "sub-neighborhoods" to group together related connectors or ledgers. We have no specific recommendations at this time.
 
-### Payment Rail
+### Ledger Prefixes
```
emschwartz: I think the "ledger prefix" is the whole first part of the address, including the allocation scheme, neighborhoods, and ledger identifier

0015-ilp-addresses/0015-ilp-addresses.md
```diff
@@ -54,105 +61,102 @@ The allocation scheme is the first part of an address, which indicates how the a
...
 
-The payment rail refers to a specific way some accounts or ledgers are connected outside of ILP, such as ACH in the U.S., SEPA in Europe, or other online payment systems such as PayPal or Alipay. We recommend using an all-lowercase abbreviation for the payment rail if one is applicable to the address. Examples:
+A ledger prefix is the unique name of the ledger. When doing a local delivery, the [ilp-connector][] reference implementation uses this segment to figure out which ledger plugin to use. As a result, it's difficult (but possible) to make a connector that directly interacts with two ledgers that use the same ledger prefix.
```
emschwartz: This "difficult (but possible)" comment is confusing. Ledgers should have different identifiers

0015-ilp-addresses/0015-ilp-addresses.md
```diff
@@ -54,105 +61,102 @@ The allocation scheme is the first part of an address, which indicates how the a
...
-### Ledger Prefix
+[ilp-connector]: https://github.com/interledgerjs/ilp-connector
 
-The ledger prefix is the unique name of the ledger. When doing a local delivery, the connector typically uses this segment to figure out which ledger plugin to use. As a result, it's difficult (but possible) to make a connector between two ledgers that use the same ledger prefix.
 
-### Subledger
+### Account Identifiers
 
-Zero or more segments representing optional divisions within a ledger; for example, if a ledger is composed of partitions that run on different hardware, the partitions could be named here. If there are no fees involved in transferring between a ledger and a subledger, this can still be a "local delivery" from the connector's perspective.
+One or more segments that serve as a unique identifier of the account within the ledger. The ledger plugin maps these to accounts within a ledger. For some ledgers, a simple conversion rule may suffice; other ledgers may require a lookup table. The [five-bells-ledger-plugin][] reference implementation translates uses one full segment exactly as the account identifier.
```
emschwartz: typo "translates uses"

0015-ilp-addresses/0015-ilp-addresses.md
```diff
@@ -54,105 +61,102 @@ The allocation scheme is the first part of an address, which indicates how the a
...
 
-If fees are necessary for connecting to a subledger, payments to that subledger must be routed through a connector.
+[five-bells-ledger-plugin]: https://github.com/interledgerjs/ilp-plugin-bells
 
-### Account Identifier
+### Interactions
 
-A unique identifier of the account within the ledger. The ledger plugin maps these to accounts within a ledger. For some ledgers, a simple conversion rule may suffice; other ledgers may require a lookup table. The five-bells-ledger-plugin reference implementation translates uses this full segment exactly as the account identifier.
+Additional segments within an address. Ledgers and ledger plugins may use the interactions segment of an address when generating notifications, so programs listening for payments can respond differently based on this portion of the address. Interactions may be unique per payment or purpose.
```
emschwartz: Rather than "within an address" I would say "at the end of an address"

0015-ilp-addresses/0015-ilp-addresses.md
```diff
@@ -54,105 +61,102 @@ The allocation scheme is the first part of an address, which indicates how the a
...
 
-If a ledger has different groupings of accounts, it might represent them as additional segments that are effectively part of the account identifier.
+To prevent multiple listeners from reacting to incoming payments intended for each other, the "interactions" segment of an address should start with a segment identifying how the payment was planned.
```
emschwartz: Might be useful to note that this concept is analogous to ports in IP

0015-ilp-addresses/0015-ilp-addresses.md
mDuo13: I am inclined to agree, but that makes it hard to refer to "just the last part that identifies this ledger from the other ones in the same neighborhood"

mDuo13:
OK! Maybe this revision is the last one before the merge? ;)

0015-ilp-addresses/0015-ilp-addresses.md
```diff
@@ -0,0 +1,179 @@
...
+
+You can also use the following regular expressions to verify the same requirements:
+
+| Address Type        | Regular Expression                                     |
+|:--------------------|:-------------------------------------------------------|
+| All addresses       | `^(g|private|example|peer|self|test[1-3])\.([a-zA-Z0-9_~-]+\.)*([a-zA-Z0-9_~-]+)?$` |
+| Address prefix      | `^(g|private|example|peer|self|test[1-3])\.([a-zA-Z0-9_~-]+\.)*$` |
+| Destination address | `^(g|private|example|peer|self|test[1-3])\.([a-zA-Z0-9_~-]+\.)+[a-zA-Z0-9_~-]+$` |
+
+(Note that neither the ABNF nor the regular expressions enforce the maximum length requirement. Remember to check that total length of the address is **1023 characters or less**.)
```
emschwartz: This seems like it's asking for people to shoot themselves in the foot. Can't we just provide a regex that starts with `(?=^.{1,1023}$)`? The only issue with this is that Go's regexes don't understand lookaheads. Not sure about Java, but JS, Python, and PHP are all fine with that syntax.

0015-ilp-addresses/0015-ilp-addresses.md
```diff
@@ -0,0 +1,179 @@
...
+segment     = 1*( ALPHA / DIGIT / "_" / "~" / "-" )
+```
+
+You can also use the following regular expressions to verify the same requirements:
+
+| Address Type        | Regular Expression                                     |
+|:--------------------|:-------------------------------------------------------|
+| All addresses       | `^(g|private|example|peer|self|test[1-3])\.([a-zA-Z0-9_~-]+\.)*([a-zA-Z0-9_~-]+)?$` |
+| Address prefix      | `^(g|private|example|peer|self|test[1-3])\.([a-zA-Z0-9_~-]+\.)*$` |
+| Destination address | `^(g|private|example|peer|self|test[1-3])\.([a-zA-Z0-9_~-]+\.)+[a-zA-Z0-9_~-]+$` |
```
dappelt: I propose we change in all three regular expressions the part matching a dot from `\.` to `[.]`. The resulting regular expression is [equivalent](https://regex101.com/r/K3hnzg/2) in terms of matching the same strings, but it is less prone to copy-paste errors. 

In most programming languages, the backslash (`\`) is used as the escape character in string literals. Declaring a string literal with one of the regular expressions from above would lead to the `\.` being interpreted as an escape sequence and the `\` would be removed from the string. In consequence, the regular expression does not match the expected address format. For example:

```javascript
// javascript
const allAddresses = `^(g|private|example|peer|self|test[1-3])\.([a-zA-Z0-9_~-]+\.)*([a-zA-Z0-9_~-]+)?$`
const re = new RegExp(allAddresses)
console.log(re) // /^(g|private|example|peer|self|test[1-3]).([a-zA-Z0-9_~-]+.)*([a-zA-Z0-9_~-]+)?$/
```

With the proposed change, we would avoid the backslash being interpreted as an escape character in the string literal:

```javascript
// javascript
const allAddresses = `^(g|private|example|peer|self|test[1-3])[.]([a-zA-Z0-9_~-]+[.])*([a-zA-Z0-9_~-]+)?$`
const re = new RegExp(allAddresses)
console.log(re) // /^(g|private|example|peer|self|test[1-3])[.]([a-zA-Z0-9_~-]+[.])*([a-zA-Z0-9_~-]+)?$/
```

PR #161, opened 2017-02-22
fix: fix fulfillment typo

sentientwaffle:
The `incoming_fulfill` event should emit a `fulfillment` `String`, not `Buffer`.

(see https://github.com/interledger/rfcs/pull/160)


justmoon:
Agree with this PR because it reflects what's implemented. Doesn't mean I'm not open to arguments if someone thinks we should change this API to be all buffers.

PR #162, opened 2017-02-24
Differentiate between throwing and rejecting

michielbdejong:
The following two functions are not the same, the spec should be specific about that:
```
function thrower() {
  if (problem) {
    throw new Error('oops');
  }
  return Promise.resolve('success');
}
function rejecter() {
  if (problem) {
    return Promise.reject(new Error('oops'));
  }
  return Promise.resolve('success');
}
```


michielbdejong:
It's a bit ambiguous where the definition of the `Transfer` class defines what the ledger plugin should do, but assuming this is talking about functions that return promises, I used reject and not throw there.

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
```diff
@@ -233,7 +233,7 @@ For a detailed description of these properties, please see [`OutgoingTransfer`](
 <code>ledgerPlugin.sendMessage( **message**:[OutgoingMessage](#outgoingmessage) ) ⇒ Promise.&lt;null></code>
 
 Plugin must be connected, otherwise the promise should reject. Send a ledger-local message.
-If there is a problem with the structure or validity of the message, then `sendMessage` should throw an error in the form of a rejected promise.
+If there is a problem with the structure or validity of the message, then `sendMessage` should rejects with an error.
```
sharafian: should reject with an error, instead of should rejects with.

michielbdejong: fixed

Issue #146, opened 2017-01-08
Proposal: we need an Interledger Packet -- not just an ILP Header


emschwartz:
In https://github.com/interledger/rfcs/issues/127, https://github.com/interledger/rfcs/issues/138, and https://github.com/interledger/rfcs/issues/137, https://github.com/interledgerjs/ilp-connector/pull/290 I have been pushing for us to define an Interledger Packet, as opposed to treating the ILP header and "extra headers" as separate.

This case boils down to:
- Enabling end-to-end protocols
- Making ILP easier to grok

**Enabling end-to-end protocols**
In https://github.com/interledger/rfcs/issues/43#issuecomment-271187813 I laid out various possibilities for Interledger Transport Layer protocols.

In order to make it easy to define and implement such protocols, we need a common way to define extra headers AND we need to ensure that the extra headers will be relayed by connectors. Right now the connector [only relays the ILP header](https://github.com/interledgerjs/ilp-connector/pull/290), meaning that additional protocol headers would need to be included in a data field in the ILP header, which doesn't seem like the best approach in a JSON format.

In addition to defining Transport Layer protocols built on ILP, I think we will also want to build Application Layer protocols directly on top of ILP. The Simple Payment Setup Protocol is a bit of an odd case because it uses the Internet stack + an ILP transport protocol, instead of being built purely on top of the ILP stack. However, I could envision us building application layer protocols that are entirely built on ILP (e.g. a way to include identity or other payment metadata in a payment sent with a sender-initiated transport layer protocol).

If connectors want to define other headers to pass along on a single hop (as mentioned by @justmoon in https://github.com/interledgerjs/ilp-connector/pull/290#pullrequestreview-15379504), they could include those alongside the ILP Packet in the transfer memo, or they could add them to the ILP packet's headers.

**Grokking Interledger**
The Internet Protocol can be quickly understood by looking at the IP Address and the IP Packet. Without understanding any of the other details it makes sense that it is a protocol for transmitting packets to given addresses.

Similarly, it would be great to be able to understand ILP as a protocol for transmitting Interledger Packets to a given Interledger Address. I want to be able to point to a _thing_ like an address and a packet format and say "this is Interledger".

I think it would also be nice to have a single format for encoding the packet, because it would mean one less thing for implementors to think about and one fewer place for bugs or vulnerabilities to arise. 

[...]

PR #163, opened 2017-03-06
Update architecture spec & misc updates

justmoon:
Resolves https://github.com/interledger/rfcs/issues/146


0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -194,39 +199,13 @@ Transport layer protocols are responsible for coordinating the different transfe
...
+#### Privately Shared Key (PSK)
 
-##### As a Sub-Protocol
+<span class="show alert alert-warning">**TODO:** Describe PSK.</span>
 
-ATP can be used as a sub-protocol for part of a UTP transaction. This happens when a connector decides to create an ATP transfer as the next hop. It takes on both the risk of being unable to pass on the receipt and the risk of the notaries failing. This makes sense to do if the next connector quotes a better price for an ATP transfer than for a UTP transfer.
+#### Interledger Payment Request (IPR)
 
-Similarly, an ATP transaction can turn into a UTP transaction when a connector decides to use UTP for the next hop. It takes on the risk of not being able to pass on the receipt to the notaries. This makes sense if ATP is not available.
+<span class="show alert alert-warning">**TODO:** Describe IPR.</span>
```
emschwartz: Shouldn't we just link to the IPR RFC?

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -98,47 +100,46 @@ Ledger protocols are responsible for executing the individual transfers that con
...
+* Prepared -- Funds are held.
 * Executed -- The transfer has completed.
 * Rejected -- The transfer has been canceled (and funds returned to the sender.)
 
-<span class="show alert alert-info">**Hint:** Escrow is the financial equivalent of a [two-phase commit](http://foldoc.org/two-phase%20commit).</span>
+<span class="show alert alert-info">**Hint:** Authorization holds are the financial equivalent of a [two-phase commit](http://foldoc.org/two-phase%20commit).</span>
 
-
-The ledger MUST be able to verify a simple SHA-256 hashlock cryptographic escrow condition in order to automatically trigger the release of the escrow. It must also allow a timeout to automatically roll back an escrowed transfer.
+The ledger MUST be able to release the held funds to the receiver upon receiving a 32-byte preimage whose SHA-256 hash matches a value provided by the sender. It MUST also support releasing the funds back to the sender after a timeout.
```
sharafian: For a ledger to be supported, it should accept ONLY 32-byte pre-images. Otherwise a connector could stand to lose money, if they have an outgoing transfer on a ledger with a larger fulfillment size than the ledger of the corresponding incoming transfer.

emschwartz: I'd say that it can accept more (in general) but it needs a way to ensure that for a given transfer the fulfillment will only be considered valid if it is exactly 32 bytes (e.g. like the cost limit in Crypto Conditions)

sharafian: fair enough. rather than saying the ledger supports only 32-byte preimages, we should say you need a way to construct a transfer that can only be fulfilled by a 32-byte preimage

sharafian: is it possible to do that on bitcoin with some kind of bitcoin scripts magic?

stevenroose: @sharafian yeah. It's basically an HTLC like Lightning uses them. Except it's not from a two-party escrow address but just a one-to-one HTLC.

adrianhopebailie: @sharafian if we want to leave the door open for other sizes then we should say that the ledger must allow a client to specify the expected size of the fulfillment when requesting the transfer. If that is not provided it must be assumed to be 32 bytes.

0001-interledger-architecture/0001-interledger-architecture.md
justmoon: We'll definitely link it, but the previous version of this doc had a short paragraph for each protocol / component, which was quite useful I thought.

0001-interledger-architecture/0001-interledger-architecture.md
emschwartz: @stevenroose I know Lightning uses HTLCs with 32-byte preimages, but do you know if the scripts they use _guarantee_ that the preimage could not be longer than 32 bytes?

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -98,47 +100,46 @@ Ledger protocols are responsible for executing the individual transfers that con
...
 
 The ledger MUST support attaching a short message or *memo* to each transfer.
 
 #### Full Support
```
emschwartz: Should we mention speed here? If you want to fully support ILP the ledger should be able to prepare and execute transfers in ~1 second or less.

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -98,47 +100,46 @@ Ledger protocols are responsible for executing the individual transfers that con
 
 #### Minimal Support
 
-For minimal Interledger support, the ledger MUST have the ability for funds to be transferred from one account to another. This enables OTP transactions to pass through this ledger. It also allows third parties to act as escrow providers in order to enable UTP and ATP transactions on the ledger.
+For minimal Interledger support, the ledger MUST have the ability for funds to be transferred from one account to another. All of the more advanced ledger layer functionality, such as escrowed transfers, messaging, etc. can be implemented on top of this type of basic ledger.
 
 #### Basic Support
```
emschwartz: What about mentioning notifications?

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -98,47 +100,46 @@ Ledger protocols are responsible for executing the individual transfers that con
...
-It MUST support all cryptographic condition types with the status "recommended".
-
-It MUST support memos to be added for the credited and debited accounts separately and SHOULD support fairly large memo sizes.
+It MUST support memos up to 65535 bytes.
 
-It MUST support a way to discover connectors connected to the ledger.
+It MUST support sending an authenticated message of up to 65535 bytes to the holder of another account on the ledger.
 
-It MUST support a way to look up a fulfillment by condition hash. It SHOULD automatically reject new transfers (that have not been prepared yet) that have an execution condition for which the ledger already knows the fulfillment. This aids in [UTP error recovery](#error-recovery).
+It MUST support a way to look up a fulfillment by condition hash. It SHOULD automatically reject new transfers (that have not been prepared yet) that have an execution condition for which the ledger already knows the fulfillment. This aids in [error recovery](#error-recovery).
```
adrianhopebailie: We should mention that it is necessary for a ledger protocol to define an addressing scheme that complies with the ILP Addressing scheme such that accounts in the ledger can be addressed using ILP Addresses.

See: https://github.com/interledger/rfcs/issues/31#issuecomment-286469564

0001-interledger-architecture/0001-interledger-architecture.md
emschwartz: What about enabling off-ledger, ledger-secured payment channels? That might be for more-than-full support

0003-interledger-protocol/0003-interledger-protocol.md
```diff
@@ -172,67 +131,66 @@ As with the [internet protocol](https://tools.ietf.org/html/rfc791#section-2.3),
...
-| destinationAddress | IlpAddress | Address corresponding to the destination account |
-| destinationAmount | IlpAmount | Amount the destination account should receive, denominated in the asset of the destination ledger |
-| condition | IlpCondition | Execution condition for the transfers
-| expiresAt | IlpTimestamp | Maximum expiry time of the last transfer that the recipient will accept |
-
-#### version
-
-    INTEGER(0..255)
+| amount | UInt64 | Amount the destination account should receive, denominated in the asset of the destination ledger |
+| account | Address | Address corresponding to the destination account |
```
michielbdejong: Why the change from 'address' to 'account'? Will this same change be made in the [IPR](https://github.com/interledger/rfcs/blob/master/0011-interledger-payment-request/0011-interledger-payment-request.md#json) format?

sharafian: IPR is (probably) going to use the data format from the new version of PSK once we switch to the binary ILP packet format.

Paraphrasing @justmoon 's reasoning on `account` vs. `address`, the account is the actual thing being represented, while an address is merely the data format of it (like calling the amount "number")

mDuo13:
From https://github.com/interledger/java-ilp-core/pull/18#discussion_r106468355 —

It's confusing that ILP, or the Interledger Protocol (Interledger Project?), is actually a protocol suite containing a protocol specifically named "Interledger Protocol" (IL-RFC-3). We should strongly consider renaming IL-RFC-3 to make this less confusing. I don't have any good candidates for what else to call it, though.

adrianhopebailie:
> It's confusing that ILP, or the Interledger Protocol (Interledger Project?), is actually a protocol suite containing a protocol specifically named "Interledger Protocol" (IL-RFC-3)

This RFC is called the "Interledger architecture" (i.e. not the Interledger Protocol architecture). Maybe the problem is that we use the acronym ILP too liberally when it should only be used to describe the Interledger Protocol (i.e. IL-RFC-3).

i.e. The Interledger Architecture defines a number of protocol layers including the Interledger Protocol (ILP) layer.

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -98,47 +100,46 @@ Ledger protocols are responsible for executing the individual transfers that con
...
 
-The ledger MUST provide cryptographic escrow. Escrow means that transfers can have four states:
+The ledger MUST provide cryptographic authorization holds. During an authorization hold, money is put aside for a specific transfer until that transfer's outcome has been decided.
+
+Transfers using authorization holds can be in four distinct states:
 
-![enter image description here](https://lh3.googleusercontent.com/-QHsehrPTXgM/Vu_zs1dLWPI/AAAAAAAAEys/zchHKJj_MQs/s0/Transfer%252520States.png "Transfer States")
+![Transfers start in the proposed state, transition to the prepared state and finally to the executed state, which is final. Proposed and prepared transfers can also transition to the rejected state, which is final as well.](../shared/graphs/transfer-states.svg)
 
 * Proposed -- Nothing has happened yet.
```
emschwartz: Can we take out the `Proposed` state?

0001-interledger-architecture/0001-interledger-architecture.md
emschwartz: Also, what about account notifications?

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -98,47 +100,46 @@ Ledger protocols are responsible for executing the individual transfers that con
...
-#### Simple Ledger Protocol (SLP)
+#### Five Bells Ledger Protocol (5BLP)
 
-Simple Ledger Protocol (SLP) is a RESTful, JSON-based protocol that was developed specifically to provide the minimum functionality required for full Interledger support.
+Five Bells Ledger Protocol (5BLP) is a RESTful, JSON-based protocol that was developed specifically to provide the minimum functionality required for full Interledger support.
 
-A reference implementation of a ledger using SLP can be found [here](https://github.com/interledger/five-bells-ledger).
+A reference implementation of a ledger using 5BLP can be found [here](https://github.com/interledger/five-bells-ledger).
 
 #### Blockchain Protocols (e.g. Bitcoin)
```
emschwartz: Remove old mention of OTP/UTP in the bitcoin section

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -48,7 +48,7 @@ A basic objective of the Interledger design is to tolerate a wide range of ledge
```
emschwartz: This spec should no longer say "**This is a strawman proposal.** Choices made in this document are merely intended as a starting point." at the top

0001-interledger-architecture/0001-interledger-architecture.md
justmoon: Agree that we can remove `Proposed`, but in its place we should add a name for the initial state, i.e. where the ledger does not know about the transfer yet.

0001-interledger-architecture/0001-interledger-architecture.md
justmoon: > do you know if the scripts they use guarantee that the preimage could not be longer than 32 bytes?

Yes. In fact [their Bitcoin script](https://github.com/lightningnetwork/lightning-rfc/blob/master/03-transactions.md#received-htlc-outputs) checks that the preimage is **exactly** 32 bytes, so 31 bytes is disallowed in Lightning.

Relevant snippet is:
```
OP_SIZE 32 OP_EQUAL
```

This means: "Add the size [of the preimage] to the stack, add 32 to the stack, check that the top two things on the stack are equal."

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -1,11 +1,9 @@
 # Interledger Architecture
 
-This document outlines the Interledger architecture and explains how the different layers relate to each other.
+This document outlines the Interledger architecture and protocol suite and explains how the different layers relate to each other. The [Interledger Protocol (ILP)](#interledger-protocol-ilp) is one layer in the Interledger Architecture.
```
justmoon: Saw the same wording in the index. Minor issue, but "architecture and protocol suite" reads weird to me, since the decision to split it into multiple protocol layers is just a feature of the architecture. So it's kinda like saying: "I like this car and sports car." Not really wrong, just a bit strange way to say it.

Like the clarifying sentence though on the relationship between ILP and Interledger architecture.

Would lower case "architecture".

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -217,15 +229,24 @@ Application layer protocols deal with the exchange of payment details and associ
 
 The Simple Payment Setup Protocol (SPSP) is an application layer protocol for negotiating payment details. SPSP handles account and amount discovery, condition creation, quoting and setup. SPSP uses Webfinger ([RFC 7033](https://tools.ietf.org/html/rfc7033)) and an HTTP-based protocol for querying account and amount details, [ILQP](#interledger-quoting-protocol-ilqp) for quoting, and [ILP](#interledger-protocol-ilp) for payment execution.
 
-The protocol is described in [IL-RFC 9](../0009-simple-payment-setup-protocol/).
+The protocol is described in [IL-RFC 9](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md).
```
justmoon: Note that we have to update the script that renders these specs for interledger.org. Currently, this link will link you to the unrendered `.md` file.

(But it is better to fix the link for Github users' benefit and then just rewrite it back as part of the rendering process for the website.)

0003-interledger-protocol/0003-interledger-protocol.md
```diff
@@ -172,78 +131,66 @@ As with the [internet protocol](https://tools.ietf.org/html/rfc791#section-2.3),
...
 ```
 
-Care must be taken in mapping interledger addresses to local ledger accounts. Examples of address mappings may be found in "Address Mappings" ((TODO)).
+More information about ILP addresses can be found in the [ILP Address Specification](../0015-ilp-addresses/).
+
+The mapping from addresses to local accounts on a ledger is defined by the ledger protocol.
 
 ### Connectors
 
 Connectors implement the interledger protocol to forward payments between ledgers. Connectors also implement the [Connector to Connector Protocol (CCP)](../0010-connector-to-connector-protocol/) to coordinate routing and other interledger control information.
```
mDuo13: connectors also implement (or include a module that implements) the ledger protocol of the ledgers where they have accounts

0003-interledger-protocol/0003-interledger-protocol.md
```diff
@@ -172,78 +131,66 @@ As with the [internet protocol](https://tools.ietf.org/html/rfc791#section-2.3),
...
 
-![Interledger is an overlay across ledgers](assets/ilp-ledger-relation.png)
-
 ## Specification
 
-### ILP Header Format
+### ILP Payment Format
```
mDuo13: Change this to "ILP Payment Packet Format"

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -98,53 +98,62 @@ Ledger protocols are responsible for executing the individual transfers that con
...
 
-It MUST support a way to discover connectors connected to the ledger.
+The ledger MUST support a way to look up a fulfillment by condition hash. It SHOULD automatically reject new transfers (that have not been prepared yet) that have an execution condition for which the ledger already knows the fulfillment. This aids in [error recovery](#error-recovery).
 
-It MUST support a way to look up a fulfillment by condition hash. It SHOULD automatically reject new transfers (that have not been prepared yet) that have an execution condition for which the ledger already knows the fulfillment. This aids in [UTP error recovery](#error-recovery).
+The ledger MUST support preparing, executing, and rejecting transfers in 1 second or less.
+
+The ledger MUST define an [ILP Address](../0015-ilp-addresses/0015-ilp-addresses.md) prefix and scheme such that accounts on the ledger can be addressed using canonical ILP addresses.
+
+The ledger MUST support off-ledger, ledger-secured payment channels.
```
emschwartz: I think we should actually just take that out for now. The role of payment channels in ILP is something that we need to think more about and it's strange to call it a requirement for a ledger when we don't have a solid description of what exactly a ledger would need to do to fulfill that.

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -98,53 +98,62 @@ Ledger protocols are responsible for executing the individual transfers that con
...
-#### Simple Ledger Protocol (SLP)
+#### Five Bells Ledger Protocol (5BLP)
 
-Simple Ledger Protocol (SLP) is a RESTful, JSON-based protocol that was developed specifically to provide the minimum functionality required for full Interledger support.
+Five Bells Ledger Protocol (5BLP) is a RESTful, JSON-based protocol that was developed specifically to provide the minimum functionality required for full Interledger support.
 
-A reference implementation of a ledger using SLP can be found [here](https://github.com/interledger/five-bells-ledger).
+A reference implementation of a ledger using 5BLP can be found [here](https://github.com/interledger/five-bells-ledger).
 
 #### Blockchain Protocols (e.g. Bitcoin)
```
emschwartz: Why? All the other things in that section have "(e.g. ...)" in the header

PR #165, opened 2017-03-09
docs(0004,0008) RejectionMessage; use integer amounts

sentientwaffle:


0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
```diff
@@ -460,12 +461,17 @@ ILP Address prefix of the ledger that this transfer is going through on.
 #### amount
 <code>**amount**:String</code>
 
-A decimal amount, represented as a string. MUST be positive. The supported precision is defined by each ledger plugin and can be queried by the host via [`getInfo`](#getinfo). The ledger plugin MUST throw an `InsufficientPrecisionError` if the given amount exceeds the supported level of precision.
+An integer amount, represented as a string. MUST be positive. The supported precision is defined by each ledger plugin and can be queried by the host via [`getInfo`](#getinfo). The ledger plugin MUST throw an `InsufficientPrecisionError` if the given amount exceeds the supported level of precision.
```
emschwartz: Is this precision issue still relevant with the integer amounts?

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
```diff
@@ -400,8 +400,9 @@ left undefined (but not any other false-y value) if unused.
 | `String` | [from](#from) | ILP Address of the source account |
 | `String` | [to](#to) | ILP Address of the destination account |
 | `String` | [ledger](#ledger) | ILP Address prefix of the ledger |
-| `String` | [amount](#amount) | Decimal transfer amount |
-| `Object` | [data](#data) | Data packet or memo to be sent with the transfer, starts with an ILP header |
+| `String` | [amount](#amount) | Integer transfer amount |
+| `String` | [ilp](#ilp) | Base64-encoded ILP header |
```
emschwartz: It should be ILP Packet instead of header

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
```diff
@@ -460,12 +461,17 @@ ILP Address prefix of the ledger that this transfer is going through on.
...
 
-A decimal amount, represented as a string. MUST be positive. The supported precision is defined by each ledger plugin and can be queried by the host via [`getInfo`](#getinfo). The ledger plugin MUST throw an `InsufficientPrecisionError` if the given amount exceeds the supported level of precision.
+An integer amount, represented as a string. MUST be positive. The supported precision is defined by each ledger plugin and can be queried by the host via [`getInfo`](#getinfo). The ledger plugin MUST throw an `InsufficientPrecisionError` if the given amount exceeds the supported level of precision.
+
+#### ilp
+<code>**ilp**:String</code>
+
+An [ILP header](../0003-interledger-protocol/), denoting the payment's final destination.
 
 #### data
```
emschwartz: What is `data` going to be used for now that we have a separate `ilp` section?

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
justmoon: No.

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
justmoon: Nothing right now, and it should probably be removed. I do think there might be a place for having a place to put data that is meant for the next connector but isn't necessarily added by the original sender and therefore can't be in the ILP packet. (Not sure what I would call that though - we can probably come up with something better than `data`?)

For example: Let's say as a connector I want next-hop connectors to notify me when they've prepared their outgoing transfer. Maybe I want to use that information for risk management purposes or whatever. So it would be nice if I could attach something to the transfer which says "please notify me when you've prepared this".

Perhaps this new field can be just another buffer and we define the exact format later?

adrianhopebailie: ILP Header > ILP Packet

> we can probably come up with something better than data
memo?

> Perhaps this new field can be just another buffer and we define the exact format later?
+1

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
```diff
@@ -155,7 +155,7 @@ The mapping from the ILP address to the local ledger address is dependent on the
 #### getBalance
 <code>ledgerPlugin.getBalance() ⇒ Promise.&lt;String></code>
 
-Return a decimal string representing the current balance. Plugin must be connected, otherwise the promise should reject.
+Return a integer string representing the current balance. Plugin must be connected, otherwise the promise should reject.
```
sappenin: Minor nit.  Change `Return a integer` to `Return an integer`?

sentientwaffle:
@justmoon @emschwartz updated

PR #168, opened 2017-03-14
ASN.1: Add final ILP Payment Packet format (and drafts of ILQP packets)

justmoon:


asn1/GenericTypes.asn
```diff
@@ -0,0 +1,66 @@
+GenericTypes
+DEFINITIONS
+AUTOMATIC TAGS ::=
+BEGIN
+
+-- Small Integers
+Int8 ::= INTEGER (-128..127)
```
adrianhopebailie: Do we need all of these types anymore? I appreciate the value of the comments for documenting how the design was finalized but perhaps we can put that somewhere else to keep this clean?

asn1/InterledgerTypes.asn
```diff
@@ -0,0 +1,165 @@
...
+-- A standard interledger address
+Address ::= IA5String
+    (FROM
+        ( hyphen
+        | period
+        | "0".."9"
+        | "A".."Z"
+        | underscore
+        | "a".."z"
+        | tilde )
```
emschwartz: Really not very important, but why are the categories ordered in this way? I would expect alphanumeric characters to come before the symbols

asn1/InterledgerTypes.asn
```diff
@@ -0,0 +1,165 @@
...
+    )
+    (SIZE (1..1023))
+
+-- --------------------------------------------------------------------------
+
+-- Amounts are represented as IEEE 754 double-precision floating point.
```
emschwartz: This section needs to be replaced with an explanation of the choice of UInt64's and the fact that those represent the minimum divisible unit on the destination ledger

asn1/InterledgerTypes.asn
```diff
@@ -0,0 +1,165 @@
...
+-- applied to. Since binary floating point math is much more efficient, it
+-- makes sense to do all of the calculations consistently in binary and only
+-- round to decimals for amounts that humans will actually see.
+--
+-- Note that subnormal amounts as well as special amounts other than positive
+-- zero (-Infinity, +Infinity, -0, NaN, ...) are not allowed in Interledger.
+
+Amount ::= UInt64
+
+-- Selection of other options considered:
```
emschwartz: I think this can be taken out

asn1/InterledgerTypes.asn
```diff
@@ -0,0 +1,165 @@
...
+-- })
+
+-- Amount ::= SEQUENCE {
+--  mantissa UInt64,
+--  exponent Int8
+-- }
+
+-- --------------------------------------------------------------------------
+
+-- Different ledgers may use different rounding parameters
```
emschwartz: Can this section be taken out?

asn1/InterledgerTypes.asn
```diff
@@ -0,0 +1,165 @@
...
+--
+-- The curve is expressed as a series of points given as <x, y> coordinates.
+--
+-- Each x coordinate corresponds to an input amount, each y corresponds to an
+-- output amount. Each point must have larger x and y values than the previous
+-- point. All coordinates must be greater than zero.
+--
+-- The only amounts
+
+LiquidityCurve ::= SEQUENCE OF SEQUENCE {
```
emschwartz: Why do we call this a liquidity curve instead of a rate curve?

asn1/InterledgerTypes.asn
```diff
@@ -0,0 +1,165 @@
...
+-- Liquidity curves describe the relationship between input and output amount
+-- for a given path between a given pair of ledgers.
+--
+-- The curve is expressed as a series of points given as <x, y> coordinates.
+--
+-- Each x coordinate corresponds to an input amount, each y corresponds to an
+-- output amount. Each point must have larger x and y values than the previous
+-- point. All coordinates must be greater than zero.
+--
+-- The only amounts
```
emschwartz: Oh the suspense! "The only amounts"...what?

asn1/InterledgerTypes.asn
```diff
@@ -0,0 +1,165 @@
...
+-- 8601.
+
+Timestamp ::= GeneralizedTime
+
+-- --------------------------------------------------------------------------
+
+-- Liquidity curves describe the relationship between input and output amount
+-- for a given path between a given pair of ledgers.
+--
+-- The curve is expressed as a series of points given as <x, y> coordinates.
```
emschwartz: If `x` and `y` need to be explained in this context, what do you think about just renaming them to something like `inputAmount` and `outputAmount`?

asn1/InterledgerTypes.asn
justmoon: They are ordered by value, i.e. their order in the ASCII table. E.g. `'A' == 0x41`, so it comes before `'_' == 0x5F`, which in turn comes before `'a' == 0x61`

asn1/InterledgerTypes.asn
justmoon: The rate is the slope of this curve, so the "rate curve" would be the derivative of this curve. Open to another name if anyone has ideas.

asn1/InterledgerTypes.asn
```diff
@@ -0,0 +1,165 @@
...
+--
+-- Each x coordinate corresponds to an input amount, each y corresponds to an
+-- output amount. Each point must have larger x and y values than the previous
```
mDuo13: Are you sure both x and y have to be larger? Can you not have a "curve" that plateaus in the middle, e.g.:

```
   y₃ ┨         ╭┄
      ┃        ╱
      ┃       ╱
y₁,y₂ ┨  ╭╌╌╌╯
      ┃ ╱
   y₀ ┨╯
      ┗━━┳━━━┳━━┳━━
     x₀ x₁   x₂ x₃
```

(in this case y₁=y₂ but x₂>x₁)

asn1/GenericTypes.asn
justmoon: I like having a full suite of generic types, in case we need them in the future. And I like having them in the repo to avoid external dependencies. So my vote is to keep this. It's in a separate file, so I don't see how it would be distracting.

asn1/InterledgerProtocol.asn
```diff
@@ -0,0 +1,31 @@
+InterledgerProtocol
+DEFINITIONS
+AUTOMATIC TAGS ::=
+BEGIN
+
+IMPORTS
+    UInt64
```
justmoon: I was thinking about that. If there is an amount type, we should use it everywhere obviously. But if `Amount` is just `UInt64` I think it's more readable if we just put `UInt64` everywhere. For complex types, sure, define it once. But if it's just a primitive type, don't make me go to another file just to find out what it is. What do you think?

asn1/InterledgerQuotingProtocol.asn
```diff
@@ -2,40 +2,76 @@ InterledgerQuotingProtocol DEFINITIONS AUTOMATIC TAGS ::=
...
+    requestId UInt128,
+    -- Curve describing the liquidity (relationship between input and output
+  -- amounts) for the quoted route
+    liquidity LiquidityCurve,
+    -- Common prefix of all addresses for which this liquidity curve applies
+    -- SHOULD end with a period "." (but clients MUST NOT enforce this)
+    destinationPrefix Address,
+    -- How long the sender should put money on hold (in milliseconds)
+    sourceHoldDuration UInt32
+    -- TODO: An exact quote expiry doesn't seem necessary...
```
justmoon: Will change to include expiry date.

asn1/InterledgerQuotingProtocol.asn
```diff
@@ -2,40 +2,76 @@ InterledgerQuotingProtocol DEFINITIONS AUTOMATIC TAGS ::=
...
-	-- provided in the request.
-	requestedAmount Amount,
-	sourceHoldDuration IlpHoldDuration
+-- Quote with a specific source or destination amount
+QuoteBySourceAmountRequest ::= SEQUENCE {
+    requestId UInt128,
+    destinationAccount Address,
+    sourceAmount Amount,
+    -- How much time the receiver needs to fulfill the payment (in milliseconds)
+    destinationHoldDuration UInt32
```
justmoon: Not sure if we should care about that use case. You will always want to hold it for the minimum amount of time. If you don't know how long the recipient needs, you can probably make a reasonable guess or just ask them.

It just doesn't make much sense why you would choose the destination hold based on the sender's willingness to hold. It's like buying paint based on how much they have at the store instead of estimating how much you need.

asn1/InterledgerTypes.asn
```diff
@@ -0,0 +1,165 @@
...
+hyphen IA5String ::= "-"
+period IA5String ::= "."
+underscore IA5String ::= "_"
+tilde IA5String ::= "~"
+
+-- A standard interledger address
+Address ::= IA5String
+    (FROM
+        ( hyphen
+        | period
```
justmoon: I don't think that's possible in ASN.1, but am happy to be proven wrong by someone with better ASN.1-fu.

asn1/InterledgerTypes.asn
justmoon: In that case, the price is zero/infinity. My initial intuition was that that could result in some mathematical funkiness. But I haven't been able to think of a case. In crypto, you always disallow zero basically everywhere, so it's become my default habit.

The current implementation does allow equal y coordinates, but no equal x coordinates. That means that inverting a curve may give you an invalid curve. Not sure if that matters since we don't really invert curves.

I'd be ok to write what the current implementation does into the spec, but would caution us to think about this point some more.

mDuo13: I think it makes sense that you may have a plateau in the curve. (I haven't thought too deeply into this, but that might be meaningful for assets that have some weird external costs involved in transferring them, or some sort of transition point between transfer methods.) I'd just reword it to say that the x values must always be greater than the previous and the y values must be equal or greater.

justmoon:
@adrianhopebailie @emschwartz @mDuo13 Updated to address your comments.

I used `UInt64` as the amount format everywhere rather than giving it an alias. It saves readers from having to look up what `Amount` is.

That meant taking out the explanation of why we chose `UInt64` as the amount because there wasn't an obvious place to put it. (Didn't want to clutter the ILP Packet format with a long-winded, slightly off-topic explanation of the amount data type.) I think that's fine, we can explain it in the ILP RFC instead.

asn1/InterledgerProtocol.asn
```diff
@@ -7,8 +7,7 @@ IMPORTS
     UInt64
     FROM GenericTypes
 
-    Address,
-    Amount
+    Address
     FROM InterledgerTypes
 
     MetadataEntry
```
emschwartz: `MetadataEntry` should be removed

Issue #152, opened 2017-01-30
RFC9: What is SPSP?


sharafian:
Right now, the version of SPSP [in the RFCS](https://github.com/interledger/rfcs/blob/master/0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md) isn't the same as the version of SPSP being used in production (in the ILP-Kit).

The ILP Kit's implementation rolls the query step and the optional webfinger lookup into an HTTP get that returns the query information as a webfinger resource.

We should have a discussion over what SPSP should look like.

## Open Questions

0. How should the SPSP methods be divided up?
0. How should invoices work?
0. What HTTP method should setup be?
0. Can you query with a URL instead of a webfinger `acct`?
0. Should SPSP use [SSP](https://github.com/interledger/rfcs/issues/43#issuecomment-274274862) or [IPR](https://github.com/interledger/rfcs/blob/master/0011-interledger-payment-request/0011-interledger-payment-request.md)?

## Use Case

SPSP is for the case where:

0. The sender and receiver don't have a pre-established data channel.
0. The receiver has an HTTP server or has an HTTP server acting on their behalf.
0. The receiver wants to accept exactly one payment.
0. One or two HTTP round trips per payment are acceptable.

From an architectural point of view:

0. The receiver wants to perform logic at setup-time, not at fulfill-time.
0. The receiver wants to accept payments in a stateless way.
0. The HTTP server that gives out IPR packets has the same HMAC key as the receivers that fulfill payments, but the HTTP server doesn't necessarily have ledger credentials.
0. The client doesn't need any crypto libraries (for IPR).

# Proposal

## SPSP flow (from the client's point of view)

I've chosen to split the SPSP flow into two phases.
Whether an API chooses to split the flow up into these two phases is an implementation detail.

### Quote-Query Phase

0. GET webfinger endpoint corresponding to receiver (`alice@example.com` -> `https://example.com/.well-known/webfinger?resource=acct:alice@example.com`, OR `https://example.com/receiverid/12414` -> `https://example.com/.well-known/webfinger?resource=https://example.com/receiverid/12414`)
0. (If source-amount quoted) Using the ILP address from query result, make an ILQP quote to get destination amount.
0. (If dest-amount quoted) Using the ILP address from query result, make an ILQP quote to get source amount.

Now you have all the information required to set up and send the payment. A client implementation could represent this as a quote-query object that can be validated by client-side logic. A uuid on the quote-query object could make the send-setup phase idempotent.

[...]

PR #166, opened 2017-03-13
Docs: add PSK details spec

sharafian:
Starting write-up for PSK details format, similar to the HTTP request format. From @emschwartz and @stefan

PR #169, opened 2017-03-21
RFC-9: SPSPv2 (switch to PSK)

emschwartz:
* use binary ILP packet format
* use PSK instead of IPR
* remove the payee/invoice distinction
* switch to integer amounts
* receiver info can be used to send multiple payments
* "receiver" now only refers to the entity receiving the payment, not the SPSP endpoint
* switch Webfinger rel to https://interledger.org/rel/spsp/v2

Depends on https://github.com/interledger/rfcs/pull/166
Resolves https://github.com/interledger/rfcs/issues/152

### DIFF
--- a/0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
+++ b/0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
@@ -2,17 +2,17 @@
 
 ## Preface
 
-This document describes the Simple Payment Setup Protocol (SPSP), a basic protocol for exchanging payment information between senders and recipients to set up an Interledger payment.
+This document describes the Simple Payment Setup Protocol (SPSP), a basic protocol for exchanging payment information between senders and receivers to set up an Interledger payment. SPSP uses the [Pre-Shared Key (PSK)](../0016-pre-shared-key/0016-pre-shared-key.md) transport protocol for condition generation and data encoding.
 
 ## Introduction
 
 ### Motivation
 
-The Interledger Protocol does not specify how payment details, such as the ILP Packet or Crypto Condition, should be exchanged between the sender and recipient. SPSP is a minimal protocol that uses HTTP for communicating these details.
+The Interledger Protocol does not specify how payment details, such as the ILP Packet or Crypto Condition, should be exchanged between the sender and receiver. SPSP is a minimal protocol that uses HTTPS for communicating these details.
 
 ### Scope
 
-SPSP provides for exchanging basic payment details needed by a sender to confirm the details of and set up an ILP payment. It is intended for use by end-user applications.
+SPSP provides for exchanging basic receiver details needed by a sender to set up an ILP payment. It is intended for use by end-user applications.
 
 ### Interfaces
 
@@ -22,89 +22,48 @@ SPSP messages MUST be exchanged over HTTPS.
 
 ### Operation
 
-Any SPSP recipient will create a receiving HTTP endpoint called the *receiver*. The sender can query this endpoint to get information about the type of payment that can be made to this receiver. The sender also uses the receiver endpoint to set up the ILP payment.
+Any SPSP receiver will run an SPSP server and expose an HTTPS endpoint called the SPSP Endpoint. The sender can query this endpoint to get information about the type of payment that can be made to this receiver. The sender can set up and send multiple ILP payments using the details provided by the receiver.
 
 ### Definitions
 
-#### SPSP Client
-
-The sender application that uses SPSP to interact with the SPSP Server.
-
-#### SPSP Server
-
-The server used on the recipient's side to handle SPSP requests.
-
-#### Receiver
-
-The specific HTTP endpoint on the SPSP used for setting up a payment.
+* **SPSP Client** - The sender application that uses SPSP to interact with the SPSP Server
+* **SPSP Server** - The server used on the receiver's side to handle SPSP requests
+* **SPSP Endpoint** - The specific HTTPS endpoint on the SPSP server used for setting up a payment
+* **PSK Module** - Software included in the SPSP Client and Server that implements the [Pre-Shared Key](../0016-pre-shared-key/0016-pre-shared-key.md) protocol
 
 ## Overview
 
 ### Relation to Other Protocols
 
-SPSP is used for exchanging payment information before an ILP payment is initiated. It uses the receiver generated [Interledger Payment Requests](../0011-interledger-payment-request) over HTTP.
+SPSP is used for exchanging payment information before an ILP payment is initiated. The sender and receiver use the [Pre-Shared Key (PSK)](../0016-pre-shared-key/0016-pre-shared-key.md) transport protocol to generate the condition and fulfillment. The receiver generates the shared secret to be used in PSK and communicates it to the sender over HTTPS.
 
 ### Model of Operation
 
-#### Fixed Destination Amount
+We assume that the sender knows the receiver's SPSP endpoint (see [Appendix A: (Optional) Webfinger Discovery](#appendix-a-optional-webfinger-discovery)).
 
-SPSP may be used by the sender to send a payment such that the recipient receives a specific amount of their chosen currency or asset type. The model of operation is illustrated with the following example:
-
-We assume that the sender knows the receiver endpoint (see [Appendix A: (Optional) Webfinger Discovery](#appendix-a-optional-webfinger-discovery)).
-
-1. The sender's SPSP client queries the receiver endpoint.
-2. The receiver endpoint responds with the receiver info, including the receiver's currency code.
-3. The sender chooses an amount of the receiver's asset to deliver.
-4. The sender's SPSP client submits the payment information, including the destination amount, to the receiver endpoint.
-5. The receiver endpoint responds with an [Interledger Payment Request](../0011-interledger-payment-request), which includes the execution condition.
-6. The sender's SPSP client uses its ILP module to get a quote in their currency or asset type for the ILP transfer.
-7. The sender accepts the quote.
-8. The sender's SPSP client uses its ILP module to initiate the ILP payment.
-9. The receiver's ILP module registers the incoming transfer held pending the fulfillment of the Crypto Condition. It validates that the transfer matches the packet and regenerates the condition fulfillment using the method recommended in the [Interledger Payment Request](../0011-interledger-payment-request) spec. The ILP module submits the fulfillment to execute the transfer and claim the funds.
-10. The sender's SPSP client receives a notification from its ILP module that the transfer has been executed, including the condition fulfillment from the recipient, and notifies the sender that the payment is completed.
-
-#### Fixed Source Amount
-
-SPSP may be used by the sender to send a payment of a fixed amount of the sender's chosen currency or asset type. This is illustrated with the following example:
-
-We assume that the sender knows the receiver endpoint (see [Appendix A: (Optional) Webfinger Discovery](#appendix-a-optional-webfinger-discovery)).
-
-1. The sender's SPSP client queries the receiver endpoint.
-2. The receiver endpoint responds with the receiver info, including the recipient's ILP address.
-3. The sender choses an amount of their currency or assets to send.
-4. The sender's SPSP client uses its ILP module to quote how much of the recipient's currency or asset type will be delivered to the recipient's ILP address for the given source amount.
-5. The sender accepts the quote.
-6. The sender's SPSP client submits the payment information, including the destination amount, to the receiver endpoint.
-7. The receiver endpoint responds with an [Interledger Payment Requests](../0011-interledger-payment-request), which includes the execution condition.
-8. The sender's SPSP client uses its ILP module to create a transfer **using the chosen source amount, NOT by quoting the payment request**, and attaches the ILP packet to the transfer.
-9. The receiver's ILP module registers the incoming transfer held pending the fulfillment of the Crypto Condition. It validates that the transfer matches the packet and regenerates the condition fulfillment using the method recommended in the [Interledger Payment Request](../0011-interledger-payment-request) spec. The ILP module submits the fulfillment to execute the transfer and claim the funds.
-10. The sender's SPSP client receives a notification from its ILP module that the transfer has been executed, including the condition fulfillment from the recipient, and notifies the sender that the payment is completed.
-
-#### Invoice
-
-1. The sender's SPSP client queries the receiver endpoint.
-2. The receiver endpoint responds with the receiver info, including the invoice status, amount, and currency code.
-3. The sender's SPSP client submits the sender's info, including the sender's ILP address to the receiver endpoint.
-4. The receiver endpoint responds with an [Interledger Payment Request](../0011-interledger-payment-request) corresponding to the destination amount.
-5. The sender's SPSP client uses its ILP module to get a quote in their currency or asset type for the ILP transfer.
-6. The sender accepts the quote.
-7. The sender's SPSP client uses its ILP module to initiate the ILP transfer.
-8. The receiver's ILP module registers the incoming transfer held pending the fulfillment of the Crypto Condition. It validates that the transfer matches the packet and regenerates the condition fulfillment using the method recommended in the [Interledger Payment Request](../0011-interledger-payment-request) spec. The ILP module submits the fulfillment to execute the transfer and claim the funds.
-9. The sender's SPSP client receives a notification from its ILP module that the transfer has been executed, including the condition fulfillment from the recipient, and notifies the sender that the payment is completed.
+1. The sender's SPSP Client queries the receiver's SPSP Endpoint.
+2. The SPSP Endpoint responds with the receiver info, including the receiver's ILP address and the shared secret to be used in PSK.
+3. The sender constructs an ILP payment using the receiver's ILP address.
+4. The sender uses PSK to generate the payment condition and format additional data intended for the reciever to be sent with the payment.
+5. The sender prepares a local transfer to a connector with the condition and ILP packet attached.
+6. The receiver's PSK module registers the incoming transfer, parses and validates the ILP packet and the incoming transfer.
+7. The receiver MAY submit the incoming payment to an external system for review to ensure that the funds are wanted.
+8. If the payment is expected, the receiver's PSK module submits the condition fulfillment to the ledger to execute the transfer. If not, the PSK module rejects the incoming transfer.
+9. If the receiver executed the transfer, the sender's SPSP client receives a notification from its ILP module that the transfer has been executed, including the condition fulfillment from the receiver, and notifies the sender that the payment is completed. If the receiver rejected the transfer, the sender's SPSP client receives a notification with the receiver-specified error message detailing why the payment was rejected.
 
 ## Specification
 
-The receiver endpoint is a URI used by the sender to query information about the sender and set up payments. The receiver URI MAY contain query string parameters. The sender SHOULD treat the URI as opaque.
+The SPSP endpoint is a URI used by the sender to query information about the sender and set up payments. The SPSP endpoint URI MAY contain query string parameters. The sender SHOULD treat the URI as opaque.
 
-The receiver endpoint MUST respond to HTTP `GET` and `POST` requests in the following manner:
+The SPSP Endpoint MUST respond to HTTPS `GET` requests in the following manner:
 
-### Query (`GET <receiver>`)
+### Query (`GET <SPSP Endpoint>`)
 
-The sender queries the receiver endpoint to get information about the type of payment that can be made to this receiver:
+The sender queries the SPSP endpoint to get information about the type of payment that can be made to this receiver:
 
 #### Request
 ``` http
-GET /api/receivers/bob HTTP/1.1
+GET /api/spsp/bob HTTP/1.1
 Host: red.ilpdemo.org
 Accept: application/json
 ```
@@ -115,12 +74,20 @@ HTTP/1.1 200 OK
 Content-Type: application/json
 
 {
-  "type": "payee",
-  "account": "ilpdemo.red.bob",
-  "currency_code": "USD",
-  "currency_symbol": "$",
-  "name": "Bob Dylan",
-  "image_url": "https://red.ilpdemo.org/api/receivers/bob/profile_pic.jpg"
+  "destination_account": "example.ilpdemo.red.bob",
+  "shared_secret": "6jR5iNIVRvqeasJeCty6C-YB5X9FhSOUPCL_5nha5Vs",
+  "maximum_destination_amount": "100000",
+  "minimum_destination_amount": "10",
+  "ledger_info": {
+    "currency_code": "USD",
+    "currency_symbol": "$",
+    "scale": 2,
+    "precision": 10
+  },
+  "receiver_info": {
+    "name": "Bob Dylan",
+    "image_url": "https://red.ilpdemo.org/api/spsp/bob/profile_pic.jpg"
+  }
 }
 ```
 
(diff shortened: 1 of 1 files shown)


0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -146,38 +121,22 @@ The SPSP Client understands the following Cache-Control directives:
...
+| `maximum_destination_amount` | Integer String | Maximum amount, denoted in the minimum divisible units of the ledger, the receiver will accept |
+| `minimum_destination_amount` | Integer String | Minimum amount, denoted in the minimum divisible units of the ledger, the receiver will accept |
+| `ledger_info` | Object | Details about the destination ledger |
+| `ledger_info.currency_code` | String | Currency code to identify the receiver's currency. Currencies that have [ISO 4217](https://en.wikipedia.org/wiki/ISO_4217) codes should use those. Sender UIs SHOULD be able to render non-standard codes |
+| `ledger_info.currency_symbol` | String | Symbol for the receiver's currency intended for display in the sender's UI (e.g. `"$"` or `"shares"`). Sender UIs SHOULD be able to render non-standard symbols |
+| `ledger_info.scale` | Integer | The scale of the amounts on the destination ledger (e.g. an amount of `"1000"` with a scale of `2` translates to `10.00` units of the destination ledger's currency) |
+| `ledger_info.precision` | Integer | The maximum precision supported by the ledger |
+| `receiver_info` | Object | Optional information about the receiver |
```
sharafian: should mention that arbitrary extra fields can be added here

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -36,67 +36,34 @@ The server used on the recipient's side to handle SPSP requests.
...
 
-The specific HTTP endpoint on the SPSP used for setting up a payment.
+The specific HTTPS endpoint on the SPSP used for setting up a payment.
 
 ## Overview
 
 ### Relation to Other Protocols
 
-SPSP is used for exchanging payment information before an ILP payment is initiated. It uses the receiver generated [Interledger Payment Requests](../0011-interledger-payment-request) over HTTP.
+SPSP is used for exchanging payment information before an ILP payment is initiated. The sender and recipient use the [Pre-Shared Key transport protocol(../0016-pre-shared-key/0016-pre-shared-key.md) to generate the condition and fulfillment. The recipient generates the shared secret to be used in PSK and communicates it to the sender over HTTPS.
```
mDuo13: Is SPSP totally tied to PSK, or could you substitute another transport layer instead, e.g. in case you want a non-repudiable fulfillment?

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -36,67 +36,34 @@ The server used on the recipient's side to handle SPSP requests.
 
 #### Receiver
 
-The specific HTTP endpoint on the SPSP used for setting up a payment.
+The specific HTTPS endpoint on the SPSP used for setting up a payment.
```
mDuo13: This "definitions" section might read better as a list of bullets with bolded terms rather than extra sub-headers, e.g.

- **SPSP Client:** The sender application that uses SPSP to interact with the SPSP Server.
- **SPSP Server:** The server used on the recipient's side to handle SPSP requests.
- **Receiver:** The specific HTTPS endpoint on the SPSP used for setting up a payment.

Also, I think the Receiver definition is missing a word here: "on the SPSP _Server_" probably. I'd reword the sentence overall:

> **Receiver:** The HTTPS endpoint provided by the SPSP Server, which the SPSP Client calls to set up a payment.

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
mDuo13: You should formally introduce the acronym PSK the first time you mention Pre-Shared Key transport protocol. Also, the link has a typo (missing the closing square bracket). Change to:

```
The sender and recipient use the [Pre-Shared Key transport protocol (PSK)](../0016-pre-shared-key/0016-pre-shared-key.md) to generate the condition and fulfillment.

File: 0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md [hunk 4]
```diff
@@ -36,67 +36,34 @@ The server used on the recipient's side to handle SPSP requests.
...
-8. The sender's SPSP client uses its ILP module to create a transfer **using the chosen source amount, NOT by quoting the payment request**, and attaches the ILP packet to the transfer.
-9. The receiver's ILP module registers the incoming transfer held pending the fulfillment of the Crypto Condition. It validates that the transfer matches the packet and regenerates the condition fulfillment using the method recommended in the [Interledger Payment Request](../0011-interledger-payment-request) spec. The ILP module submits the fulfillment to execute the transfer and claim the funds.
+2. The receiver endpoint responds with the receiver info, including the recipient's ILP address and the shared secret to be used in PSK.
+3. The sender constructs an ILP payment using the recipient's ILP address.
+4. The sender uses PSK to generate the payment condition and format additional data to be sent with the payment.
+5. The sender uses their ILP module to initiate the ILP payment.
+6. The receiver's ILP module registers the incoming transfer held pending the fulfillment of the condition.
```
mDuo13: Should probably introduce the idea of an "ILP Module" before this point. It's not clear what said module contains, what its scope is, or what level it operates on. I'm thinking it's a piece of software built into the SPSP Client that handles the ILP-layer interactions?

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -146,38 +121,22 @@ The SPSP Client understands the following Cache-Control directives:
...
-| `currency_code` | String | Currency code to identify the receiver's currency. Currencies that have [ISO 4217](https://en.wikipedia.org/wiki/ISO_4217) codes should use those. Sender UIs SHOULD be able to render non-standard codes |
-| `currency_symbol` | String | Symbol for the receiver's currency intended for display in the sender's UI (e.g. `"$"` or `"shares"`). Sender UIs SHOULD be able to render non-standard symbols |
-| `name` | String | Full name of the individual, company or organization the receiver represents |
-| `image_url` | HTTPS URL | URL that a picture of the recipient can be fetched from. The image MUST be square and SHOULD be 128x128 pixels. |
+| `destination_account` | [ILP Address](../0015-ilp-addresses/0015-ilp-addresses.md) | ILP Address of the receiver's account |
+| `shared_secret` | 32 bytes, [base64-url encoded](https://en.wikipedia.org/wiki/Base64#URL_applications) | The shared secret to be used in the [Pre-Shared Key protocol](../0016-pre-shared-key/0016-pre-shared-key.md) |
+| `maximum_destination_amount` | Integer String | Maximum amount, denoted in the minimum divisible units of the ledger, the receiver will accept |
```
mDuo13: You might want to explain what kinds of concerns would control the receiver's decision to set the `maximum_destination_amount` and `minimum_destination_amount` values.

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -146,38 +121,22 @@ The SPSP Client understands the following Cache-Control directives:
...
+| `ledger_info` | Object | Details about the destination ledger |
+| `ledger_info.currency_code` | String | Currency code to identify the receiver's currency. Currencies that have [ISO 4217](https://en.wikipedia.org/wiki/ISO_4217) codes should use those. Sender UIs SHOULD be able to render non-standard codes |
+| `ledger_info.currency_symbol` | String | Symbol for the receiver's currency intended for display in the sender's UI (e.g. `"$"` or `"shares"`). Sender UIs SHOULD be able to render non-standard symbols |
+| `ledger_info.scale` | Integer | The scale of the amounts on the destination ledger (e.g. an amount of `"1000"` with a scale of `2` translates to `10.00` units of the destination ledger's currency) |
+| `ledger_info.precision` | Integer | The maximum precision supported by the ledger |
+| `receiver_info` | Object | Optional information about the receiver |
+| `receiver_info.name` | String | _(OPTIONAL)_ Full name of the individual, company or organization the receiver represents |
+| `receiver_info.image_url` | HTTPS URL | _(OPTIONAL)_ URL that a picture of the recipient can be fetched from |
```
mDuo13: Suggested rewording:

> URL where the sender can get a picture representing the recipient.

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
emschwartz: SPSP is totally tied to PSK. We realized that the combination of HTTPS and receiver-generated conditions would not give you non-repudiability anyway.

To get non-repudiability, you need the receiver to sign an IPR using a well-known key, such that you can show the signed payment request and corresponding fulfillment to other people and they can verify the proof that you fulfilled your debt to the receiver.

While it's totally possible to design an application layer protocol like this (and we might make one in the not so distant future), it's unnecessary for common use cases and it's definitely not simple. For normal payments, depending on the security of TLS is fine, and having proof that the recipient got the money (even if it's not proof you could convince other people with) is more than enough.

With PSK, getting the fulfillment from the transfer being executed is proof enough that the recipient got the money as long as a) the receiver didn't give the secret to anyone else (which would be stupid because then the receiver wouldn't get the money) b) the TLS connection was not man-in-the-middled c) the sender didn't use the secret to fulfill the transfer.

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -191,106 +150,21 @@ Content-Type: application/json
...
-  "type": "invoice",
-  "account": "ilpdemo.red.amazon.111-7777777-1111111",
-  "currency_code": "USD",
-  "currency_symbol": "$",
-  "amount": "10.40",
-  "status": "unpaid",
-  "invoice_info": "https://www.amazon.com/gp/your-account/order-details?ie=UTF8&orderID=111-7777777-1111111"
-}
-```
+  * The sender MAY use the `ledger_info.scale` to determine the integer amount to include in the ILP packet
```
justmoon: Because of the "MAY", it sounds like there is some other way to determine the integer amount? I think this could use some clarification.

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -191,106 +150,21 @@ Content-Type: application/json
...
+  * The sender SHOULD NOT set an amount greater than the `maximum_destination_amount` or the `ledger_info.precision` or one lower than the `minimum_destination_amount`, as this will be rejected by the receiver
```
justmoon: > an amount greater than the `maximum_destination_amount` or the `ledger_info.precision`

Sounds like `ledger_info.precision` is a max amount instead of a number of digits. I would explicitly call out the difference, e.g. "an amount greater than the `maximum_destination_amount` or with a precision greater than `ledger_info.precision`."

Maybe split the precision thing into another sentence.

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
emschwartz: Added to the definitions

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -146,38 +115,22 @@ The SPSP Client understands the following Cache-Control directives:
...
+| `ledger_info` | Object | Details about the destination ledger |
+| `ledger_info.currency_code` | String | Currency code to identify the receiver's currency. Currencies that have [ISO 4217](https://en.wikipedia.org/wiki/ISO_4217) codes should use those. Sender UIs SHOULD be able to render non-standard codes |
+| `ledger_info.currency_symbol` | String | Symbol for the receiver's currency intended for display in the sender's UI (e.g. `"$"` or `"shares"`). Sender UIs SHOULD be able to render non-standard symbols |
+| `ledger_info.scale` | Integer | The scale of the amounts on the destination ledger (e.g. an amount of `"1000"` with a scale of `2` translates to `10.00` units of the destination ledger's currency) |
+| `ledger_info.precision` | Integer | The maximum precision supported by the ledger |
+| `receiver_info` | Object | _(Optional)_ Arbitrary additional information about the receiver. This field has no schema and the receiver may include any fields they choose. The field names listed below are recommended merely for interoperability purposes. |
```
sharafian: The `receiver_info` object should be required, but all fields in it should be optional.

emschwartz: If it's an empty object what's the point of including it at all?

sharafian: it makes the response easier to handle (it removes a check to make sure `receiver_info` exists before referencing its fields), and it makes it easier to validate (if someone were to typo `receiver_info`, for instance, it would get caught)

emschwartz: I think you'll need that check anyway, but SPSP servers should include the empty object

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -318,3 +187,4 @@ Content-Type: application/json
 }
 ```
 [Try this request](https://red.ilpdemo.org/.well-known/webfinger?resource=acct%3Abob%40red.ilpdemo.org)
```
michielbdejong: bob@red.ilpdemo.org doesn't exist, alice@red.ilpdemo.org does, though. You could either change this link or create the bob account so that it starts working :)

emschwartz: bob@red.ilpdemo.org now exists

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -22,81 +22,42 @@ SPSP messages MUST be exchanged over HTTPS.
 
 ### Operation
 
-Any SPSP recipient will create a receiving HTTP endpoint called the *receiver*. The sender can query this endpoint to get information about the type of payment that can be made to this receiver. The sender also uses the receiver endpoint to set up the ILP payment.
+Any SPSP recipient will create a receiving HTTP endpoint called the *receiver*. The sender can query this endpoint to get information about the type of payment that can be made to this receiver. The sender can set up and send multiple ILP payments using the details provided by the receiver.
```
sharafian: Referring to `recipient` and `receiver` as different things is confusing

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -22,81 +22,42 @@ SPSP messages MUST be exchanged over HTTPS.
...
-
-The server used on the recipient's side to handle SPSP requests.
-
-#### Receiver
-
-The specific HTTP endpoint on the SPSP used for setting up a payment.
+* **SPSP Client** - The sender application that uses SPSP to interact with the SPSP Server
+* **SPSP Server** - The server used on the recipient's side to handle SPSP requests
+* **Receiver** - The specific HTTPS endpoint on the SPSP server used for setting up a payment
+* **ILP Module** - Software included in the SPSP Client and Server that sends and receives Interledger payments
```
sharafian: This is confusing because the `ilp` module includes an SPSP client. Why not refer to it as a ledger plugin?

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
sharafian: Why is the SPSP `rel` called `receiver` and not `spsp`? That made more sense when there were several endpoints.

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -22,81 +22,42 @@ SPSP messages MUST be exchanged over HTTPS.
...
-5. The receiver endpoint responds with an [Interledger Payment Request](../0011-interledger-payment-request), which includes the execution condition.
-6. The sender's SPSP client uses its ILP module to get a quote in their currency or asset type for the ILP transfer.
-7. The sender accepts the quote.
-8. The sender's SPSP client uses its ILP module to initiate the ILP payment.
-9. The receiver's ILP module registers the incoming transfer held pending the fulfillment of the Crypto Condition. It validates that the transfer matches the packet and regenerates the condition fulfillment using the method recommended in the [Interledger Payment Request](../0011-interledger-payment-request) spec. The ILP module submits the fulfillment to execute the transfer and claim the funds.
+2. The receiver endpoint responds with the receiver info, including the recipient's ILP address and the shared secret to be used in PSK.
+3. The sender constructs an ILP payment using the recipient's ILP address.
+4. The sender uses PSK to generate the payment condition and format additional data to be sent with the payment.
```
sharafian: What should the "additional data" look like? Should it be utf8-encoded JSON, or arbitrary binary data to be decided by the specific application?

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -22,81 +22,42 @@ SPSP messages MUST be exchanged over HTTPS.
...
+5. The sender uses their ILP module to initiate the ILP payment.
+6. The receiver's ILP module registers the incoming transfer held pending the fulfillment of the condition.
+7. The receiver's PSK module parses and validates the ILP packet and the incoming transfer.
+8. The receiver may submit the incoming payment to an external system for review to ensure that the funds are wanted.
```
sharafian: Steps 7-9 are included in the PSK flow, so they shouldn't be re-described here as part of SPSP.

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -22,81 +22,42 @@ SPSP messages MUST be exchanged over HTTPS.
...
+9. The receiver's PSK module submits the condition fulfillment using the ILP module.
 10. The sender's SPSP client receives a notification from its ILP module that the transfer has been executed, including the condition fulfillment from the recipient, and notifies the sender that the payment is completed.
```
sharafian: Are there any SPSP-specific error messages or behaviors that should be put into SPSP clients?

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
emschwartz: This is answered below in the Payment Setup part

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
emschwartz: I think those are worth including in here because they're things you would need to handle when implementing an SPSP server. Just because the current implementation of the PSK client does these things and provides a callback that you'd use to insert the review step, it could also be implemented as separate steps that you need to call from the SPSP server.

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
emschwartz: Changed

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
emschwartz: I think we may need to take another pass at these specs once we define the ILP error format more. I would say that SPSP should have JSON error messages maybe with an `id` and `message` field or something like that, but I'm not sure if we need to define the SPSP errors.

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -191,106 +143,22 @@ Content-Type: application/json
...
+  * The sender uses the `ledger_info.scale` to determine the integer amount to include in the ILP packet (e.g. if they want the recipient to receive 10 units on their ledger, a `ledger_info.scale` value of 2 would indicate they should set the `amount` to 1000). In cases where the sender knows the integer amount the receiver should get, for example if that amount was requested by the receiver out of band, the sender does not need to use the `ledger_info.scale`
+  * The sender SHOULD NOT set an amount greater than the `maximum_destination_amount` or lower than the `minimum_destination_amount`, as this will be rejected by the receiver
+  * The sender SHOULD NOT set an amount with a precision greater than the `ledger_info.precision`, as this will be rejected by the destination ledger
 
-#### Request
+* The `data` is encoded using the [Pre-Shared Key protocol](../0016-pre-shared-key/0016-pre-shared-key.md):
 
-##### For a Payee
+  * The key is the `shared_secret` provided by the receiver
+  * Additional data may be included in the PSK details. SPSP data MUST be JSON (and implementations SHOULD set the `Content-Type: application/json` header)
```
sharafian: It should be specified that `Content-Type` is a private header.

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -22,83 +22,43 @@ SPSP messages MUST be exchanged over HTTPS.
...
-7. The sender accepts the quote.
-8. The sender's SPSP client uses its ILP module to initiate the ILP payment.
-9. The receiver's ILP module registers the incoming transfer held pending the fulfillment of the Crypto Condition. It validates that the transfer matches the packet and regenerates the condition fulfillment using the method recommended in the [Interledger Payment Request](../0011-interledger-payment-request) spec. The ILP module submits the fulfillment to execute the transfer and claim the funds.
+1. The sender's SPSP Client queries the receiver's SPSP Endpoint.
+2. The SPSP Endpoint responds with the receiver info, including the recipient's ILP address and the shared secret to be used in PSK.
+3. The sender constructs an ILP payment using the recipient's ILP address.
+4. The sender uses PSK to generate the payment condition and format additional data intended for the reciever to be sent with the payment.
+5. The sender prepares a local transfer to a connector with the condition and ILP packet attached.
+6. The receiver registers the incoming transfer held pending the fulfillment of the condition.
```
sharafian: What does this mean?

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -191,106 +142,22 @@ Content-Type: application/json
...
-  "sender_identifier": "alice@blue.ilpdemo.org"
-}
-```
+* The `data` is encoded using the [Pre-Shared Key protocol](../0016-pre-shared-key/0016-pre-shared-key.md):
 
-| Field | Type | Description |
-|---|---|---|
-| `sender_identifier` | String | Identifier of the sender |
+  * The key is the `shared_secret` provided by the receiver
+  * Additional data may be included in the PSK details. SPSP data MUST be JSON and a private PSK header SHOULD be included with `Content-Type: application/json`
```
sharafian: Is it worth mentioning `Content-Length` as well?

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -311,10 +178,11 @@ Content-Type: application/json
   "subject": "acct:bob@red.ilpdemo.org",
   "links": [
     {
-      "rel": "https://interledger.org/rel/receiver",
+      "rel": "https://interledger.org/rel/spsp",
```
sharafian: The current SPSP client in `ilp` looks for `https://interledger.org/rel/spsp/v1`. Should we change it there, or here?

emschwartz: Technically it's v2 now...

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -22,83 +22,42 @@ SPSP messages MUST be exchanged over HTTPS.
...
-
-1. The sender's SPSP client queries the receiver endpoint.
-2. The receiver endpoint responds with the receiver info, including the invoice status, amount, and currency code.
-3. The sender's SPSP client submits the sender's info, including the sender's ILP address to the receiver endpoint.
-4. The receiver endpoint responds with an [Interledger Payment Request](../0011-interledger-payment-request) corresponding to the destination amount.
-5. The sender's SPSP client uses its ILP module to get a quote in their currency or asset type for the ILP transfer.
-6. The sender accepts the quote.
-7. The sender's SPSP client uses its ILP module to initiate the ILP transfer.
-8. The receiver's ILP module registers the incoming transfer held pending the fulfillment of the Crypto Condition. It validates that the transfer matches the packet and regenerates the condition fulfillment using the method recommended in the [Interledger Payment Request](../0011-interledger-payment-request) spec. The ILP module submits the fulfillment to execute the transfer and claim the funds.
+1. The sender's SPSP Client queries the receiver's SPSP Endpoint.
```
mDuo13: Now that we've gotten rid of the "Receiver Endpoint" maybe we should replace all instances of "receiver" with "recipient" for consistency.

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -22,83 +22,42 @@ SPSP messages MUST be exchanged over HTTPS.
...
+2. The SPSP Endpoint responds with the receiver info, including the recipient's ILP address and the shared secret to be used in PSK.
+3. The sender constructs an ILP payment using the recipient's ILP address.
+4. The sender uses PSK to generate the payment condition and format additional data intended for the reciever to be sent with the payment.
+5. The sender prepares a local transfer to a connector with the condition and ILP packet attached.
+6. The receiver's PSK module registers the incoming transfer, parses and validates the ILP packet and the incoming transfer.
+7. The receiver may submit the incoming payment to an external system for review to ensure that the funds are wanted.
+8. The receiver's PSK module submits the condition fulfillment to the ledger to execute the transfer.
```
mDuo13: We should clarify in this step (or an extra step before it?) regarding what happens in the case where the review in step 7 determines that the recipient wants to reject the payment.

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -22,83 +22,42 @@ SPSP messages MUST be exchanged over HTTPS.
...
-8. The receiver's ILP module registers the incoming transfer held pending the fulfillment of the Crypto Condition. It validates that the transfer matches the packet and regenerates the condition fulfillment using the method recommended in the [Interledger Payment Request](../0011-interledger-payment-request) spec. The ILP module submits the fulfillment to execute the transfer and claim the funds.
+1. The sender's SPSP Client queries the receiver's SPSP Endpoint.
+2. The SPSP Endpoint responds with the receiver info, including the recipient's ILP address and the shared secret to be used in PSK.
+3. The sender constructs an ILP payment using the recipient's ILP address.
+4. The sender uses PSK to generate the payment condition and format additional data intended for the reciever to be sent with the payment.
+5. The sender prepares a local transfer to a connector with the condition and ILP packet attached.
+6. The receiver's PSK module registers the incoming transfer, parses and validates the ILP packet and the incoming transfer.
+7. The receiver may submit the incoming payment to an external system for review to ensure that the funds are wanted.
```
mDuo13: "may"→"MAY"

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -22,83 +22,42 @@ SPSP messages MUST be exchanged over HTTPS.
...
+8. The receiver's PSK module submits the condition fulfillment to the ledger to execute the transfer.
 9. The sender's SPSP client receives a notification from its ILP module that the transfer has been executed, including the condition fulfillment from the recipient, and notifies the sender that the payment is completed.
 
 ## Specification
 
-The receiver endpoint is a URI used by the sender to query information about the sender and set up payments. The receiver URI MAY contain query string parameters. The sender SHOULD treat the URI as opaque.
+The SPSP endpoint is a URI used by the sender to query information about the sender and set up payments. The SPSP endpoint URI MAY contain query string parameters. The sender SHOULD treat the URI as opaque.
 
-The receiver endpoint MUST respond to HTTP `GET` and `POST` requests in the following manner:
+The SPSP Endpoint MUST respond to HTTPS `GET` requests in the following manner:
```
mDuo13: Does "the following manner" include any error cases? Perhaps we should define a few, like a 503 error for when some or all data is temporarily unavailable.

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -191,106 +142,23 @@ Content-Type: application/json
...
-| Field | Type | Description |
-|---|---|---|
-| `amount` | Decimal String | _Destination amount_ the receiver will receive |
-| `sender_identifier` | String | Identifier of the sender |
-| `memo` | String | Message for the recipient linked to the payment |
+### Payment Setup
 
-##### For an Invoice
+The sender uses the receiver details to create the ILP packet:
+* The `account` in the ILP packet is the `destination_account` provided by the receiver
```
mDuo13: Add a blank line between the lead-in and the start of the bulleted list. Some Markdown parsers don't recognize a list if it's not separated from a paragraph by a blank line.

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -191,106 +142,23 @@ Content-Type: application/json
...
-{
-  "sender_identifier": "alice@blue.ilpdemo.org"
-}
-```
+* The `data` is encoded using the [Pre-Shared Key protocol](../0016-pre-shared-key/0016-pre-shared-key.md):
 
-| Field | Type | Description |
-|---|---|---|
-| `sender_identifier` | String | Identifier of the sender |
+  * The key is the `shared_secret` provided by the receiver
```
mDuo13: As above, nested lists must be indented 4 spaces more than the parent list for most Markdown parsers to recognize them as such.

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -191,106 +142,23 @@ Content-Type: application/json
...
+  * Additional data may be included in the PSK details. SPSP data MUST be JSON.
+  * Private PSK headers SHOULD be included with `Content-Type: application/json` and `Content-Length: <byte length of data>`
```
mDuo13: "be included with"→"include"

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -146,40 +113,24 @@ The SPSP Client understands the following Cache-Control directives:
...
-| `name` | String | Full name of the individual, company or organization the receiver represents |
-| `image_url` | HTTPS URL | URL that a picture of the recipient can be fetched from. The image MUST be square and SHOULD be 128x128 pixels. |
-
-If this receiver is not available, an error can be generated at this stage:
+| `destination_account` | [ILP Address](../0015-ilp-addresses/0015-ilp-addresses.md) | ILP Address of the receiver's account |
+| `shared_secret` | 32 bytes, [base64-url encoded](https://en.wikipedia.org/wiki/Base64#URL_applications) | The shared secret to be used in the [Pre-Shared Key protocol](../0016-pre-shared-key/0016-pre-shared-key.md) |
+| `maximum_destination_amount` | Integer String | Maximum amount, denoted in the minimum divisible units of the ledger, the receiver will accept. This amount may be determined by the ledger's maximum account balance or transfer amount. If the receiver expects a specific amount to be delivered, this may be set equal to the `minimum_destination_amount` |
```
mDuo13: Perhaps it's worth (re-)mentioning why we use integer strings instead of native JSON numbers:

> **Note:** Currency amounts are denominated as integer strings instead of native JSON numbers to avoid losing precision during JSON parsing. Applications MUST represent these numbers in a data type that has precision equal or greater than an unsigned 64-bit integer.

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
emschwartz: The only one currently defined is the 404, I thought you normally don't define 500 level errors because those aren't supposed to happen

mDuo13: It's good to clarify in advance what kinds of errors someone might get, even the ones that aren't "supposed" to happen. The reality is, errors do happen, so it's best to be ready for them.

That said, I'm OK with adding those in a later revision.

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -2,109 +2,68 @@
...
-### Query (`GET <receiver>`)
+### Query (`GET <SPSP Endpoint>`)
 
-The sender queries the receiver endpoint to get information about the type of payment that can be made to this receiver:
+The sender queries the SPSP endpoint to get information about the type of payment that can be made to this recipient:
 
 #### Request
 ``` http
-GET /api/receivers/bob HTTP/1.1
+GET /api/recipients/bob HTTP/1.1
```
justmoon: Maybe `/api/spsp/bob` would be a good path to use for the example?

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -115,12 +74,20 @@ HTTP/1.1 200 OK
...
-  "currency_symbol": "$",
-  "name": "Bob Dylan",
-  "image_url": "https://red.ilpdemo.org/api/receivers/bob/profile_pic.jpg"
+  "destination_account": "example.ilpdemo.red.bob",
+  "shared_secret": "6jR5iNIVRvqeasJeCty6C-YB5X9FhSOUPCL_5nha5Vs",
+  "maximum_destination_amount": "100000",
+  "minimum_destination_amount": "10",
+  "ledger_info": {
+    "currency_code": "USD",
+    "currency_symbol": "$",
```
michielbdejong: wouldn't it be cleaner to return just a currency code, and leave out the currency symbol?

emschwartz: How would you display custom or lesser-known currencies?

michielbdejong: ilp-kit and other node-based UI applications should start using some standard registry like https://www.npmjs.com/package/currency-symbol-map, and people can then send pull requests to that npm module to make their custom currencies have fancy symbols. If there is a need, we can create a cross-language registry somewhere. But IMHO, making the UI appearance part of the protocol breaks separation of concerns.

Issue #138, opened 2016-12-29
RFC 11: IPRs should be Interledger Packets


emschwartz:
If we [formalize the ILP packet JSON format](https://github.com/interledger/rfcs/issues/137), we should replace the Interledger Payment Request format with a normal ILP packet. Right now we have too many different formats that express basically the same information but in slightly different ways (different names for the same data fields and a slightly different JSON hierarchy).

I would propose that the IPR format look like:

```json
{
  "account": "us.usd.wf.bob",
  "amount": "10",
  "paymentRequest": {
    "id": "d3c11acf-6232-46c4-8104-015c983a0ff3",
    "expiresAt": "2016-12-29T15:09:20.734Z",
    "condition": "cc:0:3:47DEQpj8HBSa-_TImW-5JCeuQeRkm5NMpJWZG3hSuFU:0",
    "memo": "This is for that thing"
  }
}
```

The payment request header would include the condition but the spec would say that sending clients SHOULD remove the condition before sending the payment, because it is redundant information.

Receivers would include the fields listed above (minus the `condition`) in the hash they use for the condition. This means that no other headers would be included in the MAC. I think we could also make the payment request have a string memo (as opposed to including arbitrary JSON) because it's simpler and because the main use case for this is to include some minimal extra information for the receiver to identify what the incoming payment was for.

sharafian:
The only issue I can think of with this is that you could have a condition and expiry in the `payment_request` that don't match the condition and expiry of the local transfer.

If the ILP packet you send is strongly encouraged to be a modified version of this Payment Request ILP Packet, isn't that _effectively_ a third format?

emschwartz:
The expiry in the payment request is a different one than that in the local transfer. The receiver already should use the one in the payment request (that they set originally), not the one in the transfer. If the one in the transfer is too short they'll try to submit the fulfillment and it will fail. If it's too long it doesn't matter. One of the reasons the expiry in the payment request is included in the MAC though is to make sure that no one can tamper with this detail and make the receiver accept the money after they originally said they would.

If the condition can be in the payment request header I would suggest saying that the receiver MUST remove the condition. They need to remove it for hashing the details anyway, and they should only care about the condition of the transfer.

I think that saying one field of the packet is optional is way better than having a completely separate format. Right now the IPR spec has this `additional_headers` field, but we haven't needed that yet and didn't actually figure out a good way to include other headers. If you were sending a whole ILP packet, we wouldn't need to figure out a separate way to include headers.

PR #170, opened 2017-03-22
RFC 11: Change IPR to use ILP Packet and PSK

emschwartz:
depends on https://github.com/interledger/rfcs/pull/166, https://github.com/interledger/rfcs/pull/175
* change the IPR format to only include a packet and condition
* replace JSON format with binary
* recommend using a PSK implementation to generate the payment request

resolves https://github.com/interledger/rfcs/issues/138

### DIFF
--- a/0011-interledger-payment-request/0011-interledger-payment-request.md
+++ b/0011-interledger-payment-request/0011-interledger-payment-request.md
@@ -2,21 +2,23 @@
 
 ## Preface
 
-This document specifies the Interledger Payment Request (IPR) data format. It is intended for use in protocols in which the recipient of an Interledger payment first communicates a request for payment to the sender.
-
-This document provides an example flow to illustrate how IPR could be used in a higher level protocol. This document also details a method for generating Interledger transfer conditions that enable recipients to use the condition as an integrity check on incoming transfers and packets, without maintaining state about all outstanding payment requests.
+This document specifies the Interledger Payment Request (IPR) transport protocol. It is an end-to-end protocol in which the recipient of an Interledger payment first communicates a request for payment to the sender. This document also details how the recipient can verify that incoming payments match previously created requests without storing a history of all outstanding requests.
 
 ## Introduction
 
 ### Motivation
 
-The Interledger Protocol uses holds based on Crypto Conditions and expiries to ensure a sender's funds are delivered to the destination or returned to the sender's account.
+The Interledger Protocol uses holds based on cryptographic conditions and expiries to ensure a sender's funds are delivered to the destination or returned to the sender's account.
 
-A common approach is for conditions to be generated and fulfilled by the receivers of Interledger transfers. This document specifies one such approach that enables receivers to verify that an incoming transfer matches a payment request they previously created, without the receiver needing to keep a record of all outstanding payment requests.
+One approach is for conditions to be generated and fulfilled by the receivers of Interledger payments. When the sender of the payment receives the fulfillment, they can be certain that the recipient received the payment.
+
+Receiver-generated conditions are also useful in building non-repudiable application layer protocols. The recipient would generate the condition and cryptographically sign a statement that the preimage of the specified hash indicates the sender has settled a particular debt. The recipient would only disclose the hash preimage upon receipt of payment. Once the sender gets the preimage, they would be able to demonstrate to 3rd parties that they paid the recipient using the preimage and the original signed statement.
+
+In cases where non-repudiability is unnecessary, the [Pre-Shared Key](../0016-pre-shared-key/0016-pre-shared-key.md) transport protocol may be preferrable because it does not require end-to-end communication for each payment.
 
 ### Scope
 
-This document defines a suggested format for representing payment requests and an approach for receiver-generated conditions. Protocols built on top of Interledger MAY use the format provided here, though they can use any format they choose instead.
+This document defines a suggested format for representing payment requests and an approach for receiver-generated conditions. Protocols built on top of IPR MAY use the format provided here, though they can use any format they choose instead.
 
 While this document also provides a recommendation for generating transfer conditions, conditions are often opaque to all parties aside from the receiver. Receivers MAY use any condition they choose, however this specification includes best practices that receivers SHOULD take into consideration.
 
@@ -26,14 +28,13 @@ This document does **not** specify how payment requests are communicated from a 
 
 Interledger Payment Requests are intended to be used in higher-level protocols as follows:
 
-1. Recipient creates the payment request and generates the condition using one of algorithms outlined in [Condition Generation](#condition-generation).
-2. Recipient communicates the payment request to the sender.
-3. Sender generates an ILP Packet from the payment request.
-4. Sender prepares a local ledger transfer to a connector, held pending the execution condition given in the payment request and with the ILP Packet attached.
-5. Connectors forward the payment until it reaches the recipient.
-6. Recipient receives a notification of a held transfer with an ILP Packet attached.
-7. Recipient checks the transfer amount against the ILP Packet and checks the payment request has not expired.
-8. Recipient regenerates the condition and fulfillment from the ILP Packet attached to the incoming transfer, using the same algorithm as before, and fulfills the transfer's condition. If the condition generated does not match the one in the transfer it means that the ILP Packet has been tampered with and the transfer goes unfulfilled.
+1. Recipient creates the ILP Packet and generates the condition using one of algorithms outlined in [Condition Generation](#condition-generation).
+2. Recipient communicates the ILP Packet and condition to the sender.
+3. Sender prepares a local ledger transfer to a connector, held pending the execution condition provided by the recipient and with the ILP Packet provided by the recipient attached.
+4. Connectors forward the payment until it reaches the recipient.
+5. Recipient receives a notification of a held transfer with an ILP Packet attached.
+6. Recipient checks the transfer amount against the ILP Packet and checks the payment request has not expired.
+7. Recipient regenerates the condition and fulfillment from the ILP Packet attached to the incoming transfer, using the same algorithm as before, and fulfills the transfer's condition. If the condition generated does not match the one in the transfer it means that the ILP Packet has been tampered with and the transfer goes unfulfilled.
 
 ## Payment Request Specification
 
@@ -41,55 +42,17 @@ Interledger Payment Requests are intended to be used in higher-level protocols a
 
 ```json
 {
-    "address": "ilpdemo.red.bob.b9c4ceba-51e4-4a80-b1a7-2972383e98af",
-    "amount": "10.25",
-    "condition": "cc:0:3:hd5x8kpaDDLQu-KqMyCrlsg5QJ9g9qaFr9ytTwqyCsw:32",
-    "expires_at": "2016-08-16T12:00:00Z",
-    "data": {
-        "re": "dinner the other night"
-    },
-    "additional_headers": "asdf98zxcvlknannasdpfi09qwoijasdfk09xcv009as7zxcv"
+  "packet": "AYGyAAAAAAAAAAo6ZXhhbXBsZS5pbHBkZW1vLmJsdWUuYm9iLkR0azB1Y3U4OHFJQm1iRFEwelhxa2lkTWZhOHRQakJZUW1QU0svMS4wCk5vbmNlOiBvdHotT25NbThURHo3NThGYWY4cXRRCkVuY3J5cHRpb246IGFlcy0yNTYtZ2NtIEdFRUR0MXk4NmlkSVFMa2xWbnpiRWcKCuXPle38YB02KVAHPtVNqRgP7Ok6D-QRAA",
+  "condition": "oHeVy7WAEhrJF3U4NeLt5QT6Fr3L2YeBCmB88Xu9sEY"
 }
 ```
 
 | Field | Type | Short Description |
 |:--|:--|:--|
-| `address` | ILP Address | Request-specific address |
-| `amount` | Decimal String | Amount requested by the recipient |
-| `condition` | Crypto Condition | Execution condition for the payment |
-| `expires_at` | [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) Timestamp | Expiry of the request |
-| `data` | Object | Data to be included in the ILP Packet |
-| `additional_headers` | Base64 String | Additional headers for the ILP Packet |
+| `packet` | [Base64-Url](https://en.wikipedia.org/wiki/Base64#URL_applications) Encoded ILP Packet | ILP Payment packet including the destination address and amount |
+| `condition` | 32 Bytes, Base64-Url Encoded | Execution condition for the payment |
 
-#### address
-
-The ILP Address the payment is to be routed to. This SHOULD be a request-specific address in the form `<account address>.<request ID>`, such as `ilpdemo.red.bob.b9c4ceba-51e4-4a80-b1a7-2972383e98af`.
-
-When using the recommended [Condition Generation](#condition-generation) method, account addresses MUST include a request-specific portion to ensure that conditions are unique per-request.
-
-#### amount
-
-The amount requested by the recipient, such as `10.25`, denoted in the assets of the recipient's ledger.
-
-The amount MUST be rounded based on the precision supported by the recipient's ledger.
-
-Recipients SHOULD ignore incoming transfers whose amounts are less than the amount specified in the payment request. Recipients MAY ignore incoming transfers whose amounts are greater than requested.
-
-#### condition
-
-The Crypto Condition to be used as the execution condition for the payment. See [Condition Generation](#condition-generation) for recommendations the recipient MAY use for creating the condition.
-
-#### expires_at
-
-The timestamp when the request expires. Recipients SHOULD NOT fulfill the conditions of incoming transfers that arrive after the expiry has passed. Senders MAY use the request expiry to determine the expiry of the transfer they put on hold for the first connector.
-
-#### data
-
-An arbitrary JSON object that should be included in the ILP Packet's user data header.
-
-#### additional_headers
-
-Additional binary headers, encoded as a base64 string, that should be included in the ILP Packet. The sender SHOULD treat these as opaque.
+In IPR the sender only uses the `account` and `amount` from the ILP packet and MUST treat the `data` as opaque. The sender MUST NOT modify the ILP packet at all, or the recipient will reject the payment.
 
 ### Binary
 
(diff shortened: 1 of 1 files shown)


0011-interledger-payment-request/0011-interledger-payment-request.md
```diff
@@ -26,70 +28,31 @@ This document does **not** specify how payment requests are communicated from a
...
-7. Recipient checks the transfer amount against the ILP Packet and checks the payment request has not expired.
-8. Recipient regenerates the condition and fulfillment from the ILP Packet attached to the incoming transfer, using the same algorithm as before, and fulfills the transfer's condition. If the condition generated does not match the one in the transfer it means that the ILP Packet has been tampered with and the transfer goes unfulfilled.
+1. Recipient creates the ILP Packet and generates the condition using one of algorithms outlined in [Condition Generation](#condition-generation).
+2. Recipient communicates the ILP Packet and condition to the sender.
+3. Sender prepares a local ledger transfer to a connector, held pending the execution condition provided by the recipient and with the ILP Packet provided by the recipient attached.
+4. Connectors forward the payment until it reaches the recipient.
+5. Recipient receives a notification of a held transfer with an ILP Packet attached.
+6. Recipient checks the transfer amount against the ILP Packet and checks the payment request has not expired.
```
sharafian: If the data format that the receiver uses is left undefined, it seems strange to describe the behavior in steps 6 and 7.

0011-interledger-payment-request/0011-interledger-payment-request.md
```diff
@@ -2,21 +2,23 @@
...
+
+Receiver-generated conditions are also useful in building non-repudiable application layer protocols. The recipient would generate the condition and cryptographically sign a statement that the preimage of the specified hash indicates the sender has settled a particular debt. The recipient would only disclose the hash preimage upon receipt of payment. Once the sender gets the preimage, they would be able to demonstrate to 3rd parties that they paid the recipient using the preimage and the original signed statement.
 
-A common approach is for conditions to be generated and fulfilled by the receivers of Interledger transfers. This document specifies one such approach that enables receivers to verify that an incoming transfer matches a payment request they previously created, without the receiver needing to keep a record of all outstanding payment requests.
+In cases where non-repudiability is unnecessary, the [Pre-Shared Key](../0016-pre-shared-key/0016-pre-shared-key.md) transport protocol may be preferrable because it does not require end-to-end communication for each payment.
```
sharafian: This should give a clearer definition of what cases you want IPR, and should recommend that PSK should be used unless you really know what you're doing.

0011-interledger-payment-request/0011-interledger-payment-request.md
emschwartz: I'd put that in the same category as including the way the receiver generates the shared secret in PSK. It's not technically required but it's a _very good_ idea and should be followed unless you really know what you're doing.

0011-interledger-payment-request/0011-interledger-payment-request.md
```diff
@@ -99,42 +62,17 @@ Additional binary headers, encoded as a base64 string, that should be included i
...
+Recipients MUST include a payment request-specific nonce in the ILP packet to ensure that conditions are unique per-request.
 
-This is how a [PREIMAGE-SHA-256 Crypto Condition](../0002-crypto-conditions) MAY be generated using an HMAC of a JSON-encoded Interledger Payment Request:
+Recipients SHOULD include an expiry date so that they do not accept incoming payments that are paid too long after their creation.
 
-1. The recipient takes the `address`, `amount`, `expires_at`, and `data` fields from the payment request. Any fields that are used for routing are probably not important to perform an integrity check upon, because they may be modified in transit and there is no further use for them once the ILP packet is received attached to an incoming transfer.
-2. The recipient uses a canonical encoding, such as [Canonical JSON](https://www.npmjs.com/package/canonical-json) to create a stringified version of `1.`
-3. The recipient uses a secret key to create an HMAC of `2.`
-4. The recipient uses the digest of `3.` as the preimage of a [PREIMAGE-SHA-256 Crypto Condition](../0002-crypto-conditions).
+### Using a PSK Implementation
```
justmoon: What do these three headlines mean? It sounds like I can use either JSON, Binary or a PSK Implementation, but then the actual text doesn't fit that at all.

If the packet is generated by PSK and you need the packet to get the condition for the payment request, I'd probably reorder these sections.

0011-interledger-payment-request/0011-interledger-payment-request.md
```diff
@@ -99,42 +62,17 @@ Additional binary headers, encoded as a base64 string, that should be included i
...
 
-The following snippet shows how this can be done in JavaScript:
+It is RECOMMENDED that recipients use an implementation of the [Pre-Shared Key](../0016-pre-shared-key/0016-pre-shared-key.md) protocol to generate the packet and condition for IPR.
```
justmoon: If I don't use PSK, what would I do instead? How do I decide? Why is PSK recommended?

0011-interledger-payment-request/0011-interledger-payment-request.md
```diff
@@ -99,42 +62,17 @@ Additional binary headers, encoded as a base64 string, that should be included i
...
 
-```js
-const crypto = require('crypto') // Node.js crypto
-const stringify = require('canonical-json') // https://www.npmjs.com/package/canonical-json
-const cc = require('five-bells-condition') // https://www.npmjs.com/package/five-bells-condition
+The recipient should follow BOTH the steps normally followed by the recipient AND sender. The recipient should generate the secret in the normal fashion but MUST NOT share it. The recipient then follows the steps the sender would do in a PSK payment to create the packet and condition from the secret and the payment parameters.
```
justmoon: > The recipient should follow BOTH the steps normally followed by the recipient AND sender.

This sentence kinda makes it sound more complicated than it actually is, but I'm not sure how I would reword it.

0011-interledger-payment-request/0011-interledger-payment-request.md
```diff
@@ -99,42 +62,17 @@ Additional binary headers, encoded as a base64 string, that should be included i
...
-    "re": "dinner the other night"
-  }
-})
+For details, see the [Pre-Shared Key Pseudocode](../0016-pre-shared-key/0016-pre-shared-key.md#pseudocode).
 
-const hmac = crypto.createHmac('sha256', RECIPIENT_HMAC_KEY)
-hmac.update(requestString)
-const cryptoCondition = new cc.PreimageSha256()
-cryptoCondition.setPreimage(hmac.digest())
+Note the recipient's use of a PSK implementation does not matter to the sender, because the sender treats the `data` in the ILP packet as opaque and uses the condition from the recipient without needing to know how the fulfillment is generated.
```
justmoon: I think this paragraph should follow right after the statement that PSK is "RECOMMENDED", because it gives some important background as to why it isn't *required*.

0011-interledger-payment-request/0011-interledger-payment-request.md
```diff
@@ -99,42 +62,17 @@ Additional binary headers, encoded as a base64 string, that should be included i
 
 It is RECOMMENDED that recipients use a Message Authentication Code (MAC) of the details of their payment request in the condition so that the condition can be used as an integrity check on incoming transfers. This allows a recipient to ensure that a transfer matches a request they previously generated, without the recipient needing to maintain a log of all outstanding payment requests.
 
-Recipients MUST include a payment request-specific component in the account ILP Address to ensure that conditions are unique per-request.
-
-### Preimage Condition from JSON
+Recipients MUST include a payment request-specific nonce in the ILP packet to ensure that conditions are unique per-request.
```
justmoon: This sounds like something I have to worry about in addition to using PSK, but doesn't using PSK take care of this?

0011-interledger-payment-request/0011-interledger-payment-request.md
```diff
@@ -99,42 +62,17 @@ Additional binary headers, encoded as a base64 string, that should be included i
 
 It is RECOMMENDED that recipients use a Message Authentication Code (MAC) of the details of their payment request in the condition so that the condition can be used as an integrity check on incoming transfers. This allows a recipient to ensure that a transfer matches a request they previously generated, without the recipient needing to maintain a log of all outstanding payment requests.
```
justmoon: It's not so much that you're adding a MAC, it's that you're using the condition (which you need to pick anyway) *as* a MAC. So I think something like "you should use the condition as a mac by making the fulfillment a hash of the packet" would be easier to understand.

0011-interledger-payment-request/0011-interledger-payment-request.md
```diff
@@ -2,21 +2,23 @@
...
 
-A common approach is for conditions to be generated and fulfilled by the receivers of Interledger transfers. This document specifies one such approach that enables receivers to verify that an incoming transfer matches a payment request they previously created, without the receiver needing to keep a record of all outstanding payment requests.
+In most cases, non-repudiability is unnecessary, as it is sufficient for the sender to get proof that the recipient received the payment even if that proof cannot be used to convince 3rd parties. In such cases, the [Pre-Shared Key](../0016-pre-shared-key/0016-pre-shared-key.md) transport protocol is recommended because it does not require end-to-end communication for each payment.
```
mDuo13: "for each payment"→"for every payment"

mDuo13: "3rd parties"→"third parties"

0011-interledger-payment-request/0011-interledger-payment-request.md
```diff
@@ -26,70 +28,31 @@ This document does **not** specify how payment requests are communicated from a
...
 
-1. Recipient creates the payment request and generates the condition using one of algorithms outlined in [Condition Generation](#condition-generation).
-2. Recipient communicates the payment request to the sender.
-3. Sender generates an ILP Packet from the payment request.
-4. Sender prepares a local ledger transfer to a connector, held pending the execution condition given in the payment request and with the ILP Packet attached.
-5. Connectors forward the payment until it reaches the recipient.
-6. Recipient receives a notification of a held transfer with an ILP Packet attached.
-7. Recipient checks the transfer amount against the ILP Packet and checks the payment request has not expired.
-8. Recipient regenerates the condition and fulfillment from the ILP Packet attached to the incoming transfer, using the same algorithm as before, and fulfills the transfer's condition. If the condition generated does not match the one in the transfer it means that the ILP Packet has been tampered with and the transfer goes unfulfilled.
+1. Recipient creates the ILP Packet and generates the condition using one of algorithms outlined in [Condition Generation](#condition-generation).
```
mDuo13: Let's make this more proper English. "_The_ recipient creates..." (and same on the following steps)

0011-interledger-payment-request/0011-interledger-payment-request.md
```diff
@@ -26,70 +28,31 @@ This document does **not** specify how payment requests are communicated from a
...
+2. Recipient communicates the ILP Packet and condition to the sender.
```
mDuo13: Am I mistaken, or is this data structure containing the "ILP Packet and condition" called the "IPR"?

0011-interledger-payment-request/0011-interledger-payment-request.md
```diff
@@ -26,70 +28,31 @@ This document does **not** specify how payment requests are communicated from a
...
+3. Sender prepares a local ledger transfer to a connector, held pending the execution condition provided by the recipient and with the ILP Packet provided by the recipient attached.
```
mDuo13: Would be great to link to an explanation of "connectors" in the Architecture RFC or somewhere. The term doesn't make much sense if you're not familiar with Interledger.

0011-interledger-payment-request/0011-interledger-payment-request.md
mDuo13: Maybe it's better, instead of recommending PSK, to define and recommend a system that is very, very similar to PSK.

0011-interledger-payment-request/0011-interledger-payment-request.md
```diff
@@ -2,21 +2,23 @@
 
 ## Preface
 
-This document specifies the Interledger Payment Request (IPR) data format. It is intended for use in protocols in which the recipient of an Interledger payment first communicates a request for payment to the sender.
-
-This document provides an example flow to illustrate how IPR could be used in a higher level protocol. This document also details a method for generating Interledger transfer conditions that enable recipients to use the condition as an integrity check on incoming transfers and packets, without maintaining state about all outstanding payment requests.
+This document specifies the Interledger Payment Request (IPR) transport protocol. It is an end-to-end protocol in which the recipient of an Interledger payment first communicates a request for payment to the sender. This document also details how the recipient can verify that incoming payments match previously created requests without storing a history of all outstanding requests.
```
mDuo13: > ## Preface
> This document specifies the Interledger Payment Request (IPR) transport protocol. It is an end-to-end protocol...

I suggest you drop the "Preface" header and change the opening sentences to read:

> The Interledger Payment Request (IPR) transport protocol is an end-to-end protocol...

0011-interledger-payment-request/0011-interledger-payment-request.md
sharafian: The reason that the IPR spec recommends PSK is that in actual implementation, IPR and PSK are 99% identical. The only difference is that the `generatePskParams` is called per payment in IPR, and at protocol setup for PSK.

emschwartz:
@mDuo13 I agree with @sharafian that it doesn't make sense to redefine the data format and steps since they would mostly duplicate what is written in the PSK spec. I tried to be clearer about why PSK is being used and how the receiver should actually do it. Let me know if you still think we should copy over the steps and pseudocode from PSK.

@justmoon I reordered the sections, let me know if you think that works better.

Also, since the JSON format now has only 2 fields and both of them are binary, I think we might want to get rid of the JSON representation and just make a binary format that puts the condition first and the packet right after it. @justmoon do you want to make an ASN.1 definition or should we just say you concatenate them (you'd know that the condition is the first 32 bytes)?

0011-interledger-payment-request/0011-interledger-payment-request.md
```diff
@@ -1,140 +1,85 @@
...
+### Using a PSK Implementation
 
-#### amount
+The [Pre-Shared Key](../0016-pre-shared-key/0016-pre-shared-key.md) protocol uses the payment condition as a MAC of the payment details by setting the fulfillment to be an HMAC of the ILP packet with a key derived from a secret shared between the sender and receiver. In IPR there is no shared secret and the condition is only generated by the receiver. Nevertheless, an IPR receiver can use the same methods to generate the condition and regenerate the fulfillment on receipt of the incoming transfer.
 
-The amount requested by the recipient, such as `10.25`, denoted in the assets of the recipient's ledger.
+To use a PSK implementation, the receiver follows these steps:
 
-The amount MUST be rounded based on the precision supported by the recipient's ledger.
+1. The receiver MAY derive a receiver secret and address as detailed in [PSK: Shared Secret Generation](../0016-pre-shared-key/0016-pre-shared-key.md#shared-secret-generation) but the receiver DOES NOT share the secret key.
```
mDuo13: "DOES NOT"→"MUST NOT"

0011-interledger-payment-request/0011-interledger-payment-request.md
```diff
@@ -1,140 +1,85 @@
...
+2. The receiver creates the payment packet and condition _using the steps normally executed by the sender_: [PSK: Payment Creation](../0016-pre-shared-key/0016-pre-shared-key.md#1-payment-creation).
+3. The receiver communicates the payment request, including the packet and condition, to the sender.
+4. If the receiver used the derived secret from `1.`, they regenerate that secret from the packet as specified in [PSK: Shared Secret Regeneration](../0016-pre-shared-key/0016-pre-shared-key.md#shared-secret-regeneration).
```
mDuo13: Instead of writing `1.` in code brackets, just say "step 1" here.

0011-interledger-payment-request/0011-interledger-payment-request.md
```diff
@@ -1,140 +1,85 @@
...
+5. The receiver generates the fulfillment from the secret and packet as specified in [PSK: Payment Fulfillment](../0016-pre-shared-key/0016-pre-shared-key.md#2-payment-fulfillment).
 
-Recipients SHOULD ignore incoming transfers whose amounts are less than the amount specified in the payment request. Recipients MAY ignore incoming transfers whose amounts are greater than requested.
+### Alternative Methods and General Best Practices
 
-#### condition
+Receivers MAY use any method they choose for generating the condition, such as using a random fulfillment for each request. However, this also requires storing every outstanding fulfillment and packet in a database.
 
-The Crypto Condition to be used as the execution condition for the payment. See [Condition Generation](#condition-generation) for recommendations the recipient MAY use for creating the condition.
+A PSK implementation will already provide these features, but receivers that choose a different method for generating the conditions SHOULD consider these recommendations:
```
mDuo13: It's not clear what "these features" means here—it almost reads like a PSK implementation includes storing every outstanding fulfillment in a database. Also, avoid future tense ("will provide") unless you're talking about actual future events.

I suggest rephrasing this section as follows:

> Receivers MAY use any method they choose for generating the condition, such as using a random fulfillment for each request. The receiver MUST be able to identify and verify the ILP Packet for every outstanding payment and produce the corresponding fulfillment. A PSK implementation handles these requirements without storing every fulfillment and packet in a database, by using the condition as a Message Authentication Code (MAC) of the payment request.
> 
> Receivers SHOULD follow these guidelines if implementing a custom system:
>
> - Recivers SHOULD include a nonce in the ILP Packet to distinguish it from other packets that would otherwise be identical. Each payment request SHOULD have a unique nonce.
> - Receivers SHOULD include an expiry date in the ILP Packet, so they do not accept incoming packets that are paid much later than their creation date.

0011-interledger-payment-request/0011-interledger-payment-request.md
```diff
@@ -1,140 +1,85 @@
...
+|:--|:--|:--|
+| `packet` | [Base64-Url](https://en.wikipedia.org/wiki/Base64#URL_applications) Encoded ILP Packet | ILP Payment packet including the destination address and amount |
+| `condition` | 32 Bytes, Base64-Url Encoded | Execution condition for the payment |
 
-Additional binary headers, encoded as a base64 string, that should be included in the ILP Packet. The sender SHOULD treat these as opaque.
+In IPR the sender only uses the `account` and `amount` from the ILP packet and MUST treat the `data` as opaque. The sender MUST NOT modify the ILP packet at all, or the receiver will reject the payment.
 
 ### Binary
 
 **TODO**
```
mDuo13: Don't merge this request while this **TODO** is still here.

emschwartz:
@mDuo13's comments have been addressed so I'm going to go ahead and merge this

Issue #47, opened 2016-07-12
RFC-4: Make it clear that this document is JS specific


justmoon:
Some people were confused by IL-RFC 4, because it looks like a standard for all ledger plugins, but several design choices are in reality pretty specific to JavaScript.

Let's make it more clear that RFC-4 is specific to JavaScript.

PR #173, opened 2017-03-23
Clarify that plugin interface is Javascript-specific

emschwartz:
resolves https://github.com/interledger/rfcs/issues/47


sharafian:
Should the filename be changed as well? We might also want to move this to interledgerjs instead of RFCs.

emschwartz:
Changing the filename will break a bunch of links so I don't think we should do that. Also, I think it's fine to have this doc in here, since it's served as our only definition of the ledger interface for a while. At some point we should add another RFC that describes the interface in more generic terms.

Issue #150, opened 2017-01-19
Proposal: Define ILP Errors


emschwartz:
Right now we have the concept of a "rejection reason" when transfers are rejected, but those messages are neither well-defined nor relayed backwards by connectors. 

We should define:
- ILP error codes/types
- Whether additional information can be added (e.g. for debugging purposes)
- Which error codes are final and which indicate a payment should be retried

Here's a first stab at some error cases:

**Connector rejected payment**
- ledger unresponsive
- insufficient liquidity
- insufficient timeout window
- unable to route payment
- temporarily overloaded
- ledger doesn't accept crypto condition
- receiver let payment time out (don't know why)
- invalid packet

**Receiver rejected payment**
- insufficient amount
- invalid packet
- other error (don't retry) - do we want to include a message?
- account doesn't exist (?)
- account cannot accept payment (over transaction or balance limit?)

These cannot necessarily be "trusted", because a connector relaying the message back could modify it, but it still seems helpful for development purposes at least.

adrianhopebailie:
Do we want to include regulatory stuff?
Maybe just a generic `transfer not permitted` or similar?

emschwartz:
ILP [451](https://en.wikipedia.org/wiki/HTTP_451)

sharafian:
I think ILP error messages should also have some indication of what component is rejecting the transfer. A rejecting coming from the ILP Kit API is different from a rejection coming from the ILP Connector, and in the case where both are listening on an account it can help diagnose problems.

adrianhopebailie:
> I think ILP error messages should also have some indication of what component is rejecting the transfer.

Where possible this would be great but I would be wary of making it a hard requirement until we are certain it can be done (i.e. the information is always available at the end-point where the error is handled)

emschwartz:
From @justmoon:
> Let's make sure all ILP rejections always include the ILP address and component type (connector, ledger, plugin, receiver, application-layer, app) of whatever component rejected it.

justmoon:
@sentientwaffle wrote up a proposal: https://gist.github.com/sentientwaffle/ce5af28f16c26f35c453a7b7078e4edf

justmoon:
I like in HTTP that we can look at the first digit of the error code to figure out what to do:

* 1xx Informational --> Just to let you know...
* 2xx Success --> You're good!
* 3xx Redirection --> Go over here!
* 4xx Client Error --> You messed up, probably not worth retrying unless you change the request you made.
[...]

PR #174, opened 2017-03-23
Add errors to Interledger Protocol spec

emschwartz:
resolves https://github.com/interledger/rfcs/issues/150
related to https://github.com/interledger/rfcs/pull/175

### DIFF
--- a/0003-interledger-protocol/0003-interledger-protocol.md
+++ b/0003-interledger-protocol/0003-interledger-protocol.md
@@ -143,7 +143,18 @@ The mapping from addresses to local accounts on a ledger is defined by the ledge
 
 ### Connectors
 
-Connectors implement the Interledger Protocol to forward payments between ledgers. Connectors implement (or include a module that implements) the ledger protocol of the ledgers on which they hold accounts. Connectors also implement the [Connector to Connector Protocol (CCP)](../0010-connector-to-connector-protocol/) to coordinate routing and other interledger control information.
+Connectors implement the interledger protocol to forward payments between ledgers and relay errors back along the path. Connectors implement (or include a module that implements) the ledger protocol of the ledgers on which they hold accounts.
+
+Connectors also implement the [Connector to Connector Protocol (CCP)](../0010-connector-to-connector-protocol/) to coordinate routing and other interledger control information.
+
+### Errors
+
+Errors may be generated at any point as an Interledger payment is being prepared or by the receiver. Connectors that are notified of an outgoing transfer being rejected MUST reject the corresponding incoming transfer with the same error.
+
+Connectors SHOULD include their ILP address in the [`forwardedBy`](#forwardedby) field in the error. Connectors SHOULD NOT modify errors
+in any other way.
+
+See below for the [ILP Error Format](#ilp-error-format) and [ILP Error Codes](#ilp-error-codes).
 
 ## Specification
 
@@ -192,7 +203,105 @@ Amount in discrete units of the receiving ledger's asset type. Note that the amo
 
 Arbitrary data that is attached to the payment. The contents are defined by the transport layer protocol.
 
+### ILP Error Format
+
+Here is a summary of the fields in the ILP error format:
+
+| Field | Type | Short Description |
+|:--|:--|:--|
+| code | IA5String | [ILP Error Code](#ilp-error-codes) |
+| name | IA5String | [ILP Error Code Name](#ilp-error-codes) |
+| triggeredBy | Address | ILP address of the entity that originally emitted the error |
+| forwardedBy | SEQUENCE OF Address | ILP addresses of connectors that relayed the error message |
+| triggeredAt | Timestamp | Time when the error was initially emitted |
+| data | OCTET STRING | Error data provided for debugging purposes |
+
+#### code
+
+    IA5String (SIZE(3))
+
+Error code. For example, `F00`. See [ILP Error Codes](#ilp-error-codes) for the list of error codes and their meanings.
+
+#### name
+
+    IA5String
+
+Error name. For example, `Bad Request`. See [ILP Error Codes](#ilp-error-codes) for the list of error codes and their meanings.
+
+Implementations of ILP SHOULD NOT depend on the `name` instead of the `code`. The name is primarily provided as a convenience to facilitate debugging by humans. If the `name` does not match the `code`, the `code` is the definitive identifier of the error.
+
+#### triggeredBy
+
+[ILP Address](#account) of the entity that originally emitted the error. This MAY be an address prefix if the entity that originally omitted the error is a ledger.
+
+#### forwardedBy
+
+[ILP Addresses](#account) of the connectors that relayed the error message.
+
+#### triggeredAt
+
+    Timestamp ::= GeneralizedTime
+
+Date and time when the error was initially emitted.
+
+#### data
+
+    OCTET STRING (SIZE(0..8192))
+
+Error data provided primarily for debugging purposes. Systems that emit errors SHOULD include additional explanation or context about the issue.
+
+Protocols built on top of ILP that define behavior for certain errors SHOULD specify the encoding format of error `data`.
+
+Unless otherwise specified, `data` SHOULD be encoded as UTF-8.
+
+### ILP Error Codes
+
+Inspired by [HTTP Status Codes](https://tools.ietf.org/html/rfc2616#section-10), ILP errors are categorized based on the intended behavior of the caller when they get the given error.
+
+#### F__ - Final Error
+
+Final errors indicate that the payment is invalid and should not be retried unless the details are changed.
+
+| Code | Name | Description |
+|---|---|---|
+| **F00** | **Bad Request** | Generic sender error. |
+| **F01** | **Invalid Packet** | The ILP packet was syntactically invalid. |
+| **F02** | **Unreachable** | There was no way to forward the payment, because the destination ILP address was wrong or the connector does not have a route to the destination. |
+| **F03** | **Invalid Amount** | The amount is invalid, for example it contains more digits of precision than are available on the destination ledger or the amount is greater than the total amount of the given asset in existence. |
+| **F04** | **Insufficient Destination Amount** | The receiver deemed the amount insufficient, for example you tried to pay a $100 invoice with $10. |
+| **F05** | **Wrong Condition** | The receiver generated a different condition and cannot fulfill the payment. |
+| **F06** | **Unexpected Payment** | The receiver was not expecting a payment like this (the memo and destination address don't make sense in that combination, for example if the receiver does not understand the transport protocol used) |
+| **F07** | **Cannot Receive** | The receiver is unable to accept this payment due to a constraint. For example, the payment would put the receiver above its maximum account balance. |
+| **F99** | **Application Error** | Reserved for application layer protocols. Applications MAY use names other than `Application Error`. |
+
+#### T__ - Temporary Error
+
+Temporary errors indicate a failure on the part of the receiver or an intermediary system that is unexpected or likely to be resolved soon. Senders SHOULD retry the same payment again, possibly after a short delay.
+
+| Code | Name | Description |
+|---|---|---|
+| **T00** | **Internal Error** | A generic unexpected exception. This usually indicates a bug or unhandled error case. |
+| **T01** | **Ledger Unreachable** | The connector has a route or partial route to the destination but was unable to reach the next ledger. Try again later. |
+| **T02** | **Ledger Busy** | The ledger is rejecting requests due to overloading. Try again later. |
+| **T03** | **Connector Busy** | The connector is rejecting requests due to overloading. Try again later. |
+| **T04** | **Insufficient Liquidity** | The connector would like to fulfill your request, but it doesn't currently have enough money. Try again later. |
+| **T05** | **Rate Limited** | The sender is sending too many payments and is being rate-limited by a ledger or connector. If a connector gets this error because they are being rate-limited, they SHOULD retry the payment through a different route or respond to the sender with a `T03: Connector Busy` error. |
+| **T99** | **Application Error** | Reserved for application layer protocols. Applications MAY use names other than `Application Error`. |
+
+#### R__ - Relative Error
+
+Relative errors indicate that the payment did not have enough of a margin in terms of money or time. However, it is impossible to tell whether the sender did not provide enough error margin or the path suddenly became too slow or illiquid. The sender MAY retry the payment with a larger safety margin.
+
+| Code | Name | Description
+|---|---|---|
+| **R00** | **Transfer Timed Out** | The transfer timed out, meaning the next party in the chain did not respond. This could be because you set your timeout too low or because something look longer than it should. The sender MAY try again with a higher expiry, but they SHOULD NOT do this indefinitely or a malicious connector could cause them to tie up their money for an unreasonably long time. |
+| **R01** | **Insufficient Source Amount** | Either the sender did not send enough money or the exchange rate changed before the payment was prepared. The sender MAY try again with a higher amount, but they SHOULD NOT do this indefinitely or a malicious connector could steal money from them. |
+| **R02** | **Insufficient Timeout** | The connector could not forward the payment, because the timeout was too low to subtract its safety margin. The sender MAY try again with a higher expiry, but they SHOULD NOT do this indefinitely or a malicious connector could cause them to tie up their money for an unreasonably long time. |
+| **R99** | **Application Error** | Reserved for application layer protocols. Applications MAY use names other than `Application Error`. |
+
 ## Appendix A: ASN.1 Module
+
+See [ASN.1 Definitions](../asn1/InterledgerProtocol.asn).
 
 ## Appendix B: IANA Considerations
 
--- a/0001-interledger-architecture/0001-interledger-architecture.md
+++ b/0001-interledger-architecture/0001-interledger-architecture.md
@@ -139,6 +139,8 @@ The ledger MUST support preparing, executing, and rejecting transfers in 1 secon
 
 The ledger MUST define an [ILP Address](../0015-ilp-addresses/0015-ilp-addresses.md) prefix and scheme such that accounts on the ledger can be addressed using canonical ILP addresses.
 
+It MUST support receivers (and ONLY receivers) rejecting incoming transfers. Receivers MUST be able to submit a rejection message of up to 65535 bytes. The sender of the transfer MUST be notified that the transfer has been rejected and the notification MUST include the rejection message.
+
 ### Example Protocols
 
 #### Five Bells Ledger Protocol (5BLP)


emschwartz:
Oops, I just saw @earizon's [comment](https://github.com/interledger/rfcs/issues/150#issuecomment-288421094).

Let me know what you all think of this! Happy to discuss the format more.

I don't think we should use JSON errors on the lowest level of the protocol because that would require a parser in a lot of components, but I agree that errors are meant largely for debugging and partially for applications that will handle certain ones automatically.

0017-ilp-errors/0017-ilp-errors.md
```diff
@@ -0,0 +1,73 @@
...
+Relative errors indicate that the payment did not have enough of a margin in terms of money or time. However, it is impossible to tell whether the sender did not provide enough error margin or the path suddenly became too slow or illiquid. The sender MAY retry the payment with a larger safety margin.
+
+| Code | Name | Description
+|---|---|---|
+| **R01** | **Transfer Timed Out** | The transfer timed out, i.e. the next party in the chain did not respond. This could be because you set your timeout too low or because something look longer than it should. Try again with a higher expiry. |
+| **R02** | **Insufficient Source Amount** | Either you didn't send enough money or there wasn't enough liquidity. Try again with a higher sending amount. |
+| **R03** | **Insufficient Timeout** | The connector could not forward the payment, because the timeout was too low to subtract its safety margin. Try again with a higher expiry. |
+
+**R04** - **R50** are reserved for future use.
+**R51** - **R99** are for application-defined errors.
```
sharafian: I don't think this makes sense. If the same error codes are being used by many different applications for different purposes, then they're not very helpful. You might accidentally send to the wrong application, and then get back a confusing error code.

0017-ilp-errors/0017-ilp-errors.md
```diff
@@ -0,0 +1,73 @@
...
+
+ILP connectors MUST relay errors when they are notified of a transfer being rejected.
+
+Connectors SHOULD NOT modify the error codes or messages except in the following way: connectors MAY append their ILP addresses, separated from the error message by a newline (`\n`), to assist with debugging/tracing.
+
+## Data Format
+
+ILP errors are UTF-8 encoded strings that are a maximum of 65535 bytes long. Error messages that are too long for a ledger to handle SHOULD be truncated to the first 65535 bytes.
+
+Errors consist of a three character code, followed by a colon (`:`) and an arbitrary message set by the entity that raised the error. Transport and application layer protocols MAY define error messages that have special significance in their context.
```
sharafian: Should the error name be included anywhere?

0017-ilp-errors/0017-ilp-errors.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Error Propagation
+
+ILP connectors MUST relay errors when they are notified of a transfer being rejected.
+
+Connectors SHOULD NOT modify the error codes or messages except in the following way: connectors MAY append their ILP addresses, separated from the error message by a newline (`\n`), to assist with debugging/tracing.
+
+## Data Format
+
+ILP errors are UTF-8 encoded strings that are a maximum of 65535 bytes long. Error messages that are too long for a ledger to handle SHOULD be truncated to the first 65535 bytes.
```
sharafian: Are error codes ever going to be wrapped in another object, like the data of an ILP packet? If so, we probably want to make this limit lower.

0017-ilp-errors/0017-ilp-errors.md
```diff
@@ -0,0 +1,73 @@
...
+
+### S__ - Sender Error
+
+Sender errors indicate that the payment is invalid and should not be retried unless the details are changed.
+
+| Code | Name | Description |
+|---|---|---|
+| **S00** | **Bad Request** | Generic sender error. |
+| **S01** | **Invalid Packet** | The ILP packet was syntactically invalid. |
+| **S02** | **Unreachable** | There was no way to forward the payment, because the destination ILP address was wrong or unreachable. |
```
sharafian: Would it be a good idea to have a temporary error for an unreachable destination? There could be problems due to route broadcasting that make a destination unreachable, which will resolve themselves over time.

0017-ilp-errors/0017-ilp-errors.md
```diff
@@ -0,0 +1,73 @@
...
+| **S05** | **Wrong Condition** | The receiver generated a different condition and cannot fulfill the payment. |
+| **S06** | **Unexpected Payment** | The receiver was not expecting a payment like this (the memo and destination address don't make sense in that combination, for example if the receiver does not understand the transport protocol used) |
+| **S07** | **Cannot Receive** | The receiver is unable to accept this payment due to a constraint, e.g. a limit was exceeded. |
+
+**S08** - **S50** are reserved for future use.
+**S51** - **S99** are for application-defined errors.
+
+### T__ - Temporary Error
+
+Temporary errors indicate a failure on the part of the receiver or an intermediary system that is unexpected or likely to be resolved soon. Senders SHOULD retry the same payment again, possibly after a short delay.
```
sharafian: Temporary errors could have a recommended timeout carried with them, depending on the error. Although that might cause problems if someone manipulates it.

0017-ilp-errors/0017-ilp-errors.md
```diff
@@ -0,0 +1,73 @@
...
+**T51** - **T99** are for application-defined errors.
+
+### R__ - Relative Error
+
+Relative errors indicate that the payment did not have enough of a margin in terms of money or time. However, it is impossible to tell whether the sender did not provide enough error margin or the path suddenly became too slow or illiquid. The sender MAY retry the payment with a larger safety margin.
+
+| Code | Name | Description
+|---|---|---|
+| **R01** | **Transfer Timed Out** | The transfer timed out, i.e. the next party in the chain did not respond. This could be because you set your timeout too low or because something look longer than it should. Try again with a higher expiry. |
+| **R02** | **Insufficient Source Amount** | Either you didn't send enough money or there wasn't enough liquidity. Try again with a higher sending amount. |
```
sharafian: It isn't safe behavior to automatically retry with a higher source amount. A malicious connector could take advantage of it.

0017-ilp-errors/0017-ilp-errors.md
emschwartz: I was thinking you might want to include an ILP packet in them, or something like that. But in any case, I'm not sure what the right limit is.

0017-ilp-errors/0017-ilp-errors.md
emschwartz: We could make it:
a) `<Error Code>: <Error Name>: <Error Message>`
b) `<Error Code>: <Error Name>\n
<Error Message>`
c) something else

0017-ilp-errors/0017-ilp-errors.md
emschwartz: That's `T01`

0017-ilp-errors/0017-ilp-errors.md
emschwartz: If one application sends to a completely different one I think it's more likely you'd get an error that means the application didn't understand what you sent.

That said, I wasn't sure about what to do with the unused codes so I put this in there to start a conversation about it. Happy to take it out if we don't like it.

0017-ilp-errors/0017-ilp-errors.md
sharafian: I thought `T01` was a ledger issue, not a routing one? Maybe the description could be tweaked to include both cases.

0017-ilp-errors/0017-ilp-errors.md
sentientwaffle: Should the error include a `triggered_at` timestamp? What about the address of the receiver/ledger/connector that generated the error.

0017-ilp-errors/0017-ilp-errors.md
emschwartz: You could do it within limits. I'll make a note about that

0017-ilp-errors/0017-ilp-errors.md
emschwartz: I think both of those would be nice to have.

This makes me think we should use the HTTP data format for errors, so it would be something like:
```http
ILP/1.0 S00: Bad Request
Triggered-By: example.ledger.bob
Forwarded-By: example.connector.connie,example.connector.chloe
Date: 2017-03-23T00:00:00Z
Content-Type: application/json

{
  "id": "SomeApplicationError",
  "message": "Oops..."
}
```

Thoughts?

emschwartz: This could be handled by https://github.com/interledger/rfcs/pull/174#discussion_r107709655

0017-ilp-errors/0017-ilp-errors.md
emschwartz: Actually I think connectors need to respond with `S02` if they don't know how to get to the address. They can't assume that they'll know how to find it later.

You're right that `T01` is for when the connector can't communicate with the ledger.

0017-ilp-errors/0017-ilp-errors.md
sentientwaffle: Instead of `Date` I think it should be called `Triggered-At` to distinguish the time the original error occurred from when it was last forwarded.

earizon: What happens for asynchronous error events and/or non-HTTP protocols?. I'm thinking of a wallet-client with just a websocket connection to the ILP-ledger. Such wallet-client sends an HTTP REST request with an "accept payment". The initial HTTP request is validated and forwarded (no-error is returned). The final recipient notice an S?? error later on (wrong condition for example) and wants to propagate it to the originating wallet. For the client to be notified all the information must be encapsulated in an (JSON?) object since the websocket channel has no notion of request headers.

Mabye will more conservative to define first the information the error message (serializable object in OOP terminology) must carry and define then, for each type of transport (syncrhonous HTTP request/response, websockets, gRPC, ...), the codification used to transmit it (header + json, json alone, any other future format, ...)? (or maybe I'm completly lost)

emschwartz: Sorry, to clarify, these are not actual HTTP errors. My suggestion was to define ILP errors using the HTTP header and data format (in the same way we used it for [PSK](https://github.com/interledger/rfcs/blob/master/0016-pre-shared-key/0016-pre-shared-key.md#data-format)).

Errors would be propagated through ILP connectors (even if they are not using HTTP to talk to one another) via the "rejection message", which a sender or connector would get back instead of the ILP condition fulfillment.

justmoon:
Agree on the semantics - regarding the syntax:

We've been using binary (OER) on the ILP layer so far and since ILP errors are on the ILP layer (although they are used by higher level protocols), they should probably be binary/OER as well.

0017-ilp-errors/0017-ilp-errors.md
```diff
@@ -0,0 +1,73 @@
...
+| **S01** | **Invalid Packet** | The ILP packet was syntactically invalid. |
+| **S02** | **Unreachable** | There was no way to forward the payment, because the destination ILP address was wrong or unreachable. |
+| **S03** | **Invalid Amount** | The amount is invalid, e.g. it contains more digits of precision than are available on the destination ledger or the amount is greater than the total amount of the given asset in existence. |
+| **S04** | **Insufficient Destination Amount** | The amount was insufficient (e.g. you tried to pay a $100 invoice with $10). |
+| **S05** | **Wrong Condition** | The receiver generated a different condition and cannot fulfill the payment. |
+| **S06** | **Unexpected Payment** | The receiver was not expecting a payment like this (the memo and destination address don't make sense in that combination, for example if the receiver does not understand the transport protocol used) |
+| **S07** | **Cannot Receive** | The receiver is unable to accept this payment due to a constraint, e.g. a limit was exceeded. |
+
+**S08** - **S50** are reserved for future use.
+**S51** - **S99** are for application-defined errors.
```
mDuo13: (Was going to comment here, but I'll move it down to Ben's discussion on the same idea below.)

0017-ilp-errors/0017-ilp-errors.md
```diff
@@ -0,0 +1,73 @@
...
+
+### T__ - Temporary Error
+
+Temporary errors indicate a failure on the part of the receiver or an intermediary system that is unexpected or likely to be resolved soon. Senders SHOULD retry the same payment again, possibly after a short delay.
+
+| Code | Name | Description |
+|---|---|---|
+| **T00** | **Internal Error** | Like HTTP 500. Something threw an exception, that's all we know. Try again later after we've had time to fix it. |
```
mDuo13: Suggested rephrase:

> A generic unexpected exception. This usually indicates a bug or unhandled error case.

0017-ilp-errors/0017-ilp-errors.md
```diff
@@ -0,0 +1,73 @@
...
+
+Relative errors indicate that the payment did not have enough of a margin in terms of money or time. However, it is impossible to tell whether the sender did not provide enough error margin or the path suddenly became too slow or illiquid. The sender MAY retry the payment with a larger safety margin.
+
+| Code | Name | Description
+|---|---|---|
+| **R01** | **Transfer Timed Out** | The transfer timed out, i.e. the next party in the chain did not respond. This could be because you set your timeout too low or because something look longer than it should. Try again with a higher expiry. |
+| **R02** | **Insufficient Source Amount** | Either you didn't send enough money or there wasn't enough liquidity. Try again with a higher sending amount. |
+| **R03** | **Insufficient Timeout** | The connector could not forward the payment, because the timeout was too low to subtract its safety margin. Try again with a higher expiry. |
+
+**R04** - **R50** are reserved for future use.
```
mDuo13: Maybe we should add an error code for when people are explicitly rate-limited? (Equivalent to [Twitter's "420 Enhance Your Calm" error](https://dev.twitter.com/overview/api/response-codes))

0017-ilp-errors/0017-ilp-errors.md
mDuo13: Do we have a rule or registration process for keeping application-defined errors different from one another? With only 49 error codes to work with, I suspect it's very likely different applications will pick the same error codes to mean different things. With the way error codes are passed back through the chain, it seems like there's not much room for apps to namespace their own errors.

I don't think applications necessarily need to define their own top-level error codes if they have a structured field in the body where they can provide sub-codes, but maybe it would be helpful to have just one code set aside for application-defined errors.

earizon:
I created a gist with a "mental map" about what I think must be included in the Error Protocol (And error handling). Far away from complete but could be useful:

https://gist.github.com/earizon/f62198c54429a4d43de1ebb486f5f6b2

0017-ilp-errors/0017-ilp-errors.md
```diff
@@ -0,0 +1,73 @@
...
+
+Sender errors indicate that the payment is invalid and should not be retried unless the details are changed.
+
+| Code | Name | Description |
+|---|---|---|
+| **S00** | **Bad Request** | Generic sender error. |
+| **S01** | **Invalid Packet** | The ILP packet was syntactically invalid. |
+| **S02** | **Unreachable** | There was no way to forward the payment, because the destination ILP address was wrong or unreachable. |
+| **S03** | **Invalid Amount** | The amount is invalid, e.g. it contains more digits of precision than are available on the destination ledger or the amount is greater than the total amount of the given asset in existence. |
+| **S04** | **Insufficient Destination Amount** | The amount was insufficient (e.g. you tried to pay a $100 invoice with $10). |
```
emschwartz: S04 is only for the receiver to emit and means "the packet and transfer got here just fine, but it's less than you need to pay me for me to accept it".
S03 means the amount was just invalid
R02 is used when the individual transfer amount isn't enough for the connector to pass it on

emschwartz:
@earizon Thanks for the write up!

What is the error id for?

> How can the ledger extract the needed info to associate the error received with the internal transacctions to rollback?

The way errors are passed back along the chain is that connectors or receiver can reject the incoming transfer with the error message attached (in lieu of submitting the fulfillment). I added a note about that being a requirement for the ledger layer [here](https://github.com/interledger/rfcs/pull/174/commits/e4483888ee8f4fa3945ed2c8ed13b0d5c568bcf4).

> An ILP entity (sender, receiver, ...) can be interested in subcribing (receiving error notificitions) to some error caused by some other ILP/non-ILP entity

My assumption is that all the parties involved in a payment, and only them, would be "subscribed" to get an error notification. When an individual transfer is rejected with an error message, the sender and receiver of that transfer will be notified. If the sender was a connector, they are supposed to relay the error message back along the path by rejecting the incoming transfer that was intended for them with the same error message. I added a bit more description of this to the spec.

Regarding the idea of including a `solution` field in the error, wouldn't the solution for each error type almost always be the same?

emschwartz:
@justmoon @sharafian @mDuo13 I switched the error format to use the ASN.1/OER format @justmoon [wrote up](https://github.com/interledger/rfcs/pull/175), took out the mention of "reserved codes", addressed your other comments, and added the error codes to RFC-3.

0003-interledger-protocol/0003-interledger-protocol.md
```diff
@@ -172,77 +131,171 @@ As with the [internet protocol](https://tools.ietf.org/html/rfc791#section-2.3),
...
+
+    Timestamp ::= GeneralizedTime
+
+Date and time when the error was initially emitted.
+
+#### data
+
+    OCTET STRING (SIZE(0..8192))
+
+Error data provided for debugging purposes. Protocols built on top of ILP SHOULD specify the encoding format of error `data`.
```
mDuo13: Maybe it would be easier and more self-contained if the protocol contained a content-type field to specify how the data is encoded?

0003-interledger-protocol/0003-interledger-protocol.md
```diff
@@ -172,77 +131,171 @@ As with the [internet protocol](https://tools.ietf.org/html/rfc791#section-2.3),
...
+| **T05** | **Rate Limited** | The sender is sending too many payments and is being rate-limited by a ledger or connector. |
+| **T99** | **Application Error** | Reserved for application layer protocols. |
+
+#### R__ - Relative Error
+
+Relative errors indicate that the payment did not have enough of a margin in terms of money or time. However, it is impossible to tell whether the sender did not provide enough error margin or the path suddenly became too slow or illiquid. The sender MAY retry the payment with a larger safety margin.
+
+| Code | Name | Description
+|---|---|---|
+| **R01** | **Transfer Timed Out** | The transfer timed out, i.e. the next party in the chain did not respond. This could be because you set your timeout too low or because something look longer than it should. The sender MAY try again with a higher expiry, but they SHOULD NOT do this indefinitely or a malicious connector could cause them to tie up their money for an unreasonably long time. |
```
mDuo13: Why no **R00**?

0003-interledger-protocol/0003-interledger-protocol.md
```diff
@@ -172,77 +131,171 @@ As with the [internet protocol](https://tools.ietf.org/html/rfc791#section-2.3),
...
+| **R02** | **Insufficient Source Amount** | Either you didn't send enough money or there wasn't enough liquidity. The sender MAY try again with a higher amount, but they SHOULD NOT do this indefinitely or a malicious connector could steal money from them. |
+| **R03** | **Insufficient Timeout** | The connector could not forward the payment, because the timeout was too low to subtract its safety margin. The sender MAY try again with a higher expiry, but they SHOULD NOT do this indefinitely or a malicious connector could cause them to tie up their money for an unreasonably long time. |
+| **R99** | **Application Error** | Reserved for application layer protocols. |
 
-3. The sender and receiver MAY appoint a mutually trusted third-party which has an account on the local ledger. The sender performs a regular transfer into a neutral third-party account. In the first step, funds are transfered into the account belonging to the neutral third-party.
 ### Payment Channels
```
mDuo13: Don't leave an empty "Payment Channels" header in place here.

earizon:
@emschwartz
> What is the error id for?

I was thinking of it as a global (or "global enough") unique identifier to help filtering and comparing log traces between different machines in case of forensic analysis of failures.

> Regarding the idea of including a solution field in the error, wouldn't the solution for each error type almost always be the same?

 From my own experience, for some type of errors the solution field will always be the same. For some other errors where the external context can vary widely the error type could not suffice or a given component can have extra information to help fix (or at least explain) the issue.  Some examples I'm thinking about:

- "Insufficient Source Amount" most probably doesn't need any extra "solution" field, since it's quite obvious from the error code what the problem is and how to fix it.

- For any  "Time Out" related error, a connector can suggest a prefered new timeout (because it has extra-information about peers or routes).

- For an "Unreachable" error,  a ledger/connector can also provide extra information, like "us.ledger1. has been moved to us.ledger2., please change the ILP Address and try again", or "us.ledger1. is temporaly down due to maintenance work. The scheduled restart is 4:00am UTC", ...

- For "Invalid Packet" a possible solution could be "The packet is using a deprecated format, update to new ILP packet version X" or "field X in packet must include the recipient amount using format ...."

- ....

As you see, the solution provided depends a lot on "how smart" a given piece of software is about its running context. Some simple implementations will simply return no data in this field, while others can be really smart in providing hints and guides to help fix the problem ASAP

0003-interledger-protocol/0003-interledger-protocol.md
emschwartz: I don't think this is necessary because all higher level protocols will want to standardize that anyway, and there's no reason to support multiple content types at the same time.

Note that this is very similar to how there's no content type specified for the data in an ILP Payment Packet.

emschwartz:
@earizon I don't think we should include a UUID for errors but your other suggestion makes me think this spec is still missing something.

Ever since I heard the comment from @justmoon that "protocols are only finished when there's nothing left to take out", I've been thinking that we shouldn't include any fields unless they're absolutely necessary for the functioning of the specific protocol in question.

Regarding error-specific UUIDs, I think they are primarily for a somewhat rare case that I don't think merits including them in the protocol definition. Namely the case is when you want to track an error across multiple systems. The default assumption we've been using is that most ledgers and connectors will be operated by independent parties. In this case, you wouldn't normally have visibility into the logs of multiple operators, and I don't think it'll be worth it in most cases to call up another operator to ask whether they passed on the same error message for a given transaction (since that's what the spec says they're supposed to do anyway). Also, since ILP payments don't include IDs by default, I don't think it makes sense to include them in the errors either. Errors should be stored with the transfer they were used to reject so each ledger should have the error stored alongside the original ILP packet, which may have a unique ID you could use for this purpose.

Regarding the details on how to solve a specific error, I was thinking that the `data` field in the error would be set by the transport protocol. However that doesn't make sense in all cases, especially when it's an intermediary connector emitting the error in the first place. I need to think a little more about your idea of having a "suggested resolution" field in addition to the description of the error itself.

emschwartz:
Open questions for this PR:

1. Should errors include the error names? If so, what do you do if the name is something other than what the standard tells you to expect?
2. What should intermediary connectors put in the `data` field if they emit an error (i.e. an error that is on the Interledger layer, not on the transport layer) and how should they encode that data? Is that meant to be machine-readable or just human readable?
3. Should errors include a field for recommended fix aside from the specific description of the problem that caused the error? If so, should that be machine readable or just human readable?

@earizon @justmoon @mDuo13 @sentientwaffle

justmoon:
> Should errors include the error names?

Yes.

> If so, what do you do if the name is something other than what the standard tells you to expect?

Ignore it.

(Your implementation doesn't need to even have a table with the "standard" mappings, it should just print what it got, code and message.)

> What should intermediary connectors put in the data field if they emit an error (i.e. an error that is on the Interledger layer, not on the transport layer) and how should they encode that data? Is that meant to be machine-readable or just human readable?

Generally, it should be only human-readable. Although I could certainly see us encountering a case in the future where we go: "Gee, it sure would be nice if we could pass back some data with this error."

For now, I would poke the beehive by saying: the `data` field should always be empty. That should surface whatever cases we have where people feel strongly that including data would be beneficial. Once we have some examples, we can look at it in context.

> Should errors include a field for recommended fix aside from the specific description of the problem that caused the error? If so, should that be machine readable or just human readable?

No, definitely not. Errors aren't the right place to put documentation.

If you want a great user experience, let's create a directory on interledger.org, e.g. `https://interledger.org/errors/` where we describe and explain each error and then let's make implementations print a URL like `https://interledger.org/errors/S01?mesage=...&data=...` along with the error itself. That page can then provide contextual documentation on how to fix it.

emschwartz:
> For now, I would poke the beehive by saying: the data field should always be empty. That should surface whatever cases we have where people feel strongly that including data would be beneficial. Once we have some examples, we can look at it in context.

These errors could be emitted by an intermediary connector, in which case the data format wouldn't be specified by the transport protocol:
**F01	Invalid Packet** - validation errors might be helpful
**F03	Invalid Amount** - how much precision does the ledger support (note: we might not need this error anymore now that we have integer amounts, I think this was more applicable with floats)
**T01	Ledger Unreachable** - @earizon suggested if there's scheduled maintenance or something like that it could tell you when it's going to be back up
**R01	Insufficient Source Amount** - maybe it would suggest the amount it would do it for, instead of making you guess?
**R02	Insufficient Timeout** - maybe it would tell you how long you need to put it on hold for?

> Generally, it should be only human-readable. Although I could certainly see us encountering a case in the future where we go: "Gee, it sure would be nice if we could pass back some data with this error."

You mean just for errors generated on the Interledger layer, right? Transport and application protocols may want to define specific errors and formats that can be handled programmatically (e.g. unsupported cipher suite, supported suites include:...)

emschwartz:
One more question: should ILQP use the same ILP errors?

justmoon:
> One more question: should ILQP use the same ILP errors?

Unless someone has a reason not to - I'd say definitely, yeah.

earizon:
@emschwartz 
> Errors should be stored with the transfer they were used to reject so each ledger should have the error stored alongside the original ILP packet, which may have a unique ID you could use for this purpose.

I agree. Looks definetly an extra unique ID is useless.


@emschwartz @justmoon 
>"  Should errors include a field for recommended fix aside from the specific description of the problem that caused the error? If so, should that be machine readable or just human readable?"

>> No, definitely not. Errors aren't the right place to put documentation."

The specific description of the problem that caused the error has nothing to do with documentation. It has to do with the running context that is specific and unique to each ledger or connector. I mention a few examples in a previous comment. 
An example of daily life. I go to the bakery at 11:00am, and it's closed when I expected it to be open. The standard contemplated error would be a "ERROR-CLOSED".  The specific description of the problem would be "This bakery is closed until 12:00am" or "Due to some remodeling work we have moved to Street ....".  A specific context-aware description can help humans take a sensible decision. It has more to do with final-user-experience. It can be very frustrating to final users not knowing why they can't buy a ticket with an app, when usually they have no problems. 

Notice also that standard error codes are not magical and will not help to fix many errors even if they are very computer friendly and easy to implement/parse. Take for example the (very) well defined error codes of POSIX/UNIX calls to the OS. (http://man7.org/linux/man-pages/man3/errno.3.html). Even if operating system theory is well studied and understood for decades, most C/C++ software, when executing Unix calls will limit to check if the return code was cero (no error) or not. In the later case they log a warn with the errno coude reseting any related resources or just exiting. I just took a look at apache httpd daemon,  and around 50% of errno checks are just logged and ignored, considering they are not harmful, around 20% are just compared  to be equal or distinct from cero, and the remaining 30% are actually compared against a well defined error code to retry, reconnect, .... The theory about error handling is very controversial. Nobody agree because nobody agree about what an error really is.  For example take the arguments in favor / against checked exceptions. Does it makes sense to use checked exceptions to capture errors at compile time as Java does?. Intuition and common sense says it does, but practice says it doesn't. (In fact, the very successful Spring library just limits to wrap compile-time checked exceptions in runtime unchecked exceptions). 

My "political" position is that many errors (Unreachable, Cannot Receive, ...) will reflect a failure of the whole service, not just of a particular client request and that software will not be able to take intelligent decisitions to restablish the service limiting to reject client requests until an external human operator resets the system. This is why a consider so important to provide context-aware details of the error if possible. It can improve the user-experience and easify life of system administrators (now devOps).

emschwartz:
@earizon How about this: https://github.com/interledger/rfcs/pull/174/commits/c6feabddc7888e7a82bd69f67d83d402ff6cf896?

I agree that that kind of debugging information is helpful. We can switch the implementation to put the same details that are currently included in the JSON error `message` field (see [here](https://github.com/interledgerjs/ilp-connector/blob/master/src/models/payments.js#L56)) into the error's `data` field.

Let me know what you think

earizon:
@emschwartz That will suffice from my point of view.

0003-interledger-protocol/0003-interledger-protocol.md
```diff
@@ -243,10 +254,105 @@ Not all ledgers support held transfers. In the case of a ledger that doesn't, th
...
+
+    IA5String (SIZE(3))
+
+Error code. For example, `F00`. See [ILP Error Codes](#ilp-error-codes) for the list of error codes and their meanings.
+
+#### name
+
+    IA5String
+
+Error name. For example, `Bad Request`. See [ILP Error Codes](#ilp-error-codes) for the list of error codes and their meanings.
```
mDuo13: As mentioned in the previous PR, we should clarify that programs SHOULD NOT depend on the exact name instead of the code. (If the two don't match the existing definitions, the code is the definitive identifier of the error; the name is a convenience to facilitate debugging by humans.)

0003-interledger-protocol/0003-interledger-protocol.md
```diff
@@ -243,10 +254,105 @@ Not all ledgers support held transfers. In the case of a ledger that doesn't, th
...
+
+#### triggeredBy
+
+[ILP Address](#account) of the entity that originally emitted the error.
+
+#### forwardedBy
+
+[ILP Addresses](#account) of the connectors that relayed the error message.
```
mDuo13: Should we clarify that the this could be an address prefix, for example if the entity that originally emitted the error is a ledger?

0003-interledger-protocol/0003-interledger-protocol.md
```diff
@@ -243,10 +254,105 @@ Not all ledgers support held transfers. In the case of a ledger that doesn't, th
...
+Temporary errors indicate a failure on the part of the receiver or an intermediary system that is unexpected or likely to be resolved soon. Senders SHOULD retry the same payment again, possibly after a short delay.
+
+| Code | Name | Description |
+|---|---|---|
+| **T00** | **Internal Error** | A generic unexpected exception. This usually indicates a bug or unhandled error case. |
+| **T01** | **Ledger Unreachable** | The connector has a route or partial route to the destination but was unable to reach the next ledger. Try again later. |
+| **T02** | **Ledger Busy** | The ledger is rejecting requests due to overloading. Try again later. |
+| **T03** | **Connector Busy** | The connector is rejecting requests due to overloading. Try again later. |
+| **T04** | **Insufficient Liquidity** | The connector would like to fulfill your request, but it doesn't currently have enough money. Try again later. |
+| **T05** | **Rate Limited** | The sender is sending too many payments and is being rate-limited by a ledger or connector. |
```
mDuo13: Not just "the sender", but other entities (for example, a connector in the middle of the path) might be the one being rate limited. It might also be a limit on payments received, not sent, right?

0003-interledger-protocol/0003-interledger-protocol.md
```diff
@@ -243,10 +254,105 @@ Not all ledgers support held transfers. In the case of a ledger that doesn't, th
...
+| **T99** | **Application Error** | Reserved for application layer protocols. Applications MAY use names other than `Application Error`. |
+
+#### R__ - Relative Error
+
+Relative errors indicate that the payment did not have enough of a margin in terms of money or time. However, it is impossible to tell whether the sender did not provide enough error margin or the path suddenly became too slow or illiquid. The sender MAY retry the payment with a larger safety margin.
+
+| Code | Name | Description
+|---|---|---|
+| **R00** | **Transfer Timed Out** | The transfer timed out, i.e. the next party in the chain did not respond. This could be because you set your timeout too low or because something look longer than it should. The sender MAY try again with a higher expiry, but they SHOULD NOT do this indefinitely or a malicious connector could cause them to tie up their money for an unreasonably long time. |
```
mDuo13: Avoid Latin abbreviations like "i.e."—in this case, I suggest you rephrase:

> The transfer timed out, meaning the next party in the chain did not respond.

By the way, when you say "did not respond," do you mean they did not produce the fulfillment, they did not prepare another transfer on the next ledger, or they didn't message back to confirm what they've done? It's probably useful to be clearer in what context we're expecting a response here.

0003-interledger-protocol/0003-interledger-protocol.md
emschwartz: Good point that a connector might also be rate-limited.

I don't think it's actually safe to institute a rate limit on the receiving side. If there were, I could prevent you from receiving money just by sending a load of transfers to your account (with conditions I know you can't fulfill so it doesn't cost me anything), and then the ledger would start failing transfers to you.

0003-interledger-protocol/0003-interledger-protocol.md
emschwartz: Added a comment about that to `emittedBy`. I don't think it would be worth having ledgers include their prefixes in the `forwardedBy` array

0003-interledger-protocol/0003-interledger-protocol.md
emschwartz: Added

PR #175, opened 2017-03-24
ASN.1: ILP Errors and IPR

justmoon:
This is just a first attempt. LMK your feedback.

Not sure about the additional data in the errors. Rather than having a content type, I was thinking of just making it JSON by convention. I don't think computers should ever really look at this field. They should only use the code. The `data` is really just for human developers. So making it an UTF8String and using JSON by convention seems perfectly fine.

### DIFF
--- a/asn1/InterledgerPaymentRequest.asn
+++ b/asn1/InterledgerPaymentRequest.asn
@@ -1,0 +1,21 @@
+InterledgerProtocol
+DEFINITIONS
+AUTOMATIC TAGS ::=
+BEGIN
+
+IMPORTS
+    UInt8
+    UInt256
+    FROM GenericTypes
+;
+
+InterledgerPaymentRequest ::= SEQUENCE {
+    -- IPR Version
+    version UInt8,
+    -- ILP Payment Packet
+    packet OCTET STRING,
+    -- Cryptographic condition
+    condition UInt256
+}
+
+END
--- a/asn1/InterledgerPaymentRequest.asn
+++ b/asn1/InterledgerPaymentRequest.asn
@@ -12,10 +12,10 @@ IMPORTS
 InterledgerPaymentRequest ::= SEQUENCE {
     -- IPR Version
     version UInt8,
-    -- ILP Payment Packet
-    packet OCTET STRING,
     -- Cryptographic condition
-    condition UInt256
+    condition UInt256,
+    -- ILP Payment Packet
+    packet OCTET STRING
 }
 
 END
--- a/asn1/InterledgerProtocol.asn
+++ b/asn1/InterledgerProtocol.asn
@@ -7,14 +7,15 @@ IMPORTS
     UInt64
     FROM GenericTypes
 
-    Address
+    Address,
+    Timestamp
     FROM InterledgerTypes
 
     MetadataEntry
     FROM MetadataEntry
 ;
 
-InterledgerProtocolPaymentMessage ::= SEQUENCE {
+InterledgerProtocolPayment ::= SEQUENCE {
     -- Amount which must be received at the destination
     amount UInt64,
     -- Destination ILP Address
@@ -25,6 +26,23 @@ InterledgerProtocolPaymentMessage ::= SEQUENCE {
     extensions SEQUENCE {
         ...
     }
+}
+
+InterledgerProtocolError ::= SEQUENCE {
+    -- Standardized error code
+    code IA5String (SIZE (3)),
+    -- Corresponding error name
+    codeName IA5String,
+    -- Participant that originally emitted the error
+    triggeredBy Address,
+    -- Participants that forwarded the error
+    -- TODO: This could be more trouble than it's worth - could get really big etc.
+    forwardedBy SEQUENCE OF Address,
+    -- Time of emission
+    triggeredAt Timestamp,
+    -- Additional data
+    contentType IA5String,
+    data OCTET STRING (SIZE (0..8192))
 }
 
 END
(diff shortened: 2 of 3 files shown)


asn1/InterledgerPaymentRequest.asn
```diff
@@ -0,0 +1,21 @@
...
+    FROM GenericTypes
+;
+
+InterledgerPaymentRequest ::= SEQUENCE {
+    -- IPR Version
+    version UInt8,
+    -- ILP Payment Packet
+    packet OCTET STRING,
+    -- Cryptographic condition
+    condition UInt256
```
emschwartz: Shouldn't the condition come first since it's the fixed-length value?

asn1/InterledgerPaymentRequest.asn
```diff
@@ -0,0 +1,21 @@
...
+}
```
emschwartz: We _might_ want to add an expiry to this format to let the sender know how long they have to execute it. The packet in the IPR may have an expiry in the data field but the current spec says the sender must treat the data part of the packet as opaque.

asn1/InterledgerProtocol.asn
```diff
@@ -27,4 +28,21 @@ InterledgerProtocolPaymentMessage ::= SEQUENCE {
     }
 }
 
+InterledgerProtocolError ::= SEQUENCE {
+    -- Standardized error code
+    code IA5String (SIZE (3)),
+    -- Corresponding error name
+    codeName IA5String,
```
emschwartz: If this is being transmitted in binary what's the point of including the name? You need a parser anyway so that parser might as well give you the human-readable name

asn1/InterledgerProtocol.asn
```diff
@@ -27,4 +28,21 @@ InterledgerProtocolPaymentMessage ::= SEQUENCE {
...
+    -- Participant that originally emitted the error
+    triggeredBy Address,
+    -- Participants that forwarded the error
+    -- TODO: This could be more trouble than it's worth - could get really big etc.
```
emschwartz: If connectors did actually include their addresses, you could use a failed payment as a trace route (not sure if that's a good thing or bad)

emschwartz:
> Not sure about the additional data in the errors. Rather than having a content type, I was thinking of just making it JSON by convention. 

I don't think this format needs the content type but I also don't think we should make it JSON by convention. I think it makes most sense for the error format to be determined by higher-level protocols. For an application protocol that uses JSON it would definitely make sense to have JSON errors. For one that doesn't, I don't think it would make as much sense.

> I don't think computers should ever really look at this field.

The gap I see right now is on the transport layer. I think it's very possible that transport layer protocols might define specific behavior when you get certain errors. The issue with the current setup is that you might get something like an `S06: Unexpected Payment` but if there isn't a good place to put a machine-readable explanation like `Unsupported Encryption Algorithm. Supported Algorithms: ...`, you can't handle that case automatically.

asn1/InterledgerProtocol.asn
emschwartz: Also should this field just be called `name` instead?

emschwartz: What should implementations do if the `codeName` doesn't match the expected one?

earizon:
I was just starting to warn-up with ASN.1 and java encoding/decoding but not even having typed my first line of code I got stuck with some serious problems. Most importantly, It doesn't look to exist any good opensource library for ASN.1 in the wild. Googling around that's all I have found:
https://www.openmuc.org/asn1/              No support for OER.
https://sourceforge.net/projects/jac-asn1/ No support for OER.
https://sourceforge.net/projects/jasn/     Licensed, No support for OER
http://www.oss.com/ (Nokalva)                 Licensed
https://www.obj-sys.com/products/asn1c/    Licensed

Some "stackoverflow-ers" looks also quite pesimistic about the current status of ASN.1. Extracted from:
http://stackoverflow.com/questions/5964269/asn-1-encoding-decoding

answer 1:
""" ...for a long time there was no full ASN.1 implementation available, at least for the public. ...
 And to make matters worse, ASN.1, in particular it highly redundant encoding schemes (there are at least 3 different ways to encode strings in ASN.1) used to be the cause for several security issues in the last years, due to the problems it caused in properly processing x509 certificates.... 
So to make a long story short: ASN.1 is broken and should be avoided in new designs. It's only major widespread use outside of telephone networks is x509 which is broken, too. Thus I'd not use it. Use JSON, BSON, Protocol Buffers, Netstrings or something sane.
"""

answer 2:
"""ASN.1 has become something of a niche, used for X.509-related data and about nothing else.  You might want to look at Google Protocol Buffers instead."""

(There are also some positive comments defending it)

asn1/InterledgerProtocol.asn
justmoon: > If this is being transmitted in binary what's the point of including the name? You need a parser anyway so that parser might as well give you the human-readable name

There may be new names your implementation isn't aware of. 

> Also should this field just be called name instead?

:+1:

> What should implementations do if the codeName doesn't match the expected one?

Show the name from the error. I'm not sure implementations *reading* the error would even consult a lookup table for the error names. The name is just for the developer's benefit and it makes sense to have it be generated by the party that generates the error for the reason above.

asn1/InterledgerProtocol.asn
justmoon: Great observation. I think it would be good for us to have a traceroute mechanism. If we can repurpose some existing mechanism and not have to create a dedicated one that would be awesome.

Just like IP traceroute, if you don't want people to be able to traceroute your address, you can leave this field empty when generating errors or not generate an error.

PR #176, opened 2017-03-25
RFC-3: ILP Packet explanation

mDuo13:
Depends on: #174.

Further explanation of the Interledger Protocol Packet, plus some other context cleanup to IL-RFC-3.


0003-interledger-protocol/0003-interledger-protocol.md
```diff
@@ -147,61 +145,62 @@ Connectors implement the interledger protocol to forward payments between ledger
...
 
-    UInt64 ::= INTEGER (0..18446744073709551615)
+#### Life-Cycle of the ILP Packet
 
-Amount in discrete units of the receiving ledger's asset type. Note that the amount is counted in terms of the smallest indivisible unit on the receiving ledger.
+The exact use of the ILP Packet depends on the transport protocol you use. In [IPR][], the receiver generates and communicates the packet to the sender. In [PSK][], the sender generates the packet according to agreed-upon rules with a pre-shared secret.
+[IPR]: ../0011-interledger-payment-request/0011-interledger-payment-request.md
 
-#### account
+When the sender prepares a transfer to start the payment, the sender attaches the ILP Packet to the transfer, in the memo field if possible. If a ledger does not support attaching the entire ILP Packet to a transfer as a memo, users of that ledger can transmit the ILP Packet using another system, but MUST be able to correlate transfers and ILP Packets. ***Rome's note: Doesn't this also have some sort of trust implications, like the system for correlating transfers must be trusted to the same extent as the ledger because it could be used to steal money if it's malicious?***
```
emschwartz: Other parts of the protocol depend on having an authenticated communication channel between parties that are transacting. If you sent the memo over such a channel with a reference to the transfer ID, there wouldn't be a trust issue.

0003-interledger-protocol/0003-interledger-protocol.md
```diff
@@ -147,61 +145,62 @@ Connectors implement the interledger protocol to forward payments between ledger
 
 Connectors also implement the [Connector to Connector Protocol (CCP)](../0010-connector-to-connector-protocol/) to coordinate routing and other interledger control information.
 
-### Errors
+### Packet
```
emschwartz: I think we should refer to it as the ILP Payment Packet, because there are other types of ILP (related) packets. That's what @justmoon did in the ASN.1 description

0003-interledger-protocol/0003-interledger-protocol.md
```diff
@@ -147,61 +145,62 @@ Connectors implement the interledger protocol to forward payments between ledger
...
-        ( hyphen
-        | period
-        | "0".."9"
-        | "A".."Z"
-        | underscore
-        | "a".."z"
-        | tilde )
-      )
-      (SIZE (1..1023))
+When the receiver sees the incoming prepared transfer, the receiver reads the ILP Packet to confirm the details of the packet. The receiver confirms that the amount from the ILP Packet matches the amount actually delivered by the transfer. The receiver decodes the data of the ILP Packet and matches the condition to the packet. (Depending on the transport protocol, the packet's data MAY contain the condition.) The receiver MUST confirm the integrity of the ILP Packet, for example with a [hash-based message authentication code (HMAC)](https://en.wikipedia.org/wiki/Hash-based_message_authentication_code). If the receiver finds the transfer acceptable, the receiver releases the fulfillment for the transfer, which can be used to execute all the prepared transfers.
```
emschwartz: > (Depending on the transport protocol, the packet's data MAY contain the condition.)

We decided a while back that this doesn't make sense, because it's redundant and provides no additional value. Instead, it's more likely that the condition includes the MAC of the packet.

0003-interledger-protocol/0003-interledger-protocol.md
```diff
@@ -147,61 +145,62 @@ Connectors implement the interledger protocol to forward payments between ledger
...
+See below for the [ILP Error Format](#ilp-error-format) and [ILP Error Codes](#ilp-error-codes).
+
+
+
+## Specification
+
+### ILP Packet
+
+
+The ILP Packet is represented as binary data. In addition to being faster to transmit and encode/decode, the binary format of the packet gives it a canonical format so it can be hashed consistently. The formal definition of the ILP Packet is the [ASN.1 Interledger Protocol module](../asn1/InterledgerProtocol.asn). The ILP Packet consists of the following data, in order:
```
emschwartz: We probably want to mention that the specific encoding we use is canonical OER

emschwartz:
Based on @earizon's [comment here](https://github.com/interledger/rfcs/pull/175#issuecomment-289477083), it would be good to include more of an explanation of why we're using ASN.1/OER

0003-interledger-protocol/0003-interledger-protocol.md
```diff
@@ -147,61 +145,62 @@ Connectors implement the interledger protocol to forward payments between ledger
...
 
-#### amount
+The type value `1` indicates an ILP Packet. The type values `2` through `7` indicate ILQP Packets.
 
-    UInt64 ::= INTEGER (0..18446744073709551615)
+#### Life-Cycle of the ILP Packet
 
-Amount in discrete units of the receiving ledger's asset type. Note that the amount is counted in terms of the smallest indivisible unit on the receiving ledger.
+The exact use of the ILP Packet depends on the transport protocol you use. In [IPR][], the receiver generates and communicates the packet to the sender. In [PSK][], the sender generates the packet according to agreed-upon rules with a pre-shared secret.
+[IPR]: ../0011-interledger-payment-request/0011-interledger-payment-request.md
```
mDuo13: The PSK is also a link (the reference is on line 203)

mDuo13:
Continued from https://github.com/interledger/rfcs/pull/176#issuecomment-289485436 so it doesn't get lost: It sounds like ASN.1/OER is a bit of a [controversial topic](https://github.com/interledger/rfcs/pull/175#issuecomment-289477083) so we should provide more explanation for why we're using it.

We need a hashable binary format, and ANS.1/OER gives one way to do that, but we don't have as much discussion of why we picked that over other such formats (e.g. Google's Protocol Buffers, BSON, Netstrings, etc.)

emschwartz:
@mDuo13 do you want to update this or close it out?

mDuo13:
Let's just close it. The doc it was updating has changed dramatically since this was written, so it's not as useful as it once was.

PR #177, opened 2017-03-27
docs(0004,0009): Rename scale to currencyScale, deprecate currencySymbol, deprecate precision

michielbdejong:
Since we switched to integers on transfers and ledger balances, `scale` is always zero, so that's useless in ledgerInfo. However, `currencyScale` is needed now. At the same time, deprecated `currencySymbol` since symbols, like colors, and human languages are a UI thing and not a concern of the ledger plugin. IMHO, the plugin interface should only deal with machine-readable data, not with human-computer interface.

### DIFF
--- a/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
+++ b/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
@@ -132,10 +132,8 @@ Retrieve some metadata about the ledger. Plugin must be connected, otherwise the
 ```json
 {
   "prefix": "us.fed.some-bank.",
-  "precision": 10,
-  "scale": 4,
   "currencyCode": "USD",
-  "currencySymbol": "$",
+  "currencyScale": 4,
   "connectors": [ "us.fed.some-bank.chloe" ]
 }
 ```
@@ -155,7 +153,9 @@ The mapping from the ILP address to the local ledger address is dependent on the
 #### getBalance
 <code>ledgerPlugin.getBalance() ⇒ Promise.&lt;String></code>
 
-Return an integer string representing the current balance. Plugin must be connected, otherwise the promise should reject.
+Return a (base-ten) integer string (`..., '-3', '-2', '-1', '0', '1', '2', '3', ...`) representing the current balance, in the ledger's base unit. For example, on a ledger with `currencyCode` 'USD' and `currencyScale` 6,
+the base unit would be micro-dollars.
+A balance of '1230000' should then be interpreted as equivalent to 1.23 US dollars. The maximum and minimum balance are up to the ledger to determine. Plugin must be connected, otherwise the promise should reject.
 
 #### getFulfillment
 <code>ledgerPlugin.getFulfillment( transferId ) ⇒ Promise.&lt;String></code>
@@ -399,7 +399,7 @@ left undefined (but not any other false-y value) if unused.
 | `String` | [from](#from) | ILP Address of the source account |
 | `String` | [to](#to) | ILP Address of the destination account |
 | `String` | [ledger](#ledger) | ILP Address prefix of the ledger |
-| `String` | [amount](#amount) | Integer transfer amount |
+| `String` | [amount](#amount) | Integer transfer amount, in the ledger's base unit |
 | `String` | [ilp](#ilp) | Base64-encoded ILP packet |
 | `Object` | [noteToSelf](#notetoself) | Host-provided memo that should be stored with the transfer |
 | `String` | [executionCondition](#executioncondition) | Cryptographic hold condition |
@@ -459,7 +459,7 @@ ILP Address prefix of the ledger that this transfer is going through on.
 #### amount
 <code>**amount**:String</code>
 
-An integer amount, represented as a string. MUST be positive.
+An integer amount, represented as a string of base-ten digits. MUST be `>= 0` and `< 2^64`.
 
 #### ilp
 <code>**ilp**:String</code>
@@ -590,10 +590,8 @@ Metadata describing the ledger. This data is returned by the [`getInfo`](#getinf
 | Type | Name | Description |
 |:--|:--|:--|
 | `String` | [prefix](#prefix) | The plugin's ILP address prefix |
-| `Number` | [precision](#precision) | Total number of digits allowed |
-| `Number` | [scale](#scale) | Digits allowed after decimal |
-| `String` | [currencyCode](#currencycode) | ISO three-letter currency code |
-| `String` | [currencySymbol](#currencysymbol) | UTF8 currency symbol |
+| `String` | [currencyCode](#currencyCode) | [ISO 4217](https://en.wikipedia.org/wiki/ISO_4217) three-letter currency code |
+| `Number` | [currencyScale](#currencyScale) | Integer `(..., -2, -1, 0, 1, 2, ...)`, such that one of the ledger's base units equals `10^-<currencyScale> <currencyCode>` |
 | `String[]` | [connectors](#connectors) | ILP addresses of recommended connectors |
 
 ### Fields
@@ -605,25 +603,21 @@ The ledger plugin's ILP address prefix. This is used to determine whether a give
 
 The prefix may be configured, automatically detected, or hard-coded, depending on the ledger. For example, a Bitcoin ledger plugin may have the address hard-coded, while a [`five-bells-ledger`](https://github.com/interledger/five-bells-ledger) would use an API call to get the prefix.
 
-#### precision
-<code>**precision**:Number</code>
-
-The total number of digits (base 10) of precision allowed by the ledger.
-
-#### scale
-<code>**scale**:Number</code>
-
-The number of digits allowed after the decimal point.
-
 #### currencyCode
 <code>**currencyCode**:String</code>
 
-The ISO 4217 currency code (if any) used by the ledger.
+The [ISO 4217](https://en.wikipedia.org/wiki/ISO_4217) currency code (if any) used by the ledger. A custom all-caps three-letter code, not used by ISO 4217, otherwise.
+Ledger administrators who choose a custom currency code MAY request a custom currency symbol for their chosen currency code be listed by software modules that map currency codes to currency symbols,
+for instance on node package manager (npm) in the case of JavaScript.
+To translate an integer amount or balance from the ledger, the currencyCode by itself is not enough. It has to be used in combination with the currencyScale (below) to determine how many
+of the ledger's base units correspond to one currency unit.
 
-#### currencySymbol
-<code>**currencySymbol**:String</code>
+#### currencyScale
+<code>**currencyScale**:String</code>
 
-The currency symbol as one or more UTF8 characters.
+The order of magnitude to express one full currency unit in ledger's base units. For instance, if the integer values represented on the ledger are to be interpreted as
+dollar-cents (for the purpose of settling a user's account balance, for instance), then the ledger's
+currencyCode is `USD` and its currencyScale is `2`.
 
 #### connectors
 <code>**connectors**:String[]</code>
--- a/0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
+++ b/0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
@@ -80,9 +80,7 @@ Content-Type: application/json
   "minimum_destination_amount": "10",
   "ledger_info": {
     "currency_code": "USD",
-    "currency_symbol": "$",
-    "scale": 2,
-    "precision": 10
+    "currency_scale": 2
   },
   "receiver_info": {
     "name": "Bob Dylan",
@@ -123,9 +121,7 @@ The response body is a JSON object that includes basic account details necessary
 | `minimum_destination_amount` | Integer String | Minimum amount, denoted in the minimum divisible units of the ledger, the receiver will accept. This amount may be determined by the minimum transfer amount of the ledger. If the receiver expects a specific amounts to be delivered, this may be set equal to the `maximum_destination_amount` |
 | `ledger_info` | Object | Details about the destination ledger |
 | `ledger_info.currency_code` | String | Currency code to identify the receiver's currency. Currencies that have [ISO 4217](https://en.wikipedia.org/wiki/ISO_4217) codes should use those. Sender UIs SHOULD be able to render non-standard codes |
-| `ledger_info.currency_symbol` | String | Symbol for the receiver's currency intended for display in the sender's UI (e.g. `"$"` or `"shares"`). Sender UIs SHOULD be able to render non-standard symbols |
-| `ledger_info.scale` | Integer | The scale of the amounts on the destination ledger (e.g. an amount of `"1000"` with a scale of `2` translates to `10.00` units of the destination ledger's currency) |
-| `ledger_info.precision` | Integer | The maximum precision supported by the ledger |
+| `ledger_info.currency_scale` | Integer | The scale of the amounts on the destination ledger (e.g. an amount of `"1000"` with a scale of `2` translates to `10.00` units of the destination ledger's currency) |
 | `receiver_info` | Object | Arbitrary additional information about the receiver. This field has no schema and the receiver may include any fields they choose. The field names listed below are recommended merely for interoperability purposes. |
 | `receiver_info.name` | String | _(OPTIONAL)_ Full name of the individual, company or organization the receiver represents |
 | `receiver_info.image_url` | HTTPS URL | _(OPTIONAL)_ URL where the sender can get a picture representation of the receiver |
@@ -153,9 +149,8 @@ The sender uses the receiver details to create the ILP packet:
 * The `account` in the ILP packet is the `destination_account` provided by the receiver
 * The `amount` is determined by the sender:
 
-    * The sender uses the `ledger_info.scale` to determine the integer amount to include in the ILP packet (e.g. if they want the receiver to receive 10 units on their ledger, a `ledger_info.scale` value of 2 would indicate they should set the `amount` to 1000). In cases where the sender knows the integer amount the receiver should get, for example if that amount was requested by the receiver out of band, the sender does not need to use the `ledger_info.scale`
+    * The sender uses the `ledger_info.currency_scale` to determine the integer amount to include in the ILP packet (e.g. if they want the receiver to receive 10 units of `ledger_info.currency_code` on their ledger, a `ledger_info.currency_scale` value of 2 would indicate they should set the `amount` to 1000). In cases where the sender knows the integer amount the receiver should get, for example if that amount was requested by the receiver out of band, the sender does not need to use `ledger_info.currency_code` and `ledger_info.currency_scale`
     * The sender SHOULD NOT set an amount greater than the `maximum_destination_amount` or lower than the `minimum_destination_amount`, as this will be rejected by the receiver
-    * The sender SHOULD NOT set an amount with a precision greater than the `ledger_info.precision`, as this will be rejected by the destination ledger
 
 * The `data` is encoded using the [Pre-Shared Key protocol](../0016-pre-shared-key/0016-pre-shared-key.md):
 


michielbdejong:
Made corresponding changes in doc 0009 as well.

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
```diff
@@ -155,7 +154,8 @@ The mapping from the ILP address to the local ledger address is dependent on the
 #### getBalance
 <code>ledgerPlugin.getBalance() ⇒ Promise.&lt;String></code>
 
-Return an integer string representing the current balance. Plugin must be connected, otherwise the promise should reject.
+Return an integer string representing the current balance, in the ledger's base unit. For example, if the ledger's `currencyCode` is 'EUR' and its `currencyScale` is 5, then
```
sentientwaffle: Right now `plugin.getBalance()` returns an integer, why change that behavior?

emschwartz: I think this is just clarifying what the integer represents

sentientwaffle: Ah, I see.

michielbdejong: yes. I think in the switch to integers, it changed from a `String` that allows `parseFloat`, to a `String` that allows `parseInt`. At the same time, its absolute value was multiplied by `Math.pow(10, scale)`. This docs change in this PR only adds some extra clarifying text about the new situation, it doesn't change the behavior.

sharafian: We should clarify what "base unit" means, exactly, e.g. on a `USD` ledger with a scale of 6, the base unit would be micro-dollars.

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
```diff
@@ -133,9 +133,8 @@ Retrieve some metadata about the ledger. Plugin must be connected, otherwise the
 {
   "prefix": "us.fed.some-bank.",
   "precision": 10,
```
sharafian: Why are we changing scale to currencyScale, but not changing precision to currencyPrecision? I'm not so sure I agree with the renaming in any case, because the scale and precision are ledger properties, not currency properties. `currencyCode` is a description of the ledger's currency code, not a description of the ledger's currency, if that makes sense. Something like `amountScale` and `amountPrecision` might make a little more sense, but I think might be more verbose than we need.

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
```diff
@@ -591,9 +591,8 @@ Metadata describing the ledger. This data is returned by the [`getInfo`](#getinf
 |:--|:--|:--|
 | `String` | [prefix](#prefix) | The plugin's ILP address prefix |
 | `Number` | [precision](#precision) | Total number of digits allowed |
-| `Number` | [scale](#scale) | Digits allowed after decimal |
-| `String` | [currencyCode](#currencycode) | ISO three-letter currency code |
-| `String` | [currencySymbol](#currencysymbol) | UTF8 currency symbol |
+| `String` | [currencyCode](#currencyCode) | ISO three-letter currency code |
+| `Number` | [currencyScale](#currencyScale) | Integer, such that `1 <currencyCode>` equals `10^<currencyScale>` ledger units
```
sharafian: `ledger units` -> `of the ledger's base units`

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
michielbdejong: after discussing, @sharafian agreed that the rename to `currencyScale` is defendable.

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
michielbdejong: We could mention explicitly that `currencyScale` can be negative, zero, or positive. Do we also want to support non-integer currency scales, actually? Then the ledger administrator can make the payout of their settlement method(s) fluctuate with the market.

sharafian: I think non-integer scales would be too confusing. But we should definitely support negative scales.

sharafian:
Should https://github.com/interledger/rfcs/issues/179 get included in this PR too?

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
```diff
@@ -590,10 +590,8 @@ Metadata describing the ledger. This data is returned by the [`getInfo`](#getinf
 | Type | Name | Description |
 |:--|:--|:--|
 | `String` | [prefix](#prefix) | The plugin's ILP address prefix |
-| `Number` | [precision](#precision) | Total number of digits allowed |
-| `Number` | [scale](#scale) | Digits allowed after decimal |
-| `String` | [currencyCode](#currencycode) | ISO three-letter currency code |
-| `String` | [currencySymbol](#currencysymbol) | UTF8 currency symbol |
+| `String` | [currencyCode](#currencyCode) | ISO three-letter currency code |
+| `Number` | [currencyScale](#currencyScale) | Integer `(..., -2, -1, 0, 1, 2, ...)`, such that `1 <currencyCode>` equals `10^<currencyScale>` of the ledger's base units
```
sharafian: Maybe swap this around so `1 base unit equals 10^(-1 * <currencyScale>) <currencyCode>`. In the current phrasing, a negative scale would imply that a fractional amount of base units would be used to represent one <currencyCode>, which isn't possible.

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
```diff
@@ -590,10 +590,8 @@ Metadata describing the ledger. This data is returned by the [`getInfo`](#getinf
 | Type | Name | Description |
 |:--|:--|:--|
 | `String` | [prefix](#prefix) | The plugin's ILP address prefix |
-| `Number` | [precision](#precision) | Total number of digits allowed |
-| `Number` | [scale](#scale) | Digits allowed after decimal |
-| `String` | [currencyCode](#currencycode) | ISO three-letter currency code |
-| `String` | [currencySymbol](#currencysymbol) | UTF8 currency symbol |
+| `String` | [currencyCode](#currencyCode) | ISO three-letter currency code |
```
sharafian: add which ISO number here

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
```diff
@@ -605,25 +603,17 @@ The ledger plugin's ILP address prefix. This is used to determine whether a give
...
-#### scale
-<code>**scale**:Number</code>
-
-The number of digits allowed after the decimal point.
-
 #### currencyCode
 <code>**currencyCode**:String</code>
 
-The ISO 4217 currency code (if any) used by the ledger.
+The ISO 4217 currency code (if any) used by the ledger. A custom all-caps three-letter code, not used by ISO 4217, otherwise. Ledger administrators who choose a custom currency code MAY request their chosen currency code be listed by software modules that map currency codes to currency symbols, for instance on node package manager (npm) in the case of JavaScript. The currencyCode is only meaningful in combination with the currencyScale (below).
```
sharafian: Should have a link for ISO 4217

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
```diff
@@ -605,25 +603,17 @@ The ledger plugin's ILP address prefix. This is used to determine whether a give
...
 
-#### currencySymbol
-<code>**currencySymbol**:String</code>
+#### currencyScale
+<code>**currencyScale**:String</code>
 
-The currency symbol as one or more UTF8 characters.
+The order of magnitude to express one full currency unit in ledger units. For instance, if the integer values represented on the ledger are to be interpreted as
```
sharafian: ledger units -> ledger's base units

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
```diff
@@ -605,25 +603,17 @@ The ledger plugin's ILP address prefix. This is used to determine whether a give
 
 The prefix may be configured, automatically detected, or hard-coded, depending on the ledger. For example, a Bitcoin ledger plugin may have the address hard-coded, while a [`five-bells-ledger`](https://github.com/interledger/five-bells-ledger) would use an API call to get the prefix.
 
-#### precision
```
sharafian: If the per-ledger precision is being taken out, we should specify what the expected behavior for precision is.

justmoon:
Makes sense. One can that we've kicked down the road several times already: Is SPSP only for currencies? Only ISO currencies? All currencies? How about company-issued currencies? How about loyalty points? Stocks?

I think we can - with reasonable effort - generalize SPSP to support any type of asset, mostly by using more neutral field names.

@michielbdejong @emschwartz Thoughts?

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
adrianhopebailie: Limited to ISO currencies?

adrianhopebailie:
> I think we can - with reasonable effort - generalize SPSP to support any type of asset, mostly by using more neutral field names.

The W3C Payment Request spec solves this by defining a `currencySystem` param which defaults to `urn:iso:std:iso:4217`. See https://www.w3.org/TR/payment-request/#paymentcurrencyamount-dictionary

This follows a lot of discussion about separation of data and UI concerns (i.e. solving the same problem)

michielbdejong:
I think ledger_info in spsp end-point replies has two roles:

### 1) let the sender choose a destinationAmount to make the receiver happy
The receiver is announcing, for instance, "if you send me 1000 units on that ledger, that will make me 10 USD happy". It ties in with the real-world debt that is settled or created by the interledger payment. For this, it's only necessary that people who want to pay you know what you mean. So using a custom currency (e.g. "cups of coffee") is totally fine for that. Also, the spec says the sender might not need to use the ledger info, if they already determined an (integer) amount in ledger units out-of-band.

### 2) tell the sender roughly what sourceAmount to expect
This works in combination with a publicly quoted exchange rate to the source ledger's currency. The sender and the receiver cooperate here, so that they are not paying an extortionate transaction fee. For compatibility, it makes sense to restrict this to currencies whose exchange rates are easy to determine.

If you say your ledger is in cups of coffee, the sender has no easy way of distinguishing a 0.1% connector fee from a 0.5% connector fee. The difference between the fees is a factor 5, but the difference in sourceAmount is only a factor 0.004. This difficulty comes from the fact that the sender has no way to see what the spread is (the forward fee + the backward fee is how much they would lose if the receiver sent all the money back to them in a second transfer). Connector.land could play a role in publishing such data, but for now it's easiest to peg each ledger to a publicly quoted currency, and avoid announcing a custom currency or not announcing a currency at all.

If we wouldn't have ledger info, then interledger would still work, but it would be harder for senders to know when a sourceAmount too high, and thus easier for connectors to charge extortionate fees.

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
michielbdejong: Addressed now in the description of sendTransfer `amount` parameter.

emschwartz:
> I think we can - with reasonable effort - generalize SPSP to support any type of asset, mostly by using more neutral field names.

How about just replacing the word "currency" with "asset"?

adrianhopebailie:
Maybe the **Simple** Payments Setup Protocol should stick to ISO 4217 currencies for now?

The ledger plugin interface is harder to pin down especially since we already have plugins for non-ISO4217 currencies.

emschwartz:
Why? I don't think you'd do anything differently except choose a slightly different word

adrianhopebailie:
> Why? I don't think you'd do anything differently except choose a slightly different word

What if two ledgers have the same asset and use different codes to identify it? e.g. One ledger adaptor for Bitcoin reports XBT and another reports BTC. The connector has no idea that these are the same asset.

emschwartz:
The information in SPSP is only for the sender. The asset/currency code is just for UI purposes (though it's also not safe to depend on it). Connectors only see the ILP address and amount determined by the sender.

michielbdejong:
@emschwartz 
> How about just replacing the word "currency" with "asset"?

That would work less well than the current situation, because of https://github.com/interledger/rfcs/pull/177#issuecomment-290078358

emschwartz:
I would very much agree with @justmoon that we should not restrict Interledger or the protocols we build on top of it to the current set of well-known currencies. It's fine to have extra checks when you are dealing with common currencies, but we should not build with that assumption.

> If all ledgers just say they are denominated in 'assets', rather than referring to well-known currency codes, it would become easier for connectors to charge extortionate transaction fees.

That's a fundamental issue with assets that are not commonly known, but that's just a downside of using such assets. We should not restrict people on account of that.

In an interledger world the way you determine the value of some unknown asset is by quoting. The way the system deals with connectors that charge extortionate fees is with competition. If you only have one route, of course they can charge you too much for it (like we see today). And if you have a small and very illiquid asset it's likely that the rates won't be very good. But if you choose to use such an asset that's your concern (there may be other benefits we might not be taking into account).

> If you say your ledger is in cups of coffee, the sender has no easy way of distinguishing a 0.1% connector fee from a 0.5% connector fee.

Interledger already has this design choice baked into it because the protocol does not offer a way to differentiate between the exchange rate and fees added on top of it. Ultimately it doesn't really matter to the sender because if you only have one route, you have the connector's rate or nothing. If you have multiple routes you can ask each for a quote and you just go with the best one. It doesn't matter what part of the cost is the "fee" because all you really care about is how much that translates into in your currency.

justmoon:
Let's talk concrete options.

##### Option 1: ISO 4217 currencies only

I'd say this one is pretty much a non-starter because we care a lot about Bitcoin, Ethereum, Ripple and other non-ISO currencies.

##### Option 2: ISO 4217-esque currency codes

Includes ISO 4217 currencies, but adds BTC (Bitcoin), XRP (Ripple), ETH (Ethereum), ETC (Ethereum Classic), maybe also things like XAU (gold), XAG (silver) and others.

Example LedgerInfo excerpt:

``` js
{
  "prefix": "us.fed.some-bank.",
  "currencyCode": "USD",
  "currencyScale": 4,
  "connectors": [ "us.fed.some-bank.chloe" ]
}
```

##### Option 3: Any asset type

We define a set of properties that the receiver has to provide to describe the asset they are using. We also make sure that that ISO 4217 currencies at least are standardized, so that the sender can recognize them if they know them.

Example LedgerInfo excerpt:

``` js
{
  "prefix": "us.fed.some-bank.",
  "assetSystem": "urn:iso:std:iso:4217",
  "assetCode": "USD",
  "assetScale": 4,
  "connectors": [ "us.fed.some-bank.chloe" ]
}
```

----

It seems like Option 3 addresses @michielbdejong's [use cases](https://github.com/interledger/rfcs/pull/177#issuecomment-290078358) when the sender recognizes the `assetSystem` and `assetCode`, but still allows people to use SPSP for other use cases as well.

And it provides a clear avenue for anyone who wants to standardize new types of assets for both sides to understand. E.g. if I want to add support for stocks, I just need to make an IL-RFC that says stocks use `assetSystem = '...'` and define how the codes map to the actual stocks and then any sender that wants to can render it accordingly.

michielbdejong:
@emschwartz 
> That's a fundamental issue with assets that are not commonly known
> In an interledger world the way you determine the value of some unknown asset is by quoting. 

I disagree. That vision sort-of works, but I think we can do better. Pegging your not-commonly-known asset to a well-known currency using ledger info makes it easier for senders to bargain with connectors. If you want to give people more freedom in how they peg their ledger, I would say, support float currencyScale, not support any asset type.

I vote for @justmoon's option 2, with the side-note that the important feature for inclusion should be  that a unit-of-value used as a `currencyCode` **should be publicly traded**, and a second side-note that if you want to just tell senders 'the ledger unit of my ledger is just "units".', then you can do that by leaving out ledgerInfo altogether. The sender's UI can then display the destinationAmount in just "destination ledger units".

sharafian:
I don't think we should _ever_ do anything with the destination amount or asset type, from the sender side. As a sender, it never matters to me whether the connector is charging a high fee or the receiver is charging more.  All that should matter is that I'm paying out some amount and getting some goods/services in return. If the exchange isn't worth it to me, I won't make it. Of course a receiver will want to be on a very liquid ledger, because more of the money being sent to it will actually end up in its account.

The asset type should really just be used as a shortcut for displaying amounts. It's not necessary at all, because a sophisticated connector would probably have rates by ledger instead of by asset anyways. Mainly it just makes it easier to write plugin-agnostic applications that want to display amounts in a nice way.

I think option 3 is the most versatile, but might be overkill. Why not just have "currencyCode" (or "assetCode") and use it for display purposes, and then have connectors enter the asset of their ledgers manually (as they already do). The connector backend can decide what format it wants the asset name in.

michielbdejong:
After reading http://bitcoin.stackexchange.com/questions/35543/nxt-what-is-the-difference-between-an-asset-and-a-currency I'm not so sure if I like the rename from 'currencyCode' to 'assetCode'.

I still think currencyCode and currencyScale play a very important role if the sender and the receiver team up against the connector. For instance, if you and I are friends, and I want to send money from my European bank account to your US bank account, I want to know how much transaction fee "we" are paying.

However, I can also see the point which several people made, where SPSP is used to buy for instance stocks in a company, where announcing three-letter currency code makes no sense, and "the market decides" what the exchange rate is. As a sender, I need to use other ways to judge if the best source amount is a good deal or not.

So I would propose a variation on option 3, which balances these two ways of using SPSP:

```js
{
  "prefix": "us.fed.some-bank.",
  "assetSystem": "urn:iso:std:iso:4217",
  "currencyCode": "USD",
  "currencyScale": 4,
  "connectors": [ "us.fed.some-bank.chloe" ]
}
```

michielbdejong:
Moved the 'any asset' discussion to #182.

@emschwartz is it OK if I merge the current changeset and leave #182 open as a follow-up? We can continue the discussion there, but splitting it up this way, would allow me to get on with the [changes in other repos](https://github.com/interledgerjs/ilp-kit/pull/278#issuecomment-290405357)

To recap, the scope of this PR would then stay restricted to:

* rename `scale` to `currencyScale`/`currency_scale`
* deprecate `currency_symbol`/`currencySymbol`
* deprecate `precision`

emschwartz:
Works for me. I don't feel that strongly about the "currency" vs "asset" field names actually. The only thing I do feel strongly about is that that field must just be treated as a string and senders must be able to handle strings they don't understand.

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
```diff
@@ -155,7 +153,9 @@ The mapping from the ILP address to the local ledger address is dependent on the
 #### getBalance
 <code>ledgerPlugin.getBalance() ⇒ Promise.&lt;String></code>
 
-Return an integer string representing the current balance. Plugin must be connected, otherwise the promise should reject.
+Return an integer decimal string (`..., '-3', '-2', '-1', '0', '1', '2', '3', ...`) representing the current balance, in the ledger's base unit. For example, on a ledger with `currencyCode` 'USD' and `currencyScale` 6,
```
emschwartz: "integer decimal" doesn't make sense, just integer

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
```diff
@@ -605,25 +603,17 @@ The ledger plugin's ILP address prefix. This is used to determine whether a give
...
-#### scale
-<code>**scale**:Number</code>
-
-The number of digits allowed after the decimal point.
-
 #### currencyCode
 <code>**currencyCode**:String</code>
 
-The ISO 4217 currency code (if any) used by the ledger.
+The [ISO 4217](https://en.wikipedia.org/wiki/ISO_4217) currency code (if any) used by the ledger. A custom all-caps three-letter code, not used by ISO 4217, otherwise. Ledger administrators who choose a custom currency code MAY request their chosen currency code be listed by software modules that map currency codes to currency symbols, for instance on node package manager (npm) in the case of JavaScript. The currencyCode is only meaningful in combination with the currencyScale (below).
```
emschwartz: Might be worth explaining why the currencyCode is only meaningful in combination with the currencyScale

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
michielbdejong: ok, will change it to "(base-ten)" to avoid confusion with the "digit behind the decimal point" meaning of the word "decimal"

michielbdejong:
@emschwartz @justmoon can this be merged?

justmoon:
I'd like to have some discussion about deprecating `currency_symbol` before we do so, so if you're happy to take that out for now, I'm ok with merging the other changes.

Basically, `currency_symbol` allows the sender to display a currency symbol to the user, even if they don't understand the currency code of the recipient. I completely agree with your theoretical argument that the symbol is a UI concern and I'd like to take it out for that reason. But unless/until we come up with a cleaner way to create the same user-facing benefit (senders can see a useful currency symbol even for assets they don't understand) I'm against taking out this feature.

AFAIK ILP Kit is currently using the `currency_symbol`, so taking it out will break that, no?

michielbdejong:
As far as I know, the only node with a custom currency code is mmk.interledger.network, and they are not currently using the `currency_symbol` feature:
screen shot 2017-04-07 at 07 54 38

I switched ilp-kit to [using https://www.npmjs.com/package/currency-symbol-map in the API layer](https://github.com/interledgerjs/ilp-kit/blob/19439418043dfe0480ebd74d657e142d976930a1/api/test/utils.test.js#L167), so the UI layer of ilp-kit doesn't notice the change, but if you want I can revert that.

michielbdejong:
@justmoon would it be ok with you if we just display `currencyCode.toUpperCase()` for such ledger-specific currencies, as well as for XAU, BTC, XRP, etc? There have been proposals for a symbol for Bitcoin (for instance, Ƀ), but I can't think of a symbol for XAU or XRP.

justmoon:
If the feature is not being used I guess I don't feel strongly about breaking it. So let's take it out for now and just plan to discuss it as part of #182.

michielbdejong:
Thanks!

Issue #180, opened 2017-03-29
establish names for hops etc.


michielbdejong:
We have a lot of ambiguity in how we use some ILP-related words. For instance, if you have one connector in a payment, is that a one-hop or a two-hop payment? Is the connector also a receiver and a sender? What is the difference between a payment and a transfer? I would propose we add a sort of 'glossary' section to one of the RFCs. After discussion with @sharafian on the way to lunch today, my suggestion for terminology would be as follows:

Suppose a human being named Alice wants to send money to a human being named Bob. This makes Alice the `sender` and makes Bob the `receiver` of the `payment`. If Alice and Bob both have an `account` on the same `ledger`, they can do a `one-hop` payment, involving just one `transfer`, on that one ledger. The `amount` of the transfer is measured in the ledger's `base unit`, which can for instance be micro-dollars.

If they don't have accounts on a same ledger, but they do have accounts on different ledgers, they can do a `multi-hop` payment, consisting of multiple `transfers`. The first transfer in the chain is the `source` transfer, the last one is the `destination` transfer, and if there are more than two, then the inbetween ones are `intermediate` transfers. The parties in the middle are called `connectors`.

Alice now is the `source sender`. The first connector's account on the source ledger is the `source receiver`. The last connector's account on the destination ledger is the `destination sender`, and Bob is the `destination receiver`.

sappenin:
Also, there's still some ambiguity in my mind between the person sending a payment (i.e., an account holder, which I think you've defined as a `sender`) and the institution that payment is coming from (e.g., the sender's bank).

In the banking world it seems like the terms `originator` and `beneficiary` are commonly used to refer to institutions/organizations that actually own the ledgers that a `sender` and `receiver` are utilizing.

So, perhaps the terms `originator` and `beneficiary` are used in concert with the word `institution` to represent entities that own a ledger, and the terms `sender` and `receiver` are used to denote the account holders?  Not sure if there are more commonly used words that would be better here.

michielbdejong:
> institution that payment is coming from (e.g., the sender's bank).

that would be the 'source ledger', right? but if you think it's better, we can use 'originator' instead of 'source', and then use 'beneficiary' instead of 'destination'. Only issue with that is that SPSP and ILQP already use the terms source/destination now.

michielbdejong:
totally agree that we should add our various TLAs to this glossary.

so, should it have its own RFC number?

sappenin:
[...]

PR #181, opened 2017-03-29
Overhaul architecture overview, separate ledger requirements

emschwartz:
RFC 1 missed or did not emphasize a lot of key details of the interledger architecture, while repeating certain content and including too much detail about the ledger requirement. this moves the ledger requirements to a separate RFC and restructures the architecture overview to highlight the key aspects of interledger that newcomers should understand

Based on feedback from @professorhantzen


0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -1,252 +1,135 @@
...
+This document outlines the Interledger architecture and the functions of each of the layers of the stack.
 
-This document is intended to provide an overview over the Interledger architecture and protocol suite.
+The Interledger architecture is heavily inspired by the Internet architecture described in [RFC 1122](https://tools.ietf.org/html/rfc1122), [RFC 1123](https://tools.ietf.org/html/rfc1123) and [RFC 1009](https://tools.ietf.org/html/rfc1009).
 
-## The Interledger Architecture
+## The Interledger Model
 
-All electronic transfers have to be recorded in stateful systems. Otherwise the same instance of an asset could be sent to two different destinations, essentially duplicating the asset. This is also known as a double-spend. We call these stateful systems *ledgers*.
+[diagram]
```
emschwartz: We need a diagram here that ideally shows a sender, 2 connectors, 3 ledgers, and a receiver

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -1,252 +1,135 @@
...
-    The recipient generates the condition during the quoting phase, and the sender cannot know the fulfillment until the recipient uses it to execute the payment. This is useful for building non-repudiable application layer protocols.
+## Interledger Protocol Suite
 
-#### Interledger Layer
+The Interledger stack is separated into four layers:
 
-All Interledger transport protocols use the [Interledger Protocol (ILP)](#interledger-protocol-ilp) to communicate with connectors about transfer requests. This may include requesting a quote or requesting a transfer on another ledger.
-
-The Interledger layer defines a standard way to refer to ledgers, accounts and amounts. This is used in routing as well as to try and make quotes comparable.
+![Interledger architecture layers](../shared/graphs/interledger-model.svg)
```
emschwartz: The SVG doesn't render because of https://github.com/github/markup/issues/556, we should either find a workaround or render a PNG file as well from the graph. The only issue with doing that is the node libraries that do that include PhantomJS, which is a pretty heavy dependency just for rendering 1-2 PNG files.

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -1,252 +1,135 @@
 # Interledger Architecture
 
-This document outlines the Interledger architecture and explains how the different layers relate to each other. The [Interledger Protocol (ILP)](#interledger-protocol-ilp) is one layer in the Interledger architecture.
+The Interledger protocol suite provides for secure payments across multiple assets on different ledgers.
```
sappenin: I like the usage of these grouping terms like `Interledger Architecture` and `Interledger Protocol Suite`, but I'm not sure the document clearly differentiates between the two terms.  Are these meant to be synonymous, or is “Interledger Protocol Suite” a component of the “Interledger Architecture”, which actually has more to it than just the protocol suite?

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -1,252 +1,135 @@
...
 
-The Interledger architecture is heavily inspired by the Internet architecture described in [RFC 1122](https://tools.ietf.org/html/rfc1122), [RFC 1123](https://tools.ietf.org/html/rfc1123) and [RFC 1009](https://tools.ietf.org/html/rfc1009).
+The [Interledger Protocol (ILP)](#interledger-layer) is the core of the Interledger protocol suite.
```
sappenin: I still find our colloquial usage of `ILP` to be unclear.  Let's assume there's some proper name for the collection of all Interledger protocols (the suite).  One of those protocols (the thing in RFC-3) is called “ILP”, or “Interledger Protocol”.  If you put it all together, it reduces to the following: there’s a collection of Interledger protocols, one of which is the "Interledger Protocol.”

Wouldn’t it make more sense to start calling the thing defined by RFC-3 something else, like `ILPC`, which expanded would read `Interledger Protocol Core` or just `Interledger Core Protocol`?

(I'm open to the argument that this naming "ship" has already sailed, but figured I'd give it another try)

professorhantzen: As an outsider coming to this - I agree.  "ILP" seems to be the widely used term to reference the entire suite/stack. It's certainly the term used by everyone I know outside the dev circle who discusses, well... ILP.  The current situation that one layer of the stack also shares the same name as the whole stack is confusing to newcomers and a hinderance to unambiguous discussion.  The ILPC distinction suggested here is a good solution, I would take it further and drop the "protocol".  We know it's a protocol already given the suite would be/is referred to as the Interledger Protocol, so we could simply refer to that layer as Interledger Core.

So, the Interledger Protocol (ILP) is the entire stack of protocols, one layer of which is Interledger Core, or ILC.

emschwartz:
> I still find our colloquial usage of ILP to be unclear. Let's assume there's some proper name for the collection of all Interledger protocols (the suite). One of those protocols (the thing in RFC-3) is called “ILP”, or “Interledger Protocol”. If you put it all together, it reduces to the following: there’s a collection of Interledger protocols, one of which is the "Interledger Protocol.”

@sappenin @professorhantzen I added a note that tries to clarify the colloquial use of the term ILP. Does that help?

Some of that confusion comes from how closely the architecture (and terminology) mirrors that of the Internet. The Internet Protocol is the part that defines the IP address and packet format, but people often describe the whole stack (and even the Web) as the Internet or using the Internet Protocol / IP.

Technically it is correct to call only the part that specifies the address and packet format and the flow the "Interledger Protocol" and the stack is the "Interledger stack" or "Interledger protocol suite", but I wouldn't expect that distinction to make it into normal conversation.

Another way to look at it is: if you start taking away layers of the stack and saying "I don't want to use those", at what point is what you've got no longer using the _Interledger Protocol_? I'd say it's when you're no longer using what's defined in IL-RFC 3.

sappenin:
> Another way to look at it is: if you start taking away layers of the stack and saying "I don't want to use those", at what point is what you've got no longer using the Interledger Protocol? I'd say it's when you're no longer using what's defined in IL-RFC 3.

@emschwartz That's a helpful distinction.  Perhaps consider adding more of your rationale to the doc, like some adaptation of your paragraph above?  Might be helpful for future readers to know a bit more about why the colloquialism exists.

emschwartz:
> Perhaps consider adding more of your rationale to the doc, like some adaptation of your paragraph above? Might be helpful for future readers to know a bit more about why the colloquialism exists.

I tried adding another sentence to that intro with that explanation but it felt like too much explanation, given that the intended audience presumably doesn't understand Interledger very well yet.

sappenin:
That's a fair point - perhaps that "more explanation" might make sense in an FAQ or maybe in the glossary that @michielbdejong is proposing in #180.

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -1,252 +1,135 @@
...
 
-### Interledger Protocol Suite
+"The Interledger" is a conceptual network made up of independent and diverse ledgers linked by connectors. Each account "on the Interledger" is part of a particular ledger, but they may transact with others by sending interledger payments through different ledgers and connectors. Like "the Internet", the Interledger is not a single network but is comprised of multiple interconnected networks.
 
-The Interledger architecture is separated into four layers:
+## Interledger Security
 
-![Interledger architecture layers](../shared/graphs/interledger-model.svg)
+**Interledger uses *conditional transfers* to secure payments across multiple hops and even through untrusted connectors.**
+Senders are guaranteed proof that the receiver got the payment or their money back.
```
mDuo13: Without a blank line between, this is interpreted by most Markdown parsers as a hard-wrapped continuation of the previous sentence's paragraph. You should probably set it off with a separate line, and maybe even make it into bullet points.

Also, we can strengthen this sentence:

> Senders are guaranteed proof that the receiver got the payment or their money back.

Suggested rephrase:

> Senders are guaranteed cryptographic proof that the receiver got the payment or their money back, no matter how many ledgers and connectors are in between.

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -1,252 +1,135 @@
...
 
+<span class="show alert alert-info">**Hint:** Conditional transfers or *authorization holds* are the financial equivalent of a [two-phase commit](http://foldoc.org/two-phase%20commit).</span>
 
-#### Application Layer
+### Conditional Transfers
 
-The application layer is the top layer of the Interledger protocol suite. Protocols on this layer are responsible for negotiating the key properties of a payment:
+Each local transfer is first *prepared* and then either *executed* or *rejected*.
+When a transfer is prepared, the ledger puts the funds of the source account on hold with a *cryptographic condition* and *timeout*.
+If the condition is fulfilled before the timeout, the transfer is executed and the funds are transferred.
```
mDuo13: Again, this should either be bullet points (numbered list?), a separate paragraph with a blank line between it and the previous one, or actually a continuation of the previous line.

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -1,252 +1,135 @@
...
+If the timeout is reached, the transfer expires and the ledger returns the funds to the source account automatically.
 
-* Source Account
-* Destination Account
-* Destination Amount
+Inspired by the [Lightning Network](http://lightning.network), Interledger uses the digest of the [SHA-256](https://en.wikipedia.org/wiki/SHA-2) hash function as the condition for transfers. The fulfillment is a valid 32-byte preimage for the hash specified when the transfer was prepared. Ledgers are responsible for validating fulfillments. [Transport Layer](#transport-layer) protocols are used by the sender and receiverto generate the condition for a particular payment.
```
mDuo13: typo: "receiverto"

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -1,252 +1,135 @@
...
-
-### Example Protocols
-
-#### Five Bells Ledger Protocol (5BLP)
-
-Five Bells Ledger Protocol (5BLP) is a RESTful, JSON-based protocol that was developed specifically to provide the minimum functionality required for full Interledger support.
+See [IL-RFC 17](../0017-ledger-requirements/0017-ledger-requirements.md) for a full description of the ledger layer requirements.
 
-A reference implementation of a ledger using 5BLP can be found [here](https://github.com/interledger/five-bells-ledger).
+Most implementations of Interledger use a plugin architecture to abstract the differences between different ledgers. For an example of this, see [IL-RFC 4](../0004-ledger-plugin-interface/0004-ledger-plugin-interface), which defines the interface for the Javascript implementation.
```
mDuo13: Broken link: `../0004-ledger-plugin-interface/0004-ledger-plugin-interface` is missing the `.md` on the end

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -1,252 +1,135 @@
...
+The Interledger layer is responsible for facilitating payments across multiple ledgers. It is comprised of two key components that are used together: the Interledger Protocol (ILP) and the Interledger Quoting Protocol (ILQP).
 
-For example, Bitcoin supports SHA-256 hashlocked escrow transfers which means it can participate in ILP Interledger transactions. Bitcoin's [BIP-65](https://github.com/bitcoin/bips/blob/master/bip-0065.mediawiki) proposal provided the timeouts required for Basic level support.
+The [Interledger Protocol (ILP)](../0003-interledger-protocol/0003-interledger-protocol.md) is the core of the Interledger stack and defines a standard address packet format that instruct connectors where to send a payment.
 
-#### Legacy Protocols (e.g. ACH, ISO 20022)
+[Interledger Addresses](../0015-interledger-addresses/0015-interledger-addresses.md) provide a ledger-agnostic way to address ledgers and accounts. Interledger addresses are dot-separated strings that contain prefixes to group ledgers. An example address might look like:
```
mDuo13: Broken link: IL-RFC 15 is named "ILP Addresses", not "Interledger Addresses"

(I think the name ILP is technically accurate here, because the addressing is used within the core Interledger Protocol itself?)

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -1,252 +1,135 @@
...
 
-#### Blockchain Protocols (e.g. Bitcoin)
+### Interledger Layer
 
-Blockchains are distributed, peer-to-peer systems that provide consensus over a single shared state. Any blockchain that supports escrowed funds transfers is in principle capable of acting as a ledger connected to the Interledger.
+The Interledger layer is responsible for facilitating payments across multiple ledgers. It is comprised of two key components that are used together: the Interledger Protocol (ILP) and the Interledger Quoting Protocol (ILQP).
 
-For example, Bitcoin supports SHA-256 hashlocked escrow transfers which means it can participate in ILP Interledger transactions. Bitcoin's [BIP-65](https://github.com/bitcoin/bips/blob/master/bip-0065.mediawiki) proposal provided the timeouts required for Basic level support.
+The [Interledger Protocol (ILP)](../0003-interledger-protocol/0003-interledger-protocol.md) is the core of the Interledger stack and defines a standard address packet format that instruct connectors where to send a payment.
```
mDuo13: Instead of "standard address packet format" I would just say "packet format"

mDuo13:
Do you think it's useful to have an "extended" version of the protocol diagram? For example:

extended interledger protocol suite with N connectors and ledgers

0001-interledger-architecture/0001-interledger-architecture.md
emschwartz: I actually intended for that to be a single paragraph

0001-interledger-architecture/0001-interledger-architecture.md
emschwartz: That phrase was meant to read "standard address and packet format"

emschwartz:
I'm going to merge this PR now. If anyone has more comments we can add further changes to the docs later. cc @justmoon

PR #183, opened 2017-03-31
Add RFC 18: Connector Risk Mitigations

emschwartz:
depends on https://github.com/interledger/rfcs/pull/181


0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -71,11 +71,14 @@ For more details on the flow, see the [Interledger Protocol specification](../00
...
 
-### Error Recovery and Connector Risk
+### Connector Risk and Mitigation
 
-An ILP execution chain can be interrupted if a connector fails to deliver the condition fulfillment. In this case, the sender thinks the transaction failed and retries it using the same condition as the original transaction. If the retry goes through the same connector that failed the first time, that connector can claim the incoming funds without sending a second outgoing transfer.
+Interledger protects senders from risk but connectors accept some risk in exchange for the revenue they generate from facilitating payments. In the Interledger payment flow, connectors' outgoing transfers are executed before their incoming transfers. Once each connector is notified that the outgoing transfer has been executed, they have a window of time to deliver the fulfillment and execute the incoming transfer. Connectors that fail to deliver the fulfillment in time may lose money.
```
michielbdejong: Interledger protects senders from losing money, but a sender who pick connectors at random still runs risk to have money tied up without result. the sender is also exposed to the ledger that stores their money running off, lying about the sender's remaining balance, or going bankrupt

michielbdejong: For Interledger payments, the source ledger has the option of protecting the sender, because the source ledger can detect if the first connector broke their promise. But the actual "refund guarantee" itself is given by the source ledger, not by the protocol. Right?

michielbdejong: nit: 'Once each connector' -> 'Once a connector in the chain' (both are correct, but I think 'a' makes it clearer that this is staggered)

emschwartz: It's kind of both. The protocol stipulates that ledgers should provide conditional transfers. If you use it as intended, the ledger provides part of the guarantee and the protocol's payment flow provides the rest of the guarantee that the receiver gets the funds or the sender gets their money back.

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -71,11 +71,14 @@ For more details on the flow, see the [Interledger Protocol specification](../00
...
+
+If some transfers in an Interledger payment are executed and others expire, the receiver will think the payment was completed but the sender will think the whole payment failed. Senders may retry payments that expire and they should use the same condition as the original payment. If the second attempt takes the same route, the connector that failed the first time can complete the payment by submitting the fulfillment without sending another outgoing transfer. Connectors should prefer senders (or aggregators of senders such as wallet services) and connectors that retry payments through the same route, because they increase the likelihood of completing such payments. Connectors may incentivize this behavior by offering better rates to parties known to retry failed payments through the same route.
```
michielbdejong: nit: 'will think' -> 'may think' -> because they may also think correctly ;) especially if they communicate with each other out-of-band

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -71,11 +71,14 @@ For more details on the flow, see the [Interledger Protocol specification](../00
...
+
+If some transfers in an Interledger payment are executed and others expire, the receiver will think the payment was completed but the sender may think the whole payment failed. Senders may retry payments that expire and they should use the same condition as the original payment. If the second attempt takes the same route, the connector that failed the first time can complete the payment by submitting the fulfillment without sending another outgoing transfer. Connectors should prefer senders (or aggregators of senders such as wallet services) and connectors that retry payments through the same route, because they increase the likelihood of completing such payments. Connectors may incentivize this behavior by offering better rates to parties known to retry failed payments through the same route.
```
michielbdejong: add a sentence explaining that the sender could also just ask the receiver out-of-band whether the money was delivered. in itself, buying this Boolean piece of information from the connector may not always be worth it for the sender, unless the transfer amount was smaller than the incentive, and out-of-band communication with the receiver is hard.

emschwartz: I think that's overkill for the architecture overview. The point of this paragraph is to say that retries are possible, helpful, and may be incentivized. In any case, connectors should use all of the mitigations included in the other RFC to avoid having this happen. In some ways, why the sender doesn't retry is irrelevant because the connector loses money anyway (whether it was an attack, the sender heard the receiver got the funds out of band, or something else)

0001-interledger-architecture/0001-interledger-architecture.md
michielbdejong: still, I think it's clear to say the refund guarantee is ultimately given by the ledger, and there are no mechanisms for claiming your refund if the ledger doesn't want to to give it to you. there is no way to 'let the network be the judge'. maybe under civil law you could claim your money back from the ledger, depending on their ToS? but then it's the law that gives the guarantee, not the Interledger network or the Interledger protocol.

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
```diff
@@ -0,0 +1,40 @@
...
+
+**Monitoring is a must! Even connectors that implement all of these strategies should monitor their transaction patterns and use warnings or kill switches to avoid losing money in the case of an unexpected attack.**
+
+## Fulfillment Failure
+
+The main risk connectors face in Interledger is being unable to fulfill the incoming transfer once their outgoing transfer has been executed.
+
+Possible mitigations include:
+
+* **Connecting Reliable Ledgers** - connectors should choose the ledgers they connect in part based on the reliability and speed of fulfillment notifications and the time the ledgers require to process fulfillments once they are submitted
```
michielbdejong: interesting point, as a connector, you depend on not just your incoming ledger but also your outgoing ledger.

0001-interledger-architecture/0001-interledger-architecture.md
emschwartz: I ended up taking out the mention of the sender from this section, because this is really just about the connectors

michielbdejong:
Great doc! My main comment is that this describes the risks of a connector that connects 2 or more *public* ledgers. We should also describe the risks taken by connectors like the ones currently existing on the public Interledger, where a connector is owned by exactly one user (the administrator of the ilp-kit it is part of), and only connect this person's *personal* ledger, through trustlines, with the connectors of other ilp-kit admins.

I would say for such connectors the risks are:
* peer gets compromised and does not consider it to be their fault. E.g. Alice trusts Bob, Bob trusts Charlie. Charlie becomes evil, and steals from Alice, through Bob. It is now up to Bob to pay Alice back, because it were his sub-optimal sysadmin skills that led to not only Charlie owing money to Bob, but also Bob owing money to Alice.
* peer did not take peer limit amount seriously. When they wrote 'peer limit 10,000 euros', they didn't really mean it. This may lead to real debts and misunderstanding, especially if rippled payments go over this trustline.
* trustlines naturally build up disputed balance, albeit, hopefully, at a very slow pace. this money is not necessarily lost (although it can be, see next point), but it is something to be aware of.
* late source payment for rippled payment. Alice, Bob and Charlie are all trustworthy, but due to network outage, Bob loses money while rippling a payment from Alice to Charlie. This is the 'sender and connector are friends' case, which is already described in the docs, but from a different perspective. Here, it's not "connectors should choose to connect ledgers that perform well", it's just "our common goal is to keep the network healthy and not make money out of other participants' failures or misfortune".
* most importantly, trustlines really are trustlines, and you stand to lose all the money for which you trust peers, when that peer stops using their server (their liquidity falls to zero, and your money is locked up there), or stops being friendly, or goes bankrupt (they would want to pay you, but due to unforeseen circumstances, have to ask you to please waive the debt), or ultimately, in case of human death of a network participant, their inheritance may not pay out what the deceased person owed you (you are probably more sad about the person dying than about the money they owed you, but still, it's a financial risk to park your assets in the care of fellow human beings)

emschwartz:
@michielbdejong I think a lot of those things would be worth including in a spec for trustlines (which we should put together). Most of those aren't connector risks so much as the downside of the trustline model specifically.

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
```diff
@@ -0,0 +1,40 @@
...
+* **Redundant Instances** - connectors should run redundant processes and machines to increase the difficulty of interfering with their processing
+* **Avoiding Public Endpoints** - connectors may reduce attack vectors by using communication channels and protocols that do not entail having a publicly-accessible server (for example using [WebRTC](https://webrtc.org/) or [Virtual Private Networks](https://en.wikipedia.org/wiki/Virtual_private_network) to communicate with peers
+
+## Liquidity Exhaustion
+
+Attackers could tie up connectors' liquidity by preparing payments through them that the attacker knows will fail, for example if the attacker sends themselves numerous payments and then does not fulfill any of them.
+
+Possible mitigations include:
+
+* **Managing Percentage of Liquidity on Hold** - connectors should monitor the percentage of their total liquidity that is on hold at any given time and may reject incoming payment requests if they would cause too much of the connectors' funds to be tied up
```
dappelt: Not sure if this is a suitable mitigation. To maximize profits, a connector would want to put all his liquidity to work, not just a percentage of it. 

Instead of capping the total liquidity put on hold at a fixed percentage, a better approach is to limit the amount of liquidity that is put on hold per sender. If a connector deals the first time with a given sender, it is in his interest to put rather smaller amounts on hold for him. Once a sufficient number of transfers have been fulfilled the connector might increase the maximum amount he puts on hold for this sender, essentially increasing the senders "payment bandwidth".

This somewhat resembles the [slow start](https://en.wikipedia.org/wiki/TCP_congestion_control#Slow_start) mechanism in TCP.

dappelt:
I wonder if another category of connector risk is routing. Maybe we want to bookmark the topic of routing-related connector risk and come back to it once it is mature enough. cc @momerath42

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -71,11 +71,14 @@ For more details on the flow, see the [Interledger Protocol specification](../00
 
 <span class="show alert alert-info">**Note:** Interledger only supports Universal mode as described in the whitepaper. Atomic mode can be used by adjacent subsets of participants in an Interledger payment if desired, but this is not part of the standard.</span>
 
-### Error Recovery and Connector Risk
+### Connector Risk and Mitigation
 
-An ILP execution chain can be interrupted if a connector fails to deliver the condition fulfillment. In this case, the sender thinks the transaction failed and retries it using the same condition as the original transaction. If the retry goes through the same connector that failed the first time, that connector can claim the incoming funds without sending a second outgoing transfer.
+Interledger connectors accept some risk in exchange for the revenue they generate from facilitating payments. In the Interledger payment flow, connectors' outgoing transfers are executed before their incoming transfers. Once each connector is notified that the outgoing transfer has been executed, they have a window of time to deliver the fulfillment and execute the incoming transfer. Connectors that fail to deliver the fulfillment in time may lose money.
```
mDuo13: As a style rule, use "after" instead of "once" when you don't mean "one time"

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -71,11 +71,14 @@ For more details on the flow, see the [Interledger Protocol specification](../00
...
+
+If some transfers in an Interledger payment are executed and others expire, the receiver will think the payment was completed but the sender may think the whole payment failed. Senders may retry payments that expire and they should use the same condition as the original payment. If the second attempt takes the same route, the connector that failed the first time can complete the payment by submitting the fulfillment without sending another outgoing transfer. Connectors should prefer senders (or aggregators of senders such as wallet services) and connectors that retry payments through the same route, because they increase the likelihood of completing such payments. Connectors may incentivize this behavior by offering better rates to parties known to retry failed payments through the same route.
```
mDuo13: > Senders may retry payments that expire and they should use the same condition as the original payment.

Are these "may" and "should" the same meaning as RFC2119's MAY and SHOULD?

Also, I recommend splitting this into two sentences:

> Senders MAY retry payments that expire. If they do, they SHOULD use the same condition as the original payment.

mDuo13: > Connectors may incentivize this behavior by offering better rates to parties known to retry failed payments through the same route.

Is that not a discriminatory fee that makes routing inconsistent?

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
mDuo13: This is a full sentence, so it should be capitalized and end in a period (same goes for other bullets in this section):

> connectors should choose the ledgers they connect in part based on the reliability and speed of fulfillment notifications and the time the ledgers require to process fulfillments once they are submitted

Also, this is another case for "once"→"after"

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
```diff
@@ -0,0 +1,40 @@
...
+**Monitoring is a must! Even connectors that implement all of these strategies should monitor their transaction patterns and use warnings or kill switches to avoid losing money in the case of an unexpected attack.**
+
+## Fulfillment Failure
+
+The main risk connectors face in Interledger is being unable to fulfill the incoming transfer once their outgoing transfer has been executed.
+
+Possible mitigations include:
+
+* **Connecting Reliable Ledgers** - connectors should choose the ledgers they connect in part based on the reliability and speed of fulfillment notifications and the time the ledgers require to process fulfillments once they are submitted
+* **Adjusting Transfer Expiry Window** - connectors determine the window of time they require between the incoming and outgoing transfer expires and they should set this window such that they are confident that they will be able to deliver the fulfillment in time even if the outgoing transfer is executed at the last possible moment
```
mDuo13: I recommend splitting this into two sentences:

> Connectors determine the window of time they require between when the incoming and outgoing transfers expire. Connectors should set this window such that they are confident that they will be able to deliver the fulfillment in time even if the outgoing transfer is executed at the last possible moment.

(Also, I rephrased the expiration time bit for better agreement)

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
```diff
@@ -0,0 +1,40 @@
...
+* **Denial of Service Protections** - connectors should take standard [Denial of Service](https://en.bitcoin.it/wiki/Weaknesses#Denial_of_Service_.28DoS.29_attacks) attack prevention mechanisms to prevent attackers from overloading the connectors' servers with network traffic
```
mDuo13: I don't think it's right to say, "take... mechanisms"—you can take steps or create mechanisms.

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
```diff
@@ -0,0 +1,40 @@
...
+* **Prioritizing Fulfillments** - connectors should prioritize fulfilling transfers over preparing new ones or responding to quote requests and may even have separate processes or machines responsible for those different behaviors
+* **Redundant Instances** - connectors should run redundant processes and machines to increase the difficulty of interfering with their processing
+* **Avoiding Public Endpoints** - connectors may reduce attack vectors by using communication channels and protocols that do not entail having a publicly-accessible server (for example using [WebRTC](https://webrtc.org/) or [Virtual Private Networks](https://en.wikipedia.org/wiki/Virtual_private_network) to communicate with peers
+
+## Liquidity Exhaustion
+
+Attackers could tie up connectors' liquidity by preparing payments through them that the attacker knows will fail, for example if the attacker sends themselves numerous payments and then does not fulfill any of them.
```
mDuo13: "sends themselves numerous payments" makes it sound like the payments are completing. I recommend splitting the example into a separate sentence and rephrasing:

> Attackers could tie up connectors' liquidity by preparing payments through them that the attacker knows will fail. For example, an attacker could prepare numerous payments to itself and then not fulfill any of them.

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
```diff
@@ -0,0 +1,40 @@
...
+
+Possible mitigations include:
+
+* **Managing "Payment Bandwidth"** - connectors should monitor the percentage of their total liquidity that is on hold at any given time for a given peer or customer and may reject incoming payment requests if that party exceeds their allocated "payment bandwidth". Connectors may allocate less bandwidth to unknown or untrustworthy senders. Note that ILP Packets do not contain the source account so connectors should apply this logic to the immediate peer or customer that they receive the payments from
```
mDuo13: Probably worth mentioning that they can do it on the basis of the destination (which works in the case of an attacker performing a naive version of the example attack).

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
```diff
@@ -0,0 +1,40 @@
...
+* **Preferring Smaller Payments** - smaller payments place a smaller percentage of the connector's liquidity on hold and thus each one carries less risk that the payment being prepared and then failing would create significant opportunity cost
+* **Blacklisting Senders and Receivers** - connectors may refuse to facilitate payments from certain sources or to certain destinations if their rate of failed payments is unusually high
+
+## Exchange Rate Volatility
+
+Once connectors prepare their outgoing transfers, they are committed to the payment even if the exchange rate between the assets fluctuates.
+
+Possible approaches include:
+
+* **Building a Buffer into the Rate** - connectors may add a buffer to their expected exchange rate to account for some movement in the price, and they may add a premium for especially volatile currencies
```
mDuo13: Rather than "building a buffer into the rate", the term of art is probably more like, "Accounting for Slippage"

0001-interledger-architecture/0001-interledger-architecture.md
emschwartz: Good point. I think you could do this by having multiple connector instances and only peering the better rate one with people that implement such policies.

PR #184, opened 2017-04-04
docs(0019): Glossary

michielbdejong:


9999-glossary/9999-glossary.md
```diff
@@ -0,0 +1,272 @@
+# Glossary
+
+## Account
+An administrative item that acts as a bucket of assets on a ledger.
+
+## Application layer
+The mechanisms on top of the transport layer which embed it into an application, such as the
```
sappenin: Link to `transport layer`?

justmoon:
We should assign a proper RFC number for this glossary, I believe 19 is available?

0019-glossary/0019-glossary.md
```diff
@@ -0,0 +1,272 @@
...
+useful when the cost of resolving every small difference in accounting is much higher than the value of the differences.
+A transfer whose outcome is disputed among the ledger's administering authorities, can be resolved as 'disputed', in which case
+the sender's balance is decreased, and instead of increasing the receiver's balance, the sender's "disputed balance" is increased.
+Disputed balance can disappear over time when disputed transfers happen in the other direction, or when settled by out-of-band
+communication between the authorities who administer the ledger.
+Note that although individual ledgers can support disputed transfers, an interledger transaction as a whole can never be disputed;
+its final state can only be either 'executed' or 'canceled'.
+
+## Distributed ledger
+A ledger which is operated by a group of entities.
```
emschwartz: > A ledger that is operated by a group of entities and/or run on multiple servers. This term is used somewhat interchangeably with "Blockchain" or "Decentralized Ledger".

0019-glossary/0019-glossary.md
```diff
@@ -0,0 +1,272 @@
...
+
+## Executed
+One of the final states an interledger transactions can end up in.
```
emschwartz: > A final state of a transfer indicating that the funds were transferred to the account of the immediate recipient.

0019-glossary/0019-glossary.md
```diff
@@ -0,0 +1,272 @@
...
+
+## Final state
+The state an interledger transaction ends up in. Has to be either
+[canceled](#canceled) or [executed](#executed).
+
+## Forward action
```
emschwartz: Maybe it would be better to use the terms "Preparation Phase" and "Execution Phase", since that's how we normally describe this and it's more descriptive of what's going on than "forward" and "backward"

0019-glossary/0019-glossary.md
```diff
@@ -0,0 +1,272 @@
...
+## Incoming (transfer, ledger, amount)
+Relative to a connector, the transfer/ledger/amount that is located inbetween the connector and the sender, directly adjacent to the connector.
+For a receiver, the adjacent transfer/ledger/amount is always incoming.
+A sender does not have an incoming transfer/ledger/amount.
+
+## Interledger (in "Let's use Interledger for that!")
+The [Interledger protocol stack](#interledger-protocol-stack).
+
+## interledger (in "An interledger")
+Any network of two or more ledgers connected via the Interledger protocol stack. See also [Interledger (The)](#interledger-the)
```
emschwartz: I think this is more the adjective when you're describing transactions that go across multiple ledgers

0019-glossary/0019-glossary.md
```diff
@@ -0,0 +1,272 @@
...
+
+## Interledger (in "The Interledger")
+The main public network of ledgers connected via the Interledger protocol stack. See [ConnectorLand](https://connector.land) for statistics about it.
+
+## Interledger Addresses
+Interledger addresses provide a ledger-agnostic way to address ledgers and accounts. Interledger addresses are dot-separated strings that contain prefixes to group ledgers. An example address might look like `g.us.acmebank.acmecorp.sales.199` or `g.crypto.bitcoin.1BvBMSEYstWetqTFn5Au4m4GFg7xJaNVN2`.
+When initiating an Interledger payment, the sender attaches an ILP packet to the local transfer to the connector. The packet is a binary message that includes the destination account, destination amount, and additional data for the receiver. The packet is relayed by connectors and attached to each transfer that comprises the payment. In some cases, ledger protocols may define alternative ways to communicate the packet.
```
emschwartz: This description of the packet should be removed from this definition

0019-glossary/0019-glossary.md
```diff
@@ -0,0 +1,272 @@
...
+Note that anyone can claim to have a certain Interledger address, there is no registry of them.
+Whether or not a transaction ends up at the intended receiver is ultimately safe-guarded by the hashlock condition,
+not by enforcement of address ownership.
+See [IL-RFC 15](../0015-ilp-addresses/0015-ilp-addresses.md)
+
+## Interledger Architecture
+The [Interledger protocol stack](interledger-protocol-stack), and the design principles behind it.
+
+## Interledger layer
+The lowest layer of the Interledger protocol stack, consisting of
```
emschwartz: There's an interesting open question about whether the ledger layer is part of the "Interledger protocol stack"

0019-glossary/0019-glossary.md
```diff
@@ -0,0 +1,272 @@
...
+The binary data packet that is forwarded along all hops from sender to receiver, and which provides the receiver
+with metadata which the receiver may need to be willing and able to fulfill the transaction.
+It also contains the destination address and destination amount, on which connectors will base their routing decisions -
+both in choosing the onward path, and in deciding whether their incoming amount was big enough, in comparison to the destination
+amount mentioned in the ILP packet, to even merit an [forward action](#forward-action) at all. See also [routing table](#routing-table)
+
+## Interledger Payment Request (IPR)
+In the Interledger Payment Request (IPR) protocol, the receiver generates the payment details and condition. The receiver does not share the secret used to generate the condition and fulfillment with the sender or anyone else, but the sender must ask the recipient to generate and share a condition before sending each payment. IPR is primarily useful for building non-repudiable application layer protocols, in which the sender's posession of the fulfillment proves to third parties that the sender has paid the receiver for a specific obligation.
```
emschwartz: "A Transport Layer protocol in which..."

0019-glossary/0019-glossary.md
```diff
@@ -0,0 +1,272 @@
...
+Each account has a [balance](#balance), which is the amount of the ledger's assets the account holds. Account balances may be positive or negative, representing assets or liabilities.
+
+## Liquidity
+Maximum amount which can be moved back and forth over a route.
+
+## Local transaction
+A transaction that contains just one [transfer](#transfer) and zero [hops](#hop), thus staying within the same ledger.
+
+## Outgoing (transfer, ledger, amount)
+Relative to a connector, the transfer/ledger/amount that is located inbetween the connector and the receiver, directly adjacent to the connector.
```
emschwartz: Rather than describing this in terms of adjacency, I think outgoing transfers would be easier to understand as the transfers of assets from a sender or connector to the sender or another connector.

0019-glossary/0019-glossary.md
```diff
@@ -0,0 +1,272 @@
...
+For a sender, the adjacent transfer/ledger/amount is always outgoing.
+A receiver does not have an outgoing transfer/ledger/amount.
+
+## Payment
+Roughly synomymous to [transaction](#transaction), but with emphasis on the context of the negotiation between the sender and receiver. For instance,
```
emschwartz: We have tried to refer to Interledger movements of value as "payments" to differentiate between the interledger things and the local ledger things. "Transaction" seemed a bit too broad for that

michielbdejong:
One of the main questions is how we count hops. I was about to say that a same-ledger transaction is one hop, but then changed it to say it's zero hops, because:
* there is no 'hopping' movement if you just add a transaction in the books of one ledger
* we already have a name for what one ledger does: transfer, so making hop a synonym of that would be superfluous
* we don't really have a name for what one connector does, so 'hop' could be used for this (the only other word I can think of is 'connection'
* that's how the internet does it (a connection within a LAN is 'local', a connection that goes through one router is one hop).

@justmoon what do you think?

0019-glossary/0019-glossary.md
```diff
@@ -0,0 +1,272 @@
...
+a transaction may be paying a debt, or a purchase, it may be a pull payment or a push payment. All these aspects which put a transaction in a wider
+context and link it to real-world interactions, contracts, and motivations between parties make a transaction into a payment. A
+[cyclic transaction](#cyclic-transaction) is generally not considered a payment, because the sending party of such a transaction is also the receiving party.
+
+## Payment channel
+A payment channel is similar to a trustline, but each party parks an asset on an "underlying ledger", equal to the other party's maximum balance.
+Apart from disputes canceling each other out, and out-of-band settlement, disputes can be settled by each party claiming the other party's parked
```
emschwartz: This definition makes very little sense to me. How about:

> A method used by two parties on the same ledger to make multiple transfers such that the ledger enforces the final net positions without the ledger being informed of each individual transfer. One or both parties will generally escrow some of their assets on the ledger and transfers are made by exchanging "claims" that update the net balance. When one of the parties wishes to close the channel, they submit the final balance to the ledger. Different implementations have different rules about which party is allowed to close the channel, the consequences of doing so, and the mechanisms for settling disputes. For more details see [Bitcoin Payment Channels](https://en.bitcoin.it/wiki/Payment_channels). Note that payment channels are similar to [trustlines](#trustlines) except the ledger is used to escrow one or both parties' assets and provide clear resolution mechanisms for disputes over the final net positions.

0019-glossary/0019-glossary.md
```diff
@@ -0,0 +1,272 @@
...
+asset. This way, two strangers can run a ledger with properties similar to a trustline, but without fully trusting each other.
+
+## Peer balance
```
emschwartz: Ugh this is confusing because this is unrelated to "Peering"

0019-glossary/0019-glossary.md
```diff
@@ -0,0 +1,272 @@
...
+## Forward action
+For a sender, the forward action consists in creating an outgoing transfer.
+When a connector is notified that an incoming transfer has reached the 'prepared' state, its forward action is to propose a transfer on the next ledger,
+with the same hashlock condition, and the same ILP packet attached.
+For a receiver, there is no forward action.
+
+## Hop
+One connector-facilitated connection between one ledger and the next.
+The number of hops is always one less than the number of transfers, so a one-transfer ("local") transaction is a zero-hop transaction.
+
```
emschwartz: Can you add a definition for "Hold"?

0019-glossary/0019-glossary.md
```diff
@@ -0,0 +1,272 @@
...
+## Peer balance
+See [trustline](#trustline).
+
+## Peering
+Connectors "peer" with one another to exchange information used to determine the best route for a payment, and to route it (connect its chain of
+transfers through [forward](#forward-action) and [backward](#backward-action) action.
+A peering relationship is usually implemented using either a [trustline](#trustline) or a [payment channel](#payment-channel)..
+
+## Prepared
+The second state of an interledger transaction, after [proposed](#proposed), and
```
emschwartz: > The state of a transfer in which the source account's funds are put on hold by the ledger until either the expiry is reached or the condition is fulfilled. See Conditional Transfers.

0019-glossary/0019-glossary.md
```diff
@@ -0,0 +1,272 @@
...
+before its [final state](#final-state).
+
+## Pre-Shared Key (PSK)
+The recommended transport layer for most use cases.
+In Pre-Shared Key (PSK) protocol, the sender and receiver use a shared secret to generate the payment condition, authenticate the ILP packet, and encrypt application data. Using PSK, the sender is guaranteed that fulfillment of their transfer indicates the receiver got the payment, provided that no one aside from the sender and receiver have the secret and the sender did not submit the fulfillment. See [IL-RFC 16](../0016-pre-shared-key/0016-pre-shared-key.md).
+
+## Proposed
```
emschwartz: I think this term is kind of deprecated. It's the state of a transfer before funds are put on hold. In the first version of Interledger that was actually being used, but then we got rid of it and decided that you can only create transfers when you're authorized and then the transfer will automatically be prepared.

0019-glossary/0019-glossary.md
```diff
@@ -0,0 +1,272 @@
...
+The initial state of an interledger transaction, before [prepared](#prepared).
+
+## Receiver
+The party who is the beneficiary (payee) of a transfer or a transaction (payment).
```
emschwartz: This is still ambiguous as to whether this just means the beneficiary of an interledger payment or also refers to the entity that gets each individual transfer

0019-glossary/0019-glossary.md
```diff
@@ -0,0 +1,272 @@
...
+
+## Route
+A list of [transfers](transfer) and the [hops](#hop) between them.
```
emschwartz: This can refer to both the "route" an actual payment has taken or the theoretical way that a future payment could go. The former is sometimes also called the path

0019-glossary/0019-glossary.md
```diff
@@ -0,0 +1,272 @@
...
+The party who is the initiator (payer) of a transfer or a transaction (payment).
+
+## Simple Payment Setup Protocol (SPSP)
+SPSP uses Webfinger ([RFC 7033](https://tools.ietf.org/html/rfc7033)) and an HTTPS-based protocol for communicating account, amount, and Pre-Shared Key details from the receiver to the sender.
+See [IL-RFC 9](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md).
+
+## Source (transfer, ledger, amount)
+The transfer/ledger/amount directly adjacent to the sender.
+
+## Transaction
```
emschwartz: Payment is better than transaction

0019-glossary/0019-glossary.md
```diff
@@ -0,0 +1,272 @@
...
+The movement of assets from one party to another. A transaction consists of one or more [ledger transfers](#transfer). Roughly synonymous with
+[payment](#payment).
+
+# Transfer
+The movement of assets on one single ledger. Multiple transfers can be chained together into one multi-hop [transaction](#transaction).
```
emschwartz: Inconsistent use of "multi-hop"

0019-glossary/0019-glossary.md
```diff
@@ -0,0 +1,272 @@
...
+Also sometimes called 'ledger transfer'.
+
+## Transport layer
+The middle layer of the Interledger protocol stack, which can be either
```
emschwartz: It's most important to describe the layer in terms of what it does. In this case it's determining the condition and encoding of the data in the ILP packet, and what details the sender and receiver need to discuss beforehand

0019-glossary/0019-glossary.md
```diff
@@ -0,0 +1,272 @@
...
+either [Pre-Shared Key](#psk) or [Interledger Payment Request](#ipr). It is called the transport layer because
+it provides the end-to-end hashlock condition which ultimately connects the hops that chain the individual
+ledger transfers together, into one Interledger transaction.
+
+## Trustline
```
emschwartz: I think this should be described as a credit relationship between the two parties in which each keep track of their own balance

0019-glossary/0019-glossary.md
```diff
@@ -0,0 +1,248 @@
...
+a non-physical quantity of currency, a non-physical claim to a resource, a reputation for keeping one's word, someone else's intention to obey the
+terms of a contract, etc. Some types of assets can be more easily traded on a ledger than others. Also, for some asset pairs it is easier to determine
+a fair exchange rate than for others.
+
+## Balance
+The amount of a ledger's assets held by a given account. This concept is not as clearly defined as it may seem at first glance.
+When part of an account's balance is not readily usable, it is up to the ledger implementation to decide whether an account's
+"balance" includes only the readily usable balance, or all of it, or some number inbetween. Some ledgers may give a more detailed
+report, specifying amounts for, for instance, outgoing transfers which have not yet
+completed, incoming transfers which have not yet completed, and for some ledgers even disputed balance.
```
emschwartz: This definition seems overly complicated. The additional behavior you describe is not currently implemented anywhere and it's not recommended either

0019-glossary/0019-glossary.md
```diff
@@ -0,0 +1,248 @@
...
+
+## Execution Phase
+The second phase of an ILP payment, in which the fulfillment is passed back from the receiver, via the connector(s), to the sender.
+
+## Final State
+The state an interledger payment ends up in. Has to be either
+[rejected](#rejected) or [executed](#executed).
+
+## Fulfillment
+See [conditional transfer](#conditional-transfer).
```
emschwartz: Should be "A 32-byte value used to trigger the execution of a transfer. In most Interledger payments, the fulfillment is known only to the receiver (or in the case of the Pre-Shared Key protocol it is known to the sender and the receiver). For more details see conditional transfer"

0019-glossary/0019-glossary.md
```diff
@@ -0,0 +1,248 @@
...
+
+## Hold
+The state of part of the sender's, connector's, and/or receiver's account balance, during which that balance is not readily available, in relation to
```
emschwartz: This is not really a state. When the transfer is `prepared` the money is on hold. It's confusing to use the word state differently here

0019-glossary/0019-glossary.md
```diff
@@ -0,0 +1,248 @@
...
+
+## Hop
+One of the transfers that make up a payment. Specifically, a 'multi-hop' payment is one involving more than one ledger, and thus at least one connector.
+Likewise, 'best hop', as used in the ilp-connector source code when making routing decisions, is best interpreted as 'best next transfer'.
+
+## IL-RFC
+Interledger request for comments. A document describing a part of the protocol stack, or a topic related to it, such as this glossary. The official list of IL-RFCs is maintained in
+[gh:interledger/rfcs](https://github.com/interledgers/rfcs).
+
+## Incoming (Transfer, Ledger, Amount)
```
emschwartz: Wouldn't you also describe the destination transfer as an incoming transfer from the perspective of the receiver? I don't think this term has meaning in general, it's relative to the party you're talking about

0019-glossary/0019-glossary.md
```diff
@@ -0,0 +1,248 @@
...
+The transfer/ledger/amount where a connector plays the 'receiver' role. See also [Outgoing](#outgoing-transfer-ledger-amount).
+
+## Interledger (in "Let's use Interledger for that!")
+The [Interledger protocol stack](#interledger-protocol-stack).
+
+## interledger (in "An interledger network")
+Of a network of two or more ledgers, indicing that it uses the Interledger protocol stack. See also [Interledger (The)](#interledger-the)
```
emschwartz: typo "indicing"

Also, I still think lower case interledger is just an adjective that refers in general to transactions that cross multiple ledgers. It doesn't necessarily indicate you are using the Interledger protocol stack

0019-glossary/0019-glossary.md
```diff
@@ -0,0 +1,248 @@
...
+## Interledger Module
+The part of a software application that processes ILP payments.
+Analoguous to the network card of an internet-connected computer.
+
+## Interledger Packet
+The binary data packet that is forwarded along all hops from sender to receiver, and which provides the receiver
+with metadata which the receiver may need to be willing and able to fulfill the payment.
+It also contains the destination account's address and destination amount, on which connectors will base their routing decisions -
+both in choosing the onward path, and in deciding whether their incoming amount was big enough, in comparison to the destination
+amount mentioned in the ILP packet, to even merit an [forward action](#forward-action) at all. See also [routing table](#routing-table)
```
emschwartz: remove the reference to "forward action"

0019-glossary/0019-glossary.md
```diff
@@ -0,0 +1,248 @@
...
+See [IL-RFC 11](../0011-interledger-payment-request/0011-interledger-payment-request.md).
+
+# Interledger Protocol (ILP)
+The core of the Interledger protocol suite, described in [IL-RFC 3](../0003-interledger-protocol/0003-interledger-protocol.md).
+Colloquially the whole Interledger stack is sometimes referred to as "ILP". Technically, however, the Interledger Protocol is only one layer in the
+[Interledger protocol stack](#interledger-protocol-stack).
+
+## Interledger protocol stack
+The stack consisting of the [Interledger layer](#interledger-layer),
+a choice of [transport layers](#transport-layer), and an [application layer](#application-layer).
```
emschwartz: I think the stack refers to the whole set of protocols, not just when you choose one from each layer

0019-glossary/0019-glossary.md
```diff
@@ -0,0 +1,248 @@
...
+Maximum amount which can be moved back and forth over a route.
+
+## Local Payment
+A payment that contains just one [transfer](#transfer) and zero [hops](#hop), thus staying within the same ledger.
+
+## Outgoing (Transfer, Ledger, Amount)
+The transfer/ledger/amount for which a connector plays the sender role. See also [Incoming](#incoming-transfer-ledger-amount).
+
+## Payment
+The movement of assets from one party to another. A payment consists of one or more [ledger transfers](#transfer). Roughly synonymous with
```
emschwartz: In our code this is used mostly to refer to _interledger_ movements of value, to distinguish those from local transfers

0019-glossary/0019-glossary.md
```diff
@@ -0,0 +1,248 @@
...
+
+## Payment channel
+A method used by two parties on the same ledger to make multiple transfers such that the ledger enforces the final net positions without the ledger being informed of each individual transfer. One or both parties will generally escrow some of their assets on the ledger and transfers are made by exchanging "claims" that update the net balance. When one of the parties wishes to close the channel, they submit the final balance to the ledger. Different implementations have different rules about which party is allowed to close the channel, the consequences of doing so, and the mechanisms for settling disputes. For more details see [Bitcoin Payment Channels](https://en.bitcoin.it/wiki/Payment_channels). Note that payment channels are similar to [trustlines](#trustlines) except the ledger is used to escrow one or both parties' assets and provide clear resolution mechanisms for disputes over the final net positions.
+
+A payment channel is similar to a trustline, but each party parks an asset on an "underlying ledger", equal to the other party's maximum balance.
```
emschwartz: This seems redundant with the previous paragraph

0019-glossary/0019-glossary.md
```diff
@@ -0,0 +1,248 @@
...
+Apart from disputes canceling each other out, and out-of-band settlement, disputes can be settled by each party claiming the other party's parked
+asset. This way, two strangers can run a ledger with properties similar to a trustline, but without fully trusting each other.
+
+## Peer balance
+See [trustline](#trustline).
+
+## Peering
+Connectors "peer" with one another to exchange information used to determine the best route for a payment, and to route it (connect its chain of
+transfers during the [preparation phase](#preparation-phase) and [execution phase](#execution-phase).
+A peering relationship is usually implemented using either a [trustline](#trustline) or a [payment channel](#payment-channel)..
```
emschwartz: Might be worth mentioning the analogy with ISP peering and linking to the wikipedia article about it

0019-glossary/0019-glossary.md
```diff
@@ -0,0 +1,248 @@
...
+
+## Preparation Phase
+The first phase of an ILP payment, in which the ILP packet is passed from the sender, via the connector(s), to the receiver.
```
emschwartz: Also, money is put on hold. That is the most important thing to mention about this phase!

0019-glossary/0019-glossary.md
```diff
@@ -0,0 +1,248 @@
...
+
+## Receiver
+The party who is the beneficiary (payee) of a transfer or a payment (transaction).
+NB: When Alice sends a payment through a connector to Bob,
+it's unclear whether by "the receiver" you mean the receiver of the transfer
+(i.e., the connector), or the receiver of the payment (i.e., Bob), so use this
+term carefully.
+
+## Rejected
+One of the final states an interledger payment can end up in.
```
emschwartz: ...that means the funds have been returned to the immediate sender.

0019-glossary/0019-glossary.md
```diff
@@ -0,0 +1,248 @@
...
+
+## Route
+A list of [transfers](transfer) and the [hops](#hop) between them.
```
emschwartz: "list" -> "set"

0019-glossary/0019-glossary.md
michielbdejong: it's from https://github.com/interledgerjs/ilp-plugin-bells/issues/57 but you're right, the `it is up to the ledger implementation to decide whether an account's "balance" includes only the readily usable balance, or all of it, or some number inbetween` is enough, i'll remove the more detailed sentence.

michielbdejong:
Thanks a lot for all the review work! :)

PR #185, opened 2017-04-05
Docs: add dump() to LPI store

sharafian:
Required by https://github.com/interledgerjs/ilp-plugin-virtual/pull/57 . Adds a `dump()` method to the plugin store that returns the plugin store as an object in memory. Plugin virtual uses this to cache its store and perform operations on it synchronously, to avoid race conditions. Implemented in the connector's plugin store in https://github.com/interledgerjs/ilp-connector/pull/326


emschwartz:
What happens if this object is enormous? Normally you don't want to dump the contents of something thats basically a database into memory. I'd expect it to have something more like a `.keys` function instead

sharafian:
`plugin-virtual` uses every key-value pair to calculate its balance, so it would have to take the result of `keys` and make a single database query for every single key in the store.

Issue #191, opened 2017-04-11
Update generate_web.js  to remove *.md from links in RFCs and ensure they are relative


adrianhopebailie:
Links in the RFCs should be relative and not include the *.md suffix.

Example, change:
`https://github.com/interledger/rfcs/0017-ledger-requirements/0017-ledger-requirements.md` or 
`../0017-ledger-requirements/0017-ledger-requirements.md` to 

`../0017-ledger-requirements/`

emschwartz:
👎  because then the experience of reading them on Github is terrible. If the tool we're using to render them on the website can't find them because of that, we should update the tool

adrianhopebailie:
> because then the experience of reading them on Github is terrible

I think that's a bit strong.

We should be prioritizing the website experience over the Github experience anyway. Let's investigate what it will take to update the website building logic and then see which makes the most sense to do.

emschwartz:
The primary audience for the specs consists of more technical people who want to dive into the details of ILP. I would expect most to want to read them on Github rather than the website. Also, we won't ever link all of the RFCs from the website so Github is definitely going to be the primary place to read them.

adrianhopebailie:
Looks like a small addition to https://github.com/interledger/rfcs/blob/master/scripts/generate_web.js will be enough to re-write the links:

```javascript
$('a').each(function(i, elem) {
  //Somebody who knows jQuery/Cheerio better than me can do a quick string manipulation
});
```

PR #192, opened 2017-04-11
Fixes relative links across all rfcs's.

ivanseidel:
This will make links compatible with both github and Interledger website

Reading the docs from github should lead to the right content as well as reading it from the website, without changing from one another.

It changes all `../xxxx-<anything>/xxxx-<anything>.md` to `../xxxx-<anything>`


ivanseidel:
Fixes #191

michielbdejong:
Should we merge this?

PR #198, opened 2017-04-24
RFC-4: Request/response semantics

justmoon:
Changes message semantics in ledger plugins to request / response semantics.

This was prompted by a problem that surfaced during implementation: When responding to a quote request with an error, how can the error be associated with the quote request since it does not contain an ID? We could have added an ID to the error format, but we realized that with transfers the association was handled by the ledger layer. A rejection is always associated with a corresponding transfer.

Most of our ledger plugins already use request/response semantics internally, e.g. via JSON-RPC. So we ended up building request/response on top of request/response.

By moving this into the ledger layer, we allow ledger plugins to take advantage of any optimizations they may be able to do like using an HTTP response to send the response message.

This PR also removes the deprecated `account` field and removes the `IncomingTransfer/OutgoingTransfer`, `IncomingMessage/OutgoingMessage` distinction, since these objects now look the same.

Finally, this also means that the `requestId` in the ILQP messages is no longer needed.

### DIFF
--- a/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
+++ b/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
@@ -22,10 +22,12 @@ This spec depends on the [ILP spec](../0003-interledger-protocol/).
 | | [**getAccount**](#getaccount) ( ) `⇒ String` |
 | | [**getBalance**](#getbalance) ( ) <code>⇒ Promise.&lt;String></code> |
 | | [**getFulfillment**](#getfulfillment) ( transferId ) <code>⇒ Promise.&lt;String></code> |
-| | [**sendTransfer**](#sendTransfer) ( transfer ) <code>⇒ Promise.&lt;null></code> |
-| | [**sendMessage**](#sendMessage) ( message ) <code>⇒ Promise.&lt;null></code> |
+| | [**sendTransfer**](#sendtransfer) ( transfer ) <code>⇒ Promise.&lt;null></code> |
+| | [**sendRequest**](#sendrequest) ( message ) <code>⇒ Promise.&lt;[Message](#class-message)></code> |
 | | [**fulfillCondition**](#fulfillcondition) ( transferId, fulfillment ) <code>⇒ Promise.&lt;null></code> |
 | | [**rejectIncomingTransfer**](#rejectincomingtransfer) ( transferId, reason ) <code>⇒ Promise.&lt;null></code> |
+| | [**registerRequestHandler**](#registerrequesthandler) ( requestHandler ) <code>⇒ null</code> |
+| | [**deregisterRequestHandler**](#deregisterrequesthandler) ( ) <code>⇒ null</code> |
 
 ###### Events
 | Name | Handler |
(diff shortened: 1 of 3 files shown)


0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
```diff
@@ -22,28 +22,33 @@ This spec depends on the [ILP spec](../0003-interledger-protocol/).
 | | [**getAccount**](#getaccount) ( ) `⇒ String` |
 | | [**getBalance**](#getbalance) ( ) <code>⇒ Promise.&lt;String></code> |
 | | [**getFulfillment**](#getfulfillment) ( transferId ) <code>⇒ Promise.&lt;String></code> |
-| | [**sendTransfer**](#sendTransfer) ( transfer ) <code>⇒ Promise.&lt;null></code> |
-| | [**sendMessage**](#sendMessage) ( message ) <code>⇒ Promise.&lt;null></code> |
+| | [**sendTransfer**](#sendtransfer) ( transfer ) <code>⇒ Promise.&lt;null></code> |
+| | [**sendRequest**](#sendrequest) ( message ) <code>⇒ Promise.&lt;null></code> |
```
sentientwaffle: `sendRequest()` should return `Promise<ResponseMessage>`.

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
```diff
@@ -253,13 +258,13 @@ Rejects with `NoSubscriptionsError` if the message cannot be delivered because t
 
 ###### Example
 ```js
-p.sendMessage({
+p.sendRequest({
   to: 'example.ledger.connector',
   data: { foo: 'bar' }
```
sentientwaffle: `data` renamed to `custom`

justmoon:
@sentientwaffle Good catches, addressed your comments.

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
```diff
@@ -253,13 +258,14 @@ Rejects with `NoSubscriptionsError` if the message cannot be delivered because t
 
 ###### Example
 ```js
-p.sendMessage({
+p.sendRequest({
   to: 'example.ledger.connector',
-  data: { foo: 'bar' }
+  ilp: '...base64url-encoded data...',
+  custom: { foo: 'bar' }
 })
```
emschwartz: This example should include handling the response

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
```diff
@@ -521,54 +576,46 @@ The `Message` class is used to describe local ledger message. All fields are req
...
-
-#### account
-<code>**account**:String</code>
-
-The ILP Address of a local account.
+| `String` | [id](#message-id) | Unique message identifier |
+| `String` | [from](#message-from) | ILP Address of the source account |
+| `String` | [to](#message-to) | ILP Address of the destination account |
+| `String` | [ledger](#message-ledger) | ILP Address prefix of the ledger |
+| `String` | [ilp](#message-ilp) | Base64-encoded ILP packet |
```
emschwartz: This should clarify that "ILP Packet" in this context does not necessarily mean ILP payment packet. I was confused looking at this at first

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
```diff
@@ -226,25 +231,25 @@ p.sendTransfer({
...
 |:--|:--|:--|
-| message | <code>[OutgoingMessage](#outgoingmessage)</code> | Properties of the message to be created |
+| message | <code>[Message](#class-message)</code> | Properties of the message to be created |
 
-When sending messages, the [to](#to) and [data](#data) fields are required.
+When sending messages, the [to](#message-to) and [ilp](#message-ilp) fields are required.
 
 ###### Returns
-**`Promise.<null>`** A promise which resolves when the message has been submitted (but not necessarily delivered). To ensure delivery, you must build a mechanism for that on top.
+**<code>Promise.<[Message](#class-message)></code>** A promise which resolves when a response message has been received.
```
emschwartz: Is the expectation that protocols built on top of this will define more status codes, including something like an HTTP 200, or will those be included in the standard set of ILP errors?

Will errors also be communicated as ILP packets?

justmoon: > Is the expectation that protocols built on top of this will define more status codes, including something like an HTTP 200, or will those be included in the standard set of ILP errors?

Not sure I understand the question correctly, but if there is an error in the ledger protocol (e.g. an HTTP 500 for an HTTP based protocol), the plugin should probably still generate a standard ILP error and return that instead.

> Will errors also be communicated as ILP packets?

Yes, errors will always be in the ILP error format as described in the ASN.

justmoon:
@michielbdejong Sounds good to me!

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
```diff
@@ -287,12 +293,31 @@ if transfer is not conditional.
...
+Set the callback which is used to handle incoming request messages. The callback expects one parameter (the request [Message](#class-message)) and returns a promise for the response [Message](#class-message).
+
+If a request handler is already set, this method throws a `RequestHandlerAlreadyRegisteredError`. In order to change the request handler, the old handler must first be removed via [`deregisterRequestHandler`](#deregisterRequestHandler). This is to ensure that handler are not overwritten by accident.
+
+#### deregisterRequestHandler
+<code>ledgerPlugin.deregisterRequestHandler( ) ⇒ null</code>
+
+Removes the currently used request handler. This has the same effect as if [`registerRequestHandler`](#registerrequesthandler) had never been called.
+
+If not request handler is currently set, this method does nothing.
```
sharafian: Grammatical error

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
sharafian: what's the error type if the other side doesn't have a handler registered?

michielbdejong:
This can be rebased and merged now! :)

adrianhopebailie:
@justmoon I think this is ready to merge

michielbdejong:
I'll rebase and merge it.

michielbdejong:
Merged in #228.

PR #199, opened 2017-04-26
Describe live situation

michielbdejong:
Two commits:
* a few words about C2C (still a minimal stub, just enough to get the reader on their way)
* describe how ILQP *currently* works on the public Interledger.

This one change is the only place where the rfcs repo was still ahead of reality, so with this, we're really switching the rfcs repo from *prescribing* ("soll") to *describing* ("ist"). From now on, we could line up RFC PRs with implementation PRs, and merge code+docs at the same time.

### DIFF
--- a/0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md
+++ b/0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md
@@ -50,8 +50,8 @@ The message formats described here are the `data` properties of an `OutgoingMess
 |:------------------------------|:------------|:------------|
 | `source_address`              | ILP Address | Address of the source's account in the source ledger. |
 | `destination_address`         | ILP Address | Address of the destination's account in the destination ledger. |
-| `source_amount`               | String      | Fixed integer amount to debit from the source's account. (Required unless `destination_amount` specified.) |
-| `destination_amount`          | String      | Fixed integer amount to credit to the destination's account. (Required unless `source_amount` specified.) |
+| `source_amount`               | String      | Fixed float amount to debit from the source's account, expressed in the ledger's currency (not in the ledger's base unit). (Required unless `destination_amount` specified.) |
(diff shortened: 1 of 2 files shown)


adrianhopebailie:
@michielbdejong we should consider a less strict binding between the code and specs IMO.

I have logged an issue to version the RFCs ( #201 ) as this has come up as requirement for external implementors.

We should eat our own dog-food on this and always specify in our code which version of the RFC we are implementing.

While I appreciate that it's pretty much the same group of contributors, work on the RFCs (Interledger Community) and work on the JS reference implementations (InterledgerJS) is being done in two separate projects.

michielbdejong:
I'm not so much talking about the reference implementation as a project, more about the public Interledger network. IMHO https://interledger.org/rfcs should define what people can expect to see when they join the network; whether that is with our code, or with their own implementation.

If we want https://interledger.org/rfcs to also describe the protocols that are used in other (private) networks, and not just the ones used on the *public* Interledger then yes, we should introduce versioning. And then we need a place where we say "Use any version you want, but the version that is currently live on the *public* Interledger is such-and-such".

michielbdejong:
With this week's work from the Lux team, we're finally close to getting the live network in sync with https://interledger.org/overview.html. I'm not opposed to further changes to the specs, but I would propose a new way of using this repo's `master` branch, namely to describe the current situation, and not to describe the propose situation.

We will of course still need a place to propose, describe, and spec out breaking changes before we implement, but maybe this place can be https://github.com/interledger/rfcs/pulls from now on, instead of https://interledger.org/overview.html?

adrianhopebailie:
> IMHO https://interledger.org/rfcs should define what people can expect to see when they join the network; whether that is with our code, or with their own implementation.

I don't agree with that entirely. The purpose of the RFCs has always been to describe the protocol. 

I agree that we should have a place that describes the current network. Perhaps that should be part of the docs for ILP Kit or should be part of https://connector.land ?

michielbdejong:
Merged into https://github.com/interledger/rfcs/pull/203

PR #200, opened 2017-04-26
[do not merge] restore ilqp to prescribing

michielbdejong:
This is for once we implemented the prescribed behavior of ILQP, and released a version of ilp-kit that uses this format, to restore the ILQP spec as it was prescribed a month ago.


michielbdejong:
Superseded by https://github.com/interledger/rfcs/pull/203

PR #203, opened 2017-04-28
C2C and ILQP to integers

michielbdejong:
Companion to https://github.com/interledgerjs/ilp-routing/pull/40

### DIFF
--- a/0010-connector-to-connector-protocol/0010-connector-to-connector-protocol.md
+++ b/0010-connector-to-connector-protocol/0010-connector-to-connector-protocol.md
@@ -1,1 +1,32 @@
-> Placeholder
+# Connector to Connector protocol
+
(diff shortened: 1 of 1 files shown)


sentientwaffle:
The quoting changes will use integers for quoting & route curves, though the former will be encoded in [ILQP packets](https://github.com/interledgerjs/ilp-packet#serializedeserialize-ilqp-quote-requestsresponses). Is that what you are referring to?

sharafian:
Yeah

michielbdejong:
@sharafian can you re-review this, now that I set the base branch to origin/master?

Basically, this PR proposes to keep the ILQP amounts in integers (this commit is not visible after I squashed it together with #199), and to add a document stub about C2C.

PR #204, opened 2017-04-28
docs(asn1): enable extensibility byte for ilp errors

justmoon:


michielbdejong:
Is this a breaking change?

emschwartz:
@michielbdejong I think so but I don't think we're actually using ILP Errors anywhere yet

sentientwaffle:
[ilp-plugin-bells](https://github.com/interledgerjs/ilp-plugin-bells/blob/f819a840d46d742510587861ab2777b046440c68/src/lib/plugin.js#L449) uses them, and ilp-connector adds itself to the forwarded-by list before relaying them.

michielbdejong:
ok, well, it will only break when a newer ilp-kit tries to send an error message, so then it doesn't break the happy path :)

michielbdejong:
ah, and: this is a docs update to match what the code already does, look:
https://github.com/interledgerjs/ilp-packet/blob/master/index.ts#L414
https://github.com/interledgerjs/ilp-packet/blob/master/index.ts#L442

emschwartz:
@justmoon what's the status of this?

michielbdejong:
I included this correction in testnet 17q4, since it's a docs clarification to match what the code already does: https://github.com/interledger/interledger/wiki/Interledger-over-BTP/_compare/cd976ee02f0e3a3175c516e6edf9876dccd7a59a...52378f2468d964a74e405d329c51ec01770ef321

I think it's safe to merge this into master now, but maybe we should leave that to @justmoon as he's the PR owner.

PR #205, opened 2017-04-29
Fix Heading Styles

sappenin:
It’s minor, but some of the headings in the [glossary](https://github.com/interledger/rfcs/blob/master/0019-glossary/0019-glossary.md) are bigger than the others (they have only a single #-sign, but should have two pound-signs.  This fixes that.

### DIFF
--- a/0019-glossary/0019-glossary.md
+++ b/0019-glossary/0019-glossary.md
@@ -41,15 +41,15 @@ A transaction where the destination account is the same account, on the same led
 when rebalancing liquidity (to enable future payments), or when rebalancing stored value (to spread risk, or to take advantage
 of changing exchange rates).
 
-# Destination Account
+## Destination Account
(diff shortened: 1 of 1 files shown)


sappenin:
@michielbdejong Not sure what the etiquette is for merging.  Do you want to merge, or should I?  And does this need any other reviewers?

michielbdejong:
Normally, you would merge if the button is green (so at least one approved review, and tests passing), unless you're external to the github org (which you're not), then the reviewer would merge.

I think the reason for this is that the author might want to be in charge of how and when their changes get merged, but it has also sometimes surprised me why there's this extra back-and-forth in the etiquette. :)

So yeah, long story short, go ahead and merge! :)

sappenin:
Sounds good, I'll go ahead and merge.  Thanks!

Issue #201, opened 2017-04-26
Put versioning information in RFCs


adrianhopebailie:
As implementations are being built of the various RFCs there is a need to reference the RFCs and specifically which version has been implemented.

## Proposal
* Add version meta-data to each RFC.
* Generate a versioned HTML file: `<version>.html` if one doesn't exist.
* If `<version>.html` exists then fail the build as the doc has changed and the version number should be updated.
* Copy `<version>.html` to `index.html` for latest version.
* Add a list of links to previous versions.

michielbdejong:
In step 3, only fail of `<version>.html` exists and is different. Building the master branch with (some) docs that didn't change should work.
In step 4, we can have extra flexibility with a separate config file that specifies which version is currently live; then, future versions can already by committed to master, and we don't break the docs everytime someone merges a branch.
Then, in step 5, we can have links to previous *and future* versions! :)

adrianhopebailie:
@michielbdejong I have started implementing, will get you to review later

PR #206, opened 2017-05-02
Add versioning to RFCs

adrianhopebailie:
Changes to the build process to extract a semver style version number from the title. A version specific file <version>.html is generated if one does not exist.

If the file exists and the content is different then a build error is reported. (Author should increment the build number).

Future change will be the addition of a linked list of previous versions.


adrianhopebailie:
> At what point in the writing process should the writer increase the version number?

Before publishing. i.e. Before submitting a PR

michielbdejong:
How do we keep track of which version should be displayed on https://interledger.org/overview.html? When/how is that updated?

adrianhopebailie:
The script currently publishes an index.html which is the latest. In future we can provide links to previous versions directly from the latest.

This resolves #201

michielbdejong:
> The script currently publishes an index.html which is the latest.

And will then show something like https://docs.docker.com/v1.13/ on our docs overview, where we can update which of the versions has the string `(current)` behind it?

adrianhopebailie:
> And will then show something like https://docs.docker.com/v1.13/ on our docs overview, where we can update which of the versions has the string (current) behind it?

This PR makes no assumptions about that :smile:

This just generates the versioned files. We can decide how we link to them in a future update

michielbdejong:
What is the relationship to https://github.com/interledgerjs/ilp-kit/blob/master/scripts/publish_web.js?

adrianhopebailie:
https://github.com/interledgerjs/ilp-kit/blob/master/scripts/publish_web.js is only run for deployment. It pushes the new files to gh-pages branch so they appear on the website.

See circle.yml

justmoon:
Hmm. I think I prefer the IETF approach of treating RFCs as final (except for errata) and introducing new (v2) protocols as a new RFC.

It can be confusing to look at a v1 spec, not realizing there is a v2, but the IETF have addressed that pretty nicely by simply linking to the new RFC, e.g.:

screenshot from 2017-05-11 12-12-38

Now, we could make a rule that you can never have a 2.x.x. You can only have 0.x.x and 1.x.x IL-RFCs. But copying what IETF does makes more sense to me. 0.x.x is an Internet-Draft and does not have an RFC number assigned at all, anyone can create one etc. Once it's approved as an RFC (1.0.0), it can't be changed, except to fix errors or improve the explanation in a non-breaking way. (1.x.x)

Highly recommend also [Rick Hickey's talk on versioning](https://www.youtube.com/watch?v=oyLBGkS5ICk) in this context. It's about software, not protocol specs, but it helped me a lot in understanding why the IETF's approach makes so much sense and why semantic versioning makes none.

adrianhopebailie:
New attempt at this using front-matter on each RFC. This is a nice way to add meta-data that is actually rendered by Github too. Each file therefor has a small front-matter section with attributes.

Example:

```
---
title: Interledger Architecture
draft: 1
deprecated: false
---
```

The publishing scripts will now generate a `draft-x.html` file for each new iteration so this can be linked to if required.

When a draft is considered stable then the draft number is changed to `FINAL`.

The template can now be updated to render differently for a numbered draft (Working Draft) as opposed to a Candidate Specification.

The process discussed via email has also been captured in CONTRIBUTING.md

michielbdejong:
It's hard to fully test the script without pushing the 'draft 1' versions to gh-pages first, but if there's an error we find later, then we can always fix it when we find it.

Current tests fail with 'No title specified for 0021-plugin-rpc-api/0021-plugin-rpc-api.md' but i guess that's easy to fix, and a good thing that it worked! :)

My only remark is that there will be two numbers involved to refer to a draft: its rfc number and its draft number. maybe we'll develop a habit of referring to doc versions as e.g. IL-RFC21.3 or IL-RFC21-draft3, we'll see!

adrianhopebailie:
> Current tests fail with 'No title specified for 0021-plugin-rpc-api/0021-plugin-rpc-api.md' but i guess that's easy to fix, and a good thing that it worked! :)

I'll rebase and push a new commit

> My only remark is that there will be two numbers involved to refer to a draft: its rfc number and its draft number. maybe we'll develop a habit of referring to doc versions as e.g. IL-RFC21.3 or IL-RFC21-draft3, we'll see!

I like RFC XXX draft Y. We'd likely only talk about RFC XXX unless we are referring to an explicit draft

PR #208, opened 2017-05-02
Update quote_response, fix #207

### LINKED ISSUE #207: clarify meaning of `destination_ledger` in quote response
The way I interpret `destination_ledger` in quote response, is that the quote is valid for *any* destination account on that entire ledger, not just for the `destination_account` from the quote request. @adrianhopebailie says (IIUC?) this cannot be generalized in this way, and the quote is just for the destination account in question (might be entirely different for other destination accounts on the same destination ledger)

The [current text](https://github.com/interledger/rfcs/blob/master/0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md#response-message-format-success) isn't very clear about it.

Which one is it?

michielbdejong:


0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md
```diff
@@ -101,13 +103,17 @@ The request must specify either `source_amount` or `destination_amount` but not
   "method": "quote_response",
   "id": "721e4126-98a1-4974-b35a-8a8f4655f934",
   "data": {
-    "source_connector_account": "mark",
```
michielbdejong: The implementation uses mark's full address, not relative to the ledger (is "relative ILP address" even a thing?)

0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md
```diff
@@ -101,13 +103,17 @@ The request must specify either `source_amount` or `destination_amount` but not
...
+    "source_connector_account":"example.eur-ledger.mark",
+    "source_ledger":"example.eur-ledger.",
+    "source_amount":"10025000000",
+    "source_expiry_duration":"6000",
+    "destination_ledger":"example.usd-ledger.",
+    "destination_amount":"9000000000",
+    "destination_expiry_duration":"5",
+    "liquidity_curve":[
+      ["10", "0"],
+      ["100000000000000000", "99799999999999"]
```
michielbdejong: The current implementation actually uses `[100000000000000000, 99799999999999]`, but I guess that's a bug? It should be `["100000000000000000", "99799999999999"]`, right?

michielbdejong:
Specifically, this PR states that the pricing should follow the same logic for all other possible destination accounts that share the same `destination_ledger` prefix.

0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md
adrianhopebailie: > is "relative ILP address" even a thing?

I would assume so BUT we should be careful about mixing that up with ledger specific account identifiers. I'm not sure which one this is?

michielbdejong:
I still fail to see what the distinction is between the ILP docs and the ilp-kit docs. If the behavior of ilp-kit doesn't match what the ILP protocol stack prescribes then that's a bug. The only exception is for optional parts of the stack (e.g. ilp-kit implements SPSP but not IPR).

IMHO, documenting a variation of ILP, C2C, ILQP, or SPSP that doesn't exist on the live network (the public Interledger) only serves to confuse everybody! :)

If the public Interledger didn't exist yet, I would agree, then we can document whatever we want without negative effects; but now that it does exist, I think we should try to document the ILP traffic on there as precisely as possible?

adrianhopebailie:
> I still fail to see what the distinction is between the ILP docs and the ilp-kit docs. If the behavior of ilp-kit doesn't match what the ILP protocol stack prescribes then that's a bug. The only exception is for optional parts of the stack (e.g. ilp-kit implements SPSP but not IPR).

We're using these RFCs to describe protocols at many levels of the stack. Everytime we define one of these protocols we need to trim it down to the bare minimum that is required for interoperability or (like in the case of the higher level protocols) be clear that the protocol we're describing is one of many at that layer.

Connector-to-connector protocols are unusual because they operate "to the side" of the stack in that they are used to exchange information between nodes outside the flow of a payment. The analogy are the [routing protocols](https://en.wikipedia.org/wiki/Routing_protocol) in the IP stack like BGP and RIP which are often referred to as layer management protocols.

What you are describing is a protocol that is used to exchange management information between a specific type of node (ILP-Kit). The data model is expressed in JSON and the document makes references to implementation specifics like `plugin.sendMessage(message)`.

I would consider this protocol to be analogous to an Interior Gateway Protocol whereas I beliee the intent for ILQP is to be more like an Exterior Gateway Protocol.

> IMHO, documenting a variation of ILP, C2C, ILQP, or SPSP that doesn't exist on the live network (the public Interledger) only serves to confuse everybody! :)

Depends who you consider "everybody" 😄 The goal for the ILP RFCs is analogous to IETF Internet Drafts. They are often forward looking or simple (as the name implies) Requests for Comments.

If we document how ILP-Kit works we can be very specific and less concerned about re-usability and standardization. We just document what is there. There is then a process to compare that with any alternatives and document what is common so we come away with a standard and a protocol with no assumptions about implementation specifics.

That's why I don't think we need variations of ILQP, we need to document what exists and give it an appropriate name. "ILP-Kit Management Protocol"?

michielbdejong:
This "ilp-kit management protocol" would feel like an ilp-kit specific fork of ILQP - even if we give it an "internal" name like that, it would feel a bit like WHATWG's fork of HTML. It would just mean there will be one less implementations of the "official" ILQP less relevant, wouldn't it?

Let's try to avoid a fork between ilp-kit and the specs. If there is disagreement between the two, then we should pick a winner, and adapt the other so that they no longer contradict each other.

adrianhopebailie:
> Let's try to avoid a fork between ilp-kit and the specs. If there is disagreement between the two, then we should pick a winner, and adapt the other so that they no longer contradict each other.

My point was that there will likely be many winners 😄 

The management protocol between nodes will be different depending on the type of nodes. Just like there are two flavours of routing protocol in the IP stack.

ILP-Kit is a very light Interledger node. Our expectation (at least mine) is that soon much more sophisticated nodes (the equivalent of autonomous systems on the Internet) will emerge and they will have a highly efficient binary protocol that they use between one another. My assumption has always been that ILQP was intended to be that protocol.

0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md
```diff
@@ -4,17 +4,15 @@ The Interledger Quoting Protocol is a method of getting quote information from a
 
 There are two consumers of the ILQP: sending clients, and other connectors.
 
-The JSON interface for the ILQP is a temporary measure. Eventually, it should be replaced completely by a binary version.
+> The JSON interface for ILQP is experimental reflects what is currently implemented on the live network, therefore it is documented here. Eventually, this protocol should be replaced completely by a binary version but since this is a connector-to-connector protocol many variations may exist in the future and it's possible this version may be retained for some connections.
```
michielbdejong: 'is experimental reflects' -> grammar error?

0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md
```diff
@@ -144,4 +144,4 @@ After getting a quote using ILQP, the client can display the quote to the user a
 
 ## Appendix A: ASN.1 Module
 
-The [InterledgerQuotingProtocol.asn](InterledgerQuotingProtocol.asn) ASN.1 module tentatively describes the binary version of ILQP. This is not final.
+The [InterledgerQuotingProtocol.asn](../asn1/InterledgerQuotingProtocol.asn) ASN.1 module tentatively describes the binary messages for a future version of ILQP. This is not final and is not yet deployed at the time of writing.
```
michielbdejong: we should check if Ivan's RegExp builds this hyperlink correctly

PR #211, opened 2017-05-05
Copy asn1 files to web during publish

adrianhopebailie:


adrianhopebailie:
> did you test that it fixes the hyperlinks?

Worked on my local but will check again when it's deployed.

Note that there are also links from RFC 3 to test

PR #213, opened 2017-05-09
feat: add optional plugin.getInfo().{min,max}Balance

michielbdejong:
Companion to https://github.com/interledgerjs/ilp-connector/pull/337


sharafian:
@michielbdejong what would your thoughts be on making this an optional field of getInfo, instead of an optional function?

momerath42:
Do we want to mention that the broadcast liquidity-curve becoming stale, and being used as the arbiter of route choice, is a temporary situation?

michielbdejong:
Added a line saying "This situation is likely to change in the future"

PR #217, opened 2017-05-11
Fixes some minor typos in the ASN1

adrianhopebailie:


asn1/InterledgerProtocol.asn
```diff
@@ -10,12 +10,9 @@ IMPORTS
     Address,
     Timestamp
     FROM InterledgerTypes
-
-    MetadataEntry
-    FROM MetadataEntry
 ;
 
-InterledgerProtocolPayment ::= SEQUENCE {
+InterledgerProtocolPaymentMessage ::= SEQUENCE {
```
justmoon: We should probably call this `InterledgerProtocolPaymentPacket`, because it's in the memo of a transfer, not sent as a message.

adrianhopebailie: Okay, just need to make sure whatever we choose we update GenericPacket to use the same.

The other instances are all `InterledgerXRequest` and `InterledgerXResponse` but `InterledgerProtocolPaymentRequest` doesn't seem appropriate...

adrianhopebailie: I reverted back to `InterledgerProtocolPayment` in `InterledgerProtocol.asn` and changed `GenericPacket.asn` to match

PR #218, opened 2017-05-11
docs(0008): ILQP binary messages

sentientwaffle:
Depends on https://github.com/interledger/rfcs/pull/198

### DIFF
--- a/0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md
+++ b/0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md
@@ -4,8 +4,6 @@ The Interledger Quoting Protocol is a method of getting quote information from a
 
 There are two consumers of the ILQP: sending clients, and other connectors.
 
-> The JSON interface for ILQP is experimental and reflects what is currently implemented on the live network, therefore it is documented here. Eventually, this protocol should be replaced completely by a binary version but since this is a connector-to-connector protocol many variations may exist in the future and it's possible this version may be retained for some connections. 
-
 ## Background and Terminology
 
 The quoting protocol returns the price of a _hypothetical_ payment through the connector being queried. However, no part of the ILQP initiates or creates an actual payment. All the information in the ILQP is non-binding and advisory. ILQP only calculates and reports the expected cost of a payment.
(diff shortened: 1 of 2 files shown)


0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md
```diff
@@ -33,108 +31,90 @@ The sending and receiving ledger are different ledgers. Otherwise, there should
 
 ## Get Quote
 
-Quotes are requested with `plugin.sendMessage(message)` and returned via `plugin.on("incoming_message")` (see the Ledger Plugin Interface for more details).
-The message formats described here are the `data` properties of an `OutgoingMessage`.
+Quotes are requested with `plugin.sendRequest(message)` and returned via `plugin.registerRequestHandler(fn)` (see the Ledger Plugin Interface for more details).
```
michielbdejong: Why doesn't sendRequest return the response async, or return a promise for it?

0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md
```diff
@@ -33,108 +31,90 @@ The sending and receiving ledger are different ledgers. Otherwise, there should
...
 
-### Request Message Format
+A quote request message's `ilp` property must be an `IlqpLiquidityRequest`, `IlqpBySourceRequest`, or `IlqpByDestinationRequest`. The response message's `ilp` must be an `IlqpLiquidityResponse`, `IlqpBySourceResponse`, or `IlqpByDestinationResponse` respectively. If an error occurs during quoting, `ilp` will be a [`IlpError`](../0003-interledger-protocol/0003-interledger-protocol.md#ilp-error-format) instead.
```
michielbdejong: `ilp` is maybe a confusing property name? Also, now that it's `sendRequest` instead of `sendMessage`, I would rename 'quote request message' to maybe 'quote request', maybe?

michielbdejong: The response message's `ilp` must be -> The response message's `ilp` property must be

And nicer to use them same order (liquidity, source, dest) a in the previous sentence

0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md
```diff
@@ -33,108 +31,90 @@ The sending and receiving ledger are different ledgers. Otherwise, there should
...
-| `method` | String          | `"quote_request"` |
-| `id`     | String          | A UUID to match this request to the corresponding response. |
-| `data`   | QuoteParameters | An Object describing the quote parameters. |
+## ILQP Packets
+### IlqpLiquidityRequest
 
-##### QuoteParameters
+| Field | Type | Short Description |
+|:--|:--|:--|
+| destinationAccount | Address | Address corresponding to the destination account |
```
michielbdejong: wouldn't it make more sense to allow any `destinationPrefix`, whether account or ledger?

sentientwaffle: If you request a `destinationPrefix`, the connector could reply that there is no single curve that corresponds to the entire curve.

Instead, requesting a `destinationAccount` means that the connector can select to appropriate `appliesToPrefix`.

0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md
```diff
@@ -33,108 +31,90 @@ The sending and receiving ledger are different ledgers. Otherwise, there should
 
 ## Get Quote
 
-Quotes are requested with `plugin.sendMessage(message)` and returned via `plugin.on("incoming_message")` (see the Ledger Plugin Interface for more details).
-The message formats described here are the `data` properties of an `OutgoingMessage`.
+Quotes are requested with `plugin.sendRequest(message)`, which returns the quote response (see the Ledger Plugin Interface for more details).
```
michielbdejong: technically, it returns *a Promise for* the quote reponse

michielbdejong: also, if wouldn't the equivalent of `sendMessage(message)` be `sendRequest(request)` or  `sendRequest(quoteRequest)`?

0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md
```diff
@@ -33,108 +31,90 @@ The sending and receiving ledger are different ledgers. Otherwise, there should
...
 
-### Request Message Format
+A quote request's `ilp` property must be an `IlqpLiquidityRequest`, `IlqpBySourceRequest`, or `IlqpByDestinationRequest`. The response's `ilp` property must be an `IlqpLiquidityResponse`, `IlqpBySourceResponse`, or `IlqpByDestinationResponse` respectively. If an error occurs during quoting, `ilp` will be a [`IlpError`](../0003-interledger-protocol/0003-interledger-protocol.md#ilp-error-format) instead.
```
michielbdejong: Wouldn't `quoteType` / `responseType` be a more appropriate name for this property than `ilp`?

0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md
michielbdejong: ah ok, makes sense

0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md
```diff
@@ -33,108 +31,90 @@ The sending and receiving ledger are different ledgers. Otherwise, there should
...
-The request must specify either `source_amount` or `destination_amount` but not both.
+| Field | Type | Short Description |
+|:--|:--|:--|
+| liquidity | LiquidityCurve | Curve describing the liquidity for the quoted route |
+| appliesToPrefix | Address | Common prefix of all addresses for which this liquidity curve applies. |
+| sourceHoldDuration | UInt32 | How long the sender should put the money on hold (in milliseconds) |
+| expiresAt | Timestamp | Maximum time where the connector expects to be able to honor this liquidity curve |
 
-##### Example request
+`LiquidityCurve` is a `SEQUENCE OF SEQUENCE { x UInt64, y UInt64 }`
```
michielbdejong: An example would be useful here; "for example, `[ [0, 0], [10, 265] ]`"

0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md
```diff
@@ -33,108 +31,90 @@ The sending and receiving ledger are different ledgers. Otherwise, there should
...
-  }
-}
-```
+### IlqpBySourceRequest
+
+| Field | Type | Short Description |
+|:--|:--|:--|
+| destinationAccount | Address | Address corresponding to the destination account |
+| sourceAmount | UInt64 | Amount the sender needs to send, denominated in the asset of the source ledger |
+| destinationHoldDuration | UInt32 | How much time the receiver needs to fulfill the payment (in milliseconds) |
```
michielbdejong: so amount now goes source->dest, but hold duration still goes dest->source? why is this?

0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md
```diff
@@ -33,108 +31,90 @@ The sending and receiving ledger are different ledgers. Otherwise, there should
...
-    "destination_expiry_duration":"5",
-    "liquidity_curve":[
-      ["10", "0"],
-      ["100000000000000000", "99799999999999"]
-    ]}
-  }
+  "ledger": "example.eur-ledger.",
+  "from": "example.eur-ledger.alice",
+  "to": "example.usd-ledger.bob",
+  "ilp": "BB8RZXhhbXBsZS5uZXh1cy5ib2IAAAACGHEaAAAAC7gA"
```
michielbdejong: wait, what happened here? This doesn't match any of the three formats described above?

0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md
```diff
@@ -33,108 +31,90 @@ The sending and receiving ledger are different ledgers. Otherwise, there should
...
-| `id`      | String | The type of error. |
-| `message` | String | Additional details about the nature of the error. |
+```js
+{
+  "ledger": "example.eur-ledger.",
+  "from": "example.usd-ledger.bob",
+  "to": "example.eur-ledger.alice",
+  "ilp": "BQ0AAAACGHEaAAAAC7gA"
+}
+```
```
michielbdejong: same question

michielbdejong:
Where are codes like https://github.com/interledgerjs/ilp-packet/blob/master/index.ts#L10 documented? Shouldn't that be in this doc as well, or is that covered by the ASN1 doc?

0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md
sentientwaffle: The `request` and `response` are both [`Message`](https://github.com/interledger/rfcs/blob/e4dab0259283a500f923f92ca4f50b18509cdc44/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md#class-message)s.

0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md
sentientwaffle: It is called `ilp` because `Transfer`s have the same [property](https://github.com/interledger/rfcs/blob/master/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md#ilp) for their ILP packet.

0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md
sentientwaffle: Its binary data, so an example wouldn't be that helpful.

0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md
michielbdejong: well, by analogy, shouldn't it be called `ilqp`, then?

0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md
michielbdejong: The fact that you had to point that out to me proves even more how important an example is here. :) So then use either JavaScript or pseudo-code, and put something like:

"this is in binary format, so for example if sending 0 delivers 0, sending 10 delivers 265, and the range between `[0, 0]` and `[10, 265]` increases linearly, then LiquidityCurve would have the value:

`"Fy534JNJ6A/dF=="` which is `Base64('4') + Base64('0') + Base64('0') + Base64('10') + Base64('265')` (the initial '4' is the number of bytes that follows, see ... for more details about binary IL(Q)P packets.

0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md
sentientwaffle: [Message#ilp](https://github.com/interledger/rfcs/blob/e4dab0259283a500f923f92ca4f50b18509cdc44/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md#messageilp) could refer to any ILP packet. Since this is the ILQP spec, I only mention the ones applicable to quoting.

0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md
sentientwaffle: I'm going to remove these request/response message examples, I don't think they add anything. Instead I'll reference [interledgerjs/ilp-packet](https://github.com/interledgerjs/ilp-packet/).

sentientwaffle:
@michielbdejong they are documented in [RFC-3](https://github.com/interledger/rfcs/blob/master/0003-interledger-protocol/0003-interledger-protocol.md#header-type-registry).

0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md
```diff
@@ -33,115 +31,68 @@ The sending and receiving ledger are different ledgers. Otherwise, there should
...
-{
-  "method": "error",
-  "id": "721e4126-98a1-4974-b35a-8a8f4655f934",
-  "data": {
-    "id": "InvalidBodyError",
-    "message": "Missing required parameter: source_address"
-  }
-}
-```
+Quotes are requested with `plugin.sendRequest(message)`, which returns a `Promise` to the quote response (see the Ledger Plugin Interface for more details).
```
emschwartz: I'm not sure this should be included in this spec, since how quotes are requested is specific to the ledger plugin interface in the implementation you're using. Maybe just put it in a more generic way that quotes are sent through a request/response mechanism exposed by the ledger plugins or ledger layer.

PR #231, opened 2017-06-30
Add Hashed-Timelock Agreement (HTLA) doc

emschwartz:


0000-hashed-timelock-agreements.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Hashed-Timelock Agreements (HTLAs)
+
+A contract is a certain type of agreement that is enforced by a third party. Hashed-Timelock Agreements (HTLAs) generalize the idea of HTLCs to include both agreements that are and are not enforced by a ledger.
+
+This mechanism enables secure Interledger payments even through ledgers that do not support conditional transfers.
+
+## HTLAs Without Ledger Support
+
+Two parties using an HTLA over a ledger that does not support hashlocks and timeouts could proceed in the following manner. The sender would send a message to the recipient telling them that they want to "prepare" a transfer with a given hashlock and timeout. The parties agree that if the recipient presents the preimage of the hash before the timeout, the transfer is executed and the sender owes the recipient the money. Debts are settled using simple transfers on the ledger.
```
mDuo13: Who decides whether the hash was presented before the timeout? Do the participants need to "sync their watches" in advance? If one side says it was sent but the other side says it wasn't received, how do they resolve the contradiction?

EDIT: I see this is discussed to some extent in the later sections. Probably still worth clarifying how the system arbitrates the issue of disagreement about time.

0000-hashed-timelock-agreements.md
```diff
@@ -0,0 +1,116 @@
...
+
+Using conditional payment channels, the sender and recipient can transact without any funds being at risk, because all disputes will be mediated by the ledger. This can be used as a mechanism to enable a greater volume of payments "through" a ledger than the ledger can natively support.
+
+***Note on ledger speed:** For a number of reasons, it is important that Interledger payments execute or fail quickly. With conditional payment channels, the recipient relies on the fact that they can redeem their funds even if their is a dispute with the sender by presenting their claims and the hashlock preimages to the ledger. However, this means that the transfer timeout must account for the ledger processing time, in case the recipient must resort to submitting the claims to the ledger. Thus conditional payment channels only work well with Interledger if the ledger can process claims in a few seconds or less.
```
mDuo13: Typo: "their is a dispute"→"there is a dispute"

mDuo13: Also, mismatched number of `*` characters here

0000-hashed-timelock-agreements.md
```diff
@@ -0,0 +1,116 @@
...
+
+When the transfer is "prepared", the sender's funds are put on hold by the ledger, pending the fulfillment of a predefined condition. The condition is a hashlock, or the digest of a cryptographic hash function, such as SHA-256 in Lightning and Interledger. The "contract" stipulates that the recipient may claim the funds by presenting a valid preimage of the hash digest before a given timeout. After the timeout, the funds are automatically returned to the sender. This is the Hashed-Timelock Contract.
+
+HTLCs are enforced by the ledger, so the parties transacting only need to trust the ledger to correctly execute the contract. However, this mechanism requires support from the ledger and only works with ledgers that implement hashlocks and timeouts.
+
+Interledger is designed to work with all ledgers, so it must support both ledgers with and without support for hashlocks and timeouts.
+
+## Hashed-Timelock Agreements (HTLAs)
+
+A contract is a certain type of agreement that is enforced by a third party. Hashed-Timelock Agreements (HTLAs) generalize the idea of HTLCs to include both agreements that are and are not enforced by a ledger.
```
adrianhopebailie: nit: s/enforced/usually enforceable/

0000-hashed-timelock-agreements.md
```diff
@@ -0,0 +1,116 @@
...
+
+In the case of a post-funded agreement, the sender can settle their debt with the recipient either for every payment, if their ledger is fast and has low fees, or once the total amount owed reaches the parties' trust/credit limit. In this case, the recipient must trust the sender to pay what they owe. Recipients limit their risk by capping the total amount of the transfers they are willing to accept from the sender that have not been paid for.
+
+For pre-funded agreements, the sender transfers either the amount of an individual payment or a bulk amount to the recipient. The recipient then deducts the amount of each "conditional transfer" executed from the pre-funded amount. In this case, the sender must trust the recipient not to steal the pre-funded amount, but this risk can obviously be limited by capping the pre-funded amount.
+
+HTLAs thus work with ledgers that support hashlocks and timeouts (as HTLCs) and with ledgers that do not.
+
+## Spectrum of HTLA Types
+
+The various types of HTLAs present a tradeoff between ledger support required and trust required between peers connecting through that ledger.
```
adrianhopebailie: Some commas would make this easier to read

0000-hashed-timelock-agreements.md
```diff
@@ -0,0 +1,116 @@
...
+
+This mechanism enables secure Interledger payments even through ledgers that do not support conditional transfers.
+
+## HTLAs Without Ledger Support
+
+Two parties using an HTLA over a ledger that does not support hashlocks and timeouts could proceed in the following manner. The sender would send a message to the recipient telling them that they want to "prepare" a transfer with a given hashlock and timeout. The parties agree that if the recipient presents the preimage of the hash before the timeout, the transfer is executed and the sender owes the recipient the money. Debts are settled using simple transfers on the ledger.
+
+In the case of a post-funded agreement, the sender can settle their debt with the recipient either for every payment, if their ledger is fast and has low fees, or once the total amount owed reaches the parties' trust/credit limit. In this case, the recipient must trust the sender to pay what they owe. Recipients limit their risk by capping the total amount of the transfers they are willing to accept from the sender that have not been paid for.
```
adrianhopebailie: > Recipients limit their risk by capping the total amount of the transfers they are willing to accept from the sender that have not been paid for.

Long sentence that requires a few re-reads to understand.

0000-hashed-timelock-agreements.md
```diff
@@ -0,0 +1,116 @@
+# Hashed-Timelock Agreements (HTLAs)
+
+> Generalization of [Hashed-Timelock Contracts (HTLCs)](https://en.bitcoin.it/wiki/Hashed_Timelock_Contracts) used to secure Interledger payments.
+
+Interledger uses conditional transfers to secure multi-hop payments. This document describes how Interledger payments can be implemented over any type of ledger, including those that do not provide support for conditional transfers. Hashlocks and timeouts can be enforced by ledgers or they can be upheld by senders and recipients themselves.
+
+Before reading this document, it is important to understand how Interledger isolates participants from risks posed by indirect connections. The tradeoffs listed here are only the concern of parties that are _directly connected_. For more on the security model, see the [Interledger Architecture](./0001-interledger-architecture/0001-interledger-architecture.md).
```
emschwartz: Would it be worth including a section in this doc about how HTLAs enable secure multi-hop payments? Or is it sufficient to point to other Interledger docs?

0000-hashed-timelock-agreements.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Conclusion: Integrating All the Ledgers
+
+Interledger provides secure multi-hop payments over any type of ledger by using bilateral Hashed-Timelock Agreements (HTLAs). This works with public and private blockchains, centralized ledgers, payment channels, and even with cash or cases where there is no ledger. There is a spectrum of HTLA types that require varying levels of ledger functionality and bilateral trust. However, because Interledger isolates participants from risk, these options and tradeoffs are only the concern of individual participants and their direct connections. No one is affected by the choices of indirect connections, so all are free to select their ledgers and connectors and use the type of agreement that suits them.
+
+_Internet "hat tip": the importance of being able to integrate **any** type of network was demonstrated by [RFC 1149: A Standard for the Transmission of IP Datagrams on Avian Carriers](https://tools.ietf.org/html/rfc1149)._
+
```
emschwartz: Would it be worth including an appendix that mentions additional integration options like Notarized Payment Channels or 3rd Party Escrow? The former isn't yet implemented anywhere and the latter isn't an ideal option so I left them out, but they are also possibilities.

0000-hashed-timelock-agreements.md
adrianhopebailie: I think it's good as is

0000-hashed-timelock-agreements.md
adrianhopebailie: I think we need a whole RFC explaining how you can upgrade to Notarized channels and atomic mode.

emschwartz: Notarized channels are actually completely separate from atomic mode

0000-hashed-timelock-agreements.md
adrianhopebailie: Okay, it's not clear that you're talking about ledgers. It sounds like you mean contracts in general.

0000-hashed-timelock-agreements.md
adrianhopebailie: Hmmm, then maybe some mention isn't a bad idea. Add as an appendix in a subsequent iteration

0000-hashed-timelock-agreements.md
```diff
@@ -0,0 +1,120 @@
...
+HTLCs are enforced by the ledger, so the parties transacting only need to trust the ledger to correctly execute the contract. However, this mechanism requires support from the ledger and only works with ledgers that implement hashlocks and timeouts.
+
+Interledger is designed to work with all ledgers, so it must support both ledgers with and without support for hashlocks and timeouts.
+
+## Hashed-Timelock Agreements (HTLAs)
+
+A contract is a certain type of agreement that is enforced by a third party. Hashed-Timelock Agreements (HTLAs) generalize the idea of HTLCs to include both agreements that are and are not enforced by a ledger. The concept was originally proposed in [this thread on the Interledger mailing list](https://lists.w3.org/Archives/Public/public-interledger/2017Jun/0009.html).
```
sharafian: > generalize the idea of HTLCs to include both agreements that are and are not enforced by a ledger.

I think it's clearer to say "to include agreements that are not enforced by a ledger"

0000-hashed-timelock-agreements.md
```diff
@@ -0,0 +1,120 @@
...
+
+&ast; **Note on ledger speed:** For a number of reasons, it is important that Interledger payments execute or fail quickly. With conditional payment channels, the recipient relies on the fact that they can redeem their funds even if there is a dispute with the sender by presenting their claims and the hashlock preimages to the ledger. However, this means that the transfer timeout must account for the ledger processing time, in case the recipient must resort to submitting the claims to the ledger. Thus conditional payment channels only work well with Interledger if the ledger can process claims in a few seconds or less.
+
+### On-Ledger Holds/Escrow (using HTLCs)
+
+* **Ledger Features Required:** Transfers with HTLCs
+* **Non-Functional Ledger Requirements:** Fast, Low Fees, High Throughput
+* **Money at Risk:** None
+* **Examples:** [Ethereum Escrow Contract](https://github.com/interledgerjs/ilp-plugin-ethereum), [XRP Escrow](https://ripple.com/build/transactions/#escrowcreate)
```
sharafian: You could link the actual solidity file here

emschwartz: Thought about it but I'd rather not in case the filename changes

0000-hashed-timelock-agreements.md
emschwartz: @adrianhopebailie @sharafian I added the appendix. Let me know what you think of it

0000-hashed-timelock-agreements.md
```diff
@@ -0,0 +1,120 @@
...
+
+Using conditional payment channels, the sender and recipient can transact without any funds being at risk, because all disputes will be mediated by the ledger. This can be used as a mechanism to enable a greater volume of payments "through" a ledger than the ledger can natively support.
+
+&ast; **Note on ledger speed:** For a number of reasons, it is important that Interledger payments execute or fail quickly. With conditional payment channels, the recipient relies on the fact that they can redeem their funds even if there is a dispute with the sender by presenting their claims and the hashlock preimages to the ledger. However, this means that the transfer timeout must account for the ledger processing time, in case the recipient must resort to submitting the claims to the ledger. Thus conditional payment channels only work well with Interledger if the ledger can process claims in a few seconds or less.
+
```
michielbdejong: "conditional payment channels only work well with Interledger if the ledger can process claims in a few seconds or less" - that seems contradictory, knowing lightning uses them on top a very slow ledger.

I would turn the statement around, and say the payment timeout needs to be at least the ledger slowness, times the number of hops (so in the case of lightning, never less than 10 minutes * numHops). In practice, it's in everybody's interest to either accept or reject a payment much faster than this worst-case limit.

0000-hashed-timelock-agreements.md
```diff
@@ -0,0 +1,120 @@
...
+To use a simple payment channel in an HTLA, the sender "prepares" the conditional transfer simply by sending a message to the recipient (including the hashlock and timeout). If the recipient presents the hash preimage before the timeout, the sender sends a new signed claim to the recipient to cover the total amount of the transfers sent thus far.
+
+The agreement between the sender and recipient dictates how disputes are handled, including cases in which the recipient thinks they submitted the preimage in time and the sender thinks otherwise. Both parties should enforce their own view of the timeout, rather than deferring to the other party.
+
+In this model, the recipient must trust the sender for the total amount of the prepared or executed transfers that is not yet covered by payment channel claims. The recipient limits their risk by capping the amount of the transfers they are willing to accept from the sender without the corresponding claims. The risk could be flipped from the recipient to the sender by sending the claims along with the messages to prepare transfers.
+
```
michielbdejong: maybe clarify: "(pre-funded instead of post-funded)"

0000-hashed-timelock-agreements.md
```diff
@@ -0,0 +1,120 @@
...
+The functionality necessary for simple payment channels is present in all major blockchains today, including Bitcoin (even without [SegWit](https://en.bitcoin.it/wiki/Segregated_Witness)), Ethereum, XRP, Zcash, and Chain.
+
```
michielbdejong: in nearly all major blockchains; notable exception is Monero

0021-hashed-timelock-agreements/0021-hashed-timelock-agreements.md
```diff
@@ -0,0 +1,139 @@
...
+
+Interledger provides secure multi-hop payments over any type of ledger by using bilateral Hashed-Timelock Agreements (HTLAs). This works with public and private blockchains, centralized ledgers, payment channels, and even with cash or cases where there is no ledger. There is a spectrum of HTLA types that require varying levels of ledger functionality and bilateral trust. However, because Interledger isolates participants from risk, these options and tradeoffs are only the concern of individual participants and their direct connections. No one is affected by the choices of indirect connections, so all are free to select their ledgers and connectors and use the type of agreement that suits them.
+
+_Internet "hat tip": the importance of being able to integrate **any** type of network was demonstrated by [RFC 1149: A Standard for the Transmission of IP Datagrams on Avian Carriers](https://tools.ietf.org/html/rfc1149)._
+
+## Appendix: Additional HTLA Types
+
+This appendix includes additional HTLA types that are possible but were not mentioned above either because they are less desirable or have yet to be implemented.
+
+### Third Party Escrow
```
sharafian: What's the difference between using a third party escrow provider and just peering with a connector you trust? If there is a difference, it might be worth explaining how the risk is different.

emschwartz: Fair point. I'll make a note about that

0021-hashed-timelock-agreements/0021-hashed-timelock-agreements.md
```diff
@@ -0,0 +1,139 @@
...
+
+To mitigate the risk of timeout disputes, the sender and recipient could agree to use a third party notary to act as the timekeeper. The sender prepares the transfer by submitting the details to the notary and the recipient (unless the notary also acts as a message broker). The recipient submits the preimage to the notary, who enforces the timeout. The notary signs a statement attesting to whether the preimage was recieved in time and gives the signed message to both the sender and recipient.
+
+Third-party notaries could be used with simple or more complex payment channels. In the simple type, the participants would agree that the notary has the final say about the timeout. In the more complex variety, the payment channels could be constructed such that the channel updates are conditional upon the signed statement from the notary. When using third-party notaries the sender and recipient must trust the notary not to be colluding with one of the parties and to not sign two conflicting statements.
```
sharafian: When you say simple and complex here, what do you mean? In a trustline the notary makes no difference in the trust, but in even an unconditional payment channel where the notary has the authority to make claims on either party's behalf, the risk of defaulting on an in-flight payment goes away (or rather, it is replaced with the risk of notary collusion/failure). I don't think you actually need "complex" channels, unless I'm misunderstanding what you mean by that

emschwartz: With simple channels you're basically just using the notary for coordination (which admittedly doesn't provide that much benefit). The only type where the notary's signature would actually make a difference is in a complex/"conditional" payment channel but one in which the condition is some multi-sig thing instead of a hashlock.

sharafian: What if we use a simple payment channel like bitcoin paychan, but the claim signature is the notary's instead of the sender? Or in the ripple case, set the `PublicKey` to the notary's public key instead of the sender.

emschwartz: That's a cool idea. I think we might want to add that as a separate type

mDuo13:
I like the edits made so far.

0021-hashed-timelock-agreements/0021-hashed-timelock-agreements.md
```diff
@@ -139,10 +139,11 @@ Third-party notaries could be used with simple or more complex payment channels.
...
+Falling in-between the guarantees of Third Party Escrow and Notarized Payment Channels, a "Third Party Payment Channel" involves escrowing money in a payment channel and giving a third party authority to create claims. To set up the payment channel, the sender funds a temporary account on the ledger as in a Simple Payment Channel. However, the third party's key is authorized to create claims on the channel balance instead of using the sender's public key.
 
-The receiver doesn't have to trust the sender to give them a claim after they present the fulfillment; the trust is instead placed in the third party. This means that the receiver doesn't have to trust the sender at all, and they only need to trust the third party for in-flight money.
+When the sender prepares a transfer, they send a message to the recipient including the hashlock and timeout. To execute the transfer, the recipient submits the preimage to the third party. If the third party gets the preimage before the timeout, they pay the recipient from the sender's funds by signing a new claim.
```
sharafian: Actually the third party is required to act like a message broker; if the receiver acts on just a message from the sender then they don't know that the sender told the third party about it. The receiver could later submit the fulfillment to the third party, only to hear that it was never even prepared.

sharafian: I guess the sender could also give their signature at prepare time and the receiver could show that to the third party when they fulfill

zmanian:
Several thoughts on improvements to the document.

I would recommend thinking of HTLA support as a route level property vs a ledger level property. It is a good question what it means for specific route to support an HTLA. Does it mean all the intermediate ledgers support an HTLA or a subset of legs? Being able to ask the routing layer for an HTLA supporting route is a pretty powerful property.

One of the most powerful properties of an HTLA is allow a computing environment that support business process management to atomically integrate with a settlement system. This allows actions like release of liability to be conditional on the execution of an ILP payment. I think we should write up something in greater detail.

emschwartz:
@zmanian This might need some more explanation and emphasis but it's actually a really important that the type of HTLA used is _not_ a route-level property. It is only a decision on a per-hop/per-ledger basis. Interledger enables a single payment to include HTLAs all across this spectrum and every party gets the security guarantees of their chosen agreement.

The way to think about it is if you have a certain HTLA agreement with me, you don't care about my agreements with other people. As long as our agreement is upheld (i.e. you prepare transfers to me, I send you preimages, you give me money for the transfers), the other agreements don't concern you at all. Either the preimage gets to you before the timeout, in which case I get the money, or it doesn't, in which case the money comes back to you.

Does that make sense? I'll try to add some more on that point to the doc tomorrow.

> business process management to atomically integrate with a settlement system

This is indeed one of the very cool implications of Interledger. Part of what led to us starting to work on Interledger was wanting a standard way to interface with the financial system for Codius smart contracts (automated business processes that could be decentralized or not, depending on the use case requirements). Might also want to have a chat with @BobWay about circular Interledger payments and @diminator about Interledger for non-financial assets.

zmanian:
With only ledger level HTLAs, implementing a payment system atomically linked to another business logic system would require breaking part of the Interledger abstractions.

If at Alice at ABC bank wants to Bob at XYZ bank as soon a title is transfered. Alice and Bob need at least 1 shared ledger which can be transversed by an ILP transaction to establish the HTLA.

emschwartz:
@zmanian I'm not sure I totally understand your point but I don't think any abstractions are being broken.

Alice has some HTLA with the connector that allows her to "transfer" 1 unit on ABC Bank's ledger. The connector has a different HTLA with Bob that allows them to "transfer" 1 unit on XYZ Bank's ledger. Neither Alice nor Bob care what counts as a transfer for the other one as long as they consider it paid. The fact that both HTLAs use the same hashlock makes the two transfers atomic(ish - Universal style, not properly Atomic).

To do a circular payment where you're exchanging money for the title to something, you'd set it up like:
```
Alice --(ABC Bank)--> Connector --(XYZ Bank)--> Bob --(Title "Ledger")--> Alice
```
Each `--(...)-->` represents a completely different type of HTLA, but all of them use the same hashlock. Each party acts as a connector and passes back the preimage once it is produced by Alice (in circular payments the receiver is also the sender).

HTLAs let you abstract away what it means to "get paid". That's just an agreement between two parties and no one else cares how that is structured as long as each individual is happy with the agreements they're directly a part of. Since all of the hashlocks will be the same, either all of the transfers will happen or none of them will (barring the connector fulfillment failure).

zmanian:
@emschwartz The way you described it makes sense to me but that format doesn't really come through in the doc. Maybe adding some of that description makes sense.

0022-hashed-timelock-agreements/0022-hashed-timelock-agreements.md
```diff
@@ -54,21 +54,36 @@ For more details on HTLAs without ledger support see the sections below on [Simp
...
 
-Each `--(...)-->` represents a different type of HTLA, but all of them use the same hashlock. Each connector forwards the payment, preparing funds with their peer, and passes back the preimage once it is produced by Bob.
+Each `--(...)-->` represents a different type of HTLA, but all of them use the same hashlock.
+
+The Interledger protocol flow (detailed below) ensures that each participant only needs to care about the HTLAs in which they are directly involved. Failure in a different part of the path -- or the selection of a seemingly unwise HTLA type -- does not affect any other agreements.
 
-This setup ensures that Alice only needs to care about whether her ledger executes the HTLC faithfully. If so, her funds will only be transferred once Bob has been paid.
+#### Interledger Protocol payment flow:
+
+1. `Alice` and `Bob` agree on the hashlock `H`. The preimage `P` is only known to `Bob`.
```
sharafian: I think each of these steps should include a link to the corresponding HTLA description

0022-hashed-timelock-agreements/0022-hashed-timelock-agreements.md
```diff
@@ -54,21 +54,36 @@ For more details on HTLAs without ledger support see the sections below on [Simp
...
+2. `Alice` prepares the transfer to `Connector 1` by creating and funding an HTLC on `Blockchain A` with hashlock `H`.
+3. `Connector 1` prepares a transfer to `Connector 2` via their shared payment channel, also using hashlock `H`.
+4. `Connector 2` prepares a transfer to `Bob` on their shared trustline using Hashlock `H`.
+5. If `Bob` produces the preimage `P` before the transfer timeout, `Connector 2` will "pay" `Bob` by increasing his balance on their trustline.
```
sharafian: This should go through the whole chain of the success case, and then have a separate list of steps describing how the failure case doesn't affect anyone except the failing link

0022-hashed-timelock-agreements/0022-hashed-timelock-agreements.md
```diff
@@ -54,21 +54,36 @@ For more details on HTLAs without ledger support see the sections below on [Simp
...
+6. If `Connector 2` sends `P` to `Connector 1` before their transfer times out, `Connector 1` will send a signed claim to pay `Connector 2`. (Note that whether or not `Connector 1` actually does this is irrelevant to the other parties in the path. It only matters to `Connector 2` whether `Connector 1` upholds their end of the agreement. The other agreements will be executed normally in either case.)
```
sharafian: It's worth saying more explicitly that "if Connector 1 refuses to give a claim to Connector 2, the payment from Connector 2 to Bob cannot be undone, and Connector 1 can still submit the preimage to Blockchain A. The risk was known to Connector 2, and **no other nodes** are affected by it."

0022-hashed-timelock-agreements/0022-hashed-timelock-agreements.md
emschwartz: I think this is overkill for this document

emschwartz:
Thanks @sharafian @mDuo13 @sappenin @zmanian @michielbdejong @adrianhopebailie @BitfuryLightning for the input!

PR #237, opened 2017-07-10
Five Bells Ledger API

mDuo13:
Replaces PR #125 and PR #90. In keeping with our current thinking, an updated "Common Ledger API" (with all our desired cosmetic / "nice-to-have" changes) may not be needed in the long run, but a formalization of the Five Bells Ledger API currently in use is helpful for those who want to build ledger interfaces that are compatible with [the existing ledger plugin](https://github.com/interledgerjs/ilp-plugin-bells).


0012-five-bells-ledger-api/0012-five-bells-ledger-api.md
```diff
@@ -0,0 +1,1290 @@
...
+
+The Five Bells Ledger API defines one way to implement sufficient functionality to operate a ledger. It is not intended to be an optimal or ideal ledger, especially since such a thing is subjective or at least hard to quantify. (Indeed, Interledger itself was conceived from the conclusion that no one ledger can serve the needs of every person in existence.) In general, this specification reflects to the existing [Five Bells Ledger reference implementation](https://github.com/interledgerjs/five-bells-ledger) so that others may implement compatible software.
+
+
+### Authorization and Authentication
+
+The ledger MUST authenticate API clients. HTTP(S) methods MUST use one of the following systems for authentication:
+
+- [HTTP Basic Auth](https://tools.ietf.org/html/rfc2617#section-2)
+- [TLS Client Certificates](https://tools.ietf.org/html/rfc5246#section-7.4.6)
```
emschwartz: Actually @dappelt recently recommended switching from HTTP Basic Auth on all endpoints to using token-based authentication on all endpoints and using the single auth_token endpoint with Basic Auth. @dappelt want to elaborate more on that?

0012-five-bells-ledger-api/0012-five-bells-ledger-api.md
```diff
@@ -0,0 +1,1290 @@
...
+
+
+A transfer object contains the fields from the following table. Some fields are _Ledger-provided_, meaning they cannot be set directly by clients. Fields that are "Optional" or "Ledger-provided" in the following table may be omitted by clients submitting transfer objects to the API, but **those fields are not optional to implement**. All fields, including the `memo` fields, must be implemented for the Five Bells Ledger API to work properly. Fields of nested objects are indicated with a dot (`.`) character; no field names contain a dot literal.
+
+| Name                   | Type                 | Description                  |
+|:-----------------------|:---------------------|:-----------------------------|
+| `id`                   | [UUID][]             | Client-provided UUID for this resource. MUST be unique within the ledger. The path to the [Get Transfer][] method includes this identifier. See [Transfer IDs][] for more information. |
```
emschwartz: I think in the API the ids are actually full URLs with UUIDs at the end

0012-five-bells-ledger-api/0012-five-bells-ledger-api.md
```diff
@@ -0,0 +1,1290 @@
...
+
+| Field        | Type    | Description                                         |
+|:-------------|:--------|:----------------------------------------------------|
+| `account`    | [URL][] | An identifier for the account to credit or debit. This MUST be an HTTP(S) URL where the client can call the [Get Account][] method. |
+| `amount`     | String  | Positive decimal amount of money to debit from or credit to this account. See [Amounts][] for formatting rules. |
+| `memo`       | Object  | _(Optional)_ Arbitrary object with additional information about this credit or debit.  |
+| `authorized` | Boolean | _(Optional, debit objects only)_ Whether this account has authorized this transfer. The ledger MUST NOT allow a request to set this value to `true` unless the client is authenticated as an Administrator or as the owner of this account. |
+
+**Note:** Ledgers MUST implement the `memo` field. ILP clients use this field to store the ILP Packet object. Specifically, the reference ILP Client puts the ILP Packet as a base64url-encoded string in the the `ilp` field of the `memo` of the credit to the Connector.
```
emschwartz: Might be worth specifying a recommended length

0012-five-bells-ledger-api/0012-five-bells-ledger-api.md
```diff
@@ -0,0 +1,1290 @@
...
+
+#### Transfer States
+[Transfer States]: #transfer-states
+
+The following state diagram shows the possible transitions between a transfer's `state` values:
+
+[![Transfer state diagram](5BL-transfer-states.png)](5BL-transfer-states.png)
+
+The [Prepare Transfer][] method creates a new transfer. Unless creator of the transfer is the only account to be debited and the account sets the `authorized` field of the debit to `true`, the transfer is created in the `proposed` state.
```
emschwartz: It's not really recommended to use the proposed state (unless you have multiple credits or debits, in which case you need to use it) because it means that unauthenticated users can create load on your database. (@dappelt)

0012-five-bells-ledger-api/0012-five-bells-ledger-api.md
```diff
@@ -0,0 +1,1290 @@
...
+
+| Name                      | Type    | Description                            |
+|:--------------------------|:--------|:---------------------------------------|
+| `name`                    | String  | Name of the account. MUST be unique per account and MUST match the regular expression `^[a-zA-Z0-9._~-]{1,256}$`. After an account is created, this field is immutable. |
+| `fingerprint`             | String  | _(Optional)_ A fingerprint of the account owner's client certificate. This field MUST be available if and only if the ledger supports client certificate authentication. MUST match the regular expression `^[0-9A-Fa-f]{2}(:[0-9A-Fa-f]{2}){19,127}$`. A ledger MAY allow the account owner to set this value. |
+| `ledger`                  | [URL][] | _(Optional)_ The path to the ledger containing this account. MUST be an HTTP(S) URL where the client can [Get Ledger Metadata][]. (When an account is created, the ledger automatically provides this value.) |
```
emschwartz: The Five Bells Ledger Plugin depends on this being there because you login with just your account URL. The fact that it's listed as optional is a bit confusing

0012-five-bells-ledger-api/0012-five-bells-ledger-api.md
```diff
@@ -0,0 +1,1290 @@
...
+
+Conditions are distributable event descriptions, and fulfillments are cryptographically verifiable messages that prove an event occurred. If you transmit a fulfillment, then everyone who has the corresponding condition can agree that the condition has been met.
+
+Crypto-conditions control the execution of conditional transfers. The Five Bells Ledger API supports conditions and fulfillments in string format.
+
+The Crypto-Conditions specification anticipates that it will need to expand to keep up with changes in the field of cryptography, so conditions always define which rules and algorithms are necessary to verify the fulfillment. Implementations can use the condition's feature list to determine if they can properly process the fulfillment, without having seen the fulfillment itself.
+
+Example condition in string format:
+
+    cc:0:3:dB-8fb14MdO75Brp_Pvh4d7ganckilrRl13RS_UmrXA:66
```
emschwartz: They now use Crypto Conditions v2, which looks different

0012-five-bells-ledger-api/0012-five-bells-ledger-api.md
```diff
@@ -0,0 +1,1290 @@
...
+##### Response Format
+
+A successful result uses the HTTP response code **200 OK** and contains a JSON object with the following fields. (A field name like `foo[].bar` indicates that the row describes the field `bar` in _each_ object contained in the array `foo`.)
+
+| Field               | Value              | Description                       |
+|:--------------------|:-------------------|:----------------------------------|
+| `currency_code`   | String    | Three-letter ([ISO 4217](http://www.xe.com/iso4217.php)) code of the currency this ledger tracks. |
+| `currency_symbol` | String    | Currency symbol to use in user interfaces for the currency represented in this ledger. For example, "$". |
+| `ilp_prefix`        | String             | The ILP Address Prefix of the ledger. This must match the definition from [IL-RFC-15: ILP Addresses](https://github.com/interledger/rfcs/blob/master/0015-ilp-addresses/0015-ilp-addresses.md). |
+| `connectors`        | Array of [URLs][URL] | Accounts belonging to _recommended_ ILP Connectors. Each member of this list MUST be an HTTP(S) URL where a client can [get the connector's ledger account][Get Account]. This list MAY be empty. |
```
emschwartz: Actually right now it's an array of objects that each contain `id` (URL) and `name` (username on ledger) fields

0012-five-bells-ledger-api/0012-five-bells-ledger-api.md
```diff
@@ -0,0 +1,1290 @@
...
+#### Example
+
+Request:
+
+```json
+PUT /transfers/3a2a1d9e-8640-4d2d-b06c-84f2cd613204
+Content-Type: application/json
+
+{
+  "id": "3a2a1d9e-8640-4d2d-b06c-84f2cd613204",
```
emschwartz: ID is submitted as a URL

0012-five-bells-ledger-api/0012-five-bells-ledger-api.md
```diff
@@ -0,0 +1,1290 @@
...
+
+**[Metadata URL Name][]:** `transfer_fulfillment`.
+
+##### Body Parameters
+
+The message body should be a [Crypto-Condition Fulfillment][] in string format.
+
+#### Response Format
+
+A successful result uses the HTTP response code **200 OK**. The message body of the response is the submitted fulfillment as plain text. A ledger MUST return a successful response if and only if the transfer was executed or canceled as a result of this request.
```
emschwartz: That's misleading. It should give you a 200 if the fulfillment was valid. Even if you submit the same fulfillment multiple times, it should give you a 200 for each of them. Following REST/HTTP conventions, it should probably give a 201 the first time its submitted (but I'm not sure if it does that now)

0012-five-bells-ledger-api/0012-five-bells-ledger-api.md
```diff
@@ -0,0 +1,1290 @@
...
+
+### WebSocket Authentication
+
+Clients MUST authenticate themselves with a token from the [Get Auth Token][] method. This token can be supplied in one of two ways:
+
+- In the `token` query parameter. For example:
+
+        wss://red.ilpdemo.org/ledger/websocket/?token=9AtVZPN3t49Kx07stO813UHXv6pcES
+
+- As a "Bearer" token in the `Authorization` HTTP header of the request (before the protocol switch to WebSocket). For example:
```
emschwartz: This is not supported by the WebSocket API in browsers. @dappelt what's your recommendation?

0012-five-bells-ledger-api/0012-five-bells-ledger-api.md
```diff
@@ -0,0 +1,1290 @@
...
+**Note:** Any fields other than `id` and `message` should be considered optional and informational. Clients MUST NOT depend on the presence of those fields. The examples in this spec contain suggestions for additional fields, but those should not be taken as requirements.
+
+### Unauthorized
+[Unauthorized]: #unauthorized
+
+The [authentication information](#authorization-and-authentication) supplied to this request was insufficient for one of the following reasons:
+
+- This method requires authentication but none was provided
+- The credentials were provided using a system not supported by this method (e.g. Basic Auth when Client Certificates are required)
+- The credentials were malformed or don't match the known account information
```
emschwartz: Some of these I think are supposed to return HTTP 403 instead of 401. @dappelt

0012-five-bells-ledger-api/0012-five-bells-ledger-api.md
dappelt: It is for the ledger computational expensive to verify that credentials passed in via Basic Auth are correct (it should take ~1 sec. for security reasons). In contrast, token-based authentication (e.g. JSON Web Tokens in fb-ledger) are much cheaper to verify for the server.

By switching from Basic Auth to token-based authentication we make it more difficult for unauthorized requests to put a substantial load on the server. Only the endpoint that hands out the auth token should be callable with HTTP Basic Auth. Furthermore, this endpoint should have a strict rate-limiting policy to prevent flooding with fake credentials.

0012-five-bells-ledger-api/0012-five-bells-ledger-api.md
dappelt: In order to make FB-ledger resilient against DoS attacks, we should try to avoid that unauthenticated requests result in database operations.

0012-five-bells-ledger-api/0012-five-bells-ledger-api.md
dappelt: > The credentials were malformed or don't match the known account information

Providing wrong credentials for a supported authentication mechanism should result in 403. See: https://en.wikipedia.org/wiki/HTTP_403#Difference_from_status_.22401_Unauthorized.22

0012-five-bells-ledger-api/0012-five-bells-ledger-api.md
dappelt: WebSocket Authentication is a bit tricky. This is what the RFC says:
> This protocol doesn't prescribe any particular way that servers can authenticate clients during the WebSocket handshake. The WebSocket server can use any client authentication mechanism available to a generic HTTP server, such as cookies, HTTP authentication, or TLS authentication. [RFC6455](https://tools.ietf.org/html/rfc6455)

There are a couple of options to authenticate clients. Which one to choose depends on the browser applications we want to support, so this should be an open discussion. Some options are:

1. **Cookie-based**: When the browser calls the endpoint `/auth_token` on the ledger, the ledger sets a cookie. The ledger does this by adding an HTTP Header to the response, e.g.: `Set-Cookie: token=some-json-web-token; HttpOnly; Path=/websocket`. When the client-side javascript opens the websocket (e.g. `new WebSocket('ws://ledger.example:3000/websocket')`), the browser automatically appends the cookie to the HTTP upgrade request, e.g.:
```
GET /websocket HTTP/1.1
User-Agent: Mozilla/5.0 ...
...
Cookie: token=some-json-web-token // auth details go here
...
Upgrade: websocket
```

From a security standpoint this option has the advantage that the token is not exposed to client-side javascript and, hence, xss vulnerabilities can't be used to steal your auth token.

2. **Websocket Subprotocol**: When opening a WebSocket with javascript, it is possible to indicate a websocket subprotocol. We could use this subprotocol to communicate auth details. For example, when client-side js opens a WebSocket like this `new WebSocket('ws://ledger.example:3000/websocket', 'token=somejsonwebtoken')`, the resulting HTTP request contains this HTTP Header: `Sec-WebSocket-Protocol: token=somejsonwebtoken`.

This option feels a bit like a hack, since the subprotocol header is not dedicated to transport auth details.

3. **Once the websocket is open**: Once the websocket connection is established, the client sends first a message with his credentials.

4. **Query String**: We could send the auth token in the query string, e.g. `new WebSocket('ws://ledger.example:3000/websocket?token=somejsonwebtoken')`. However, I think this is the least secure option. Refer to the [OAuth RFC](https://tools.ietf.org/html/rfc6750#section-5) for some more insights, quote:

> Because of the security weaknesses associated with the URI method   (see Section 5), including the high likelihood that the URL containing the access token will be logged, it SHOULD NOT be used unless it is impossible to transport the access token in the   “Authorization” request header field or the HTTP request entity-body.

Slack uses a creative approach to mitigate this problem. If a client wants to connect, they give out a "single-use" WebSocket URL that is only valid for 30 seconds. We could  easily copy this approach by expiring the auth tokens that the ledger gives out after 30 seconds. So if the token ends up being logged by a proxy, the token is useless after a short time period.

0012-five-bells-ledger-api/0012-five-bells-ledger-api.md
mDuo13: OK. I added this.

I left an exception that the ledger "MAY" accept non-token auth on other methods, since that could be useful to ledgers in secure sub-nets and private deployments.

0012-five-bells-ledger-api/0012-five-bells-ledger-api.md
mDuo13: I have no idea what length to recommend for the memo. If the ILP Packet is base64url-encoded, a little back of the envelope math tells me that the `memo` field containing an `ilp` packet should be around within 46,000 bytes.

```
   8 bytes for the amount (a 64bit UInt)
+ 1024 bytes for address
+ 32768 bytes for data
---------------------------------------
    33856 bytes
 × 4/3 for base64url-encoding (rounded up)
---------------------------------------
    45142 bytes
+ 10 bytes for the JSON object containing it (`{"ilp": "<data>"}`)
---------------------------------------
45152 bytes
```

Asking ledgers to allow the memos field to be up to 46kb seems like a lot, but maybe it's not?

0012-five-bells-ledger-api/0012-five-bells-ledger-api.md
mDuo13: I'm not sure I see how the `proposed` state connects to unauthorized users.

How's this for a recommendation that fixes the issue?

> A ledger SHOULD reject a request to create a `proposed` transfer, unless the client is authenticated as one of the accounts in the `debits` or `credits` of the transfer, or the client is authenticated as an administrator.

0012-five-bells-ledger-api/0012-five-bells-ledger-api.md
mDuo13: In the paragraph leading into this table, it says:

> Fields that are "Optional" or "Ledger-provided" in the following table may be omitted by clients submitting objects to the API, but those fields are not optional to implement.

Is that not sufficient?

0012-five-bells-ledger-api/0012-five-bells-ledger-api.md
mDuo13: The v2 Crypto-Conditions don't have a string format at all, do they?

0012-five-bells-ledger-api/0012-five-bells-ledger-api.md
mDuo13: @sentientwaffle Can you tell me whether the "Submit Fulfillment" response returns 201 the first time a fulfillment causes a transfer to be executed or canceled?

0012-five-bells-ledger-api/0012-five-bells-ledger-api.md
emschwartz: I think it might be better to take out the `optional` from there and put another small section that describes what fields are required/optional when you are doing PUT operations

0012-five-bells-ledger-api/0012-five-bells-ledger-api.md
emschwartz: https://tools.ietf.org/html/draft-thomas-crypto-conditions-02#section-10

0012-five-bells-ledger-api/0012-five-bells-ledger-api.md
sentientwaffle: Yes, the ledger returns 201 when the transfer is executed/cancelled.

See: <https://github.com/interledgerjs/five-bells-ledger/blob/5c579e34ea6bcb6cbb385da4a3ccd21b197fefab/src/controllers/transfers.js#L307>

emschwartz:
LGTM. @dappelt should also sign off on this before it's merged

emschwartz:
@mDuo13 can you add a draft number, rebase on top of master and merge this?

mDuo13:
Added version number and rebased. Not sure how to get the tests to run. Is CircleCI having an outage?

PR #238, opened 2017-07-15
fix: remove redundant "each"

sbellem:
by each each --> by each


michielbdejong:
Merged in #239, thanks @sbellem!

PR #242, opened 2017-07-18
Tweak RFCs template

adrianhopebailie:
Show draft number
Add link to latest draft


adrianhopebailie:
NOTE: The build will fail on this as the template has changed so the script thinks all drafts have been updated. This requires manual intervention once the change has been merged.

PR #251, opened 2017-07-29
docs: introduce ledger-ledger lotocol

### LINKED ISSUE #249: RFC 21: Switch to OER encoding
### Why?

- **Consistency** - ILP and ILQP are in OER so you already need a (de)serializer to use Interledger and using OER for this protocol would mean we wouldn't even have a dependency on having a JSON parser
- **Efficiency / Speed** - Better than JSON on both fronts
- **Simplicity** - We picked OER because the format is so easy to implement and use

### Why now?

1. The RPC API is more done than it appears.
2. This might be the only plugin developers build into things.
3. Because of 2., changing this protocol later will be harder than we think.

> 1. The RPC API is more done than it appears.

We have tended to switch protocols to binary when we think they are more or less done. One could argue that the RPC protocol is so new that we shouldn't do this yet and should instead [freeze it how it is now](https://github.com/interledger/rfcs/issues/234#issuecomment-315411811). However, this protocol is basically just using the formats and methods from the Ledger Plugin Interface and sending them over the wire using HTTP. The LPI has been surprisingly consistent for a long time (with the exception of the switch to `sendRequest` from `sendMessage`), so I would argue this protocol is closer to being done than it might seem.

> 2. This might be the only plugin developers build into things.

Supporting multiple plugins is one of the things that makes reimplementing Interledger or building it into apps/services complicated. It would be _much_ simpler if there were one plugin type that everyone supported. For users, that would also mean you could login to different ILP-enabled applications using a single credential.

If you want to use a different ledger plugin, you would run a simple adapter/connector that translates the RPC protocol into the specific ledger plugin protocol. This is basically just a "plugin" implemented as a standalone service instead of an internal module (@michielbdejong implemented a plugin wrapper like that called [`ilp-frog`](https://github.com/michielbdejong/ilp-frog)). We know this works because the RPC protocol is just the same ledger abstraction we've been using internally in the JS code. What's nice about that solution is that it puts the work of supporting different ledgers on the people that are using them, rather than on developers trying to build Interledger into their apps.

> 3. Because of 2., changing this protocol later will be harder than we think.

I want to start building apps with Interledger and I want other developers to do the same. If we start building things that only support the one plugin, or write new implementations of Interledger clients or connectors that don't bother with supporting multiple plugins, this will be very difficult to change later. Since I think we're pretty close to an acceptable "finished" state with this protocol, I would argue we should make this change now.

---

### Plugin RPC as the Common Ledger API

In general, I think this protocol will become Interledger's Etherenet. Back when we were discussing the ["Common Ledger API"](https://github.com/interledger/rfcs/pull/125), we were bikeshedding a ton because we didn't yet understand what interactions with ledgers would look like or have enough experience using that API. Now we have a better idea and I think the Common Ledger API will look a lot more like this protocol.

We started with the Five Bells Ledger API because we wanted an API that provided all the functionality we thought ledgers would need. But then we realized it did more than was strictly necessary and so we created the Ledger Plugin Interface to abstract away the parts we didn't need. The Plugin RPC protocol is just the Ledger Plugin Interface abstraction sent over the wire, so it makes sense that this protocol would be the minimum you need from the ledger layer to do Interledger.

Since the core Interledger stack already uses OER, I think it would make sense for this protocol (as a newcomer to the "core stack") to use the same format.

justmoon:
Binary protocol to replace RPC API. Resolves #249.

I did not pick the name.

Created with @emschwartz and @sharafian.

### DIFF
--- a/asn1/LedgerLedgerLotocol.asn
+++ b/asn1/LedgerLedgerLotocol.asn
@@ -21,19 +21,19 @@ SideProtocolData ::= SEQUENCE OF SEQUENCE {
   protocolData OCTET STRING
 }
 
-MessageRequest ::= SEQUENCE {
+PacketRequest ::= SEQUENCE {
   packet InterledgerPacket,
   -- Additional data for protocol extensibility
   sideProtocolData SideProtocolData
 }
 
-MessageResponse ::= SEQUENCE {
+PacketResponse ::= SEQUENCE {
   packet InterledgerPacket,
   -- Additional data for protocol extensibility
   sideProtocolData SideProtocolData
 }
 
-TransferRequest ::= SEQUENCE {
+PrepareRequest ::= SEQUENCE {
   transferId UInt128,
   amount UInt64,
   condition UInt256,
@@ -43,7 +43,7 @@ TransferRequest ::= SEQUENCE {
   sideProtocolData SideProtocolData
 }
 
-TransferResponse ::= SEQUENCE {
+PrepareResponse ::= SEQUENCE {
   -- Additional data for protocol extensibility
   sideProtocolData SideProtocolData
 }
@@ -82,7 +82,7 @@ TransferStateResponse ::= SEQUENCE {
   state CHOICE {
     fulfillment UInt256,
     rejectionReason InterledgerPacket,
-    pending NULL,
+    prepared NULL,
     nonExistent NULL
   },
   -- Additional data for protocol extensibility
@@ -107,10 +107,10 @@ CALL ::= CLASS {
 CallSet CALL ::= {
     -- All requests MUST use odd typeIds
     -- All responses MUST use even typeIds
-    {1 MessageRequest} |
-    {2 MessageResponse} |
-    {3 TransferRequest} |
-    {4 TransferResponse} |
+    {1 PacketRequest} |
+    {2 PacketResponse} |
+    {3 PrepareRequest} |
+    {4 PrepareResponse} |
     {5 FulfillRequest} |
     {6 FulfillResponse} |
     {7 RejectRequest} |
--- a/asn1/LedgerLedgerLotocol.asn
+++ b/asn1/LedgerLedgerLotocol.asn
@@ -89,19 +89,6 @@ TransferStateResponse ::= SEQUENCE {
   sideProtocolData SideProtocolData
 }
 
-BalanceRequest ::= SEQUENCE {
-  -- Additional data for protocol extensibility
-  sideProtocolData SideProtocolData
-}
-
-BalanceResponse ::= SEQUENCE {
-  currentBalance VarInt,
-  minBalance VarInt,
-  maxBalance VarInt,
-  -- Additional data for protocol extensibility
-  sideProtocolData SideProtocolData
-}
-
 CustomRequest ::= SEQUENCE {
   -- Additional data for protocol extensibility
   sideProtocolData SideProtocolData
@@ -130,10 +117,8 @@ CallSet CALL ::= {
     {8 RejectResponse} |
     {9 TransferStateRequest} |
     {10 TransferStateResponse} |
-    {11 BalanceRequest} |
-    {12 BalanceResponse} |
-    {13 CustomRequest} |
-    {14 CustomResponse}
+    {11 CustomRequest} |
+    {12 CustomResponse}
 }
 
 LedgerLedgerLotocolPacket ::= SEQUENCE {
--- a/asn1/LedgerLedgerLotocol.asn
+++ b/asn1/LedgerLedgerLotocol.asn
@@ -1,0 +1,151 @@
+LedgerLedgerLotocol
+DEFINITIONS
+AUTOMATIC TAGS ::=
+BEGIN
+
+IMPORTS
+    UInt8,
+    UInt32,
+    UInt64,
+    UInt128,
+    UInt256
+    FROM GenericTypes
+
+    Address,
+    Timestamp
+    FROM InterledgerTypes
+;
+
+SideProtocolData ::= SEQUENCE OF SEQUENCE {
+  protocolName IA5String,
+  protocolData OCTET STRING
+}
+
+MessageRequest ::= SEQUENCE {
+  packet InterledgerPacket,
+  -- Additional data for protocol extensibility
+  sideProtocolData SideProtocolData
+}
+
+MessageResponse ::= SEQUENCE {
+  packet InterledgerPacket,
+  -- Additional data for protocol extensibility
+  sideProtocolData SideProtocolData
+}
+
+TransferRequest ::= SEQUENCE {
+  transferId UInt128,
+  amount UInt64,
+  condition UInt256,
+  expiresAt Timestamp,
+  packet InterledgerPacket,
+  -- Additional data for protocol extensibility
+  sideProtocolData SideProtocolData
+}
+
+TransferResponse ::= SEQUENCE {
+  -- Additional data for protocol extensibility
+  sideProtocolData SideProtocolData
+}
+
+FulfillRequest ::= SEQUENCE {
+  transferId UInt128,
+  fulfillment UInt256,
+  -- Additional data for protocol extensibility
+  sideProtocolData SideProtocolData
+}
+
+FulfillResponse ::= SEQUENCE {
+  -- Additional data for protocol extensibility
+  sideProtocolData SideProtocolData
+}
+
+RejectRequest ::= SEQUENCE {
+  transferId UInt128,
+  rejectionReason InterledgerPacket,
+  -- Additional data for protocol extensibility
+  sideProtocolData SideProtocolData
+}
+
+RejectResponse ::= SEQUENCE {
+  -- Additional data for protocol extensibility
+  sideProtocolData SideProtocolData
+}
+
+TransferStateRequest ::= SEQUENCE {
+  transferId UInt128,
+  -- Additional data for protocol extensibility
+  sideProtocolData SideProtocolData
+}
+
+TransferStateResponse ::= SEQUENCE {
+  state CHOICE {
+    fulfillment UInt256,
+    rejectionReason InterledgerPacket,
+    pending NULL,
+    nonExistent NULL
+  },
+  -- Additional data for protocol extensibility
+  sideProtocolData SideProtocolData
+}
+
+BalanceRequest ::= SEQUENCE {
+  -- Additional data for protocol extensibility
+  sideProtocolData SideProtocolData
+}
+
+BalanceResponse ::= SEQUENCE {
+  currentBalance VarInt,
+  minBalance VarInt,
+  maxBalance VarInt,
+  -- Additional data for protocol extensibility
+  sideProtocolData SideProtocolData
+}
+
+CustomRequest ::= SEQUENCE {
+  -- Additional data for protocol extensibility
+  sideProtocolData SideProtocolData
+}
+
+CustomResponse ::= SEQUENCE {
+  -- Additional data for protocol extensibility
+  sideProtocolData SideProtocolData
+}
+
+CALL ::= CLASS {
+    &typeId UInt8 UNIQUE,
+    &Type
+} WITH SYNTAX {&typeId &Type}
+
+CallSet CALL ::= {
+    -- All requests MUST use odd typeIds
+    -- All responses MUST use even typeIds
+    {1 MessageRequest} |
+    {2 MessageResponse} |
+    {3 TransferRequest} |
+    {4 TransferResponse} |
+    {5 FulfillRequest} |
+    {6 FulfillResponse} |
+    {7 RejectRequest} |
+    {8 RejectResponse} |
+    {9 TransferStateRequest} |
+    {10 TransferStateResponse} |
+    {11 BalanceRequest} |
+    {12 BalanceResponse} |
+    {13 CustomRequest} |
+    {14 CustomResponse}
+}
+
+LedgerLedgerLotocolPacket ::= SEQUENCE {
+    -- One byte type ID
+    type CALL.&typeId ({CallSet}),
+    -- Allow multiple accounts to be multiplexed on one connection
+    accountId UInt32,
+    -- Used to associate requests and corresponding responses
+    -- If requestId = 0, the server MUST not send a response
+    requestId UInt32,
+    -- Length-prefixed main data
+    data CALL.&Type ({CallSet}{@type})
+}
+
+END


adrianhopebailie:
It would be useful to have a way to return an Interledger Error in reponses and also some kind of response code.

I assume that this model assumes the response code comes from the transport (e.g. HTTP, 200, 403 etc) but those codes are not really always appropriate unless were designing this API to use REST (which I don't recommend, I like the RPC pattern we've got).

The transfer life-cycle also seems a bit odd. There is an async response to a transfer request if it is rejected but not if it is accepted. What is returned in the transfer response for a successful transfer request?

asn1/LedgerLedgerLotocol.asn
```diff
@@ -0,0 +1,151 @@
...
+    {9 TransferStateRequest} |
+    {10 TransferStateResponse} |
+    {11 BalanceRequest} |
+    {12 BalanceResponse} |
+    {13 CustomRequest} |
+    {14 CustomResponse}
+}
+
+LedgerLedgerLotocolPacket ::= SEQUENCE {
+    -- One byte type ID
```
adrianhopebailie: Maybe differentiate between request and response so that response can carry an optional error and a response code.

asn1/LedgerLedgerLotocol.asn
```diff
@@ -0,0 +1,151 @@
...
+    UInt256
+    FROM GenericTypes
+
+    Address,
+    Timestamp
+    FROM InterledgerTypes
+;
+
+SideProtocolData ::= SEQUENCE OF SEQUENCE {
+  protocolName IA5String,
```
adrianhopebailie: Any examples of what this might be?

adrianhopebailie:
p.s. I suggest the name Interledger RPC

michielbdejong:
What's the difference between MessageRequest and CustomRequest? Both seem to have a similar role, of making the protocol extensible. For a ledger API, sendMessage makes sense (it's asking the ledger to send something to another account holder), but for RPC I don't think it's as logical to have. They both can be seen as a form of nested "RequestRequest", where you still have to specify what goes inside. As an illustration of this, you can see that the current [`send_request`](https://github.com/interledger/rfcs/blob/master/0021-plugin-rpc-api/0021-plugin-rpc-api.md#send-request-routing) has a 'method' field, even though in the URL it already specified `method=send_request`.

So I would say extensibility can be done through the `CallSet` registry (we can propose method 14, 15, 16, ... as backwards-compatible protocol extension), and then rename `MessageRequest` to `QuoteRequest` (maybe all 3 of them separately) and rename `CustomRequest` to `BroadcastRoutesRequest`.

michielbdejong:
If we decide that this is the way forward, then:
1) what is our deprecation schedule for ilp-kit 3.0?
2) can we commit to keeping this spec stable for a certain time?

asn1/LedgerLedgerLotocol.asn
emschwartz: They are differentiated by each `CallSet` item either being a request or a response. Note that ILP Errors count as `IlpPacket` types

asn1/LedgerLedgerLotocol.asn
```diff
@@ -0,0 +1,151 @@
...
+    -- Allow multiple accounts to be multiplexed on one connection
+    accountId UInt32,
+    -- Used to associate requests and corresponding responses
+    -- If requestId = 0, the server MUST not send a response
+    requestId UInt32,
+    -- Length-prefixed main data
+    data CALL.&Type ({CallSet}{@type})
+}
+
+END
```
emschwartz: Since there is no call to get currency code and scale, where should a ledger plugin for this protocol get that info from? We talked about getting that from some kind of webfinger lookup but does that mean the plugin for this protocol would depend on the webfinger lookup as well, or would those options have to be configured?

emschwartz:
> What's the difference between MessageRequest and CustomRequest? 

@michielbdejong `MessageRequest` includes an `IlpPacket` and `CustomRequest` doesn't. @justmoon argued for doing it this way instead of making the `IlpPacket` an optional field

> what is our deprecation schedule for ilp-kit 3.0?

This spec needs to be hashed out more, implemented as a plugin, and worked into ilp-kit before we talk about any deprecation schedules. We could also make it backwards compatible by enabling kits to peer over different plugins.

> can we commit to keeping this spec stable for a certain time?

Not before an initial version is even hashed out, but my idea with proposing using a binary format in the first place was that I think this could become a fairly long-lived ledger-layer protocol.

asn1/LedgerLedgerLotocol.asn
emschwartz: This would also be useful for sending payment channel claims

michielbdejong:
one method I'm missing is 'settle', where one party tells the other party that they think the trustline balance changed in their favor, and includes either proof of on-ledger (unconditional) payment, or a message referring to manual/out-of-band settlement, or an updated payment channel claim. it's currently in the fulfill response side-protocol-data, but I think it would be good to make it more explicit?

Another would be 'setup', where ledger A tells ledger B its own RPC endpoint. Either that, or:
A simplification would be to make two changes: 
1) sendTransferAndWait (would return the finalization),
2) getNewRoutes would be the reverse of broadcast_routes
That way all methods (getQuote, sendTransferAndWait, getNewRoutes, settle) would be sender-to-next-connector, and then you don't have to tell the next connector your own RPC endpoint, because no communication is ever initiated in the receiver-to-sender direction.

Just one more thought on backwards compatibility - we could also reduce this new protocol to an optional extension of the current protocol, where one peer can try a `send_binary` RPC call, and if it fails, fall back to the normal JSON format. It would mean all servers are still compatible, but if both server support binary, then they can benefit from its advantages. Not sure if that would be desirable, because it would merge the current and the new protocol into one ugly protocol.

emschwartz:
> one method I'm missing is 'settle'

I think that would be a good use of the `sideProtocolData`. The particular details you'd be expected to include (and even what messages you need to attach them to) are dependent on the settlement method you're using, so I wouldn't make a separate call for this.

> Another would be 'setup', where ledger A tells ledger B its own RPC endpoint

That sounds a bit unsafe to me

> 1. sendTransferAndWait (would return the finalization),

The proposal is that this new protocol would go over raw (TCP/TLS) sockets or WebSockets instead of HTTP, so there's no waiting anyway.

> 2. getNewRoutes would be the reverse of broadcast_routes

Sounds like something that should be put in the routing protocol, not in the ledger layer protocol. If we make `IlpPacket` (OER) messages for the routing protocol then they can be included in this protocol's messages.

> we could also reduce this new protocol to an optional extension of the current protocol, where one peer can try a send_binary RPC call

👎  We should handle backwards compatibility on the ledger layer by making ILP Kit support different plugins better. Also, since this would go over raw sockets or WebSockets it wouldn't really make sense to combine the two.

michielbdejong:
ok, so I think a prerequisite for adding this new protocol as a second way in which servers can peer with each other, is for servers to announce which peering method(s) they support. If settlements and route broadcasts are not defined by LLL, then a server would have to announce "I support LLL + JSON-route-broadcasts version X + XRP-paychan-settlements version Y".

emschwartz:
@michielbdejong I almost agree with that, but not quite.

| Layer | Protocols | Announce versions? |
|---|---|---|
| Ledger | LLL / JSON RPC / Plugin XRP PayChan / (...other payment channel plugin) | Connectors need to decide what ledger (plugin) they are going to peer over. This is less of a "versioning" question and more a matter of picking a ledger protocol and having the implementation support multiple options. |
| Interledger | ILP + ILQP | This is versioned but changing this version will be like upgrading from IPv4 to v6. I don't think it's worth including this in the announcements, because parties that use different versions of these protocols are pretty much on different Interledger networks |
| Routing | CCP | Agreeing on the version and/or variant of this should probably be in the peering protocol |

> This protocol competes with ilp-kit's current peering protocol.

That is only true if connectors that use this assume all other connectors use it. Peering should include an option about what ledger protocol the connectors are going to use. ILP Kit should be changed to be more flexible and support multiple plugins for peers.

michielbdejong:
ok, in any case, when you peer with someone, you want to know exactly what they support on all of those layers. we can start with adding an "LLL support" indication to ilp-kit's WebFinger announcement, and then make it use the correct plugin for each peer based on that.

asn1/LedgerLedgerLotocol.asn
```diff
@@ -0,0 +1,151 @@
...
+}
+
+MessageResponse ::= SEQUENCE {
+  packet InterledgerPacket,
+  -- Additional data for protocol extensibility
+  sideProtocolData SideProtocolData
+}
+
+TransferRequest ::= SEQUENCE {
+  transferId UInt128,
```
michielbdejong: you could also use the condition as a unique id of the transfer?

michielbdejong: ah right, i remember now, this allows friendly repeated attempts at the same payment

asn1/LedgerLedgerLotocol.asn
```diff
@@ -0,0 +1,151 @@
...
+  sideProtocolData SideProtocolData
+}
+
+TransferResponse ::= SEQUENCE {
+  -- Additional data for protocol extensibility
+  sideProtocolData SideProtocolData
+}
+
+FulfillRequest ::= SEQUENCE {
+  transferId UInt128,
```
michielbdejong: or condition

asn1/LedgerLedgerLotocol.asn
```diff
@@ -0,0 +1,151 @@
...
+  sideProtocolData SideProtocolData
+}
+
+RejectResponse ::= SEQUENCE {
+  -- Additional data for protocol extensibility
+  sideProtocolData SideProtocolData
+}
+
+TransferStateRequest ::= SEQUENCE {
+  transferId UInt128,
```
michielbdejong: here it makes sense, i guess, to have a separate id field

asn1/LedgerLedgerLotocol.asn
```diff
@@ -0,0 +1,151 @@
...
+  sideProtocolData SideProtocolData
+}
+
+BalanceRequest ::= SEQUENCE {
+  -- Additional data for protocol extensibility
+  sideProtocolData SideProtocolData
+}
+
+BalanceResponse ::= SEQUENCE {
+  currentBalance VarInt,
```
michielbdejong: We should define more precisely, i think the most generic definition would be:
* `+` total amounts which you received and peer agrees you have fulfilled in time
* `-` total amounts which you sent and peer claims to have fulfilled in time
* `-` total amounts which you sent and peer claims to still be pending
* `+` total amounts which peer agrees you sent on-ledger (unconditional/fulfilled/paychan-close)
* `-` total amounts which peer claims you received on-ledger (unconditional/fulfilled/paychan-close)
* `+` total of, for each open outgoing paychan, highest claim peer agrees you gave
* `-` total of, for each open incoming paychan, highest claim peer says they gave you

So then:
* your max send limit is currentBalance-minBalance
* your max receive limit is maxBalance-currentBalance

asn1/LedgerLedgerLotocol.asn
```diff
@@ -0,0 +1,151 @@
...
+    {12 BalanceResponse} |
+    {13 CustomRequest} |
+    {14 CustomResponse}
+}
+
+LedgerLedgerLotocolPacket ::= SEQUENCE {
+    -- One byte type ID
+    type CALL.&typeId ({CallSet}),
+    -- Allow multiple accounts to be multiplexed on one connection
+    accountId UInt32,
```
michielbdejong: does account mean "from" or "to", here? Or is it more "prefix"?

asn1/LedgerLedgerLotocol.asn
```diff
@@ -0,0 +1,151 @@
...
+    -- Used to associate requests and corresponding responses
+    -- If requestId = 0, the server MUST not send a response
+    requestId UInt32,
```
michielbdejong: doesn't this also duplicate the transferId?

asn1/LedgerLedgerLotocol.asn
emschwartz: Enforcing condition uniqueness would also prevent you from doing things like Optimistic over Universal (which might not be a good idea anyway)

asn1/LedgerLedgerLotocol.asn
emschwartz: It's more prefix

asn1/LedgerLedgerLotocol.asn
michielbdejong: I think it's reasonable to remove getInfo and getAccount. There is a lot more you also need to negotiate, like version of this protocol, version of underlying socket protocol, endpoint URL in each direction, and auth tokens. Also, it might be that currency info is only defined relatively (e.g. company stock).

asn1/LedgerLedgerLotocol.asn
emschwartz: They do slightly different things. I could make multiple calls, in which each request/response would have the same `requestId`, regarding the same transfer. First I would prepare it, which has a request and response, then it would get fulfilled, which has another request/response, and then I could check the state, etc. `requestId` is meant to correlate a specific call request with its response.

asn1/LedgerLedgerLotocol.asn
michielbdejong: another view is saying settlement can only occur through circular conditional payments to self, "into this trustline", so then this protocol can only be used for one-to-one trustlines, and you only need the first 3 items to describe the balance. Apart from LLL, we would still on-ledger escrow plugins as well as paychan plugins, but they would only be used for the settlement payments, and LLL would be used for the "main traffic".

michielbdejong:
One thing I'm missing in this protocol is a clear state machine definition of how the balance updates as a result of transfers getting finalized in the symmetric trustline case:
* I would like to be able to be able to send a custom message that changes the balance (I'm thinking of sending paychan claims or sending a reference to an unconditional on-ledger / out-of-band payment), including 'agree' and 'dispute' response codes.
* I assume message in one direction are delivered FIFO, but I would like to know if it's also impossible that two messages cross each other if they travel in opposite direction. That's relevant for communicating balance updates.
* If all messages are strictly linearly ordered, then each Fulfill message could carry an updated balance, and each Fulfill response could carry an agree/dispute response code, where dispute reason might be 'No transfer prepared for that condition' or 'New balance amount is different from previous balance + transfer amount' or 'Fulfillment incorrect for that condition' or 'Attempt to fulfill already fulfilled transfer' or 'Attempt to fulfill after cancel'
* Disagreements about timing are fundamentally unavoidable when doing Interledger over a trustline. So the protocol should have a way to deal with them: Keep track of a disputed amount. If the disagreement is in each other's favor then the disputed amount is lowered; if it's in conflicting direction, then the disputed amount is increased. That way, the state of the trustline can stay consistent indefinitely.

emschwartz:
> One thing I'm missing in this protocol is a clear state machine definition

To be fair, this is just the data format. We also need an accompanying RFC to explain how its used.

> I would like to be able to be able to send a custom message that changes the balance

Don't think anything here is stopping you from doing that.

> I assume message in one direction are delivered FIFO, but I would like to know if it's also impossible that two messages cross each other if they travel in opposite direction. That's relevant for communicating balance updates.

I'm not sure this protocol would make any guarantees about the message order. That only causes an issue though if you're playing around very close to the balance limit and are taking action based on the balance numbers your peer is reporting, right?

> If all messages are strictly linearly ordered, then each Fulfill message could carry an updated balance

This seems more complicated than necessary to me. As as user of this protocol I think I'd prefer to leave that out and just do some little reconciliation when necessary.

> Disagreements about timing are fundamentally unavoidable when doing Interledger over a trustline. So the protocol should have a way to deal with them: Keep track of a disputed amount.

Does that need to be in the protocol (on the wire format) or would that just be an implementation feature?

michielbdejong:
then we should maybe move getBalance out of scope as well

emschwartz:
I think having _some_ kind of `getBalance` call seems useful, but if there are additional fields you want to ask about in the context of reconciliation that could arguably go into a "side protocol".

michielbdejong:
What is the function of TransferResponse, FulfillResponse and RejectResponse? It would be more efficient to just have an incrementing RequestId for detecting when a message was lost, or maybe the underlying socket guarantees that it can detect and retry when messages get lost (for instance, a WebSocket can have latency and can be disconnected, but never drops messages).

emschwartz:
Acknowledging that the request was processed.

I think making every request blocking (which doing this over TCP might do anyway) seems like it would make it much less efficient than if you just fire them off asynchronously.

michielbdejong:
BalanceRequest/BalanceResponse is only useful in the asymmetric case (i.e. when only one peer remembers the trustline balance) if its meaning is strictly defined. In the symmetric case (when both peers remember their account of the trustline balance), you would want to know not only how much balance your peer is claiming, but also which events were taken into account while summing up the `currentBalance`, so that you can check whether you and your peer agree about it. If we move BalanceRequest/BalanceResponse out of scope then we don't even have to distinguish between the symmetric and the asymmetric case, so that's a lot simpler.

michielbdejong:
> Acknowledging that the request was processed.

You know your TransferRequest was processed when you get back a FulfillRequest or a RejectRequest.
You know your FulfillRequest/RejectRequest was processed when your peer does not follow up with a TransferStateRequest.

emschwartz:
> You know your TransferRequest was processed when you get back a FulfillRequest or a RejectRequest.

Actually that's not totally true. Right now the LPI `sendTransfer` promise resolves when the transfer has been accepted by the ledger. You only get the fulfillment notification or rejection when the Interledger payment has made it to the final receiver. If the message just gets dropped you might want to resend it, but you'll have to wait the whole expiry duration to figure that out if there's no acknowledgement.

Maybe we'd want to rely on TCP for resending messages? Although I think there's also a strong argument for doing this whole protocol over UDP, especially with things like quote requests, because you don't want them to block other, more time-sensitive messages.

michielbdejong:
I agree BalanceRequest/BalanceResponse can stay in, but *only* if we mathematically define the meaning of the `currentBalance` number. I propose we define it as the sum of amounts for which:
a) FulfillRequest resulted in a FulfillResponse 'OK' (and not a FulfillResponse 'too late')
b) some side protocol changed the balance in a way that is defined by that side protocol

Note that this gives the preparing party an opportunity to cheat: they can use the fulfillment, and then *still* tell the fulfiller that they were too late. When this happens, the fulfiller should temporarily lower their trust in their peer; if it happens repeatedly in one direction but never in the other direction, the trustline will eventually dry up, and this is how the fulfiller's exposure to this form of cheating is capped.

If we cannot agree an what the mathematical meaning of `currentBalance` is, then I think we should remove BalanceRequest/BalanceResponse from the core protocol, and make it a CustomRequest.

emschwartz:
Ok I'm convinced now that we should take out the balance request and response.

There are at least three different uses of this protocol, which would all want something different from that call:
- Asymmetric trustlines: the balance is whatever the stateful side says it is
- Symmetric trustlines: there are different values related to the balance that you might want to communicate about, including all of the things @michielbdejong mentioned [above](https://github.com/interledger/rfcs/pull/251#discussion_r130641448)
- Claim-based payment channels: might want to communicate all of the values used in symmetric trustlines plus others related to how much has been claimed

It seems crazy to me to take out the balance call from this protocol, but I think @michielbdejong is probably right that it's better to leave it out than have it defined differently by different implementations of this protocol. Since what is meant by "the balance" is necessarily different in different use cases, we should take it out of this spec.

asn1/LedgerLedgerLotocol.asn
```diff
@@ -0,0 +1,136 @@
...
+  -- Additional data for protocol extensibility
+  sideProtocolData SideProtocolData
+}
+
+PrepareRequest ::= SEQUENCE {
+  transferId UInt128,
+  amount UInt64,
+  condition UInt256,
+  expiresAt Timestamp,
+  packet InterledgerPacket,
```
michielbdejong: instead of including an InterledgerPacket here, with an InterledgerProtocolPayment inside it, and a useless `type` field that will always be `1`, we should maybe directly include the InterledgerProtocolPayment itself here?

emschwartz: I don't think we should do that because there may be other versions of the InterledgerProtocolPayment packet later, and we could also decide to do things like sending quote requests on transfers.

michielbdejong: ok, good point. I removed this from https://github.com/interledger/rfcs/pull/256#issuecomment-319746470 (which I am now quite starting to like as I look at it again, actually).

asn1/LedgerLedgerLotocol.asn
adrianhopebailie: > I could make multiple calls, in which each request/response would have the same requestId, regarding the same transfer

This doesn't make sense. Do you mean different requestId and same transferId?

emschwartz: Yes, sorry. Same `transferId`

asn1/LedgerLedgerLotocol.asn
```diff
@@ -0,0 +1,136 @@
...
+PacketResponse ::= SEQUENCE {
+  packet InterledgerPacket,
+  -- Additional data for protocol extensibility
+  sideProtocolData SideProtocolData
+}
+
+PrepareRequest ::= SEQUENCE {
+  transferId UInt128,
+  amount UInt64,
+  condition UInt256,
```
michielbdejong: why rename `executionCondition` to `condition`?

emschwartz:
Superseded by https://github.com/interledger/rfcs/pull/271

PR #256, opened 2017-08-02
renumber LLL packets to reuse ILP packets directly

michielbdejong:


asn1/LedgerLedgerLotocol.asn
```diff
@@ -107,18 +107,27 @@ CALL ::= CLASS {
...
-    {12 CustomResponse}
+    -- {1 reserved because this is the typeId of InterledgerProtocolPaymentPacket in
+    -- https://github.com/interledger/rfcs/edit/mj-lll-packet-types/asn1/GenericPacket.asn } |
+    {2 QuoteLiquidityRequest} |
+    {3 QuoteLiquidityResponse} |
+    {4 QuoteBySourceAmountRequest} |
+    {5 QuoteBySourceAmountResponse} |
+    {6 QuoteByDestinationAmountRequest} |
+    {7 QuoteByDestinationAmountResponse} |
+    {8 InterledgerProtocolError} |
```
michielbdejong: you can either use this type to send some error message directly to your peer, or include this error in the reason field of a RejectRequest

asn1/LedgerLedgerLotocol.asn
```diff
@@ -107,18 +107,27 @@ CALL ::= CLASS {
...
-    {11 CustomRequest} |
-    {12 CustomResponse}
+    -- {1 reserved because this is the typeId of InterledgerProtocolPaymentPacket in
+    -- https://github.com/interledger/rfcs/edit/mj-lll-packet-types/asn1/GenericPacket.asn } |
+    {2 QuoteLiquidityRequest} |
+    {3 QuoteLiquidityResponse} |
+    {4 QuoteBySourceAmountRequest} |
+    {5 QuoteBySourceAmountResponse} |
+    {6 QuoteByDestinationAmountRequest} |
+    {7 QuoteByDestinationAmountResponse} |
```
michielbdejong: Note that these would be the ILP packets as already defined, so with extensions, but without SideProtocol. This means you can't freely add fields, but only if the ILQP spec authors admit your custom extension as an official ASN1 extension for ILQP. They could, of course create a generic "SideProtocol" extension field, and that would then be an official extensions that allows ad-hoc extensions (IIUC).

asn1/LedgerLedgerLotocol.asn
```diff
@@ -107,18 +107,27 @@ CALL ::= CLASS {
...
+    {8 InterledgerProtocolError} |
+    -- skipping numbering to allow more future ILP-layer packet types
+    {254 CustomRequest} |
+    {255 CustomResponse} |
```
michielbdejong: I thought this would fit better inside the low number space than outside it, since it's "custom" and could be anything, whether Interledger-layer, or ledger-layer, or side protocol. We could even make two identical but separate CustomRequest/Response pairs, one for use on the Interledger layer and one for use on the ledger layer, but that seems like overkill.

asn1/LedgerLedgerLotocol.asn
```diff
@@ -107,18 +107,27 @@ CALL ::= CLASS {
...
+    -- rest of types are outside the ILP-layer, so including the ledger layer
+    {256 PrepareRequest} |
```
michielbdejong: My personal preference would be to number these 9-16, but since they're ledger layer and not Interledger layer, they have to go with higher numbers.
It might actually be necessary to wrap them all in "escape code" type 255, so that you get bytes:

255 ILP-packet type escape code
1      high byte of ledger-level packet type
1      low byte of ledger-level packet type
...    rest of the data, determined by ledger-level packet type.

asn1/LedgerLedgerLotocol.asn
michielbdejong: if 255 becomes the "escape code" ILP packet type then these would move to 253 and 254.

asn1/LedgerLedgerLotocol.asn
michielbdejong: If we reserve type 255 as an escape code, then it's actually safe to number these ledger-level packet types as 9-16, and then everything will just be simple for the near future, until maybe one day we *need* to use two-byte types. And even if that future comes, we will know we'll be prepared for it.

michielbdejong:
@emschwartz I now did the following:
* add CustomRequest/Response as InterledgerPacket
* add an InterledgerPacket type called "HighPacket", that can wrap packets with two-byte types into an InterledgerPacket with the special one-byte type of 255
* add the TransferProtocol for {Prepare/Fulfill/Reject/TransferState}{Request/Response}, and import those as InterledgerPacket types 9-16.
* now all you need in the LLL is PacketRequest and PacketResponse. But since the packet inside a PacketRequest is always one of InterledgerPacket's "...Request" types, and analogously for PacketResponse, you can merge those into one 'PacketMessage'
* and now the LLL call type is always 'PacketMessage', so you can get rid of that class, and put its 'packet InterledgerPacket' directly in the place of LLL-Packet's data field.
* remove the 'type' field from LLL-Packet, since its role is now served by the type field inside the packet.
* that way, there's only one packet type registry.

michielbdejong:
To give a concrete example, this is what a prepare request would look like on the wire in the current LLL proposal:
```js
{
  type: 'PrepareRequest',
  accountId: 123456677,
  requestId: 352432914,
  data: {
    transferId: 245623465,
    amount: 3456512343,
    condition: 45675623645763563742623646234,
    expiresAt: '20170721041632.999Z',
    packet: // if we apply https://github.com/interledger/rfcs/pull/251#discussion_r130942690
      amount: 1234567,
      account: 'g.tasdfasef'
      data: '2BE02BE'H,
    },
    sideProtocolData: []
  }
}
```

And this is the reshuffle I propose:
```js
{
  accountId: 123456677,
  requestId: 352432914,
  packet: { // extra level of nesting, unfortunately
    type: 'PrepareRequest',
    data: {
      transferId: 245623465,
      amount: 3456512343,
      condition: 45675623645763563742623646234,
      expiresAt: '20170721041632.999Z',
      packet: // if we apply https://github.com/interledger/rfcs/pull/251#discussion_r130942690
        amount: 1234567,
        account: 'g.tasdfasef'
        data: '2BE02BE'H,
      },
      sideProtocolData: []
    }
  }
}
```

For ILQP, the levels of nesting would stay the same - from current:
```js
{
  type: 'PacketResponse',
  accountId: 123456677,
  requestId: 352432914,
  data: {
    type: 'QuoteLiquidityResponse',
    data: {
      liquidity: [
        [
          0,
          0
        ]
      ],
      appliesToPrefix: 'g.ledger.',
      sourceHoldDuration: 2565234,
      expiresAt: '20170721041632.999Z',
      extensions: []
    },
    sideProtocolData: []
  }
}
```

To what I propose:
```js
{
  accountId: 123456677,
  requestId: 352432914,
  packet: {
    type: 'QuoteLiquidityResponse',
    data: {
      liquidity: [
        [
          0,
          0
        ]
      ],
      appliesToPrefix: 'g.ledger.',
      sourceHoldDuration: 2565234,
      expiresAt: '20170721041632.999Z',
      extensions: []
    }
  }
}
```

michielbdejong:
OK, regardless of whether the TransferProtocol gets number 9..16 or 256..263, what do you think of the idea the LLL should just be reduced to:
* accountId
* requestId
* packet

And then that packet can be ILQP, or TransferProtocol, or custom packets for e.g. querying your balance or broadcasting routes

asn1/LedgerLedgerLotocol.asn
```diff
@@ -11,30 +11,40 @@ IMPORTS
     UInt256
     FROM GenericTypes
 
+    SideProtocolDate
```
emschwartz: typo: should be data

asn1/GenericPacket.asn
```diff
@@ -35,13 +67,18 @@ PacketSet PACKET ::= {
     {6 QuoteByDestinationAmountRequest} |
     {7 QuoteByDestinationAmountResponse} |
     {8 InterledgerProtocolError}
+
+    {253 CustomRequest} |
+    {254 CustomResponse} |
```
emschwartz: Arguably if this message is going to be riding on a request or response, you just need a `CustomMessage` type like in https://github.com/interledger/rfcs/pull/260/files#diff-42c9b615e907424c7cae3feb333d8b6fR75

asn1/GenericPacket.asn
```diff
@@ -21,12 +21,44 @@ IMPORTS
...
+    &typeId UInt16 UNIQUE,
+    &Type
+} WITH SYNTAX {&typeId &Type}
+
+HighPacketSet HIGHPACKET ::= {
+    -- nothing here yet, will start filling up when the LowPacketSet is full
+}
+
+-- wraps a high packet into a low packet:
+HighPacketWrapper ::= SEQUENCE {
```
emschwartz: How are you proposing that this be used? (You don't have type ids defined for the LLL)

asn1/LedgerLedgerLotocol.asn
```diff
@@ -43,42 +53,41 @@ PrepareRequest ::= SEQUENCE {
...
+Ack ::= SEQUENCE {
+  requestIdRef UInt128
   -- Additional data for protocol extensibility
   sideProtocolData SideProtocolData
 }
 
-FulfillRequest ::= SEQUENCE {
-  transferId UInt128,
-  fulfillment UInt256,
+Response ::= SEQUENCE {
```
emschwartz: Should responses always be assumed to be free?

asn1/LedgerLedgerLotocol.asn
```diff
@@ -11,30 +11,40 @@ IMPORTS
     UInt256
     FROM GenericTypes
 
+    SideProtocolDate
+    FROM GenericPacket
+
+    InterledgerPacket
+    FROM GenericPacket
```
emschwartz: don't need `FROM GenericPacket` on two lines

asn1/LedgerLedgerLotocol.asn
```diff
@@ -11,30 +11,40 @@ IMPORTS
...
 ;
 
-SideProtocolData ::= SEQUENCE OF SEQUENCE {
-  protocolName IA5String,
-  protocolData OCTET STRING
-}
-
-PacketRequest ::= SEQUENCE {
+FreeRequest ::= SEQUENCE {
+  fromId UInt128,
```
emschwartz: Why are these `UInt128`s instead of ILP Addresses? Seems more confusing to have to do address mapping. Also, the address format is flexible enough that you could use shortened aliases if you wanted to.

See https://github.com/interledger/rfcs/pull/261

asn1/LedgerLedgerLotocol.asn
```diff
@@ -1,4 +1,4 @@
-LedgerLedgerLotocol
+LedgerLedgerLrotocol
```
emschwartz: 👎  on "Lrotocol", even more than on "Lotocol"

asn1/LedgerLedgerLotocol.asn
michielbdejong: I think it's confusing to use ILP addresses in a context where most ILP addresses would be "valid but not local to this connection, so sort of invalid anyway".`.from` and `.to` need to be local to the connection in practice, so we might as well say so in the protocol as well. See also https://github.com/interledger/rfcs/pull/261#discussion_r131582148

michielbdejong:
superseded by https://github.com/interledger/rfcs/pull/262

PR #257, opened 2017-08-02
docs: specifies allowed characters in headers and values

dappelt:


0016-pre-shared-key/0016-pre-shared-key.md
```diff
@@ -33,7 +33,9 @@ A disadvantage of PSK is that it is repudiable. Although the sender does get cry
 
 ## Data Format
 
-The PSK details are divided into two sets of headers. Each set of headers is formatted like those found in [HTTP Requests](https://tools.ietf.org/html/rfc7230#section-3.2). Application data is appended to the private headers, after a blank line. This object is then encrypted and appended to the public headers as binary data, after a blank line. Both private and public headers are parsed in the exact same way as HTTP headers. All strings are UTF-8 encoded.
+The PSK details are divided into two sets of headers. Each set of headers is formatted like those found in [HTTP Requests](https://tools.ietf.org/html/rfc7230#section-3.2) and follows the format `Header-Name: Header-Value`. Note that implementations have to validate inputs before using it as part of `Header-Name` or `Header-Value`. In particular, such inputs MUST NOT contain any line feed (i.e. `\n`).
```
emschwartz: If there is a vulnerability associated with this I would put the last sentence in bold on its own:

**Headers MUST NOT contain line feed characters such as `\n`.** Implementations that do not enforce this may be vulnerable to [Header Injection Attacks](https://en.wikipedia.org/wiki/HTTP_header_injection).

PR #258, opened 2017-08-02
docs(0004): RequestHandlerAlreadyRegisteredError

sentientwaffle:


michielbdejong:
In https://github.com/interledger/rfcs/blob/8176fd3a48fd904ff4bf4584a115b9a4ce537220/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md#registerrequesthandler it says this method 'throws' that error. That's incorrect, the method 'rejects with' that error.

sentientwaffle:
ilp-plugin-bells [throws](https://github.com/interledgerjs/ilp-plugin-bells/blob/f819a840d46d742510587861ab2777b046440c68/src/lib/plugin.js#L418), not rejects

michielbdejong:
ah, that's a bug in that plugin then. It also seems to return undefined instead of a Promise. Didn't we have a plugin test suite for the LPI somewhere? It should be able to catch this sort of devations from the spec.

sentientwaffle:
Why would it return a promise? The spec says the function should return nothing, and it's a synchronous function.

michielbdejong:
Ah sorry, my bad. I misread the line
`ledgerPlugin.registerRequestHandler( requestHandler: ( request: Message ) ⇒ Promise<Message> ) ⇒ null`
I guess the difference between `null` and `undefined` is small enough. :) All correct then!

PR #259, opened 2017-08-02
fix: include the correct class name

michielbdejong:
* [InterledgerProtocolPayment](https://github.com/interledger/rfcs/blob/master/asn1/GenericPacket.asn#L30) is what can be wrapped inside an ILP packet
* [InterledgerProtocolPayment](https://github.com/interledger/rfcs/blob/master/asn1/InterledgerProtocol.asn#L15) is also what has been defined.
* Yet [InterledgerProtocolPaymentMessage](https://github.com/interledger/rfcs/blob/master/asn1/GenericPacket.asn#L11) is what the GenericPacket doc tries to include from the InterledgerProtocol doc.

### DIFF
--- a/asn1/GenericPacket.asn
+++ b/asn1/GenericPacket.asn
@@ -8,7 +8,7 @@ IMPORTS
     VarBytes
     FROM GenericTypes
 
-    InterledgerProtocolPaymentMessage,
+    InterledgerProtocolPayment,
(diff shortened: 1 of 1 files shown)


adrianhopebailie:
I thought this had been fixed somewhere? So now I'm worried there's a dangling commit in a PR somewhere

PR #260, opened 2017-08-04
Proposal: Interledger RPC (instead of LLL)

emschwartz:
slightly different take on https://github.com/interledger/rfcs/pull/251 and https://github.com/interledger/rfcs/pull/256
based on conversations with @michielbdejong, @sharafian, @justmoon

By separating the transfer contents from whether they are part of a
bilateral (e.g. trustline) or multilateral (e.g. centralized ledger)
relationship, the transfer protocol becomes much more general and thus
makes sense to include alongside the "core" ILP data formats.

`InterledgerRpc` wraps the `InterledgerPacket` types in an envelope that
adds a `to`, `from` (either or both of which can be set to empty strings if
they are implied, as is the case in a bilateral connection), a `requestId`
to correlate requests and responses, and `sideProtocolData` for
extensibility.

Any type of `InterledgerRpc` request can become a paid request by
including a `Transfer` or `ConditionalTransfer` in it (which themselves can
contain other `InterledgerPacket`s -- `InterledgerProtocolPayment`s or any
other type of data that you might want/need to pay for).

### DIFF
--- a/asn1/GenericPacket.asn
+++ b/asn1/GenericPacket.asn
@@ -19,6 +19,17 @@ IMPORTS
     QuoteByDestinationAmountRequest,
     QuoteByDestinationAmountResponse
     FROM InterledgerQuotingProtocol
+
+    Transfer,
+    ConditionalTransfer,
+    ConditionFulfillment,
+    TransferStateRequest,
+    TransferStateResponse,
+    FROM TransferProtocol
+
+    Ack,
+    CustomMessage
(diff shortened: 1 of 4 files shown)


emschwartz:
@sharafian convinced me that it is better to have the LLL as a properly separate protocol with a different type registry so I'm closing this PR in favor of https://github.com/interledger/rfcs/pull/261

PR #262, opened 2017-08-07
LLL with ILP packet payloads

michielbdejong:
If we redefine LLL as free/paid/conditional request/response protocol, where the payload is always an ILP packet, then it makes sense to move `CustomMessage` and `SideProtocolData` to the payload (ILP packet) level.

* rename PacketRequest to FreeRequest
* add PaidRequest
* add a 'Custom' ILP packet
* remove SideProtocolData and instead use 'Custom' inside 'FreePacket' 
* get rid of transactionId in PrepareRequest in favor of requestId

### DIFF
--- a/asn1/InterledgerProtocol.asn
+++ b/asn1/InterledgerProtocol.asn
@@ -40,4 +40,9 @@ InterledgerProtocolError ::= SEQUENCE {
     data OCTET STRING (SIZE (0..8192))
 }
 
+InterledgerSideProtocolData ::= SEQUENCE OF SEQUENCE {
+  protocolName IA5String,
+  protocolData OCTET STRING
+}
+
 END
--- a/asn1/LedgerLedgerLotocol.asn
+++ b/asn1/LedgerLedgerLotocol.asn
@@ -16,87 +16,88 @@ IMPORTS
     FROM InterledgerTypes
 ;
 
-SideProtocolData ::= SEQUENCE OF SEQUENCE {
-  protocolName IA5String,
-  protocolData OCTET STRING
+FreeRequest ::= SEQUENCE {
+    -- Used to associate requests and corresponding responses
+    -- MUST be either zero and non-repeating
+    -- If requestId = 0, the server MUST not send a response
+    requestId UInt128,
+    packet InterledgerPacket
 }
 
-PacketRequest ::= SEQUENCE {
-  packet InterledgerPacket,
-  -- Additional data for protocol extensibility
-  sideProtocolData SideProtocolData
+FreeResponse ::= SEQUENCE {
+    -- Used to associate response to earlier FreeRequest
+    requestIdRef UInt128,
+    packet InterledgerPacket
 }
 
-PacketResponse ::= SEQUENCE {
-  packet InterledgerPacket,
-  -- Additional data for protocol extensibility
-  sideProtocolData SideProtocolData
+PaidRequest ::= SEQUENCE {
+    -- Used to associate requests and corresponding responses
+    -- MUST be either zero and non-repeating
+    -- If requestId = 0, the server MUST not send a response
+    requestId UInt128,
+    amount UInt64,
+    packet InterledgerPacket
+}
+
+PaidResponse ::= SEQUENCE {
+    -- Used to associate response to earlier PaidRequest
+    requestIdRef UInt128,
+    packet InterledgerPacket
 }
 
 PrepareRequest ::= SEQUENCE {
-  transferId UInt128,
-  amount UInt64,
-  condition UInt256,
-  expiresAt Timestamp,
-  packet InterledgerPacket,
-  -- Additional data for protocol extensibility
-  sideProtocolData SideProtocolData
+    -- Used to associate later response/fulfill/reject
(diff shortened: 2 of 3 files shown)


asn1/InterledgerProtocol.asn
```diff
@@ -40,4 +40,9 @@ InterledgerProtocolError ::= SEQUENCE {
     data OCTET STRING (SIZE (0..8192))
 }
 
+InterledgerSideProtocolData ::= SEQUENCE OF SEQUENCE {
+  protocolName IA5String,
+  protocolData OCTET STRING
+}
```
emschwartz: The way this is defined, you can't actually send side protocol data along with each of the standard messages. This is really just a way to send a custom message in one of the requests. I think it's useful to allow each message to be extended, in addition to allowing fully custom messages to be sent via this protocol.

Also, if one of these custom things is going to become part of the standard packet set I would rather make it a single set of `protocolName` (maybe `mimeType`) and `protocolData`, instead of the `SEQUENCE OF` that.

asn1/LedgerLedgerLotocol.asn
```diff
@@ -0,0 +1,135 @@
...
+    -- If requestId = 0, the server MUST not send a response
+    requestId UInt128,
+    amount UInt64,
+    packet InterledgerPacket
+}
+
+PaidResponse ::= SEQUENCE {
+    -- Used to associate response to earlier PaidRequest
+    requestIdRef UInt128,
+    packet InterledgerPacket
```
emschwartz: It's a little confusing to me that a `PaidResponse` would actually be free

asn1/LedgerLedgerLotocol.asn
```diff
@@ -0,0 +1,135 @@
...
+    {9 RejectRequest} |
+    {10 RejectResponse} |
+    {11 RequestStateRequest} |
+    {12 RequestStateResponse}
+}
+
+LedgerLedgerLotocolPacket ::= SEQUENCE {
+    -- One byte type ID
+    type CALL.&typeId ({CallSet}),
+    -- Allow multiple from/to accounts to be multiplexed on one connection
```
sharafian: How do you use this in the bilateral case vs. in the multilateral case? In the bilateral case, there should be one ID to determine what trustline this is, but the actual accounts on the trustline can be implied from that. Does that go in the fromId or toId? Both?

asn1/LedgerLedgerLotocol.asn
```diff
@@ -0,0 +1,135 @@
...
+    packet InterledgerPacket
+}
+
+FreeResponse ::= SEQUENCE {
+    -- Used to associate response to earlier FreeRequest
+    requestIdRef UInt128,
+    packet InterledgerPacket
+}
+
+PaidRequest ::= SEQUENCE {
```
sharafian: Why does there have to be a difference between prepare transfer and paid request? Couldn't you prepare a transfer and put a request in the side protocol data (or in the ILP packet, instead of an ILP payment packet)?

asn1/LedgerLedgerLotocol.asn
```diff
@@ -0,0 +1,135 @@
...
+PaidResponse ::= SEQUENCE {
+    -- Used to associate response to earlier PaidRequest
+    requestIdRef UInt128,
+    packet InterledgerPacket
+}
+
+PrepareRequest ::= SEQUENCE {
+    -- Used to associate later response/fulfill/reject
+    -- MUST be non-zero and non-repeating
+    requestId UInt128,
```
sharafian: I think we should separate requestId from transferId. Otherwise we can get confusing scenarios where many requests/responses are sent with the same ID, i.e. if someone tries to fulfill twice or fails to fulfill and then rejects

asn1/LedgerLedgerLotocol.asn
emschwartz: The `PaidRequest` is an unconditional transfer. It wouldn't make that much sense to put a protocol other than an Interledger Payment Packet on a conditional transfer. You would need to figure out another way for the recipient of it to know the fulfillment, which probably wouldn't provide a lot of value.

michielbdejong:
The reasons we discussed for having requestId on top of transferId are:
* payments are not an information request, they are a request for a remote action, and there is a difference between local and remote failure. A payment may succeed even if the local transfer is rejected.
* as a node that doesn't keep state, you still want to know if your fulfillment lead to an entry in the transfer log.
* some other reasons.

asn1/LedgerLedgerLotocol.asn
adrianhopebailie: A Request and a Transfer are different things (and should exist at different layers of the stack) so we MUST have separate IDs for them.

asn1/LedgerLedgerLotocol.asn
```diff
@@ -0,0 +1,135 @@
...
+IMPORTS
+    UInt8,
+    UInt32,
+    UInt64,
+    UInt128,
+    UInt256
+    FROM GenericTypes
+
+    Address,
+    Timestamp
```
sharafian: Should also import the InterledgerPacket type

PR #263, opened 2017-08-07
Rename LLL to Common Ledger Protocol

emschwartz:
(See commit messages)

Created with @sharafian @michielbdejong @dappelt

### DIFF
--- a/asn1/CommonLedgerProtocol.asn
+++ b/asn1/CommonLedgerProtocol.asn
@@ -40,7 +40,7 @@ CustomResponse ::= SEQUENCE {
 Prepare ::= SEQUENCE {
   transferId UInt128,
   amount UInt64,
-  condition UInt256,
+  executionCondition UInt256,
   expiresAt Timestamp,
   packet InterledgerPacket,
   -- Additional data for protocol extensibility
--- a/asn1/CommonLedgerProtocol.asn
+++ b/asn1/CommonLedgerProtocol.asn
@@ -72,23 +72,6 @@ RejectResponse ::= SEQUENCE {
   sideProtocolData SideProtocolData
 }
 
-TransferStateRequest ::= SEQUENCE {
-  transferId UInt128,
-  -- Additional data for protocol extensibility
-  sideProtocolData SideProtocolData
-}
-
-TransferStateResponse ::= SEQUENCE {
-  state CHOICE {
-    fulfillment UInt256,
-    rejectionReason InterledgerPacket,
-    prepared NULL,
-    nonExistent NULL
-  },
-  -- Additional data for protocol extensibility
-  sideProtocolData SideProtocolData
-}
-
 CustomRequest ::= SEQUENCE {
   -- Additional data for protocol extensibility
   sideProtocolData SideProtocolData
@@ -115,10 +98,8 @@ CallSet CALL ::= {
     {6 FulfillResponse} |
     {7 RejectRequest} |
     {8 RejectResponse} |
-    {9 TransferStateRequest} |
-    {10 TransferStateResponse} |
-    {11 CustomRequest} |
-    {12 CustomResponse}
+    {9 CustomRequest} |
+    {10 CustomResponse}
 }
 
 CommonLedgerProtocolPacket ::= SEQUENCE {
--- a/asn1/CommonLedgerProtocol.asn
+++ b/asn1/CommonLedgerProtocol.asn
@@ -124,8 +124,6 @@ CallSet CALL ::= {
 CommonLedgerProtocolPacket ::= SEQUENCE {
     -- One byte type ID
     type CALL.&typeId ({CallSet}),
-    -- Allow multiple accounts to be multiplexed on one connection
-    accountId UInt32,
     -- Used to associate requests and corresponding responses
     -- If requestId = 0, the server MUST not send a response
     requestId UInt32,
--- a/asn1/CommonLedgerProtocol.asn
+++ b/asn1/CommonLedgerProtocol.asn
@@ -1,4 +1,4 @@
-LedgerLedgerLotocol
+CommonLedgerProtocol
 DEFINITIONS
 AUTOMATIC TAGS ::=
 BEGIN
@@ -121,7 +121,7 @@ CallSet CALL ::= {
     {12 CustomResponse}
 }
 
-LedgerLedgerLotocolPacket ::= SEQUENCE {
+CommonLedgerProtocolPacket ::= SEQUENCE {
     -- One byte type ID
     type CALL.&typeId ({CallSet}),
     -- Allow multiple accounts to be multiplexed on one connection
--- a/asn1/CommonLedgerProtocol.asn
+++ b/asn1/CommonLedgerProtocol.asn
@@ -21,19 +21,23 @@ SideProtocolData ::= SEQUENCE OF SEQUENCE {
   protocolData OCTET STRING
 }
 
-PacketRequest ::= SEQUENCE {
+Ack ::= SEQUENCE {
+  ack NULL,
+  -- Additional data for protocol extensibility
+  sideProtocolData SideProtocolData
+}
+
+Response ::= SEQUENCE {
   packet InterledgerPacket,
   -- Additional data for protocol extensibility
   sideProtocolData SideProtocolData
 }
 
-PacketResponse ::= SEQUENCE {
-  packet InterledgerPacket,
-  -- Additional data for protocol extensibility
+CustomResponse ::= SEQUENCE {
   sideProtocolData SideProtocolData
 }
 
-PrepareRequest ::= SEQUENCE {
+Prepare ::= SEQUENCE {
   transferId UInt128,
   amount UInt64,
   condition UInt256,
@@ -43,42 +47,27 @@ PrepareRequest ::= SEQUENCE {
   sideProtocolData SideProtocolData
 }
 
-PrepareResponse ::= SEQUENCE {
-  -- Additional data for protocol extensibility
-  sideProtocolData SideProtocolData
-}
-
-FulfillRequest ::= SEQUENCE {
+Fulfill ::= SEQUENCE {
   transferId UInt128,
   fulfillment UInt256,
   -- Additional data for protocol extensibility
   sideProtocolData SideProtocolData
 }
 
-FulfillResponse ::= SEQUENCE {
-  -- Additional data for protocol extensibility
-  sideProtocolData SideProtocolData
-}
-
-RejectRequest ::= SEQUENCE {
+Reject ::= SEQUENCE {
   transferId UInt128,
   rejectionReason InterledgerPacket,
   -- Additional data for protocol extensibility
   sideProtocolData SideProtocolData
 }
 
-RejectResponse ::= SEQUENCE {
+Message ::= SEQUENCE {
+  packet InterledgerPacket,
   -- Additional data for protocol extensibility
   sideProtocolData SideProtocolData
 }
 
 CustomRequest ::= SEQUENCE {
-  -- Additional data for protocol extensibility
-  sideProtocolData SideProtocolData
-}
-
-CustomResponse ::= SEQUENCE {
-  -- Additional data for protocol extensibility
   sideProtocolData SideProtocolData
 }
 
@@ -88,18 +77,14 @@ CALL ::= CLASS {
 } WITH SYNTAX {&typeId &Type}
 
 CallSet CALL ::= {
-    -- All requests MUST use odd typeIds
-    -- All responses MUST use even typeIds
-    {1 PacketRequest} |
-    {2 PacketResponse} |
-    {3 PrepareRequest} |
-    {4 PrepareResponse} |
-    {5 FulfillRequest} |
-    {6 FulfillResponse} |
-    {7 RejectRequest} |
-    {8 RejectResponse} |
-    {9 CustomRequest} |
-    {10 CustomResponse}
+    {1 Ack} |
+    {2 Response} |
+    {3 CustomResponse} |
+    {4 Prepare} |
+    {5 Fulfill} |
+    {6 Reject} |
+    {7 Message} |
+    {8 CustomRequest}
 }
 
 CommonLedgerProtocolPacket ::= SEQUENCE {
--- a/asn1/CommonLedgerProtocol.asn
+++ b/asn1/CommonLedgerProtocol.asn
@@ -75,9 +75,7 @@ Message ::= SEQUENCE {
   sideProtocolData SideProtocolData
 }
 
-CustomRequest ::= SEQUENCE {
-  sideProtocolData SideProtocolData
-}
+CustomRequest ::= SideProtocolData
 
 CALL ::= CLASS {
     &typeId UInt8 UNIQUE,
@@ -98,7 +96,6 @@ CommonLedgerProtocolPacket ::= SEQUENCE {
     -- One byte type ID
     type CALL.&typeId ({CallSet}),
     -- Used to associate requests and corresponding responses
-    -- If requestId = 0, the server MUST not send a response
     requestId UInt32,
     -- Length-prefixed main data
     data CALL.&Type ({CallSet}{@type})
--- a/asn1/CommonLedgerProtocol.asn
+++ b/asn1/CommonLedgerProtocol.asn
@@ -14,6 +14,9 @@ IMPORTS
     Address,
     Timestamp
     FROM InterledgerTypes
+
+    InterledgerPacket
+    FROM GenericPacket
 ;
 
 SideProtocolData ::= SEQUENCE OF SEQUENCE {
@@ -21,11 +24,11 @@ SideProtocolData ::= SEQUENCE OF SEQUENCE {
   protocolData OCTET STRING
 }
 
-Ack ::= SEQUENCE {
-  ack NULL,
-  -- Additional data for protocol extensibility
-  sideProtocolData SideProtocolData
-}
+-- Ack and Response are the response types.
+-- When using these in a CLP packet, the requestId should match
+-- the requestId of the request they respond to.
+
+Ack ::= SideProtocolData
 
 Response ::= SEQUENCE {
   packet InterledgerPacket,
@@ -33,9 +36,14 @@ Response ::= SEQUENCE {
   sideProtocolData SideProtocolData
 }
 
-CustomResponse ::= SEQUENCE {
-  sideProtocolData SideProtocolData
-}
+-- Prepare, Fulfill, Reject, Message, and CustomRequest are the request types.
+-- They are sent through the ledger, to the connector.
+-- They will trigger one of the following responses:
+-- * no response, if requestId=0
+-- * an ilp-error, produced by the ledger (e.g. if the request is malformed)
+-- * an ilp-error, produced by the connector (e.g. if no quote was found)
+-- * an ilqp-response or other ILP-packet success response from the connector
+-- * an Ack, if there was no result or error to report
 
 Prepare ::= SEQUENCE {
   transferId UInt128,
@@ -79,12 +87,11 @@ CALL ::= CLASS {
 CallSet CALL ::= {
     {1 Ack} |
     {2 Response} |
-    {3 CustomResponse} |
-    {4 Prepare} |
-    {5 Fulfill} |
-    {6 Reject} |
-    {7 Message} |
-    {8 CustomRequest}
+    {3 Prepare} |
+    {4 Fulfill} |
+    {5 Reject} |
+    {6 Message} |
+    {7 CustomRequest}
 }
 
 CommonLedgerProtocolPacket ::= SEQUENCE {
--- a/asn1/GenericPacket.asn
+++ b/asn1/GenericPacket.asn
@@ -8,7 +8,7 @@ IMPORTS
     VarBytes
     FROM GenericTypes
 
-    InterledgerProtocolPaymentMessage,
+    InterledgerProtocolPayment,
     InterledgerProtocolError
     FROM InterledgerProtocol
 
--- a/asn1/InterledgerPaymentRequest.asn
+++ b/asn1/InterledgerPaymentRequest.asn
@@ -1,10 +1,10 @@
-InterledgerProtocol
+InterledgerPaymentRequest
 DEFINITIONS
 AUTOMATIC TAGS ::=
 BEGIN
 
 IMPORTS
-    UInt8
+    UInt8,
     UInt256
     FROM GenericTypes
 ;
--- a/asn1/InterledgerPaymentRequest.asn
+++ b/asn1/InterledgerPaymentRequest.asn
@@ -14,7 +14,7 @@ InterledgerPaymentRequest ::= SEQUENCE {
     version UInt8,
     -- Cryptographic condition
     condition UInt256,
-    -- ILP Payment Packet
+    -- InterledgerProtocolPayment
     packet OCTET STRING
 }
 
--- a/asn1/InterledgerQuotingProtocol.asn
+++ b/asn1/InterledgerQuotingProtocol.asn
@@ -1,9 +1,11 @@
-InterledgerQuotingProtocol DEFINITIONS AUTOMATIC TAGS ::=
+InterledgerQuotingProtocol
+DEFINITIONS
+AUTOMATIC TAGS ::=
 BEGIN
 
 IMPORTS
     UInt32,
-    UInt64,
+    UInt64
     FROM GenericTypes
 
     Address,


asn1/CommonLedgerProtocol.asn
```diff
@@ -124,8 +124,6 @@ CallSet CALL ::= {
 LedgerLedgerLotocolPacket ::= SEQUENCE {
```
michielbdejong: Rename here, too

emschwartz:
@sharafian and @michielbdejong please re-review. I added the commits to remove the `TransferStateRequest` and the response types corresponding to each request.

michielbdejong:
I did the same and arrived at #264, where I also did the `packet CHOICE { null NULL, packet InterledgerPacket}` which you suggested on Slack, to reduce to just 1 Request and 1 Response type.

emschwartz:
@michielbdejong: I was thinking of doing it that way but @sharafian made the very good point that including a CHOICE sends an extra byte over the wire, and it's definitely worth having an extra type defined to avoid that if it's such a minor difference.

asn1/CommonLedgerProtocol.asn
```diff
@@ -21,81 +21,53 @@ SideProtocolData ::= SEQUENCE OF SEQUENCE {
   protocolData OCTET STRING
 }
 
-PacketRequest ::= SEQUENCE {
-  packet InterledgerPacket,
+Ack ::= SEQUENCE {
+  ack NULL,
```
adrianhopebailie: In OER this will be encoded as zero bytes if there is no `sideProtocolData`. Is that the desired behaviour?

emschwartz: Yes as long as you know its an `Ack`. You really just need the type

michielbdejong:
The first thing I thought when I woke was damn, we're getting rid of OER inside JSON, and replacing it with key/value inside OER. Everything related to actual on-ledger transfers and to payment channel claim updates, as well as chaining the transfers into a sequential (or at least ordered) transfer log and calculating a current balance, as well as dispute reconciliation (deciding how a transfer affects the balance in different dispute cases), is going to go into SideProtocolData which will look like an array of key-value pairs, and unlike OER, still leaves room for field naming errors, e.g.:
```js
[
  ["transactionId", "65341"],
  ["account": "deprecated"],
  ["from": "alice"],
  ["to": "bob"],
  ["transaction_id": "oops, didn't mean to repeat this one"]
]
```
It will still be up to the side protocol to get that right. Wouldn't it be wise to be stricter about that format?

Why don't we just use the existing [data field](https://github.com/interledger/rfcs/blob/eb82f064aeaf530767791b7c979032d1d7399abd/asn1/InterledgerProtocol.asn#L40) in the case of errors  (maybe we should rename that field to extensions?), or the [extensions field](https://github.com/interledger/rfcs/blob/eb82f064aeaf530767791b7c979032d1d7399abd/asn1/InterledgerQuotingProtocol.asn#L49) in case of quotes?

asn1/CommonLedgerProtocol.asn
michielbdejong: ok, but then maybe we could just name them:
`Request`
`Response`
`RequestWithPacket`
`ResponseWithPacket`
`RequestWithSideData`
`ResponseWithSideData`
`RequestWithPacketAndSideData`
`ResponseWithPacketAndSideData`

michielbdejong: To relay last night's discussion, keeping request/response here even though the GenericPacket also already implements that is useful to a) have a single way to make it clear if the request sender requires a response or not, and b) avoid the situation where one side sends a response but the other side thinks it's a request, and then responds to it, after which both sides would get confused.

adrianhopebailie: > make it clear if the request sender requires a response or not
This is defined by the protocol. We shouldn't need a way to infer this from the request.

> avoid the situation where one side sends a response but the other side thinks it's a request, and then responds to it, after which both sides would get confused.

This is why we need a separation between logical request/response and message request/response. A logical response could be async and therefor packaged in a new message request.

emschwartz: With the messages defined here I don't see any ambiguity about the responses.

When you send any of `Prepare`, `Fulfill`, `Reject`, `Message`, you either get an `Ack` or a `Response` back. If you set the `requestId` in the `CommonLedgerProtocolPacket` to `0`, you are indicating you don't want that response. When you send a `CustomRequest` you'll probably get a `CustomResponse`, unless otherwise defined in the side protocol.

adrianhopebailie: > If you set the requestId in the CommonLedgerProtocolPacket to 0, you are indicating you don't want that response.

What's the purpose of this?

emschwartz:
@michielbdejong I think there's a good reason to put all of those things you mentioned into `sideProtocolData`. If you look at the existing plugins that use the `ilp-plugin-payment-channel-framework`, each of them implements only 1-2 extensions of the messaging protocol and the messages they send are very ledger-specific.

> actual on-ledger transfers

If these are done as settlement then there's little relation between the transfers sent over this protocol and the on-ledger transfers. If you want to add a notification to your peer to look out for a certain transfer, that's one message to add as part of a side protocol.

> payment channel claim updates

Again, you basically need one extra message for sending a claim, or you might want to attach it to the response to a `Fulfill` call.

> chaining the transfers into a sequential (or at least ordered) transfer log and calculating a current balance

Why would that be part of the messaging protocol? That's what the plugins or systems on either side will implement.

> dispute reconciliation

This is only an issue in cases where both sides are maintaining separate views on the balance. In that case, there are multiple ways to handle disputes that all call for different protocols. This ranges from using a notary to be the arbiter of timeouts, tracking disputed transfers and discussing them later, or just accepting losses and calling it even as long as it's within a "reasonable" threshold. Neither of these solutions is necessarily better and it remains to be seen which ones people will use in practice. I think it's far better to leave out these details now and let people add them in as necessary.

> It will still be up to the side protocol to get that right. Wouldn't it be wise to be stricter about that format?

Why would a side protocol be any less strict? I would imagine such side protocols often being defined on the level of all the people using a certain ledger. It would make sense to have a standard way to do Interledger over XRP, for example, and that should be written up in a spec. I still think it's useful to have the common elements pulled out so we can avoid duplicating effort and just say "this spec extends the Common Ledger Protocol v1.0".

> Why don't we just use the existing data field in the case of errors

This is already meant to express errors and when errors are sent back as `Response`s this should contain data related to what went wrong. I'm not sure what else you're suggesting.

> or the extensions field in case of quotes

OER extensions are designed to allow the protocol authors to add additional fields later that will be ignored by earlier implementations. It's not the same type of anyone-can-do-it extensibility that a key-value list gives you. If we realize all quote messages should actually have a field for X (for example, a Time to Live value, like @sharafian) suggested, we can add that using extensions. The `sideProtocolData` is much more flexible and can be used for things where we wouldn't ever want to add them to the core protocol (like ledger-specific details).

asn1/CommonLedgerProtocol.asn
emschwartz: This protocol can also be used in asymmetric cases where only one side is keeping the state. In such cases, the server may not care about whether the client responds or not to a `Fulfill` request. Setting the `requestId` to `0` basically turns it into a notification.

sharafian: http://www.jsonrpc.org/specification#notification JSON RPC has an identical mechanism

michielbdejong:
I think we have a different interpretation of "the protocol authors". CLP is a tool for the authors of more complete protocols, and they will only use CLP if they find it does what they need.

Say you and I peer with each other over the "acme-v5.4" protocol. This protocol, at this version string, has defined fields for strong typing and extensibility room for Duck typing. If I send you an extra field in the extensibility room then I don't know if you will understand it or not. In fact, I can be pretty sure that your acme-5.4 compatible software will probably just ignore those extra fields. This can be used for experimental data that piggybacks on the well-defined acme-5.4 messages. For instance, I may send you extra statistics data, and you have an additional process listening on the transcript of your acme-5.4 server, that allows you to get a view of the statistical performance of our trustline.

So at that level, I agree that an extra space for key-value data is useful.

But if we're the acme-5.4 protocol authors, we have a choice of basing acme-5.4 on CLP, or defining it from scratch. Of course we prefer to reuse what CLP already has, but we do want it to be OER all the way. It's not nice if our acme-5.4 protocol has some OER fields and some strings-inside-OER key-value fields. So we would prefer to add CLP extensions for e.g. `updatedBalance` at the OER level, not at the SideProtocol level.

In fact, as an acme-5.4 spec author, nobody is forcing me to base acme-5.4 on CLP, I'll just use it if it's a useful basis. I could even fork it for a trivial reason, like if I prefer to do `transfer.updatedBalance` instead of `transfer.extensions.updatedBalance`. Such a slight deviation from CLP would be hell for acme-5.4 implementers if they try to build acme-5.4 support on top of software modules that implement generic CLP-support, but in the end, their task is building acme-5.4 support, they don't care much whether acme-5.4 is correctly based on CLP or not.

sharafian:
Adding side protocol data is one way to extend CLP, but I agree that extending the OER is a valid way too. However, if you extend the OER you need to change the code in order to parse your extended format. This makes side protocol a more desirable way to add small extensions like claims and settlement, which could be implemented with a few changes on something like payment-channel-framework.

emschwartz:
Any ledger protocol author can definitely fork CLP, or design something completely different from scratch.

However, as we've found writing ledger plugins, it's much easier to use something like the [payment channel framework](https://github.com/interledgerjs/ilp-plugin-payment-channel-framework) than implement everything yourself. That framework will be changed to use the CLP as the base protocol and allow individual plugins to add custom messages (in the same way you can add custom RPC methods now) and custom data to existing messages (in the same way you can add custom arguments to existing RPC calls now). The benefit of allowing for multiple plugin types is that we don't have to standardize this layer. But as we've seen from experience it's way easier to implement plugins if you can reuse something like that framework that does 80% of the work for you (and avoids common gotchas).

asn1/CommonLedgerProtocol.asn
```diff
@@ -21,81 +21,53 @@ SideProtocolData ::= SEQUENCE OF SEQUENCE {
...
   -- Additional data for protocol extensibility
   sideProtocolData SideProtocolData
 }
 
-PacketResponse ::= SEQUENCE {
+Response ::= SEQUENCE {
   packet InterledgerPacket,
```
sharafian: the InterledgerPacket type hasn't been imported

asn1/CommonLedgerProtocol.asn
adrianhopebailie: We seem to be mixing layers a bit here. `Prepare`, `Fulfill` and `Reject` are logical operations. They relate to a transfer not a request/response pair, that should give us a clue. We should define their semantics separately from the semantics of matching requests and responses.

Sending a message that doesn't need a response isn't something you should have to define at the logical layer (and at the messaging layer, I'd argue it doesn't matter). If I send a request for which I don't need a response why does the server care?

If the logical interaction is async then I should get back an Ack response at the messaging layer (i.e. A request/response pair that only cover the logical request).

Then the async response can be covered in a new request/response pair at the messaging layer (initiated in the opposite direction).

Using a requestId of zero seems to have no obvious benefit it just raises a lot of questions for me:
- Are we expecting a server to not know if it's operating in an async or synchronous mode?
- Are we expecting the receiver to not understand the logical request and so not know if it must ack or respond?
- Are we expecting a client to throw an error if it gets an Ack and doesn't want it?
- How do I track down a request from the logs if all notifications have the same id (zero)?

emschwartz: If you look at the way the messages are defined, the logical operations are all defined as `Call`s. Then there is a wrapper that adds the request/response semantics.

The protocol that uses these messages should define "when you send a `Prepare`, you should expect an `Ack` and then later a `Fulfill` or `Reject` as a request".

If you get a request with a non-zero `requestId`, you must send a response or the other party will likely keep sending you the message. If the `requestId` is zero, they'll ignore your response if you send one.

> Are we expecting a server to not know if it's operating in an async or synchronous mode?

That is dependent on the protocol using these messages. In general I would expect all Message requests to be synchronous and all transfers to be asynchronous.

> Are we expecting the receiver to not understand the logical request and so not know if it must ack or respond?

They are expected to understand it. There's an argument to be made that we don't need this extra way to specify whether you need to send a response, but I think it's worth having in there.

> Are we expecting a client to throw an error if it gets an Ack and doesn't want it?

No. @sharafian already observed behavior where triggering an error when you get a message you don't expect can cause an infinite loop if the other side doesn't expect your error message.

> How do I track down a request from the logs if all notifications have the same id (zero)?

If it's a transfer notification you'd look for it by the `transferId`. For quotes you would set a `requestId` because you do want a response.

michielbdejong:
Yes, if you think of the plugin framework implementing this, and exposing direct access to the SideProtocolData, then the key-value setup does make sense.

asn1/CommonLedgerProtocol.asn
michielbdejong: I fixed a bunch of asn bugs including this one in https://github.com/interledger/rfcs/pull/267/files

justmoon:
>```
>[
>  ["transactionId", "65341"],
>  ["account": "deprecated"],
>  ["from": "alice"],
>  ["to": "bob"],
>  ["transaction_id": "oops, didn't mean to repeat this one"]
>]
>```

Just to be clear, the `sideProtocolData` expressed in JSON would actually be an object:

``` js
{
  "example-protocol-1.0": "..."
}
```

As for the data value - note that it can be arbitrary data, so it would have to be base64 if the protocol is unknown. We could add a mime type field like we did in older proposals of the ILP packet. (Could be a simple number from a fixed set of choices: 0 - `application/octet-string`, 1 - `text/plain; charset=utf-8`, 2 - `application/json`)

michielbdejong:
Other than logging/displaying the raw data I don't see what you could do with side protocol data if you know the mime type but not the meaning, so I would consider that as a sort of [XSLT](https://www.w3schools.com/xml/tryxslt.asp?xmlfile=cdcatalog&xsltfile=cdcatalog_choose) for pretty-printing side protocol data in logs.

Thanks for the pointer on array vs object, I also understand the role of side protocol data much better now, and agree with it.

I think we can merge this PR as soon as @adrianhopebailie signs off?

asn1/CommonLedgerProtocol.asn
```diff
@@ -0,0 +1,107 @@
...
+SideProtocolData ::= SEQUENCE OF SEQUENCE {
+  protocolName IA5String,
+  protocolData OCTET STRING
+}
+
+-- Ack and Response are the response types.
+-- When using these in a CLP packet, the requestId should match
+-- the requestId of the request they respond to.
+
+Ack ::= SideProtocolData
```
emschwartz: @michielbdejong Why isn't the `Ack` defined as a `SEQUENCE` like everything else? `CustomRequest` is also just a wrapper around the `SideProtocolData` but it's still defined as another `SEQUENCE`

michielbdejong: That's fixed in https://github.com/interledger/rfcs/pull/268/commits/9e0fa1be5dd4b90049c282a2c320db39ab103016

michielbdejong: It's also how clp-packet [implements it currently](https://github.com/interledgerjs/clp-packet/blob/master/index.js#L75-L80)

michielbdejong:
I thought we had agreement on #263, but the whole thing seems to be falling apart like a house of cards now. Two objections came:
* from Adrian (if I word it correctly), the messaging layer should not deal with local transfers, and the layer that does deal with those transfers, should also deal with quote requests (but not with *remote* payments)
* from Stefan (if I word it correctly), given GenericPacket (ILP/ILQP/Err) on top of CLP (request/response+accounting), there may be requests that can not be added to GenericPacket (because they are not Interledger-level), but that you do always want to have. The example in this case is `getAvailableLiquidity`.

We could either discard these objections and merge as-is, or try to accommodate them. What are we trying to build? When talking to my peer, I want to:
* `Request` send requests for information/action
* `Response` get information in response to requests/confirmation that the action was executed
* `Error` get errors reported

That's pretty basic and generic. Now we make it more money-specific, by saying CLP has built-in support for accounting a scalar, integer balance, and requests cost points (money):

* `Prepare` send requests for information/action, putting a number of points aside that I'm willing to pay
* `Fulfill` get information/confirmation in response to requests, points escrow is executed
* `Reject` get errors reported, points escrow is canceled

This is still generic, not Interledger-specific, we just added accounting. Now we add deadlines, and that messes things up a bit, because it requires an extra confirmation 'yes, your fulfill was on time'.

We discussed several ways of doing this, and in order to stay with universally familiar request/response semantics (instead of three-step request/response/confirmation), decided to add change 'Fulfill' into a proper request, and a few extra responses:

* `Request` send requests for information/action, expecting an immediate Response
* `Response` response to a request
* `Prepare` send requests for information/action, *temporarily* putting a number of points aside that I'm willing to pay
* `Fulfill` *send* information/confirmation referring to an earlier Prepare, points escrow is executed
* `Reject` *send* request to cancel points escrow, include error message
* `Ack` confirm that a Prepare/Fulfill/Reject was received and accepted successfully
* `Error` report an error in response to a Request, or to a Prepare/Fulfill/Reject

Now we have something which is still not specific to Interledger, although it's starting to look awefully similar to the five-bells-ledger API, as well as to the LedgerPluginInterface from RFC-0004.

The next question is, what requests for information and what requests for action would you send your peer? Until now, we said this aligns exactly with GenericPacket from the Interledger protocol itself: ask for a quote, ask for a remote payment, get back an ILP error. We said, you may want to ask your peer what the current points-accounting balance is, but that's a side protocol. But now we have a problem, because `getAvailableBalance` is a request that's too important to be a side protocol, and not important enough (or to be more precise, not universal enough) to become a GenericPacket. So this is where I would take a step back and ask, why does GenericPacket fulfill these two roles, of: 
1) listing packets that everybody who does Interledger needs to implement
2) a registry of requests for information and requests for action that can be sent over CLP

Maybe it's time to have a CLP request registry that's not restricted to Interledger-level request? So to fix this, we could create the `CLP-Request` and `CLP-response registries. The numbering between requests and responses is a bit tricky but would work as follows:
* 1, as well as every even number except 8, are reserved for requests
* 8, as well as every odd number except 1, are reserved for responses
* any number <=255 is reserved to exactly match the requests and responses in the InterledgerPacket registry.
* if any future InterledgerPackets break the odd/even rule, we will update these numbering rules accordingly for those numbers.

So now the `CLP-request` registry becomes:
* 1-65535, odd, except 1: reserved to avoid confusion with responses, and allow multiplexing
* 1 InterledgerProtocolPayment (only odd-numbered entry in this registry)
* 2 QuoteLiquidityRequest
* 4 QuoteBySourceAmountRequest
* 6 QuoteByDestinationAmountRequest
* 8 reserved for historical reasons
* 10-254: reserved for future request types defined in InterlegerPacket
* 256 AvailableLiquidityInformationRequest
* 258+, even: reserved for future *requests* (whether information or action).

And a `CLP-Response` registry:
* 0-65534, even, except 8: reserved to avoid confusion with requests, and allow multiplexing
* 1 reserved for historical reasons
* 3 QuoteLiquidityResponse from InterledgerPacket
* 5 QuoteBySourceAmountResponse from InterledgerPacket
* 7 QuoteByDestinationAmountResponse from InterledgerPacket
* 8 InterledgerProtocolError from InterledgerPacket (only even-number entry in this registry)
* 9-255, odd: reserved for future response types defined in InterledgerPacket
* 257 -- AvailableLiquidityInformationResponse
* 259+, odd:  reserved for future *responses* (whether success or error).

asn1/CommonLedgerProtocol.asn
emschwartz: It's strange that Ack would have a different structure than all the rest..

emschwartz:
> the whole thing seems to be falling apart like a house of cards now

I very much disagree with this characterization.

> from Adrian (if I word it correctly), the messaging layer should not deal with local transfers, and the layer that does deal with those transfers, should also deal with quote requests (but not with remote payments)

As we've discussed before, I think it's precisely the ability to do paid requests alongside unpaid ones that makes this protocol useful. Quoting is definitely not on this layer (as discussed [here](https://github.com/interledger/rfcs/pull/266#issuecomment-321300979) and [here](https://github.com/interledger/rfcs/pull/266#issuecomment-321384040)).

> from Stefan (if I word it correctly), given GenericPacket (ILP/ILQP/Err) on top of CLP (request/response+accounting), there may be requests that can not be added to GenericPacket (because they are not Interledger-level), but that you do always want to have. The example in this case is getAvailableLiquidity.

This seems like a perfect use case for the `CustomRequest`. If we find that every protocol using this base ends up adding some other request we can add those to version 1.1.

I think there's an argument for having a way to query your available liquidity but it's possible that we don't have enough information right now to make a final decision about whether it will be needed in all or most cases. So I think we can just implement it as a custom request for now and move on.

> Now we have something which is still not specific to Interledger, although it's starting to look awefully similar to the five-bells-ledger API, as well as to the LedgerPluginInterface from RFC-0004.

This is neither surprising nor a bad thing. The Five Bells Ledger API was originally designed to be a ledger API that provided everything we wanted from a ledger to do Interledger over it. However, we later realized that there were a couple of things that were either unneeded or bad design decisions:

- Crypto Conditions - replaced with simple sha256 conditions
- Separate transports for commands and notifications (HTTP + Websockets) - CLP uses one transport for both
- Multi-credit, multi-debit transfers - that was a feature we originally thought we wanted but quickly decided it was much more complicated than it was worth (which is why the Ledger Plugin Interface abstracted it away)

CLP is what you get when you take the Five Bells Ledger API and the Ledger Plugin Interface and apply everything we've learned since those were developed.

michielbdejong:
None of that really answers my point about the friction between 'request you can send over CLP' and 'interledger request'. As we discussed in person just now, a big registry that has methods from unrelated protocols would be spammable and hard to maintain. See https://github.com/interledger/rfcs/pull/269/files for a view on how currently the callset decides between Interledger protocol vs custom,and then custom has a registry based on strings.

adrianhopebailie:
In terms of everything at the internetworking layer I am concerned that we are conflating a whole bunch of stuff and trying to design a protocol stack like an API.

We're confusing encapsulation and layering with extraction of common elements into base classes.

Also I think our determination to not change the GenericPacket definitions and the InterledgerPaymentPacket are preventing us from designing this stack correctly.

For the lower layers I think CLP is great but have a few suggestions below.

The way I'd like us to look at this stack is very simply to start:

1. We have direct links between peers and we can **transfer value** and **send data** along those links. 

Those are different operations and trying to characterize them as free or paid data is just confusing (to me at least). Any layers below the internetworking layer should be 100% bilateral. Anything in the headers/framing **around** an internetworking packet MUST be assumed to be lost when the packet is re-framed for the next hop.

2. We have an internetworking layer where we have a new internetwork-scoped address space and a packet definition for encapsulating data from higher level protocols.

Packets at the internetworking layer MUST all have a common internetworking packet header which contains, at a minimum the source and destination address of the packet so that the packet can be transported across networks and the recipient is able to send a response back across networks.

We are missing the source address from our common header today which has forced us to push a whole lot of transport semantics down into the internetworking layer.

We also don't allow any but the payment packet to encapsulate higher level protocol data.

These are minor changes which I think, at worst will make the encapsulation and layering clearer.

(@michielbdejong I'm happy to keep ILQP in the internetworking layer based on the discussion yesterday but I think it requires some work first.)

3. We have no transport layer but we are looking for transport (and even application layer) semantics such as request/response matching.

(If we want these transport layer features below the internetworking layer then we must accept that they only exist bi-laterally. E.g. A requestId is only valid between 2 peered nodes.) 

In designing the CLP (a protocol at the layer below internetworking) I'd like us to re-use common patterns for messaging and request/response semantics and not pretend we know better than many years of work in this pretty generic area.

We should separate logical functions (transfer value, send data) from the request/response semantics (send a request, get a response) and plain old messaging (put a message on the wire, pick it on the other side).

Some concrete suggestions:

- Have a base packet that is just a message. i.e. It has an id a type and data. 
- Responses have a `correlationId` that is the `messageId` of the request message.
- It should be possible for the logical messages (Prepare, Fulfill, Error, Data) to be carried in either a request or response.

This is a cleaner separation of concerns which I think is more flexible and allows us to add new logical messages in future more easily.

What do you think of this:

```
CommonLedgerProtocolPacket ::= SEQUENCE {
  messageId UInt32,
  correlationId UInt32 OPTIONAL,
  -- One byte type ID
  -- Length-prefixed main data
  data CALL.&Type ({CallSet}{@type})
}

emschwartz:
Superseded by https://github.com/interledger/rfcs/pull/271

Issue #253, opened 2017-07-31
Alternative view of the Interledger layers


michielbdejong:
We think of Interledger as an ILP layer on top of a ledger layer. That means that for instance paychan settlement claims are more low-level than the sha256 hashlocks used by ILP, even if manual out-of-band settlements are not. It also means that `sendTransfer` is a first-class citizen of the RPC API, but `quote_by_source/dest/curve` and `broadcast_routes` are requests that get packaged inside other request.

But you could also think of ILP as a chain of transfers from one party to another, without defining if the next party in the chain is a ledger, an end user, or a connector.

From that perspective, I would say the **comms layer** is the lowest layer between two connectors, and is one of:
* five-bells-ledger
* json-rpc over https
* LLL
* carrier pigeons
* etc.

Then there's the **message layer**, that contains lots of different messages, for:
* "**ILP chaining**" (forward a transfer with 1) same condition, 2) equivalent amount minus your fee, 3) equivalent expiresAt minus your minMessageWindow)
* route broadcasts about ILP address prefixes, or otherwise
* various types of quote requests
* settlements via paychan claims
* unconditional settlements via other ledgers
* getBalance, getInfo, ...

And then on top of the ILP-chaining layer, the **address layer** is the single agreed ILP address space, i.e. the exact destination address string is also forwarded along the chain. But an implementation of the ILP chaining layer could support both our current address format and other ways of indicating a payment's destination.

And on top of that address layer is the end-to-end ILP **packet forwarding layer**, where not only the payment and the ILP address arrive at the destination during the prepare phase, but also the exact rest of the ILP packet that was included in the first transfer.

Not sure if we would want to do something with this, but I thought I would just write up my thoughts about it. :)

justmoon:
> That means that for instance paychan settlement claims are more low-level than the sha256 hashlocks used by ILP, even if manual out-of-band settlements are not.

I don't follow this point. Settlement claims and manual settlement are both part of the ledger layer I would say. The definition is that the ledger layer contains everything you need to transact with people on the same "ledger" as you. (The "ledger" could be a trustline.) Meanwhile, the Interledger layer contains protocols that allow you to transact with someone who isn't on your ledger. Of course, it's fine to use Interledger even for local transactions if doing so is convenient, but that doesn't change the fact that you *could* just use the ledger layer.

So as far as settlement goes - that is only between you and your peer, so clearly it's on the ledger layer.

michielbdejong:
[...]

PR #264, opened 2017-08-07
Reduce CLP to the French Five

michielbdejong:


michielbdejong:
Rebased on current version of #263

asn1/CommonLedgerProtocol.asn
```diff
@@ -21,22 +21,24 @@ SideProtocolData ::= SEQUENCE OF SEQUENCE {
   protocolData OCTET STRING
 }
 
-Ack ::= SEQUENCE {
-  ack NULL,
+Request ::= SEQUENCE {
+  packet CHOICE {
```
emschwartz: As commented in https://github.com/interledger/rfcs/pull/263#issuecomment-320779560, I think it's better to add an additional type than send an extra byte over the wire for every response.

adrianhopebailie: +1 - Having lots of types in the definition is not a bad thing.

adrianhopebailie:
I believe there is a layering issue in the set of generic packets that is making this harder than it should be. If you look at this list, the only packet that is relayed, unchanged, between nodes is the `InterledgerProtocolPayment`.

```
    {1 InterledgerProtocolPayment} |
    {2 QuoteLiquidityRequest} |
    {3 QuoteLiquidityResponse} |
    {4 QuoteBySourceAmountRequest} |
    {5 QuoteBySourceAmountResponse} |
    {6 QuoteByDestinationAmountRequest} |
    {7 QuoteByDestinationAmountResponse} |
    {8 InterledgerProtocolError}
```

This suggests that the `InterledgerProtocolPayment` should exist one layer higher in the stack and be encapsulated in a packet that is passed from node to node like the others.

We should remove `InterledgerProtocolPayment` and add `TransferRequest`, `SuccessfulTransferResponse` and `FailedTransferResponse` defined as follows:

```
TransferRequest ::= SEQUENCE {
  transferId UInt128,
  amount UInt64,
  executionCondition UInt256,
  expiresAt Timestamp,
  packet InterledgerProtocolPayment,
}

SuccessfulTransferResponse ::= SEQUENCE {
  transferId UInt128,
  fulfillment UInt256
}

FailedTransferResponse ::= SEQUENCE {
  transferId UInt128,
  reason InterledgerProtocolError
}
```

It's not clear if the ILQP requests can ever have an ILP Error response so let's assume that is only used in response to a transfer request.

Therefor the new set of Generic Packets are:  
```
    {1 TransferRequest} |
    {2 SuccessfulTransferResponse} |
    {3 FailedTransferResponse} |
    {4 QuoteLiquidityRequest} |
    {5 QuoteLiquidityResponse} |
    {6 QuoteBySourceAmountRequest} |
    {7 QuoteBySourceAmountResponse} |
    {8 QuoteByDestinationAmountRequest} |
    {9 QuoteByDestinationAmountResponse} 
```
This is the protocol layer that exists below ILP, the messages exchanged directly between nodes (connectors, senders and receivers). 

The transfer request encapsulates the `InterledgerProtocolPayment` (often referred to as the ILP Packet) along with the other required data for a transfer (expiry, condition etc).

Note that the `...Request` and `...Response` at this layer imply the logical flow not the message flow (which is captured in the next layer down, the CommonMessageProtocol). For example, a `TransferRequest` may be carried in a `RequestMessage` and the `ResponseMessage` may simply carry an `Ack`. Then the `SuccessfulTransferResponse` is returned, carried in a `RequestMessage` envelope, because it is initiated from the upstream node. 

The CommonMessageProtocol could then be reduced to:

```

Ack ::= SEQUENCE {
  -- The messageId of the request
  correlationId UInt32,
}

RequestMessage ::= SEQUENCE {
  packet InterledgerPacket,
}

ResponseMessage ::= SEQUENCE {
  -- The messageId of the request
  correlationId UInt32,
  packet InterledgerPacket,
}

CustomRequestMessage ::= SEQUENCE {
}

CustomResponseMessage ::= SEQUENCE {
  correlationId UInt32,
}

CALL ::= CLASS {
    &typeId UInt8 UNIQUE,
    &Type
} WITH SYNTAX {&typeId &Type}

CallSet CALL ::= {
    {1 Ack} |
    {2 RequestMessage} |
    {3 ResponseMessage} |
    {4 CustomRequestMessage} |
    {5 CustomResponseMessage} 
}

CommonMessageProtocolPacket ::= SEQUENCE {
    -- One byte type ID
    -- Used to associate requests and corresponding responses
    -- If messageId = 0, the server MUST not send a response
    messageId UInt32,
    -- Length-prefixed main data
    data CALL.&Type ({CallSet}{@type})
}
```

michielbdejong:
> I believe there is a layering issue in the set of generic packets that is making this harder than it should be.

I had exactly the same reaction on Friday. Evan started explaining why he wants to have it layered with separate "ILP-packet" layer and "ledger layer".

I had forgotten about this point, but now that Adrian brought it up independently, I agree that it would be simpler make the CLP layer simpler if it just deals with request-response and side-data. 

@emschwartz can you explain again why you don't want to put the prepare/fulfill/reject types next the quote/response/ILP-error types, in the GenericPacket registry?

michielbdejong:
I added back the TransferProtocol at the GenericPacket layer and now the CLP is really, really simple, just deals with linking responses to requests, and with request/response types that just predict whether it has a packet and/or SideData.

emschwartz:
> I believe there is a layering issue in the set of generic packets that is making this harder than it should be. If you look at this list, the only packet that is relayed, unchanged, between nodes is the InterledgerProtocolPayment.

That is not true. Quotes can be relayed as well (the amounts may or may not be changed but the message format stays the same) and errors definitely are.

> It's not clear if the ILQP requests can ever have an ILP Error response

The definitely can! Why wouldn't they? If you can't get a response, why not respond with an `F02: Unreachable` error?

This is turning into a discussion of how we think about the protocol layers so I'm going to respond to the general question in https://github.com/interledger/rfcs/issues/253.

adrianhopebailie:
> the amounts may or may not be changed but the message format stays the same

If anything changes it's not an end-to-end message like the InterledgerProtocolPayment. They shouldn't be in the same layer.

> If you can't get a response, why not respond with an F02: Unreachable error?

Encapsulated how? How does the sender know that the error corresponds to the request?

I have a counter-proposal to these designs which I'll put in a PR

emschwartz:
The Interledger layer is not meant to be end-to-end, it's the part that's understood and implemented by _every_ party in the path. Even in the case of the InterledgerProtocolPayment the address and amount are used by the connectors throughout the path. End-to-end means that the data is irrelevant and should be treated as completely opaque by everyone other than the sender and receiver. The Application and Transport layers, as we have currently defined them, are end-to-end.

In whatever Ledger layer protocol you use to talk to a connector we assume that there is a request/response flow (currently the quote requests and responses have IDs in them for correlation but we are leaning more towards deprecating those fields now). That's how you get the error back.

michielbdejong:
I still feel this way of organizing "CLP does the request/response, GenericPacket does the quoting/transfers" would have been more logical, and I think it would have been easier to understand for newcomers, but if you define CLP as a virtual ledger that supports Interledger, then Evan's latest proposal, with which Ben already agrees, is also defendable.

In the end, my main interest in this is to hope this is the last one in a series of rapid changes, and once we agree on CLP, we can enter relatively stable waters. I would still like to propose a freeze once this is merged, of several months where we don't change any of protocols.

This PR is now superseded by #267, which contains just a few minor edits to Evan's proposal.

PR #266, opened 2017-08-08
Add message layer 

adrianhopebailie:
Added a generic messaging layer to carry bi-lateral messages between nodes (GenericMessage.asn).

Modify logical request/response layer (GenericPacket.asn) to include transfers (InterledgerTransferProtocol.asn) and move payments up to Interledger layer (InterledgerProtocol.asn) which has only payments and error packets.

Layers are now clearly defined and encapsulated with the ledger layer having two sub-layers, a logical layer and a messaging layer.

```
Application
Transport
 - End-to-end transport specific (e.g. PSK)
Interledger
 - ILP Payment Packet/ILP Error
Ledger
 - Logical Transfer/Quote
 - Message
```

### DIFF
--- a/asn1/GenericMessage.asn
+++ b/asn1/GenericMessage.asn
@@ -1,0 +1,79 @@
+GenericMessage
+DEFINITIONS
+AUTOMATIC TAGS ::=
+BEGIN
+
+IMPORTS
+    UInt8
+    UInt128
+    UInt256
+    FROM GenericTypes
+
+    Timestamp
+    FROM InterledgerTypes
+;
+
+SideProtocolData ::= SEQUENCE OF SEQUENCE {
+  protocolName IA5String,
+  protocolData OCTET STRING
+}
+
+AckResponseMessage ::= SEQUENCE {
+  -- The messageId of the request
+  correlationId UInt32,
+}
+
+ErrorResponseMessage ::= SEQUENCE {
+  -- The messageId of the request
+  correlationId UInt32,
+  -- The error data (Could define a struct here for messaging errors?)
+  error OCTET STRING (SIZE (0..8192))
+
+  sideProtocolData SideProtocolData
+}
+
+InterledgerRequestMessage ::= SEQUENCE {
+  packet InterledgerPacket,
+  sideProtocolData SideProtocolData
+}
+
+InterledgerResponseMessage ::= SEQUENCE {
+  -- The messageId of the request
+  correlationId UInt32,
+  packet InterledgerPacket,
+  sideProtocolData SideProtocolData
+}
+
+GenericRequestMessage ::= SEQUENCE {
+  sideProtocolData SideProtocolData
+}
+
+GenericResponseMessage ::= SEQUENCE {
+  correlationId UInt32,
+  sideProtocolData SideProtocolData
+}
+
+MESSAGE ::= CLASS {
+    &typeId UInt8 UNIQUE,
+    &Type
+} WITH SYNTAX {&typeId &Type}
+
+MessageSet MESSAGE ::= {
+    {1 AckResponseMessage} |
+    {2 ErrorResponseMessage} |
+    {3 InterledgerRequestMessage} |
+    {4 InterledgerResponseMessage} |
+    {5 GenericRequestMessage} |
+    {6 GenericResponseMessage} 
+}
+
+MessagePacket ::= SEQUENCE {
+    -- Unique identifier for the message
+    messageId UInt32,
+    -- One byte type ID
+    type MESSAGE.&typeId ({MessageSet}),
+    -- Length-prefixed message data
+    data MESSAGE.&Type ({MessageSet}{@type})
+}
+
+END
--- a/asn1/GenericPacket.asn
+++ b/asn1/GenericPacket.asn
@@ -8,16 +8,18 @@ IMPORTS
     VarBytes
     FROM GenericTypes
 
-    InterledgerProtocolPayment,
-    InterledgerProtocolError
-    FROM InterledgerProtocol
+    TransferRequest,
+    SuccessfulTransferResponse,
+    FailedTransferResponse
+    FROM InterledgerTransferProtocol
 
     QuoteLiquidityRequest,
     QuoteLiquidityResponse,
     QuoteBySourceAmountRequest,
     QuoteBySourceAmountResponse,
     QuoteByDestinationAmountRequest,
-    QuoteByDestinationAmountResponse
+    QuoteByDestinationAmountResponse,
+    FailedQuoteResponse
     FROM InterledgerQuotingProtocol
 ;
 
@@ -27,14 +29,16 @@ PACKET ::= CLASS {
 } WITH SYNTAX {&typeId &Type}
 
 PacketSet PACKET ::= {
-    {1 InterledgerProtocolPayment} |
-    {2 QuoteLiquidityRequest} |
-    {3 QuoteLiquidityResponse} |
-    {4 QuoteBySourceAmountRequest} |
-    {5 QuoteBySourceAmountResponse} |
-    {6 QuoteByDestinationAmountRequest} |
-    {7 QuoteByDestinationAmountResponse} |
-    {8 InterledgerProtocolError}
+    {1 TransferRequest} |
+    {2 SuccessfulTransferResponse} |
+    {3 FailedTransferResponse} |
+    {4 QuoteLiquidityRequest} |
+    {5 QuoteLiquidityResponse} |
+    {6 QuoteBySourceAmountRequest} |
+    {7 QuoteBySourceAmountResponse} |
+    {8 QuoteByDestinationAmountRequest} |
+    {9 QuoteByDestinationAmountResponse} |
+    {10 FailedQuoteResponse}
 }
 
 InterledgerPacket ::= SEQUENCE {
--- a/asn1/InterledgerPaymentRequest.asn
+++ b/asn1/InterledgerPaymentRequest.asn
@@ -1,4 +1,4 @@
-InterledgerProtocol
+InterledgerPaymentRequest
 DEFINITIONS
 AUTOMATIC TAGS ::=
 BEGIN
--- a/asn1/InterledgerQuotingProtocol.asn
+++ b/asn1/InterledgerQuotingProtocol.asn
@@ -1,4 +1,5 @@
-InterledgerQuotingProtocol DEFINITIONS AUTOMATIC TAGS ::=
+InterledgerQuotingProtocol 
+DEFINITIONS AUTOMATIC TAGS ::=
 BEGIN
 
 IMPORTS
@@ -10,12 +11,16 @@ IMPORTS
     Timestamp,
     LiquidityCurve
     FROM InterledgerTypes
+
+    InterledgerProtocolError
+    FROM InterledgerProtocol
 ;
 
 -- Request to receive liquidity information between the current ledger and the
 -- destination account. This information is sufficient to locally quote any
 -- amount until the curve expires.
 QuoteLiquidityRequest ::= SEQUENCE {
+    quoteId UInt128,
     destinationAccount Address,
     -- How much time the receiver needs to fulfill the payment (in milliseconds)
     destinationHoldDuration UInt32,
@@ -27,6 +32,7 @@ QuoteLiquidityRequest ::= SEQUENCE {
 }
 
 QuoteLiquidityResponse ::= SEQUENCE {
+    quoteId UInt128,
     -- Curve describing the liquidity (relationship between input and output
     -- amounts) for the quoted route
     liquidity LiquidityCurve,
@@ -53,6 +59,7 @@ QuoteLiquidityResponse ::= SEQUENCE {
 
 -- Quoting with a specified source amount to determine destination amount
 QuoteBySourceAmountRequest ::= SEQUENCE {
+    quoteId UInt128,
     destinationAccount Address,
     sourceAmount UInt64,
     -- How much time the receiver needs to fulfill the payment (in milliseconds)
@@ -65,6 +72,7 @@ QuoteBySourceAmountRequest ::= SEQUENCE {
 }
 
 QuoteBySourceAmountResponse ::= SEQUENCE {
+    quoteId UInt128,
     -- Amount that will arrive at the receiver
     destinationAmount UInt64,
     -- How long the sender should put money on hold (in milliseconds)
@@ -78,6 +86,7 @@ QuoteBySourceAmountResponse ::= SEQUENCE {
 
 -- Quoting with a specified destination amount to determine source amount
 QuoteByDestinationAmountRequest ::= SEQUENCE {
+    quoteId UInt128,
     destinationAccount Address,
     destinationAmount UInt64,
     -- How much time the receiver needs to fulfill the payment (in milliseconds)
@@ -90,11 +99,22 @@ QuoteByDestinationAmountRequest ::= SEQUENCE {
 }
 
 QuoteByDestinationAmountResponse ::= SEQUENCE {
+    quoteId UInt128,
     -- Amount the sender needs to send based on the requested destination amount
     sourceAmount UInt64,
     -- How long the sender should put money on hold (in milliseconds)
     sourceHoldDuration UInt32,
 
+    -- Enable ASN.1 Extensibility
+    extensions SEQUENCE {
+        ...
+    }
+}
+
+FailedQuoteResponse ::= SEQUENCE {
+    quoteId UInt128,
+    error InterledgerProtocolError
+    
     -- Enable ASN.1 Extensibility
     extensions SEQUENCE {
         ...
--- a/asn1/InterledgerTransferProtocol.asn
+++ b/asn1/InterledgerTransferProtocol.asn
@@ -1,0 +1,38 @@
+InterledgerTransferProtocol
+DEFINITIONS
+AUTOMATIC TAGS ::=
+BEGIN
+
+IMPORTS
+    UInt64,
+    UInt128,
+    UInt256
+    FROM GenericTypes
+
+    Timestamp
+    FROM InterledgerTypes
+    
+    InterledgerProtocolPayment,
+    InterledgerProtocolError
+    FROM InterledgerProtocol
+;
+
+TransferRequest ::= SEQUENCE {
+  transferId UInt128,
+  amount UInt64,
+  executionCondition UInt256,
+  expiresAt Timestamp,
+  packet InterledgerProtocolPayment,
+}
+
+SuccessfulTransferResponse ::= SEQUENCE {
+  transferId UInt128,
+  fulfillment UInt256
+}
+
+FailedTransferResponse ::= SEQUENCE {
+  transferId UInt128,
+  error InterledgerProtocolError
+}
+
+END


michielbdejong:
@emschwartz According to your layer diagram, say we could agree that prepare/fulfill/reject are more about 'chains' than about 'value' (a ledger could do unconditional settlements below the sha256 chaining layer), then they fall within what we call Interledger, and not within what we call ledger, so it would still have been correct to add their asn1 definition [at the Interledger level](https://github.com/interledger/rfcs/blob/master/asn1/InterledgerProtocol.asn) (but I'm also OK with adding them [at the Ledger level](https://github.com/interledger/rfcs/blob/fd58ced40550d10eb202d5e4d1dcf7ba208c11c0/asn1/CommonLedgerProtocol.asn#L48-L70), as you propose)

adrianhopebailie:
> I think part of the confusion is coming from the filename `GenericPacket.asn`. Those are the packets on the Interledger layer. That file should be renamed to `InterledgerPacket.asn` (that's already what's defined in the file).

Actually I disagree. I don't think that they are part of the Interledger layer and the only reason we think of them that way is because the IPP is in there. More below...

> ILQP is on the Interledger layer, not on the Ledger layer.

Why? I don't think that is correct. From an encapsulation perspective, the ILQP messages are at the same level as Prepare, Fulfill and Reject transfers.

The fact that we have a "standard" for quoting message formats does not imply that they are end-to-end messages in the same way that an InterledgerPaymentPacket is. The way to measure this is, would it be okay for a node to modify a field in an ILQP message before forwarding it on? The answer is yes for all of the ILQP messages, and no for the IPP.

> Just because we're now thinking of building connectors that also act as ledgers doesn't mean we should break the abstractions.

That's not the case. I have suggested for some time that the current abstractions are not right but looking at the full stack in ASN.1 helps to illustrate my point.

> Ledgers convey value and data. That's it. If ledgers don't natively provide functionality for sending messages we can use the Internet for that instead, but we should assume that the ledgers of the future will provide such functionality if that turns out to be necessary.

Thinking about ledgers is the problem. Interledger is actually not about the ledgers but about the nodes in-between (as we've discovered).

The bilateral protocol I am proposing here (the logical messages currently defined in `GenericPacket.asn` and message envelopes in `GenericMessage.asn`) is for communication between nodes (connectors. senders and receivers).

Whether there is a ledger in the middle or not is not material to this abstraction. In some respects the name ledger layer is actually misleading. This layer is where value is transferred and data is exchanged directly between nodes over payment channels, trustlines, payment networks or whatever you want to call them.

This is my view of the layers:

```
         Functionality                  Layer

+-----------------------------+  --+
|   End-to-End Payment Data   |    |   End-to-End
+-----------------------------+    +- ("Application" +
|     End-to-End Security     |    |   "Transport")
+-----------------------------+  --+
|   ILP Packet | ILP Errors   |    |
+-----------------------------+    |
|        ILP Addresses        |    +- Interledger
+-----------------------------+    |
|     Sha256 HTLA Chains      |    |
+-----------------------------+  --+
| Transfer |    Quote         |    +- Ledger
+-----------------------------+  --+
| Messages | Messaging Errors |    +- Messaging
+-----------------------------+  --+
```

> One could argue that ILQP is a bilateral concern, not an Interledger concern, but I strongly disagree.

As a general point on this, an internetworking protocol like ILP depends on bi-lateral protocols that are not "in the stack". In fact, the gateway protocols (as they are called in the Internet protocol stack) are all classified as Application Layer protocols. 

These gateway protocols depend on the addressing scheme from the internetworking protocol so it makes sense that they exist in higher layers. These protocols do not exist in the internetworking layer.

> First of all, ILQP requests contain Interledger addresses, so they are clearly built on top of that part of the stack in the diagram. 

We shouldn't confuse dependancies with protocol layering they are not the same. ILQP messages do contain ILP Addresses but they are not addressed using them. Since transporting an ILQP packet is free (i.e. it isn't carried on the back of a transfer) it's possible that the route a quote packet takes is very different to the one a payment takes. 

It's very possible that a node never routes quotes and simply queries some service each time it gets a quote and responds using the data it receives.

> Second, ILQP will need to be implemented by every client library and connector. No matter how connectors choose to respond to ILQP requests (from local information or remote), they still need to understand the protocol. If you can forward ILP Payments but haven't implemented ILQP, you are not an Interledger connector. The reason to define it this way is because we need to draw a line somewhere and say "this is Interledger". The set of functionality in the Interledger layer in the diagram above is arguably the minimum you need in order to usefully interact with a system like this.

This is conflating standardization with protocol layering. It is very helpful to have a standard for the bilateral quote messages so that everyone can adopt that and save themselves figuring out how to exchange route and liquidity data.

But, just because ILQP is part of the Interledger standard doesn't make it part of the internetworking layer in the protocol stack.

I can deploy a node that peers with other nodes and we can choose not to use ILQP and this will go unnoticed to the rest of the world. But when I send a transfer I still need it to contain a valid ILP Packet that is passed on by those peers.

michielbdejong:
My summary of last night's discussion with Evan:

* Evan's current CLP proposal is directly based on the LPI (RFC 0004) in the sense that there are ledger-related methods, and then "sendMessageThroughLedger". This modeling makes sense if you always view the API of an Interledger node as "you are talking through a ledger, to a connector". It's because we look at "nodes" with our "we're connecting ledgers" glasses on. We find it easy to model many people connected by one ledger, but when a connection is one-to-one, we have to imagine a "virtual ledger" there.
* You can also imagine a person who has "network of trust relationships" glasses on. They will find it easy to think about one-to-one connections, but when a number of people are connected via a (passive) ledger, they have to imagine a "virtual connector" there, to which each user has a trust relationship, and say "I have a trustline balance at the connector that forms the imaginary center of the blockchain, and this connector forwards payments to other people", even if that ledger with which you have a one-to-one trust relation is a passive blockchain, and doesn't have a real conversation party at its center all.
* It's hard to choose which glasses we should put on, because it depends on the long term, will one-to-one (connector-first view) relations be the common case, or will hub-and-spoke (ledger-first view) relations be the common case?
* Adrian's point about messaging as a layer is actually very close to Evan's ledger-first view if you just replace "Sending Value" with "Sending Paid Message" and replace "Sending Data" with "Sending Unpaid Messages":
```
         Functionality                  Layer
+-----------------------------+  --+
|      End-to-End Setup       |    |   End-to-End
+-----------------------------+    +- ("Application" +
|  Condition and Data Format  |    |   "Transport")
+-----------------------------+  --+
| IPP | ILQP | Err | Anything |
+-----------------------------+    |
|        ILP Addresses        |    +- Interledger
+-----------------------------+    |
|     Sha256 HTLA Chains      |    |
+-----------------------------+  --+
|    Paid    |     Unpaid     |    +- Payment-Aware Messaging Hub aka Ledger
+-----------------------------+  --+
```

* Maybe a ledger should be seen as a "messaging hub that supports paid messages", and this would make the current proposal in #263 very future proof (notice I also added 'anything' next to IPP).
*As Stefan said, Adrian is probably right that ILQP is too high up there depending on how you look at it, but I think that's a relatively minor issue if we can merge "Ledger" and "Messaging" into one.

emschwartz:
I think part of our debate is about definitions but here is what I think the heart of the matter is (this is coming from the conversation @michielbdejong and I had yesterday):

There are 3 very different things we could be building here:

## Interledger as a Settlement System

This is the view that there are ledgers that keep track of value and when you want to pay someone, you make a transfer over a ledger to them. This view is why the project is called "Interledger". This is a settlement system because at the end of a transaction there is no debt between any of the parties.

In this model it makes sense that everything would be built on top of ledgers and ledger protocols. The abstraction that you want for the ledger layer would do everything you expect your ideal ledger to do, from conditional transfers to messages. That is why we have the "Ledger Plugin Interface", why having something like a "Common Ledger Protocol" makes sense, and why we think of trustlines as an unusual type of ledger.

## Interledger as a Clearing System

The second view (and I think this is where @adrianhopebailie and @michielbdejong were coming from) is that what we are primarily building is a payment clearing system. In that view, the connectors are the central actors. The normal case is that connectors have some credit relationship or trustlines between one another and Interledger payments go through those. The unusual case is where you send every transfer through some ledger.

In this view, it makes sense to standardize the clearing protocol and say that consists of transfers, fulfillments, rejections, quotes, route broadcasts, and whatever else connectors need to communicate with one another to clear payments. Those would not need to change depending on the settlement mechanism, because how you settle is completely divorced from clearing payments. The abstraction would be one that implements all of those methods. In this case, when you are dealing with a ledger you would treat it as if there were an imaginary connector in the middle that you're talking to and you'd hack the functions it doesn't provide natively on top (credit to @michielbdejong for this way of looking at it).

## The Internet with Value

The last option is to say we are not building the "Internet of Value" but the "Internet _with_ Value".

The bottom layer of this stack is the ability to send free requests, paid requests, and conditionally paid requests between connectors. On top of that there is the ILP Address namespace that is used to forward packets from multiple different protocols to the destination.

In this view, Interledger payments are just one application built on top of conditionally paid requests. You could also add the ability to relay messages. If connectors will relay messages for free, then you can build the same Internet stack we have today on top of that and we'll just have replaced IP Addresses with ILP Addresses (side note: it's an interesting question whether the possibility for address collisions would make these addresses unviable for this use case). You could also relay messages over paid requests, in which case you would have a kind of paid Internet where spam wouldn't be as much of an issue because you have to pay to send traffic through the network. And you could build systems where you want to be passing messages but you do that via Interledger payments so that everyone along the chain is incentivized to make sure the packet is delivered because they only get paid if it is.

One interesting thought on this note is that you could actually use the same ILP Packet for all of the use cases. If you're sending data primarily you could just use an ILP Packet with an amount of 0, and then the other values are just the ILP Address and the data.

One way of looking at this is that actually the main thing we've done is created a system parallel to the Internet, that is kind of a replacement for it, which uses ILP Addresses instead of IP Addresses. You could even do strange things like having DNS records resolve to both an IP and ILP address and then your browser could fetch a website either over the normal Internet or the paid Internet. An argument here is that the reason IPv6 didn't really take off was because it offered relatively minor improvements over IPv4. In contrast, ILP introduces the ability to send value as well as information and the ILP Addresses are much more flexible even than IPv6 addresses.

...

:thinking:

emschwartz:
@michielbdejong also made a good point that for both the first and the third options you would have the same bottom layer, with the one exception that you might want to add unconditional transfers.

emschwartz:
> you could actually use the same ILP Packet for all of the use cases. If you're sending data primarily you could just use an ILP Packet with an amount of 0

Instead of sending an unnecessary 8 bytes in every message it might be better to just define a different `InterledgerMessage` type that doesn't have an amount.

sharafian:
> ILQP messages do contain ILP Addresses but they are not addressed using them. Since transporting an ILQP packet is free (i.e. it isn't carried on the back of a transfer) it's possible that the route a quote packet takes is very different to the one a payment takes.
> 
> It's very possible that a node never routes quotes and simply queries some service each time it gets a quote and responds using the data it receives.

@adrianhopebailie I disagree with your assertion that ILQP and ILP are different; I explained my reasoning on the community call but I'll write it here again. My basic idea is that ILQP and ILP are both requests for some information, and the ILP address is a hint on what direction to go to get this message.

Imagine the path of a quote:

1. The sender quotes the first connector, asking how to get to `example.red.alice`. (this could be by source, destination, or liquidity; all follow the same flow)
2. The first connector doesn't know the rate to `example.red.alice`, so it forwards the quote to the second connector. If they _did_ know the rate, they could respond here without passing on.
3. The second connector has a quote cached, so they respond.
4. The first connector passes the response back to the sender
5. The sender's request for a fulfillment has been answered

This doesn't go all the way through the path, true, but the all the sender asked for was a quote response. They were _not_ asking for a message to be delivered to the recipient.

An interledger payment is not asking for a message to be delivered either, it's just asking for an answer to the payment's condition. The sender believes that only the receiver can fulfill the condition, but the protocol just passes the payment along until somebody knows the fulfillment.

1. The sender sends a transfer to the first connector, with the destination address `example.red.alice`
2. The first connector doesn't know the fulfillment, so they forward the payment to the second connector. If they _did_ somehow know the fulfillment, they could fulfill it here. (This would not break the sender's guarantees; they asked for a fulfillment and got one)
3. The second connector doesn't know the fulfillment, so they pass it on to the receiver. If they _did_ know the fulfillment, they could fulfill it here.
4. The receiver gets the transfer, knows the fulfillment, and responds.
5. The connectors pass the fulfillment back along the chain
6. The sender's request for a fulfillment has been answered

emschwartz:
Two concrete examples of what @sharafian describes with connectors knowing fulfillments without sending the payment to the destination:

- [Optimistic over Universal](https://lists.w3.org/Archives/Public/public-interledger/2017Jun/0033.html), where the condition is a well-known hash that most connectors know the preimage of without asking anyone else.
- Retries through the same path. If you're a sender and the transfer times out you can retry safely with the same condition. If the payment goes through the same path, a connector who lost money the first time because they failed to deliver the fulfillment to the previous ledger in time can fulfill the incoming transfer without doing any remote operations.

adrianhopebailie:
@sharafian I consider ILQP and ILP packets different because an ILQP packet is not "addressed" using an ILP address. (Some of the packets could be with a few tweaks but I'll look at that separately.)

The interledger layer is the internetworking layer which, like the internet layer in the Internet stack, has an address namespace and defines a packet header for ILP packets where that address is used.

If you look at the internet stack, all protocols that consider themselves part of the internet layer have an IP header.

As we have said before, the internetworking **layer** is VERY simple. That said, IP on it's own is close to useless without a bunch of other protocols that allow hosts to communicate state, error and control messages to one another. So we need ILQP (or some form of quoting protocol) but it's not part of the internetworking layer unless we change the packets to be addressed and routed using ILP Addresses.

emschwartz:
@adrianhopebailie here's another way of thinking about it.

When designing the Interledger header we looked at the fields in the IP header for inspiration (destination address, source address, time to live, checksum, protocol, options, etc). And we decided that only one of those fields was strictly necessary and that we could make do without all of the other ones. That one was the destination address.

So the proper Interledger header contains only the destination address and then data corresponding to another protocol that it carries with it towards the destination.

We could have defined Interledger payments, quotes, and messages as riding on top of the Interledger header by looking at each like this:

```
+---------------------------+
| Interledger               |
|    destinationAddress     |
|    +----------------------|
|    | ILP Payment          |
|    |    destinationAmount |
|    |    data              |
+---------------------------+
```

```
+-------------------------+
| Interledger             |
|    destinationAddress   |
|    +--------------------|
|    | ILQP by source     |
|    |    sourceAmount    |
+-------------------------+
```
et cetra.

In each of these cases, the protocol rides on top of Interledger so that the packets are conveyed towards the `destinationAddress`.

However, since the ILP header has only one field it doesn't really make sense to make it a proper header that wraps the other protocol data. Instead, we just include that one field in the other protocol's message.

By default, you must assume that quotes go end to end. Unless you think it is reasonable for nodes to have a 100% complete map of the network, you will always need to have remote quoting as a fallback. If it is the method of last resort, then everyone must implement it and be prepared to do remote quoting if they don't have an answer stored locally. And if this is the case, all other ways of answering quote requests are just optimizations on top of the fundamental assumption that quotes go all the way down the path.

emschwartz:
Another related observation:

All Internet layer protocols rely on IP addresses for routing the packets where they need to go. 

There are a bunch of packet types on the Interledger layer that do not have ILP addresses in them (ILQP responses and errors). This is because we have assumed the bottom layer provides request/response semantics. This is an interesting difference with the Internet, where you assume only one-way communication and you attach a source address so the destination party can send a message back to you.

It remains to be seen whether expecting the lowest layer to handle request/response semantics (in addition to supporting simple one-way messages) is a good idea, but it does seem like another slight tweak that could turn out to be pretty powerful.

adrianhopebailie:
@emschwartz Yes and yes! I was just writing an addendum to my response that aligns with your thinking.

1. We should have a common header for all packets
2. We should not be mixing request/response semantics into this layer. Even simple ACK based protocols in the Internet stack (like TCP) exist at higher layers.

I'd add that currently we have a packet type indicator as part of the header too so it's not just the address. I definitely prefer the packet definition's above and think we should adjust our ASN.1 to align with that.

> However, since the ILP header has only one field it doesn't really make sense to make it a proper header that wraps the other protocol data. Instead, we just include that one field in the other protocol's message.

I think stripping everything except the `destinationAddress` was perhaps a premature optimization. As we're discovering, not including the source address places some onerous requirements on the lower layers (specifically the need to keep state to support request/response semantics).

This is in contrast to the Internet stack where this is provided by higher layers.

What is the motivation for not having a source address and even source amount?

I think we can be more optimized than a completely generic IPv6 header as we have done by having a set of known packet types which are effectively like defining extensions in IPv6 but instead packaging them as different packet types.

I also think we should have a `data` payload in all packets.

michielbdejong:
Wait, you are now talking about changing the InterledgerProtocolPayment (aka the ILP packet). That's the thing that in yesterday's meeting we mentioned as the most 'frozen' part of the whole stack. Let's dial this back. Continuing this comment in #263.

emschwartz:
> I'd add that currently we have a packet type indicator as part of the header too so it's not just the address.

True. Still don't see a problem with including those two values directly in the packets of all of the Interledger layer protocols.

> I definitely prefer the packet definition's above and think we should adjust our ASN.1 to align with that.

Unfortunately that would mean a breaking change to every protocol so even if I liked the idea in principle I would very strongly oppose making such a change now.

> What is the motivation for not having a source address and even source amount?

The source amount is determined by the amount you put on hold on the first ledger. Including it in the packet is unnecessary.

As far as source address goes, it is also unnecessary for routing the packet to its destination. Each intermediary needs to remember all their outgoing prepared transfers to be able to send the fulfillment backwards. When we added the ability to reject incoming prepared transfers that also provided a way to send error messages back, which is arguably the main purpose in having a source address in the IP packet.

> I also think we should have a data payload in all packets.

The ILP Payment Packet already has. What purpose would that serve in the quote requests or responses?

The CLP has the ability to attach side protocol data to all requests, which you can use to send extra information that may be used in quoting, routing, etc. Successive connectors will need to understand that side protocol to know if they are expected to relay the data.

adrianhopebailie:
>Unfortunately that would mean a breaking change to every protocol so even if I liked the idea in principle I would very strongly oppose making such a change now.

Why? The hundreds of implementations that would need to change? Think carefully about the real cost of changing any of this stuff and you'll see it's pretty low.

The mental cost of us having to accept a change to something that we thought was finalized shouldn't be a consideration.

> The source amount is determined by the amount you put on hold on the first ledger. Including it in the packet is unnecessary.

Unless the packet is a quote by source amount? So depending on the packet type the presence or not of the amount (source or destination) is optional. Do we agree? So amount is not part of the ILP header, it's part of the packet specific header.

> As far as source address goes, it is also unnecessary for routing the packet to its destination.

The same could said for IP and yet it's there. It's purpose is not for routing the packet its to tell the recipient how to address the response(s).

> Each intermediary needs to remember all their outgoing prepared transfers to be able to send the fulfillment backwards. 

That is a major deviation from the principles of the Internet protocol. Now every node must maintain state about each packet it has routed so it knows how to route a response.

That is a transport layer consideration.

> When we added the ability to reject incoming prepared transfers that also provided a way to send error messages back, which is arguably the main purpose in having a source address in the IP packet.

Not at all. The source address is there so that any packet sent in response can be addressed to the correct sender. That can be fulfillments, payments, errors, quotes etc.

> The ILP Payment Packet already has. What purpose would that serve in the quote requests or responses?

The same reason you have a payload in the ILP Payment packet. So you can define protocols that layer on top of it. We have already had use cases where application layer data needs to be conveyed in a quote so there is clearly a use case.

I'd like to consider changing the InterledgerPacket definition to:

```
InterledgerPacket ::= SEQUENCE {
    -- Source ILP Address
    source Address,
    -- Destination ILP Address
    destination Address,
    -- One byte type ID
    type PACKET.&typeId ({PacketSet}),
    -- Type-specific header content
    headers PACKET.&Type ({PacketSet}{@type})
    -- Information for recipient (transport layer information)
    data OCTET STRING (SIZE (0..32767)),
    -- Enable ASN.1 Extensibility
    extensions SEQUENCE {
        ...
    }
}
```

Then the payment packet would be:

```
InterledgerProtocolPayment ::= SEQUENCE {
    -- Amount which must be received at the destination
    amount UInt64,
}
```

emschwartz:
@adrianhopebailie can we close this now?

PR #269, opened 2017-08-10
Strawman: add balance protocol on clp level

michielbdejong:
Currently, the callset decides between Interledger protocol vs custom,
and then custom has a registry based on strings.
This PR shows what it would look like if CLP would support more than
one protocol natively in its callset

### DIFF
--- a/asn1/CommonLedgerProtocol.asn
+++ b/asn1/CommonLedgerProtocol.asn
@@ -14,6 +14,12 @@ IMPORTS
     Address,
     Timestamp
     FROM InterledgerTypes
+
+    InterledgerPacket
+    FROM InterledgerPacket -- we rename this file from GenericPacket.asn
+
+    BalancePacket
+    FROM BalancePacket
 ;
 
 SideProtocolData ::= SEQUENCE OF SEQUENCE {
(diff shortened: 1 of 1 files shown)


asn1/CommonLedgerProtocol.asn
```diff
@@ -77,14 +83,16 @@ CALL ::= CLASS {
...
-    {7 Message} |
-    {8 CustomRequest}
+    {0 CustomRequest}
+    {1 CustomResponse} |
+    {16 InterledgerRequest} |
+    {17 InterledgerResponse} |
+    {18 InterledgerPrepare} |
+    {19 InterledgerFulfill} |
+    {20 InterledgerReject} |
+    {21 InterledgerAck} |
```
emschwartz: Why are all of these renumbered?

asn1/CommonLedgerProtocol.asn
```diff
@@ -77,14 +83,16 @@ CALL ::= CLASS {
...
+    {32 BalanceRequest} |
```
sharafian: These don't look like they're defined in this file, is that on purpose?

asn1/CommonLedgerProtocol.asn
```diff
@@ -14,6 +14,12 @@ IMPORTS
     Address,
     Timestamp
     FROM InterledgerTypes
+
+    InterledgerPacket
+    FROM InterledgerPacket -- we rename this file from GenericPacket.asn
+
+    BalancePacket
```
emschwartz: What is going to be in this `BalancePacket`? Is it just going to be the available liquidity? Or that and min, max, disputed, etc?

The reason we took the get balance call out of CLP was that you want different things for different use cases, and CLP is just the intersection of what you want in all of those uses. 

If 90+% of the protocols that extend CLP are going to extend it with a call to get available liquidity, then I think it should go in the core protocol. However, if in most cases it won't be used then it shouldn't go in there.

michielbdejong:
This PR addresses three issues:
* should it be allowed, in theory, for native protocols other than InterledgerPacket to be added to CLP in the future, and so, reserve the call set up to 127 for that, forcing ad-hoc things to go at 128 and up -> Ben: yes
* should we add a BalanceProtocol that implements getAvailableLiquidity as the second native protocol, or ad-hoc, or named? -> Ben: named or ad-hoc, not native
* bitmask for protocol vs CLP-level action. Ben: no, leave numbering as it is.

michielbdejong:
Superseded by #271

PR #271, opened 2017-08-10
docs: introduce CommonLedgerProtocol

michielbdejong:
This PR aims to find a compromise between all our different views. It differs from #263 in that it does not allow for `requestId=0`, it distinguishes `Response` from `ErrorResponse` in the CallSet, and it leaves CallSet id's reserved for non-ILP Prepare/Fulfill/Reject.

With these minor changes, CLP still has only one native protocol, namely Interledger. All other protocols have to use SideProtocolData. However, the CallSet could easily be extended with other native protocols. Additionally, it could easily be extended with conditionally paid custom requests, as well as unconditional ones.

As Ben and I discussed this afternoon, it may be a useful compromise not to define these currently unused features yet, but to accommodate for possibly adding them in the future.

### DIFF
--- a/asn1/CommonLedgerProtocol.asn
+++ b/asn1/CommonLedgerProtocol.asn
@@ -1,0 +1,114 @@
+CommonLedgerProtocol
+DEFINITIONS
+AUTOMATIC TAGS ::=
+BEGIN
+
+IMPORTS
+    UInt8,
+    UInt32,
+    UInt64,
+    UInt128,
+    UInt256
+    FROM GenericTypes
+
+    Timestamp
+    FROM InterledgerTypes
+
+    InterledgerPacket
+    FROM InterledgerPacket
+;
+
+-- Zero or more protocolData entries, each with a unique
+-- protocolName, describe protocol-specific information about
+-- unpaid and conditionally paid requests and results
+-- sent over the CommonLedgerProtocol.
+-- The protocolName of each entry describes the protocol for
+-- interpreting the data. Peers need to establish beforehand
+-- which protocols each understand, and which names they
+-- use, which can be agreed ad-hoc, except that protocolName 'ilp'
+-- is reserved for the Interledger protocol.
+-- When multiple protocolData items are sent in a request,
+-- only the first one will be decisive in triggering the
+-- response, the rest of the protocolData items are considered
+-- side protocols that piggyback on the first one.
+-- The contentType byte can be used to allow debug tools to
+-- display the contents of the protocolData without understanding
+-- it.
+
+ContentType ::= INTEGER {
+  applicationOctetString  (0),
+  textPlainUtf8           (1),
+  applicationJson         (2)
+} (0..255)
+
+ProtocolData ::= SEQUENCE OF SEQUENCE {
+  protocolName IA5String,
+  contentType ContentType,
+  data OCTET STRING
+}
+
+-- Ack, Response, and Error are the response types.
+-- When using these in a CLP packet, the requestId should match
+-- the requestId of the request they respond to.
+
+Ack ::= ProtocolData
+Response ::= ProtocolData
+Error ::= SEQUENCE {
+  rejectionReason InterledgerPacket, -- must contain an InternetProtocolError
+  --
+  protocolData ProtocolData
+}
+
+-- Prepare, Fulfill, Reject, and Message are the request types.
+-- Each request should have a unique requestId.
+
+Prepare ::= SEQUENCE {
+  transferId UInt128,
+  amount UInt64,
+  executionCondition UInt256,
+  expiresAt Timestamp,
+  --
+  protocolData ProtocolData
+}
+
+Fulfill ::= SEQUENCE {
+  transferId UInt128,
+  fulfillment UInt256,
+  --
+  protocolData ProtocolData
+}
+
+Reject ::= SEQUENCE {
+  transferId UInt128,
+  rejectionReason InterledgerPacket, -- must contain an InterledgerProtocolError
+  --
+  protocolData ProtocolData
+}
+
+Message ::= ProtocolData
+
+CALL ::= CLASS {
+    &typeId UInt8 UNIQUE,
+    &Type
+} WITH SYNTAX {&typeId &Type}
+
+CallSet CALL ::= {
+    {1 Ack} |
+    {2 Response} |
+    {3 Error} |
+    {4 Prepare} |
+    {5 Fulfill} |
+    {6 Reject} |
+    {7 Message}
+}
+
+CommonLedgerProtocolPacket ::= SEQUENCE {
+    -- One byte type ID
+    type CALL.&typeId ({CallSet}),
+    -- Used to associate requests and corresponding responses
+    requestId UInt32,
+    -- Length-prefixed main data
+    data CALL.&Type ({CallSet}{@type})
+}
+
+END


michielbdejong:
It also fixes a few syntax errors in asn documents it depends on, and renames `GenericPacket` to `InterledgerPacket`.

asn1/CommonLedgerProtocol.asn
```diff
@@ -0,0 +1,107 @@
...
+
+Ack ::= SideProtocolData
+
+InterledgerResponse ::= SEQUENCE {
+  packet InterledgerPacket,
+  -- Additional data for protocol extensibility
+  sideProtocolData SideProtocolData
+}
+
+InterledgerError ::= Response
```
adrianhopebailie: Where is Response defined?

asn1/CommonLedgerProtocol.asn
```diff
@@ -0,0 +1,107 @@
...
+  sideProtocolData SideProtocolData
+}
+
+CustomRequest ::= SideProtocolData
+
+-- Ack, InterledgerResponse, InterledgerError, and LedgerLevelError are the response types.
+-- When using these in a CLP packet, the requestId should match
+-- the requestId of the request they respond to.
+
+Ack ::= SideProtocolData
```
adrianhopebailie: I'd prefer if an Ack was simple `Ack ::= NULL`.

If there is a need to return a response that carries side protocol data then define a `LocalResponse :: = SideProtocolData`

asn1/CommonLedgerProtocol.asn
```diff
@@ -0,0 +1,107 @@
...
+  sideProtocolData SideProtocolData
+}
+
+InterledgerRequest ::= SEQUENCE {
+  packet InterledgerPacket,
+  -- Additional data for protocol extensibility
+  sideProtocolData SideProtocolData
+}
+
+CustomRequest ::= SideProtocolData
```
adrianhopebailie: I prefer the name "LocalRequest". The absence of an internetworking packet implies this request is not being relayed.

michielbdejong: Ben and I agreed to merge native protocols and named protocols into just numbered protocols. :) So there is now only 'protocolData', and the first entry in that list will usually have a `protocolId` of 1, meaning it's an InterledgerPacket.

asn1/CommonLedgerProtocol.asn
michielbdejong: There is (now) an Ack type and a Response type, which are both the same (they are also the same as Message), in that they only contain `protocolData`. The difference between Ack and Response is subtle, and arguably useless, but it's a bit like the difference between `return` and `return undefined` in JavaScript. The effect is the same, the semantics it expresses is slightly different.

asn1/CommonLedgerProtocol.asn
michielbdejong: That may have been missing from this morning's version, it's now here: https://github.com/interledger/rfcs/pull/271/files#diff-2ed9291d851bc56085df61488a4d3b10R44

michielbdejong:
OK, I'm cool with UInt16 instead of UInt8, and with removing Ack if @emschwartz agrees.

About ordering of protocols - I think only the first protocol packet in a request should affect the response, the subsequent ones should be just piggybacking side protocols that may support responding to the first one, but don't get a response themselves. So putting a getAvailableLiquidity and a QuoteByDestinationAmount together in one requests makes no sense, and when you add something like 'UpdatedPaychanClaim' packet to an 'InterledgerProtocolPayment', the actual request which results in the fulfill should come first, before the side-effect ones.

sharafian:
cool, I think UInt16 should be a good choice. As for the protocol ordering, I kind of see what you mean where it may be ambiguous which protocols in the request correspond to which protocols in the response, but I think that could be better handled by the definitions of the protocols being used. If incompatible packets are being included (like QuoteByDestinationResponse and QuoteBySourceResponse in the same packet), it should cause an error in any case.

michielbdejong:
> If incompatible packets are being included (like QuoteByDestinationResponse and QuoteBySourceResponse in the same packet), it should cause an error in any case.

Right, I guess we can assume the peers know exactly which protocols the other peer supports, so there's no reason someone would ever mess that up. Cool, i think we're good then, let's wait for review from @emschwartz and @justmoon.

asn1/CommonLedgerProtocol.asn
```diff
@@ -0,0 +1,103 @@
...
+protocolData ::= SEQUENCE OF SEQUENCE {
+  protocolId UInt8 UNIQUE,
+  data OCTET STRING
+}
+
+-- Ack, Response, and Error are the response types.
+-- When using these in a CLP packet, the requestId should match
+-- the requestId of the request they respond to.
+
+Ack ::= ProtocolData
```
adrianhopebailie: Why does an Ack carry data? Suggest making `Ack ::= NULL`

asn1/CommonLedgerProtocol.asn
```diff
@@ -0,0 +1,103 @@
...
+Response ::= ProtocolData
+Error ::= SEQUENCE {
```
adrianhopebailie: This doesn't make sense. Why would a generic error contain InterledgerError data?

michielbdejong: we said we reuse InterledgerError codes instead of inventing a new system of OER-encoded errors for CLP.

michielbdejong:
> Assumes that a transfer always has a condition

For now we just use Unpaid and paid-conditional requests, but unconditional-paid requests can be added. In fact, I was in favor of that (for completeness, mainly), I left them out because Ben and Evan didn't want to add things which we currently don't use.

> Assume a transfer response is always a fulfillment

It can be a fulfillment, meaning succes, or a rejection, meaning error. Rejections carry an error packet. Fulfillments don't have to carry anything, but are allowed to carry any packet from any protocol you want.

> Uses ILP errors

yeah, I agree it's a bit funny but it avoids having to set up a whole extra error registry.

> Has a logical id for tranfers (transferId) but not for messages

The reason for that is that messages are responded to in the response, but conditional transfers are not. I tried to convince Evan and Ben that the fulfill/reject that comes back from a Prepare is like a request/response, but they finally convinced me that a fulfill is a request in itself, where as the fulfiller you want something from the other party. That's why there are two flows:

### Message flow: 

[requestId: 1] -> Message
[requestId: 1] <- Response

or

[requestId: 1] -> Message
[requestId: 1] <- Error

and
### Conditional Transfer Flow:

[requestId: 2] -> Prepare
[requestId: 2] <- Ack
[requestId: 3] <- Fulfill
[requestId: 3] -> Ack

or

[requestId: 2] -> Prepare
[requestId: 2] <- Ack
[requestId: 3] <- Reject
[requestId: 3] -> Ack

or 
[requestId: 2] -> Prepare
[requestId: 2] <- Ack
[requestId: 3] <- Fulfill
[requestId: 3] -> Error

or

[requestId: 2] -> Prepare
[requestId: 2] <- Ack
[requestId: 3] <- Reject
[requestId: 3] -> Error

or

[requestId: 2] -> Prepare
[requestId: 2] <- Error

asn1/CommonLedgerProtocol.asn
justmoon: Whoops, Adrian commented "Why can't this be `NULL`?" - I tried to respond, then wanted to cancel responding and accidentally deleted Adrian's comment. ^^

Anyway, my comment is: I can't say that you would *never* want to include data with an ack and setting this to ProtocolData costs one byte and enables lots of use cases in the future. Maybe I want to send a balance update with the ack or the local timestamp when the message was received, or performance data like "it took 8.2ms to process this request" etc.

asn1/CommonLedgerProtocol.asn
```diff
@@ -0,0 +1,103 @@
...
+-- interpreting the data. Peers need to establish beforehand
+-- which protocols each understand, and which numbers they
+-- use, which can be agreed ad-hoc,except that protocolId 1
+-- is reserved for the Interledger protocol.
+-- When multiple protocolData items are sent in a request,
+-- only the first one will be decisive in triggering the
+-- response, the rest of the protocolData items are considered
+-- side protocols that piggyback on the first one.
+protocolData ::= SEQUENCE OF SEQUENCE {
+  protocolId UInt8 UNIQUE,
```
justmoon: I *much* prefer a string here. You're hardly losing any bytes. `ilp` for example is four bytes vs one byte. And you are so much less likely to have collisions between different vendor-specific extensions and you can see which protocol you're dealing with.

I would also include a mime type, which would be a one byte field with the options 0 = `application/octet-string`, 1 = `text/plain; charset=utf-8`, 2 = `application/json`, with more potentially being added in the future. As I've said in another thread, not including the mime type turned out to be a huge mistake in Ripple Memos, because it made it impossible to print any useful information for debugging.

Maybe one difference in our viewpoints is that I definitely expect more than 256 protocols here. Ripple is going to define its own and I'm sure most vendors will. Most settlement methods will define their own protocols here.

michielbdejong:
OK, changed protocolId to protocolName, and added the mime types.

asn1/CommonLedgerProtocol.asn
```diff
@@ -0,0 +1,108 @@
...
+-- only the first one will be decisive in triggering the
+-- response, the rest of the protocolData items are considered
+-- side protocols that piggyback on the first one.
+-- The contentType byte can be used to allow debug tools to
+-- display the contents of the protocolData without understanding
+-- it; set to 0 for `application/octet-string`, 1 for
+-- `text/plain; charset=utf-8`, and 2 for `application/json`.
+protocolData ::= SEQUENCE OF SEQUENCE {
+  protocolName IA5String,
+  contentType UInt8,
```
justmoon: You should give this a type with named values, like so:

``` asn1
ContentType ::= INTEGER {
  applicationOctetString  (0),
  textPlainUtf8           (1),
  applicationJson         (2)
} (0..255)
```

asn1/CommonLedgerProtocol.asn
```diff
@@ -0,0 +1,109 @@
...
+Ack ::= ProtocolData
+Response ::= ProtocolData
+Error ::= SEQUENCE {
+  rejectionReason InterledgerProtocolError,
+  --
+  protocolData ProtocolData
+}
+
+-- Prepare, Fulfill, Reject, and Message are the request types.
+-- Each request should have a unique requestId.
```
emschwartz: What is the expected behavior if you don't respond to a given request? Should the server keep sending you the message?

asn1/InterledgerPaymentRequest.asn
```diff
@@ -1,21 +1,27 @@
...
     -- IPR Version
     version UInt8,
     -- Cryptographic condition
     condition UInt256,
-    -- ILP Payment Packet
-    packet OCTET STRING
+    -- please note, InterledgerProtocolPayment
+    -- should not be added directly, it should
+    -- be inside its
+    -- InterledgerPacket envelope
```
emschwartz: I don't think this comment is necessary. The ILP packet _is_ the `InterledgerProtocolPayment` inside the packet wrapper. Interledger packets should never appear without that wrapper with the type and the length. Is there somewhere that isn't clear?

asn1/CommonLedgerProtocol.asn
adrianhopebailie: See @emschwartz comment below re using ILP packets outside their wrapper.

Also, this is lazy design. Those errors are for ILP. If we need errors for this protocol we should define them.

asn1/CommonLedgerProtocol.asn
michielbdejong: Hm, hadn't really thought about that much.
If Alice sends a request and Bob then I guess three things can happen:
1) Bob responds with Ack or Response
2) Bob responds with Error
3) Bob doesn't respond, or responds with a message that Alice doesn't understand.

If Bob responds multiple times, I guess only the first response counts.
If Bob doesn't respond:
- if Alice's request was a request for information, then she can just consider it as failed.
- if Alice's request was a request to update the state of a transfer, then she doesn't know if Bob processed her request, so they're now in limbo on that transfer's state. They should then use either manual reconciliation or a reconciliation protocol, to debug/resolve that.

I guess even with comments this asn can't stand on its own because of questions like this - maybe this is something we should write up in an RFC?

asn1/CommonLedgerProtocol.asn
emschwartz: I think it would make sense to use the same error schema here and define a set of CLP-specific errors (maybe with codes that start with L). Most of them are already defined in the LPI so I would  use those as a starting point

asn1/CommonLedgerProtocol.asn
emschwartz: That is precisely the purpose behind having the option to set `requestId` to 0 if you don't care about whether you get a response back.

Protocols that use these data formats should define what to do in these cases. Doesn't need to be in the ASN.1 but we will also need an RFC written up that defines how to use CLP.

asn1/CommonLedgerProtocol.asn
sharafian: do we include the envelope with typeId and length prefix for the error, though?

asn1/CommonLedgerProtocol.asn
sharafian: I agree that in an asymmetric trustline, you often want to send messages without caring about response. But Adrian raised a good point that if all notifications have ID 0 it's hard to do any logging with them. Furthermore, in a client/server relationship the server _never_ cares whether the client responds, so it can be agreed at establishment that the client doesn't respond, rather than on a per-message basis

asn1/CommonLedgerProtocol.asn
```diff
@@ -0,0 +1,114 @@
...
+
+Message ::= ProtocolData
+
+CALL ::= CLASS {
+    &typeId UInt8 UNIQUE,
+    &Type
+} WITH SYNTAX {&typeId &Type}
+
+CallSet CALL ::= {
+    {1 Ack} |
```
sharafian: Ack and Response both indicate a success here, and both have the same definition. Can't we use empty Responses in place of a special Ack type?

asn1/CommonLedgerProtocol.asn
```diff
@@ -0,0 +1,114 @@
...
+Fulfill ::= SEQUENCE {
+  transferId UInt128,
+  fulfillment UInt256,
+  --
+  protocolData ProtocolData
+}
+
+Reject ::= SEQUENCE {
+  transferId UInt128,
+  rejectionReason InterledgerProtocolError,
```
sharafian: Should this be the full ILP packet? As defined here, it's just the contents with no typeId nor length prefix.

asn1/CommonLedgerProtocol.asn
michielbdejong: I guess it's better if we include it here as well, so people don't get trained to ever send packets without envelopes. I'll change that.

asn1/CommonLedgerProtocol.asn
michielbdejong: Adrian also wants to get rid of Ack. I don't feel very strongly either way.
@emschwartz?

adrianhopebailie: I don't want to get rid of it I just don't see the point of an Ack and a Response if they both carry protocol specific response data UNLESS they have a different semantic meaning.

i.e. An Ack should never mean success, it should only mean "I got the request".

sharafian: So do you think in the case of an error, the flow should be:

Request ->
Ack <-

Error <-
Ack ->

asn1/CommonLedgerProtocol.asn
emschwartz: I _think_ it makes sense to have all errors wrapped in the same wrapper. It's a bit strange for CLP-level errors to be wrapped in an Interledger wrapper (and maybe they shouldn't be?), but then again we are expecting them to be relayed.

asn1/CommonLedgerProtocol.asn
justmoon: > in a client/server relationship the server never cares whether the client responds

How so? What if the server is configured to re-send after a timeout if the client hasn't responded?

asn1/CommonLedgerProtocol.asn
justmoon: I think we should at least write down the flows. When I send a transfer, am I correct in assuming I would get something like this:

Prepare ->
Ack <-

Reject <-
Ack ->

Prepare ->
Ack <-

Fulfill <-
Ack ->

Also, @sharafian: In your example with `Request` you mean `Message`?

asn1/CommonLedgerProtocol.asn
justmoon: My first instinct was that yes, it makes sense to use only one type of error wrapper, but then I looked at the Interledger error and I don't think it's a good fit here:

* `code` - Fine.
* `name` - Fine.
* `triggeredBy` - Unnecessary - there is only one party that could have triggered the error.
* `forwardedBy` - Unnecessary - these are local errors that can't be forwarded.
* `triggeredAt` - Fine.
* `data` - Maybe unnecessary - perhaps data associated with the error should go in `protocolData` which is more extensible.

So my suggestion would be to just:

``` asn1
Error ::= SEQUENCE {
  code IA5String (SIZE(3)),
  name IA5String,
  triggeredAt Timestamp,
  protocolData ProtocolData
}
```

Where `code` could start with `L` to indicate a **l*ocal or **l*edger error. Whatever codes we use here, they shouldn't overlap with Interledger codes.

> but then again we are expecting them to be relayed.

I disagree. When an error occurs on the ledger layer, the connector should trigger its own error, because nobody else should have to care what ledger they are using. Put differently, the connector's error should look the same no matter what ledger was used.

So if the ledger triggers `L99 - AcmePay Is Not Feeling Like It`, the connector should *not* relay it (as an L99) but rather it should emit its own `T01 - Ledger Unreachable`. They should feel free to include the ledger error in the freeform debug info if they want, because that doesn't force anyone to understand it.

michielbdejong: Oh, but if you use the `Message` clp-message to ask for a quote (or a `Prepare` clp-message to request an Interledger Payment), and get back an `Error` clp-message, then it might contain an error from the connector, or an error from the ledger.

Likewise, if you receive a `Reject` call, its rejectionReason might come from the connector or (in case of timeout) from the ledger.

michielbdejong: Therefore I think it makes sense to just use `IlpError` wrapped in an `InterledgerPacket` envelope.

asn1/CommonLedgerProtocol.asn
michielbdejong: The scenario Ben refers to is where you have 1 or more clients listening at the same time. But you're right, the server may be configured to resend all notifications if none of the clients answered.

michielbdejong:
Cool! Squashed and rebased on master, and removed asn1 corrections in other files from this PR.

@adrianhopebailie @dappelt please approve if you're ok with this being merged now.

asn1/CommonLedgerProtocol.asn
sharafian: Yeah, basically I'm wondering whether we intend a quote to be split into two RPC calls with Acks instead of the 1 RPC call that we do now

michielbdejong:
Will merge as once @adrianhopebailie approves.

asn1/CommonLedgerProtocol.asn
adrianhopebailie: Is this valid? If not, why?

Prepare ->
Reject <-

Prepare ->
Fulfill <-

This illustrates my concern about mixing the logical definitions and messaging semantics.

asn1/CommonLedgerProtocol.asn
adrianhopebailie: I agree with @justmoon here. We shouldn't be passing ILP Errors directly at this level. They should be created at the connector  and encapsulate the CLP error if that was the cause.

asn1/CommonLedgerProtocol.asn
adrianhopebailie: I think the protocol should allow both or be consistent. If we want the response to a `Prepare` to always be an `Ack` then I think the response to a `Message` should also always be an `Ack`.

adrianhopebailie:
I agree with @emschwartz that we should merge soon and get some implementation experience.

That said, I still have some concerns so feel free to override and merge anyway if you think my concerns are unfounded.

## Errors
Errors in the CLP are not ILP Errors. Re-using that packet for CLP specific errors is in-correct and re-use just because it's convenient is lazy.

@justmoon gave a good example of scenario where an error caused at the ledger layer should be conveyed in a response in this protocol but is now hard to do because the ledger must figure out how to package that as an ILP Error.

Further to that, the `Error` type can carry protocol data so it could carry an ILP Error packet in there anyway.

## Logical and Messaging concepts mixed

I keep raising this but clearly I'm not making the point clearly or I'm shouting in the wind. As @sharafian and @justmoon describe above there is a flow we expect for transfers but it seems we expect something different for messages?

Logically, when a node requests a `Prepare` then it will get either a `Fulfillment` or `Error` in response. However, we've prescribed that this MUST be done asynchronously by saying it will be:

```
Prepare ->
Ack <-
Fulfill <-
Ack ->
```

Presumably the first two messages have the same `requestId`  to allow them to be matched (what I'd call the message identifier). We also define that the `Prepare` and `Fulfill` have a `transferId` so that even though they are in different request/response message pairs they can be matched (they have a common logical identifier).

This begs the question how a quote request/response can be sent or must it always be synchronous? In the following flow there is no way to match the QuoteRequest/QuoteResponse.

```
Message(QuoteRequest) ->
Ack <-
Response(QuoteResponse) <-
Ack <-
```

So, maybe we should have a `messageId` in the `Message`, `Response` and possibly `Error` types so that there is a logical identifier for non-transfer messages?

asn1/CommonLedgerProtocol.asn
sappenin: I don't see the purpose of `Ack`, especially if `Response` always means, "Thanks, I got the request, you may get another 'request' in the future, you may not".

adrianhopebailie: The `Ack` is there for the case where the underlying transport is better served with async messages.

See my comment below on mixing logical and messaging semantics for an example of why I think there are some issue to still fix here.

sappenin: > The Ack is there for the case where the underlying transport is better served with async messages.

I must be missing something - what's the difference between `Ack` and `Response`? They seems to be the exact same thing, but from your comment is one meant to be used in an async channel (`Ack`) and the other (`Response`) meant to be used with a sync channel?

sappenin:
@adrianhopebailie above your wrote: 

> This begs the question how a quote request/response can be sent or must it always be synchronous? In the following flow there is no way to match the QuoteRequest/QuoteResponse.
> 
> Message(QuoteRequest) ->
> Ack <-
> Response(QuoteResponse) <-
> Ack <-

But why wouldn't messages look more like this instead:

Message(QuoteRequest) ->
Response <-
Message(QuoteResponse) <-
Response ->

michielbdejong:
I think we should merge now, since we've already started implementing this, both for ilp-kit and for testnet-of-testnets. We can't let those project be stalled.

I do agree with Adrian though, and as I said in https://lists.w3.org/Archives/Public/public-interledger/2017Aug/0021.html, I think we should design a better version 2, which we can launch later, for instance on 1 january 2018.

asn1/CommonLedgerProtocol.asn
dappelt: @sappenin has a good point there. Given that TCP ensures data delivery, I don't see why we need another `Ack` in the CLP layer.

adrianhopebailie: @sappenin I think they should be different or we shouldn't have both. I think we're on the same page.

@dappelt This is an application level `Ack`. It indicates that the receiving system got the message and is able to decode it. I'd consider it similar to a 202 HTTP response.

I agree it's not strictly necessary but if we want an async pattern then it is useful to have something you can reply back to the sender to indicate you got the request. I think it makes sense but should not carry any data.

dappelt: If the receiving side cannot decode the message, it will send a `Response` containing an error. Why would we need an `Ack`?

adrianhopebailie: If the `Ack` and `Response` are the same then I wonder the same thing. The only reason I see for the `Ack` is as a very lightweight positive response.

adrianhopebailie:
@michielbdejong implementing is entirely up to you. If you want to start before the spec is final then you may just have more to change when it does. Also, implementation experience is also helpful to inform the design.

I don't think we should stop working on this because we have run out of energy and want to move onto other things. There are still a number of unanswered questions. This is a simple spec so we should be able to solve them.

That said, I'd be happy to merge this and log issues against the spec as I think this discussion is becoming unwieldy. BUT, I don't think we should take that as license to ignore these issues and turn all of our attention to implementations.

asn1/CommonLedgerProtocol.asn
```diff
@@ -0,0 +1,114 @@
...
+-- is reserved for the Interledger protocol.
+-- When multiple protocolData items are sent in a request,
+-- only the first one will be decisive in triggering the
+-- response, the rest of the protocolData items are considered
+-- side protocols that piggyback on the first one.
+-- The contentType byte can be used to allow debug tools to
+-- display the contents of the protocolData without understanding
+-- it.
+
+ContentType ::= INTEGER {
```
sappenin: Can this just be an `IA5String` so we don't have CLP re-defining Content-Types?

asn1/CommonLedgerProtocol.asn
dappelt: > So if the ledger triggers L99 - AcmePay Is Not Feeling Like It, the connector should not relay it (as an L99) but rather it should emit its own T01 - Ledger Unreachable. They should feel free to include the ledger error in the freeform debug info if they want, because that doesn't force anyone to understand it.

@justmoon In your example, in what kind of packet would the relayed error `T01 - Ledger Unreachable` be wrapped? Since `T01` is an `IlpError`, I assume it would go inside a CLP `Error`, with `protocolData` carrying the ILP error? Or would it go into `protocolData` of a CLP `Response`, since the local CLP link works fine and the error occurred in a different CLP link?

asn1/CommonLedgerProtocol.asn
justmoon: This field is intended to give applications that don't know/understand the custom protocol a way to at least partially decode the data. For example. we have values to indicate whether it's binary or text data. We intentionally want a small, defined list of supported formats here, NOT support any arbitrary mime type. If we want to add a format to the list, there should be broad agreement in the Interledger community that's it is worth adding this format to a large number of implementations. If the values in this field aren't widely understood the field is not useful.

So I don't think we redefining anything - we're just defining a short list of mime types that are widely understood and since it's a small, standardized list, an enum is the correct type to use. I just realized that there is an `ENUMERATED` type in ASN.1 which would perhaps be an even better fit than `INTEGER`.

justmoon:
> I do agree with Adrian though, and as I said in https://lists.w3.org/Archives/Public/public-interledger/2017Aug/0021.html, I think we should design a better version 2, which we can launch later, for instance on 1 january 2018.

I agree that we should freeze v1 asap, but I don't think you can finalize a spec without implementation experience. (Ideally multiple implementations by different people, but given the current size of our community, I'll take a single implementation.)

What that means is that we can use this as a draft, implement it, play around with it for a bit and when we notice that we're making fewer and less urgent changes, let's call it v1 and make it final. If we want to change it after that, it'll be v2 and be separate.

adrianhopebailie:
It's been almost a week since I raised two issues regarding: 1. errors and 2. the conflation of logical and message semantics and yet most of the discussion has been about whether or not to freeze the spec 😄 

If we just resolve those two issues I'd be very supportive of freezing the spec

dappelt:
Discussing all these issues in the same PR becomes unwieldy. I propose to merge this PR as an initial draft. This draft should **not be frozen** and we should adapt it based on first implementation experiences. I think we can all agree on that (@adrianhopebailie, please approve so we can merge).

@adrianhopebailie, I agree with you on the issue around errors (see my latest [comment](https://github.com/interledger/rfcs/pull/271#discussion_r134737651), input appreciated). Also, I would like to add a 3. item to the open issue list: Given that we have a message type `Response`, we don't need `Ack`.

justmoon:
Quick point which I don't think has been said explicitly: Can a protocol appear more than once in the protocol data? I'd vote no for simplicity, but could be convinced by a strong use case.

asn1/CommonLedgerProtocol.asn
justmoon: An empty response is just as lightweight as an `Ack`, so am I right in thinking that we're coming to the conclusion that we don't need `Ack`?

michielbdejong:
Adrian wrote:
>  I'd be happy to merge this and log issues against the spec as I think this discussion is becoming unwieldy. BUT, I don't think we should take that as license to ignore these issues and turn all of our attention to implementations.

Agreed! I'll merge this so that the discussion can continue. Also, we shouldn't forget to write the actual rfc, so far this just defines the asn.1 format.

> Can a protocol appear more than once in the protocol data?

I think just once, that's what I tried to convey in https://github.com/interledger/rfcs/pull/271/files#diff-2ed9291d851bc56085df61488a4d3b10R21 but we should describe things like that in an rfc document.

asn1/CommonLedgerProtocol.asn
dappelt: Yes, I think so. I am almost done with a first implementation and I don't use `Ack` at all.

asn1/CommonLedgerProtocol.asn
sappenin: > This field is intended to give applications that don't know/understand the custom protocol a way to at least partially decode the data.

Ok, perhaps I've been viewing this field a bit differently. I was thinking this content-type would be useful for applications that _do_ understand the side-protocol...i.e., it gives those applications a mechanism to decode the data. For example, if I have a side-protocol named `aml_protocol`, then perhaps this payload is encoded as JSON, perhaps it's binary. Are you suggesting that if `aml_protocol` wants different encodings, then one should create separate protocolNames, like `aml_protocol_json` and `aml_protocol_oer`?

michielbdejong:
We can keep discussing changes like removing Ack and changing the way errors are encoded, but I settled on commit 8b64d63 for the testnet of testnets. It is documented here: https://github.com/interledger/interledger/wiki/Interledger-over-CLP and implemented here: https://github.com/interledgerjs/clp-packet/pull/4.

asn1/CommonLedgerProtocol.asn
justmoon: Revising my earlier examples:

(ProtocolData in parenthesis.)

#### Quote (successful)

``` lua
>>> Message (ilp => 0x02...) -- TYPE_ILQP_LIQUIDITY_REQUEST
<<< Response (ilp => 0x03...) -- TYPE_ILQP_LIQUIDITY_RESPONSE 
```

#### Quote (locally successful, but connector responded with an ILP error)

``` lua
>>> Message (ilp => 0x02...) -- TYPE_ILQP_LIQUIDITY_REQUEST
<<< Response (ilp => 0x08...) -- TYPE_ILP_ERROR
```

#### Quote (local error)

``` lua
>>> Message (ilp => 0x02...) -- TYPE_ILQP_LIQUIDITY_REQUEST
<<< Error (...) -- Some CLP error format we define
```


#### Successful payment

``` lua
>>> Prepare (ilp => 0x01...) -- TYPE_ILP_PAYMENT
<<< Response () -- Empty response [for now?]

<<< Fulfill ()
>>> Response ()
```

#### Failed payment

``` lua
>>> Prepare (ilp => 0x01...) -- TYPE_ILP_PAYMENT
<<< Response ()

<<< Reject ()
>>> Response ()
```

Some questions:

* Should the `Reject` message contain a rejectionReason field (as the ASN says currently) or just an ILP Error in the ProtocolData?
* Should we define a `Fulfill` format on the ILP layer which may, for example, communicate some data back to the payer? (see #279)

asn1/CommonLedgerProtocol.asn
justmoon: @dappelt Good question. I [wrote down some suggested flows](https://github.com/interledger/rfcs/pull/271/files#r135555926) that make sense to me, but interested in feedback.

asn1/CommonLedgerProtocol.asn
michielbdejong: moving this to #282

PR #272, opened 2017-08-11
fix: ILP Packet -> InterledgerProtocolPayment

michielbdejong:
I tried to push the abbreviation 'ipp' but it didn't really seem to stick.

In any case, the current situation where 'ILP Packet' can be either https://github.com/interledger/rfcs/blob/master/asn1/InterledgerProtocol.asn#L15 by itself, or wrapped inside https://github.com/interledger/rfcs/blob/master/asn1/GenericPacket.asn#L40 is way too confusing, and is more or less guaranteed to lead to bugs where people try to parse it with/without the wrapper when the opposite was intended.

### DIFF
--- a/0016-pre-shared-key/0016-pre-shared-key.md
+++ b/0016-pre-shared-key/0016-pre-shared-key.md
@@ -1,10 +1,10 @@
 ---
 title: The Pre-Shared Key Transport Protocol (PSK)
-draft: 2
+draft: 3
 ---
 # Pre-Shared Key Transport Protocol (PSK)
 
-The Pre-Shared Key (PSK) protocol is an end-to-end transport protocol, used by the sender and receiver of an ILP payment to decide on a condition and fulfillment for a payment. By default, the protocol also encrypts any additional data sent along with the payment, using [AES-256-GCM](https://en.wikipedia.org/wiki/Galois/Counter_Mode). The full ILP payment is authenticated through an [HMAC-SHA-256](https://en.wikipedia.org/wiki/Hash-based_message_authentication_code) which is used to generate the fulfillment of a PSK payment. The entirety of the PSK data, including public headers, encrypted private headers, and encrypted private data, is encoded into an octet-stream that forms the data portion of the [ILP Packet](https://github.com/interledger/rfcs/blob/master/0003-interledger-protocol/0003-interledger-protocol.md). The PSK data is authenticated via AES-256-GCM in addition to the HMAC-SHA-256 which authenticates the full ILP payment.
+The Pre-Shared Key (PSK) protocol is an end-to-end transport protocol, used by the sender and receiver of an ILP payment to decide on a condition and fulfillment for a payment. By default, the protocol also encrypts any additional data sent along with the payment, using [AES-256-GCM](https://en.wikipedia.org/wiki/Galois/Counter_Mode). The full ILP payment is authenticated through an [HMAC-SHA-256](https://en.wikipedia.org/wiki/Hash-based_message_authentication_code) which is used to generate the fulfillment of a PSK payment. The entirety of the PSK data, including public headers, encrypted private headers, and encrypted private data, is encoded into an octet-stream that forms the data portion of the [ILP Packet](https://interledger.org/rfcs/0003-interledger-protocol/draft-4.html#specification). The PSK data is authenticated via AES-256-GCM in addition to the HMAC-SHA-256 which authenticates the full ILP payment.
 
 Pseudocode for this protocol can be read [at the bottom of this spec](#pseudocode).
 
--- a/0003-interledger-protocol/0003-interledger-protocol.md
+++ b/0003-interledger-protocol/0003-interledger-protocol.md
@@ -1,6 +1,6 @@
 ---
 title: The Interledger Protocol (ILP)
-draft: 3
+draft: 4
 ---
 # Interledger Protocol (ILP)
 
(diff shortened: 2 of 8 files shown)


michielbdejong:
Especially https://github.com/interledger/rfcs/blob/master/0016-pre-shared-key/0016-pre-shared-key.md#2-payment-fulfillment is confusing since in CLP we will be including an InterledgerPacket, not an IPP. You have to open up the InterledgerPacket to get the IPP, or in other words, you're interested in `packet.data.data.headers`.

michielbdejong:
When reading LPI, 'ILP packet' should refer to InterledgerPacket in the case of Message (because there you want to allow ILQP). Which one is intended as the 'ilp' field in sendTransfer?

michielbdejong:
> The GenericPacket module is also unfortunately named because it pulls in both ILP packet types and ILQP packet types to define the set of Interledger layer packets.

yes, and this would be fine if 'GenericPacket just meant 'ILP packet *or* ILQP packet', but the unfortunate situation is that it adds a byte at the front, so if you interpret an Interledger packet as an ILP packet, everything shifts one position, and I guess the first byte of the amount would be a 1 instead of a 0.

> the ILP packets are named InterledgerProtocolPayment and InterledgerProtocolError and there is no packet for a successful response to a payment.

That's by design and unrelated. Apart from the fulfillment, no data can safely be transported back to the sender, delivery of what you pay for has to go via a different path (possibly a new Interledger payment, carrying the data in the forward direction).

> I don't entirely agree with this change although I agree with the intent.

What alternative do you see, specifically to the naming problem where 'GenericPacket/InterledgerPacket' gets confused with 'InterledgerProtocolPayment/ILP-packet'?

michielbdejong:
@sharafian so you're saying that for instance:

> The entirety of the PSK data, including public headers, encrypted private headers, and encrypted private data, is encoded into an octet-stream that forms the data portion of the ILP Packet.

https://github.com/interledger/rfcs/blob/master/0016-pre-shared-key/0016-pre-shared-key.md#pre-shared-key-transport-protocol-psk

refers to [the data field of the GenericPacket](https://github.com/interledger/rfcs/blob/master/asn1/GenericPacket.asn#L44)? Or does that
need changing to "the data portion of the data portion of the ILP Packet"?

sharafian:
The "data" of the GenericPacket is the actual contents of the ILP packet. Maybe we should go back over the ASN1 and change some of the names, because terms like GenericPacket don't appear anywhere else. So but technically yes, it's the data field of the data field of the envelope.

michielbdejong:
> technically yes, it's the data field of the data field of the envelope.

OK, I'll update IL-RFC-16 to fix that.

adrianhopebailie:
@michielbdejong , the point I am trying to make is, when you ask this:

>The entirety of the PSK data, including public headers, encrypted private headers, and encrypted private data, is encoded into an octet-stream that forms the data portion of the ILP Packet.
https://github.com/interledger/rfcs/blob/master/0016-pre-shared-key/0016-pre-shared-key.md#pre-shared-key-transport-protocol-psk

>> refers to the data field of the GenericPacket? Or does that need changing to "the data portion of the data portion of the ILP Packet"?

The answer should be, **yes**. The confusion comes from one of the deeper problems I am trying to make the case for solving.

The packets defined in `GenericPacket` should be the set of all packets sent at the Interledger layer. They should all have a "payload". (Calling it "data" is also probably also adding to the confusion)

@sharafian said:
> Maybe we should go back over the ASN1 and change some of the names, because terms like GenericPacket don't appear anywhere else.

Yes, we should fix up the core definitions as a priority before we start changing the documentation.

justmoon:
> The packets defined in GenericPacket should be the set of all packets sent at the Interledger layer.

Minor nitpick: Should it be all packets sent at the Interledger layer or should be the set of all packets defined by the ILP (which includes ILQP) standard?

I think it would be the latter, for the simple reason that we do not have the authority to manage the namespace of all possible protocols on the interledger layer.

What this would mean in practice is that you would have to signal that you're using `ilp` on the ledger layer, which is exactly where we ended up in the CLP discussion.

0016-pre-shared-key/0016-pre-shared-key.md
```diff
@@ -1,10 +1,10 @@
...
+The Pre-Shared Key (PSK) protocol is an end-to-end transport protocol, used by the sender and receiver of an ILP payment to decide on a condition and fulfillment for a payment. By default, the protocol also encrypts any additional data sent along with the payment, using [AES-256-GCM](https://en.wikipedia.org/wiki/Galois/Counter_Mode). The full ILP payment is authenticated through an [HMAC-SHA-256](https://en.wikipedia.org/wiki/Hash-based_message_authentication_code) which is used to generate the fulfillment of a PSK payment. The entirety of the PSK data, including public headers, encrypted private headers, and encrypted private data, is encoded into an octet-stream that forms the data portion inside the contents of the [ILP Packet](https://interledger.org/rfcs/0003-interledger-protocol/draft-4.html#specification). The PSK data is authenticated via AES-256-GCM in addition to the HMAC-SHA-256 which authenticates the full ILP payment.
```
michielbdejong: changed:
> 'the data portion of the [ILP Packet]'

to:

> 'the data portion inside the contents of the [ILP Packet]'

and updated the link to point to draft-4.

adrianhopebailie:
@justmoon said:

>Should it be all packets sent at the Interledger layer or should be the set of all packets defined by the ILP (which includes ILQP) standard?

> I think it would be the latter, for the simple reason that we do not have the authority to manage the namespace of all possible protocols on the interledger layer.

Actually I disagree or maybe partially disagree. We have created the ILP Standard and therefor defined the existence of a protocol stack that centers on the Interledger layer at the middle of the hour-glass.

The protocols that operate at that layer are necessarily simple and should be long lived so I'd hope they don't change often. Even if they do, we have already defined a mechanism for new packet types to be defined (using a new packet id) so we could say that even if they aren't defined in **this** ASN.1 file they should use an unused packet identifier and "the community" (because we have no formal registry today) should agree on the assignments.

Said slightly differently; even if someone else wants to come up with a new packet type, if it is an `InterledgerPacket` then it should be designed to be carried at that layer (i.e. any lower layer protocols that carry ILP packets should be able to carry it). It should also adopt an unused type id that the community agrees upon.

Conversely, if someone wants to define a new packet that is related to ILP but is not intended to be carried at the ILP layer of the stack it should not be part of this set.

michielbdejong:
@sharafian  it's now been over one month since I addressed your comments, please keep on top of [your review queue](https://github.com/pulls/review-requested)

michielbdejong:
Do you agree your suggestion of "to say that there are several types of ILP Packets, each prefixed with a type ID byte" is at least *potentially* confusion? If not, maybe it's easiest if we discuss this on gitter or slack.

sharafian:
sure, it's potentially confusing that an ILP Packet is not necessarily an ILP Payment Packet. I don't think renaming to use the names in the ASN.1 is a good idea, though, because that's the document that the least people are familiar with. Additionally, packets never exist without the envelope in any protocol or implementation. Because people already call the ILP Payment Packet the "ILP Packet", I think changing our terminology will just cause more confusion.

michielbdejong:
> packets never exist without the envelope in any protocol or implementation

Good point. OK, then in the RFCs let's only give names to packets including there envelopes.
And in the ASN.1 I can add code comment to clarify things wherever it's unclear if a class name refers to that packet as we know if from the RFCs, or to the structure of the envelope contents that such packets tend to have.

I'll go over the changes in this PR again, with that in mind. Thanks!

emschwartz:
(commits should be squashed before merging)

PR #275, opened 2017-08-15
chore: check ASN syntax in CI; generate ASN docs

sentientwaffle:
Fixes https://github.com/interledger/rfcs/issues/216

PR #276, opened 2017-08-16
fix: asn1 corrections

michielbdejong:
* require packet in IPR to be an InterledgerPacket, not just any octet string
* rename the GenericPacket.asn file to InterledgerPacket.asn, to match the class it defines
* ~~fix incorrect module name in InterledgerPaymentRequest.asn~~
* fix newlines ~~and punctuation~~ in a few places


michielbdejong:
Ah, just noticed #275, will rebase on that

PR #279, opened 2017-08-23
Add InterledgerProtocolFulfillment packet definition

adrianhopebailie:
This PR adds a new ILP packet type for carrying fulfillments.

All ILP packets are either a request or a response as ILP has request/response semantics as a core feature of the protocol however there is no response defined for a payment.

```
InterledgerProtocolPayment ->
InterledgerProtocolFulfillment <-

QuoteLiquidityRequest ->
QuoteLiquidityResponse <-

QuoteBySourceAmountRequest ->
QuoteBySourceAmountResponse <-

QuoteByDestinationAmountRequest ->
QuoteByDestinationAmountResponse <-
```

The `InterledgerProtocolError` is the generic error response for all requests.

### DIFF
--- a/asn1/InterledgerProtocol.asn
+++ b/asn1/InterledgerProtocol.asn
@@ -4,7 +4,8 @@ AUTOMATIC TAGS ::=
 BEGIN
 
 IMPORTS
-    UInt64
+    UInt64,
+    UInt256
     FROM GenericTypes
 
     Address,
@@ -17,7 +18,17 @@ InterledgerProtocolPayment ::= SEQUENCE {
     amount UInt64,
     -- Destination ILP Address
     account Address,
-    -- Information for recipient (transport layer information)
+    -- Information for recipient (application layer information)
+    data OCTET STRING (SIZE (0..32767)),
+    -- Enable ASN.1 Extensibility
+    extensions SEQUENCE {
+        ...
+    }
+}
+
+InterledgerProtocolFulfillment ::= SEQUENCE {    
+    fulfillment UInt256,
+    -- Information for recipient (application layer information)
     data OCTET STRING (SIZE (0..32767)),
     -- Enable ASN.1 Extensibility
     extensions SEQUENCE {
--- a/asn1/InterledgerPacket.asn
+++ b/asn1/InterledgerPacket.asn
@@ -34,7 +34,8 @@ PacketSet PACKET ::= {
     {5 QuoteBySourceAmountResponse} |
     {6 QuoteByDestinationAmountRequest} |
     {7 QuoteByDestinationAmountResponse} |
-    {8 InterledgerProtocolError}
+    {8 InterledgerProtocolError} |
+    {9 InterledgerProtocolFulfillment}
 }
 
 InterledgerPacket ::= SEQUENCE {
--- a/asn1/InterledgerPacket.asn
+++ b/asn1/InterledgerPacket.asn
@@ -9,7 +9,8 @@ IMPORTS
     FROM GenericTypes
 
     InterledgerProtocolPayment,
-    InterledgerProtocolError
+    InterledgerProtocolError,
+    InterledgerProtocolFulfillment
     FROM InterledgerProtocol
 
     QuoteLiquidityRequest,


dappelt:
The fulfillment is the counterpart to the condition. Since the condition does not live on the ILP layer, the fulfillment should neither.

asn1/InterledgerProtocol.asn
```diff
@@ -17,7 +18,17 @@ InterledgerProtocolPayment ::= SEQUENCE {
...
+    -- Enable ASN.1 Extensibility
+    extensions SEQUENCE {
+        ...
+    }
+}
+
+InterledgerProtocolFulfillment ::= SEQUENCE {    
+    fulfillment UInt256,
+    -- Information for recipient (application layer information)
     data OCTET STRING (SIZE (0..32767)),
```
dappelt: The possibility of sending `data` back in some kind of payment response is interesting. Do you have any use case in mind? Another question that arises is why does `data` have to travel back through ILP as opposed to more efficient out-of-band communication?

adrianhopebailie:
> The fulfillment is the counterpart to the condition. Since the condition does not live on the ILP layer, the fulfillment should neither.

Another reason why I believe the condition is in the wrong place 😄

adrianhopebailie:
@emschwartz Why are error responses passed back as ILP packets (i.e. at a different layer)?

emschwartz:
**Idea after typing up the response to your question:**

Rename the ILP Error packet to an ILP Response, add more status codes for success, and add the ability to send responses back with calls to fulfill transfers on the Ledger Layer.

(Even if this might be a good idea I'm hesitant to suggest this kind of change and would only be convinced that it's worth it if we have a couple of compelling use cases that require this functionality)

-----

**Response to your question:**

Fair question.

ILP errors are expected to be relayed through the Ledger Layer using the ability to reject incoming transfers. The ILP Error packet itself is just the additional data that rides alongside the Ledger Layer call to reject the transfer.

The most compelling reason I see for adding a fulfillment packet is to carry data other than the fulfillment itself. I'm not sure we actually have a use case for that but I'd be more convinced about this if we did.

The way I think about it is the following. As I've written in [this thread](https://lists.w3.org/Archives/Public/public-interledger/2017Aug/0047.html), I think ILP does expect the layer below it to use sha256 HTLAs and thus have a way to pass back a 32-byte fulfillment. Since each connector and the sender must be able to get a notification of the HTLA being successfully completed and since it's dangerous to rely on any mechanism outside of your HTLA to tell you the fulfillment, it doesn't make sense to define a packet format that is also expected to be passed back separately from this just to tell you the fulfillment. If there's extra data we want to put in the response, it might be worth adding this as a feature.

However, **even if we do have extra data being passed back I'm still not convinced that it's a good idea to put the fulfillment in that packet**. Since the layer underneath ILP (still called the Ledger Layer for now, even though that's causing some confusion) is expected to have a way to pass the fulfillment back along the path, the question is whether to put the fulfillment in an additional spot. The arguments against are that you're sending unnecessary bytes and, more importantly, having the fulfillment (like the condition) in two places is asking for trouble. For both the condition and the fulfillment you would need to ignore the value passed in the corresponding ILP packet and only consider what the HTLA is actually based upon. If we have the same value in two places (or at least it's supposed to be the same semantics), someone is going to implement it wrong and lose money.

It feels a little bit like this debate is premised on a desire to include all of the pieces of information relevant for ILP in a single data packet format. However, if this is just so that the protocol definition feels more complete but is actually likely to cause serious problems and a loss of money and the same information must be communicated alongside the ILP packets anyway, it seems like all downside with no practical upside for including the value again.

adrianhopebailie:
I think the argument to not change the ILP packets is pretty strong. I don't entirely agree with it but I don't think we are going to make progress while the cost of change is still weighing on people's minds.

My plan is to document what I think an ILP v2 stack should look like based upon the experience we have gained so far. I'll do the work to cost the changes to our implementations to switch to this once we have agreed on what v2 looks like (if we do).

adrianhopebailie:
After a lot of discussion we have concluded that this is not a good idea right now. 

While one can theoretically make the case that the condition and fulfillment live at the ILP layer, practically they will almost always exist at the ledger layer too and this is the most useful place for them to be when implementing an ILP node.

michielbdejong:
Continued in https://github.com/interledger/rfcs/issues/314

PR #285, opened 2017-08-29
feat: puts protocolData inside of a sequence

### LINKED ISSUE #284: [clp] Use packet.data.protocolData even if it's the only field in packet.data?
Some clp calls have a field protocolData in addition to other fields (together in a sequence called data), but for Ack, Response, and Message, current master says packet.data should immediately be the protocol data, instead of a one-element sequence *containing* protocolData.

dappelt:
fixes #284


adrianhopebailie:
Is there an encoding where this change in definition actually makes a difference to the encoded bytes? As @justmoon points out this has no impact in OER so is it really necessary?

michielbdejong:
We mirror the OER structure in the JavaScript implementation, so there it's nicer to have protocolData nested inside obj.data, always and not only sometimes.

Issue #290, opened 2017-09-04
[clp] should error responses be `Error`s or `Response`s?


michielbdejong:
Ben says `Response`. I would say `Error`. We need to pick one.

dappelt:
Assuming that the question is should an incoming  CLP `Error` be relayed to a peer as CLP `Error` or CLP `Response`, I think it should be relayed as `Response`. See also https://github.com/interledger/rfcs/pull/271#discussion_r133363793.

michielbdejong:
@dappelt Your link to #271 doesn't seem to go to a specific comment.

I think it's pretty obvious that errors in response to unpaid requests should go into `Error` call types and not `Response`, just like errors in response to conditional requests should go into `Reject` call types and not `Fulfill`, otherwise it all becomes very messy.

dappelt:
> @dappelt Your link to #271 doesn't seem to go to a specific comment.

github collapsed the specific comment when the page is loaded (we should try to avoid long discussions in a PR). This is the comment I am reffering to:

> > but then again we are expecting them to be relayed.

>I disagree. When an error occurs on the ledger layer, the connector should trigger its own error, because nobody else should have to care what ledger they are using. Put differently, the connector's error should look the same no matter what ledger was used.

> So if the ledger triggers L99 - AcmePay Is Not Feeling Like It, the connector should not relay it (as an L99) but rather it should emit its own T01 - Ledger Unreachable. They should feel free to include the ledger error in the freeform debug info if they want, because that doesn't force anyone to understand it.

dappelt:
> I think it's pretty obvious that errors in response to unpaid requests should go into Error call types and not Response

It depends on why the error happened. If the error happened on the CLP-level (e.g. received packet is not valid CLP, unsupported side-protocol, etc.), then yes, it should reply with an `Error`. If the error happens because of some ILP-related functionality (e.g. F02 Unreachable), then it should reply with a `Response`.

adrianhopebailie:
> In short, a CLP Error error indicates a problem with the local CLP link. Everything else should go into a CLP Response.

👍

michielbdejong:
That would then not be symmetric with `Reject`, where `rejectionReason` can be either CLP-level or protocol-level. So then you would have to re-introduce the distinction between cancel (with rejectionReason) and reject (with just protocolData).

Funny how history repeats itself if you keep talking about these things long enough :)

dappelt:
Can you give an example where `rejectionReason` would be CLP-level?

[...]

PR #291, opened 2017-09-05
feat: document the BTP

sharafian:
The current RFCS describe the CLP format, but this spec describes the flow and semantics of CLP. It's still a WIP; there are a couple of TODOs and the links need to be filled in.

### DIFF
--- a/00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
+++ b/00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
@@ -40,13 +40,13 @@ case. It also includes functionality which is common between many different
 ledger types, making it a good place to start from when creating a new
 protocol.
 
-This document describes the flow and messages that CLP uses, but not
+This document describes the flow and data format that CLP uses, but not
 sub-protocols. Sub-protocols include functionality like ledger metadata,
 balance, automated settlement, and dispute resolution. Some protocols are
 documented on [the wiki page]. They are carried in the protocol data of CLP
-messages.
+packets.
 
-The CLP Packet format is described exactly in the [CLP ASN.1 spec].
+The CLP packet format is described exactly in the [CLP ASN.1 spec].
 
 ## Terminology
 
@@ -56,11 +56,11 @@ The CLP Packet format is described exactly in the [CLP ASN.1 spec].
 - A **Sub-Protocol** is a protocol which isn't defined by CLP and is carried
   in the protocol data (see below).
 
-- A **CLP Connection** is a websocket connection over which CLP Packets are
+- A **CLP Connection** is a websocket connection over which CLP packets are
   sent. Websockets are used because they provide message framing and allow CLP
 to use HTTP requests for authentication.
 
-- **CLP Packets** are the messages described in this document. They are
+- **CLP Packets** are the protocol data units described in this document. They are
   formally defined in the [CLP ASN.1 spec].
 
 - **Peers** are the parties on a CLP connection. Your peer is the party on the
@@ -185,9 +185,9 @@ There are also a couple of tricky cases to handle:
 - If an unreadable CLP packet is received, no response should be sent. An unreadable CLP packet is one which is structurally invalid, i.e. terminates before length prefixes dictate or contains illegal characters.
 
 These behaviors are important for preventing accidental feedback loops.  If an
-unexpected message triggered an error, that error may be unexpected to the
+unexpected packet triggered an error, that error may be unexpected to the
 sender. The sender would reply with another unexpected error, causing an
-infinite loop. Unreadable messages must be ignored too. If an application got
+infinite loop. Unreadable packets must be ignored too. If an application got
 onto a CLP connection and spoke the wrong protocol, it would trigger an error
 from CLP. This might trigger an error from the application, and it would
 devolve into another infinite loop.
@@ -226,7 +226,7 @@ Prepare ::= SEQUENCE {
 }
 ```
 
-`Prepare` is used to create a transfer on the bilateral ledger. The message
+`Prepare` is used to create a transfer on the bilateral ledger. The packet
 data contains `transferId`, `amount`, `executionCondition`, and `expiresAt`.
 `Prepare` is a request with side effects, because it creates a transfer with
 the given details. This transfer begins in the `prepared` state.
--- a/00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
+++ b/00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
@@ -208,10 +208,10 @@ protocol name `ilp` with content-type `application/octet-stream`.
   has data to send in reply (e.g. a quote response), it is carried in the
 protocol data.
 
-- `Error` is returned if the peer is not able to process the message, or there
-  was an unexpected error and further messages should not be sent. This does
-not include ILP errors such as "No quote found," only the cases in which the
-message cannot be sent/processed at all.
+- `Error` is returned if the peer is not able to process the `Message`, or there
+  was an unexpected error and further `Message`s should not be sent. This does
+not include ILP errors such as "No quote found", only the cases in which the
+`Message` cannot be sent/processed at all.
 
 ### Prepare
 
--- a/00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
+++ b/00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
@@ -12,64 +12,75 @@ have been added.
 ### Motivation
 
 There are many different types of ledgers, and Interledger aims to interoperate
-all of them. For fast ledgers that support hashlocks, it is enough to wrap the
-ledger's API in a [Ledger Plugin]. These allow you to send interledger payments
-through anyone else on that ledger who is running your plugin.
+all of them. For fast ledgers that support hash time locks, memos, and
+messaging, it is enough to wrap the ledger's API in a [Ledger Plugin]. These
+allow you to send interledger payments through anyone else on that ledger who
+is running your plugin.
 
 In lots of scenarios, we don't have an underlying ledger that's fast enough to
-do every ILP payment on-ledger. There are also many situations when two parties
-trust one another up to an amount, and don't need any third-party ledger to
-book-keep for them. Currently, this need is filled by [ILP Plugin Virtual],
-which uses [Plugin RPC] as its ledger protocol. JSON messages are passed back
-and forth to perform transfers and send quote requests. Custom JSON messages
-can be added to extend the Plugin RPC messages, but they aren't well defined.
+do every ILP payment on-ledger. CLP can be used in these cases as long as the
+two parties trust one another (up to a limit). If their trust limit is high
+enough, they can transact without settling on an underlying ledger at all.
+
+Currently, this case is handled by [ILP Plugin Virtual], which uses [Plugin
+RPC] as its ledger protocol. JSON messages are passed back and forth to perform
+transfers and send quote requests. Custom JSON messages can be added to extend
+the Plugin RPC messages, just like sub-protocols can be added to CLP.
 
 Because the use case for a bilateral ledger protocols is so ubiquitous, the CLP
-has been invented to make the protocol extensible, efficient, and friendly to
-re-implementation.
+has been designed to be efficient, and friendly to re-implementation.
 
 ### Scope
 
-The CLP, despite its name, is just another ledger protocol. It manages
-conditional transfers, messaging requests, and allows [side protocols] for
-extensibility. You can use ILP without using CLP. CLP is intended to be a
-well-suited solution so that a new bilateral ledger protocol doesn't need to
-exist for every new use case.
+CLP manages conditional transfers, messaging requests, result/error reporting,
+and carries sub-protocols (sometimes called side-protocols) for extensibility.
+You can use ILP without using CLP. CLP is intended to be a well-suited solution
+so that a new bilateral ledger protocol doesn't need to exist for every new use
+case. It also includes functionality which is common between many different
+ledger types, making it a good place to start from when creating a new
+protocol.
 
-This document describes the flow and messages that CLP uses. It does not cover
-establishment of CLP channels, nor does it cover any side protocols.  Side
-protocols include functionality like ledger metadata, balance, automated
-settlement, and dispute resolution. 
+This document describes the flow and messages that CLP uses, but not
+sub-protocols. Sub-protocols include functionality like ledger metadata,
+balance, automated settlement, and dispute resolution. Some protocols are
+documented on [the wiki page]. They are carried in the protocol data of CLP
+messages.
 
 The CLP Packet format is described exactly in the [CLP ASN.1 spec].
 
 ## Terminology
 
-- **CLP** is the common ledger protocol, as described by this packet and the
+- **CLP** is the common ledger protocol, as described by this document and the
   ASN.1 spec.
 
-- A **Side Protocol** is a protocol which isn't defined by CLP and is carried
-  in the side protocol data (see below).
+- A **Sub-Protocol** is a protocol which isn't defined by CLP and is carried
+  in the protocol data (see below).
 
 - A **CLP Connection** is a websocket connection over which CLP Packets are
-  sent.
+  sent. Websockets are used because they provide message framing and allow CLP
+to use HTTP requests for authentication.
 
-- **CLP Packets** are the messages defined in the CLP ASN.1 spec.
+- **CLP Packets** are the messages described in this document. They are
+  formally defined in the [CLP ASN.1 spec].
 
 - **Peers** are the parties on a CLP connection. Your peer is the party on the
   other side of the CLP connection.
 
-- The **Bilateral Ledger** is the ledger which a CLP connection controls.  The
-  bilateral ledger is a persistent log of CLP packets, which can be used to
-deduce the current balance between two peers and the state of all transfers
-between them.
+- The **Bilateral Ledger** is the ledger which the peers on a CLP connection
+  are keeping track of. The bilateral ledger is a persistent log of CLP
+packets, which can be used to deduce the current balance between two peers and
+the state of all transfers between them.
 
-- **Authoritative State** means that the entire Bilateral Ledger is being
-  tracked by a party. Both peers on a CLP connection can keep authoritative
-state, which means they can get into dispute if they disagree on the state of a
-transfer. This usually happens when network latency causes the peers to
-disagree about expiries. If one party keeps authoritative state, the other
-party must trust them not to tamper with it.
+- **Authoritative State** is the authoritative view of the Bilateral Ledger's
+  state, maintained by one or both of the peers. Because both peers on a CLP
+connection can keep authoritative state, they can get into dispute if they
+disagree on the state of a transfer. This usually happens when network latency
+causes the peers to disagree about expiries. If one party keeps authoritative
+state, the other party must trust them not to tamper with it.
+
+- A request is **In-Flight** if the request has been sent out, but no response
+  has been sent yet. A transfer is **In-Flight** if it has been prepared but
+not yet fulfilled nor rejected.
 
 ## Overview
 
@@ -96,10 +107,10 @@ responses. Every CLP packet follows a common structure:
   .               |
   |               |
   +---------------+
-1 | Side Protocol |
+1 | Sub-Protocol  |
 2 | Count (5)     |
   +---------------+
-  | Side Protocol |
+  | Sub-Protocol  |
   | Data          |
   . (6)           |
   .               |
@@ -109,33 +120,57 @@ responses. Every CLP packet follows a common structure:
 ```
 
 1. **Type**: A 1-byte value describing what type of CLP packet this is.
-The [CLP ASN.1 spec] describes what the different type IDs mean.
+The values are described below, in [CLP Type IDs](#clp-type-ids).
 
-2. **Request ID**: A 4-byte value used to correlate this requests and
-responses.
+2. **Request ID**: A random 4-byte value used to correlate requests 
+and responses. This value MAY be sequential instead of random, but care must
+be taken so that duplicate IDs are never in-flight at the same time.
 
 3. **Length Prefix**: A 1-byte (if under 128) or 2-byte value, containing the
-combined length of the packet-specific data and side protocol data sections.
+combined length of the packet-specific data and protocol data sections.
 
 4. **Packet-Specific Data:** Fields specific to the type of CLP packet. Variable
 length.
 
-5. **Side Protocol Count:** Variable-length integer containing the number of
-side protocols carried by this packet.
+5. **Sub-Protocol Count:** Variable-length integer containing the number of
+sub-protocols carried by this packet.
 
-5. **Side Protocol Data:** A list of key-value pairs, where the key is a
-length-prefixed ascii string, and the value is a length-prefixed octet string.
-The key signifies the side protocol's name, and the value is the data.
+5. **Sub-Protocol Data:** A list of protocols, containing a string (the protocol's name), a 1-byte flag (containing the MIME type), and a length-prefixed octet string (containing the protocol's data). Exact description is below in [Sub-Protocol Data Format](#sub-protocol-data-format).
+
+### CLP Type IDs
+
+| ID | Type | Request/Response |
+|:--|:--|:--|
+| 1 | `Ack` | Response |
+| 2 | `Response` | Response |
+| 3 | `Error` | Response |
+| 4 | `Prepare` | Request |
+| 5 | `Fulfill` | Request |
+| 6 | `Reject` | Request |
+| 7 | `Message` | Request |
+
+### Sub-Protocol Data Format
+
+```asn1
+ContentType ::= INTEGER {
+  applicationOctetString  (0),
+  textPlainUtf8           (1),
+  applicationJson         (2)
+} (0..255)
+
+ProtocolData ::= SEQUENCE OF SEQUENCE {
+  protocolName IA5String,
+  contentType ContentType,
+  data OCTET STRING
+}
+```
 
 ## Flow
 
 CLP uses a simple RPC flow. A request-type CLP packet is sent, and a
-response-type CLP packet is sent in response with the same request ID.
-
-If a malformed CLP packet is sent, there should be no response. If a
-response-type CLP packet is sent without matching a previous request, it should
-be ignored.  If a request has already been responded to and another response is
-sent, the second response should be ignored.
+response-type CLP packet is sent in response with the same request ID. The
+request types are `Message`, `Prepare`, `Fulfill` and `Reject`, and the
+response types are `Ack` (which may be [removed]), `Response`, and `Error`.
 
 Because it would be too slow to atomically save all requestIds that are
 processed, they are not idempotent. It is the responsibility of the requestor
@@ -144,73 +179,108 @@ that no two in-flight requests are sent out with the same requestId. The
 responder should always send back a response to a request with the same
 requestId.
 
-The request types are `Message`, `Prepare`, `Fulfill` and `Reject`, and the
-response types are `Ack` (which may be [removed]), `Response`, and `Error`.
+There are also a couple of tricky cases to handle:
+
+- If an unexpected CLP packet is received, no response should be sent. An unexpected CLP packet is a response for which a request was not sent, or a response for a request which has already been responded to.
+- If an unreadable CLP packet is received, no response should be sent. An unreadable CLP packet is one which is structurally invalid, i.e. terminates before length prefixes dictate or contains illegal characters.
+
+These behaviors are important for preventing accidental feedback loops.  If an
+unexpected message triggered an error, that error may be unexpected to the
+sender. The sender would reply with another unexpected error, causing an
+infinite loop. Unreadable messages must be ignored too. If an application got
+onto a CLP connection and spoke the wrong protocol, it would trigger an error
+from CLP. This might trigger an error from the application, and it would
+devolve into another infinite loop.
 
 ### Message
 
-`Message` is used for sending information to the peer. It contains no message
-data, only side protocol data. ILQP packets are attached under the protocol
-`ilp`.
+```asn1
+Message ::= SEQUENCE {
+  protocolData ProtocolData
+}
+```
 
-#### Response
+`Message` is used for sending information to the peer. It contains no
+packet-specific data, only protocol data. [ILQP] packets are attached under the
+protocol name `ilp` with content-type `application/octet-stream`.
 
 - `Response` is returned if the peer acknowledges the `Message`. If the peer
-  has data to send in reply (e.g. a quote response), it is carried in the side
+  has data to send in reply (e.g. a quote response), it is carried in the
 protocol data.
 
 - `Error` is returned if the peer is not able to process the message, or there
-  was an unexpected error and further messages should not be sent.
+  was an unexpected error and further messages should not be sent. This does
+not include ILP errors such as "No quote found," only the cases in which the
+message cannot be sent/processed at all.
 
 ### Prepare
+
+```asn1
+Prepare ::= SEQUENCE {
+  transferId UInt128,
+  amount UInt64,
+  executionCondition UInt256,
+  expiresAt Timestamp,
+  --
+  protocolData ProtocolData
+}
+```
 
 `Prepare` is used to create a transfer on the bilateral ledger. The message
 data contains `transferId`, `amount`, `executionCondition`, and `expiresAt`.
 `Prepare` is a request with side effects, because it creates a transfer with
 the given details. This transfer begins in the `prepared` state.
 
-`transferId` is a 128-bit unique ID that references the transfer. `amount` is
-a 64-bit integer denominated in ledger base units. The ledger base units can be
-anything, so long as both parties on the bilateral ledger agree on their
-meaning. Examples of base units are "micro-XRP," "satoshi," or "nano-dollars
-settled over paypal." `executionCondition` is a 256-bit integer containing the
-SHA-256 hash of this conditional transfer's fulfillment.  `expiresAt` is a
-ASN.1 UTC GeneralizedTime containing the expiry of this transfer.
+`transferId` is a securely random 128-bit unique ID that references the
+transfer.  `amount` is a 64-bit integer denominated in ledger base units. The
+ledger base units can be anything, so long as both parties on the bilateral
+ledger agree on their meaning. Examples of base units are "micro-XRP,"
+"satoshi," or "nano-dollars settled over paypal." `executionCondition` is a
+256-bit integer containing the SHA-256 hash of this conditional transfer's
+fulfillment.  `expiresAt` is a ASN.1 UTC GeneralizedTime containing the expiry
+of this transfer. The GeneralizedTime MUST be in UTC (i.e. no timezone info).
 
-ILP payment packets are attached to the side protocol data under the protocol
-`ilp`. 
-
-#### Response
+ILP payment packets are attached to the protocol data under the protocol
+name `ilp` and the MIME type `application/octet-stream`. 
 
 - `Response` is returned if the peer acknowledges the `Prepare`. This means the
   transfer is now `prepared` and has been applied to the balance. It may carry
-side protocol data, for example a payment channel claim.
+sub-protocol data, for example a payment channel claim.
 
 - `Error` is returned if the peer does not accept the transfer. This could be
   because it is incompatible with an existing transfer with the same ID (see
 Idempotency below), or because it exceeds some constraint your peer has placed.
-The amount could be too high, or could exceeds the balance.
+The amount could be too high, or could exceeds the balance. If the transfer ever enters the `prepared` state, an `Error` should not be returned. Instead, a `Reject` call should be made.
 
 #### Idempotency
 
 If a transfer with the given ID already exists in any state, a new transfer
-should not be created. If the packet (including side protocol data) matches an
+should not be created. If the packet (including protocol data) matches an
 existing transfer exactly and that transfer is in the `prepared` state, a
-`Response` should be returned. If the packet (including side protocol data) shares
+`Response` should be returned. If the packet (including protocol data) shares
 the ID of an existing transfer but other details do not match, an `Error`
-should be returned. If the packet (including side protocol data) exactly
+should be returned. If the packet (including protocol data) exactly
 matches an existing transfer but that transfer is in the `fulfilled` state or
 the `rejected` state, an `Error` should be returned.
 
 #### Expiry
 
 After the `expiresAt` date is reached, the transfer can no longer be fulfilled.
-If one party on the CLP connection is keeping authoritative state, they should
+If one party on the CLP connection is keeping authoritative state, they MUST
 send a `Reject` request to the other party. If both parties are keeping
-authoritative state, they can independently expire the transfer (set the state
+authoritative state, they MAY independently expire the transfer (set the state
 to `rejected` and roll it back) automatically.
 
 ### Fulfill
+
+```asn1
+Fulfill ::= SEQUENCE {
+  transferId UInt128,
+  fulfillment UInt256,
+  --
+  protocolData ProtocolData
+}
+```
 
 `Fulfill` is used to change an existing transfer from the `prepared` state to
 the `fulfilled` state. The packet-specific data of `Fulfill` is made up of `transferId`
@@ -222,29 +292,36 @@ transfer is already fulfilled, no changes are applied.
 `fulfillment` is a 256-bit integer, containing the preimage of the
 `executionCondition` of the referenced transfer.
 
-#### Response
-
 A `Fulfill` request must be sent from the receiver of the referenced transfer,
 not the sender.
 
 A `Fulfill` request is successful if `transferId` references an existing
 transfer in the `prepared` state, the `fulfillment` is the SHA-256 preimage of
-the referenced transfer's `executionCondition`, and the current time is before
-the referenced transfer's `expiresAt`.
+the referenced transfer's `executionCondition`, and the current time (when the
+request is being processed) is before the referenced transfer's `expiresAt`.
 
 A `Fulfill` request is also successful if the transfer is in the `fulfilled`
 state, and this packet matches the one that fulfilled the transfer,
-including side protocols but not including requestId.
+including protocols but not including requestId.
 
-- `Response` is returned if the request is successful.  data to be passed back.
-  This means the transfer's state is now `fulfilled`, and its balance change
-has been finalized.
+- `Response` is returned if the request is successful.  This means the
+  transfer's state is now `fulfilled`, and its balance change has been
+finalized.
 
 - `Error` is returned if there is no transfer with the given ID, the
   fulfillment does not match the condition, the expiresAt has already passed,
 or the transfer has been `rejected`.
 
 ### Reject
+
+```
+Reject ::= SEQUENCE {
+  transferId UInt128,
+  rejectionReason InterledgerPacket, -- must contain an InterledgerProtocolError
+  --
+  protocolData ProtocolData
+}
+```
 
 `Reject` is used to change an existing transfer from the `prepared`
 state to the `rejected` state. The packet-specific data of `Reject` is made up
@@ -254,8 +331,6 @@ of `transferId` and `rejectionReason`.
 `rejectionReason` is an [ILP Error], containing the reason that this transfer
 was rejected.
 
-#### Response
-
 A `Reject` request must come from the receiver of the referenced transfer,
 not the sender.
 
@@ -264,7 +339,7 @@ transfer in the `prepared` state.
 
 A `Reject` request is also successful if `transferId` references an existing
 transfer in the `rejected` state, and this packet matches the one that
-rejected this transfer, including side protocols but not including requestId.
+rejected this transfer, including protocols but not including requestId.
 
 - A `Response` is returned if the request is successful. This indicates that
   the balance changes of the referenced transfer have been rolled back.
--- a/00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
+++ b/00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
@@ -1,0 +1,273 @@
+# Common Ledger Protocol (CLP)
+
+## Preface
+
+This document describes the Common Ledger Protocol (CLP), a ledger protocol for
+bilateral liquidity relationships. It is similar in function to [Plugin RPC],
+but has been rewritten to use OER instead of JSON, and extensibility features
+have been added.
+
+## Introduction
+
+### Motivation
+
+There are many different types of ledgers, and Interledger aims to interoperate
+all of them. For fast ledgers that support hashlocks, it is enough to wrap the
+ledger's API in a [Ledger Plugin]. These allow you to send interledger payments
+through anyone else on that ledger who is running your plugin.
+
+In lots of scenarios, we don't have an underlying ledger that's fast enough to
+do every ILP payment on-ledger. There are also many situations when two parties
+trust one another up to an amount, and don't need any third-party ledger to
+book-keep for them. Currently, this need is filled by [ILP Plugin Virtual],
+which uses [Plugin RPC] as its ledger protocol. JSON messages are passed back
+and forth to perform transfers and send quote requests. Custom JSON messages
+can be added to extend the Plugin RPC messages, but they aren't well defined.
+
+Because the use case for a bilateral ledger protocols is so ubiquitous, the CLP
+has been invented to make the protocol extensible, efficient, and friendly to
+re-implementation.
+
+### Scope
+
+The CLP, despite its name, is just another ledger protocol. It manages
+conditional transfers, messaging requests, and allows [side protocols] for
+extensibility. You can use ILP without using CLP. CLP is intended to be a
+well-suited solution so that a new bilateral ledger protocol doesn't need to
+exist for every new use case.
+
+This document describes the flow and messages that CLP uses. It does not cover
+establishment of CLP channels, nor does it cover any side protocols.  Side
+protocols include functionality like ledger metadata, balance, automated
+settlement, and dispute resolution. 
+
+The CLP Packet format is described exactly in the [CLP ASN.1 spec].
+
+## Terminology
+
+- **CLP** is the common ledger protocol, as described by this packet and the
+  ASN.1 spec.
+
+- A **Side Protocol** is a protocol which isn't defined by CLP and is carried
+  in the side protocol data (see below).
+
+- A **CLP Connection** is a websocket connection over which CLP Packets are
+  sent.
+
+- **CLP Packets** are the messages defined in the CLP ASN.1 spec.
+
+- **Peers** are the parties on a CLP connection. Your peer is the party on the
+  other side of the CLP connection.
+
+- The **Bilateral Ledger** is the ledger which a CLP connection controls.  The
+  bilateral ledger is a persistent log of CLP packets, which can be used to
+deduce the current balance between two peers and the state of all transfers
+between them.
+
+- **Authoritative State** means that the entire Bilateral Ledger is being
+  tracked by a party. Both peers on a CLP connection can keep authoritative
+state, which means they can get into dispute if they disagree on the state of a
+transfer. This usually happens when network latency causes the peers to
+disagree about expiries. If one party keeps authoritative state, the other
+party must trust them not to tamper with it.
+
+## Overview
+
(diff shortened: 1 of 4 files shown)


00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,262 @@
...
+and forth to perform transfers and send quote requests. Custom JSON messages
+can be added to extend the Plugin RPC messages, but they aren't well defined.
+
+Because the use case for a bilateral ledger protocols is so ubiquitous, the CLP
+has been invented to make the protocol extensible, efficient, and friendly to
+re-implementation.
+
+### Scope
+
+The CLP, despite its name, is just another ledger protocol. It manages
```
emschwartz: This makes it sound like a really inappropriate name. The reason it's called that is 1) because we think that it will be very widespread due to the fact that trustlines are such a useful way of modeling accounts in an ILP setup and 2) because it is made up of pieces that can also be common to a lot of other ledger (plugin) protocols (like all of the bilateral payment channel ones).

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,262 @@
...
+conditional transfers, messaging requests, and allows [side protocols] for
+extensibility. You can use ILP without using CLP. CLP is intended to be a
+well-suited solution so that a new bilateral ledger protocol doesn't need to
+exist for every new use case.
+
+This document describes the flow and messages that CLP uses. It does not cover
+establishment of CLP channels, nor does it cover any side protocols.  Side
```
emschwartz: It would be useful to describe some recommendations for that (and authentication) in this doc

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,262 @@
...
+
+## Terminology
+
+- **CLP** is the common ledger protocol, as described by this packet and the
+  ASN.1 spec.
+
+- A **Side Protocol** is a protocol which isn't defined by CLP and is carried
+  in the side protocol data (see below).
+
+- A **CLP Connection** is a websocket connection over which CLP Packets are
```
emschwartz: Is it only Websockets? I thought we were also going to support TLS sockets

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,262 @@
...
+  sent.
+
+- **CLP Packets** are the messages defined in the CLP ASN.1 spec.
```
emschwartz: It would probably be worth defining them in this doc (like how RFC 3 includes the definition of the packet format along with more explanation)

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+  | Data          |
+  . (6)           |
+  .               |
+  .               |
+  |               |
+  +---------------+
+```
+
+1. **Type**: A 1-byte value describing what type of CLP packet this is.
+The [CLP ASN.1 spec] describes what the different type IDs mean.
```
emschwartz: They should be defined here as well

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+
+2. **Request ID**: A 4-byte value used to correlate this requests and
```
emschwartz: Random or sequential?

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
emschwartz: It would also be good to have a section on side protocols and to give some examples of what those might look like and how you're supposed to take advantage of that extensibility mechanism

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+5. **Side Protocol Data:** A list of key-value pairs, where the key is a
+length-prefixed ascii string, and the value is a length-prefixed octet string.
+The key signifies the side protocol's name, and the value is the data.
+
+## Flow
+
+CLP uses a simple RPC flow. A request-type CLP packet is sent, and a
+response-type CLP packet is sent in response with the same request ID.
+
+If a malformed CLP packet is sent, there should be no response. If a
```
emschwartz: Might be worth mentioning why it's so important not to respond to unexpected responses

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+responder should always send back a response to a request with the same
+requestId.
+
+The request types are `Message`, `Prepare`, `Fulfill` and `Reject`, and the
+response types are `Ack` (which may be [removed]), `Response`, and `Error`.
+
+### Message
+
+`Message` is used for sending information to the peer. It contains no message
+data, only side protocol data. ILQP packets are attached under the protocol
```
emschwartz: Link to ILQP

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+  was an unexpected error and further messages should not be sent.
+
+### Prepare
+
+`Prepare` is used to create a transfer on the bilateral ledger. The message
+data contains `transferId`, `amount`, `executionCondition`, and `expiresAt`.
+`Prepare` is a request with side effects, because it creates a transfer with
+the given details. This transfer begins in the `prepared` state.
+
+`transferId` is a 128-bit unique ID that references the transfer. `amount` is
```
emschwartz: Should you use random or sequential IDs?

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+
+CLP uses a simple RPC flow. A request-type CLP packet is sent, and a
+response-type CLP packet is sent in response with the same request ID.
+
+If a malformed CLP packet is sent, there should be no response. If a
+response-type CLP packet is sent without matching a previous request, it should
+be ignored.  If a request has already been responded to and another response is
+sent, the second response should be ignored.
+
+Because it would be too slow to atomically save all requestIds that are
```
emschwartz: This could use some clarification. Transfers must be idempotent, but it doesn't matter if other requests aren't

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+`Response` should be returned. If the packet (including side protocol data) shares
+the ID of an existing transfer but other details do not match, an `Error`
+should be returned. If the packet (including side protocol data) exactly
+matches an existing transfer but that transfer is in the `fulfilled` state or
+the `rejected` state, an `Error` should be returned.
+
+#### Expiry
+
+After the `expiresAt` date is reached, the transfer can no longer be fulfilled.
+If one party on the CLP connection is keeping authoritative state, they should
```
emschwartz: Is this optional? This spec should probably use the RFC "SHOULD", "MUST" vocab

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+A `Fulfill` request is successful if `transferId` references an existing
+transfer in the `prepared` state, the `fulfillment` is the SHA-256 preimage of
+the referenced transfer's `executionCondition`, and the current time is before
+the referenced transfer's `expiresAt`.
+
+A `Fulfill` request is also successful if the transfer is in the `fulfilled`
+state, and this packet matches the one that fulfilled the transfer,
+including side protocols but not including requestId.
+
+- `Response` is returned if the request is successful.  data to be passed back.
```
emschwartz: Sentence fragment: "data to be passed back"

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+transfer in the `prepared` state.
+
+A `Reject` request is also successful if `transferId` references an existing
+transfer in the `rejected` state, and this packet matches the one that
+rejected this transfer, including side protocols but not including requestId.
+
+- A `Response` is returned if the request is successful. This indicates that
+  the balance changes of the referenced transfer have been rolled back.
+
+- An `Error` is returned if the request was not successful. _TODO: create
```
emschwartz: Yup, that's an important TODO

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
sharafian: Transfers are idempotent with respect to the transferId, but that's different from RPC-level idempotency

emschwartz: I understood that after reading through the whole doc but while reading this section I thought it was saying that nothing was idempotent

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
sharafian: How detailed should the definition be? As the spec is right now, it describes the fields included in each of the messages and their types, it just doesn't go into the precise ordering of the fields like the ASN.1

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
sharafian: Right now the implementations only use websockets. Should every implementation implement both, or is one of them optional, or are they both optional? We should probably look more into the performance gain of TLS as compared to websockets

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
adrianhopebailie: Yeh, it seems odd to not return a response to a bad request.

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+- `Response` is returned if the peer acknowledges the `Message`. If the peer
+  has data to send in reply (e.g. a quote response), it is carried in the side
+protocol data.
+
+- `Error` is returned if the peer is not able to process the message, or there
+  was an unexpected error and further messages should not be sent.
+
+### Prepare
+
+`Prepare` is used to create a transfer on the bilateral ledger. The message
```
adrianhopebailie: Can we call this PrepareTransfer?

Verbosity in descriptions is really helpful to anyone reading this for the first time and it adds nothing to the encoded data.

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
adrianhopebailie: This ambiguity comes from the fact that requestId and transferId are at different layers in this stack.

`requestId` is at the very bottom layer of the CLP stack and is simply for matching request/responses. There is no application logic that looks at these, in fact they could be invisible to the code that processes the actual message as long as it gets the correct response for its request.

`transferId` is related to business logic (a layer up in the CLP stack) where idempotency is enforced.

I think it would be less confusing if messages also had an ID that was relevant to the business logic layer.

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+settlement, and dispute resolution. 
+
+The CLP Packet format is described exactly in the [CLP ASN.1 spec].
+
+## Terminology
+
+- **CLP** is the common ledger protocol, as described by this packet and the
+  ASN.1 spec.
+
+- A **Side Protocol** is a protocol which isn't defined by CLP and is carried
```
adrianhopebailie: I hate this name 😄 

CLP is basically empty without these protocols and they are not carried "on the side" they are carried inside CLP. This is regular encapsulation so we should just use standard terminology: **protocol** (or if we have to __sub-protocol__)

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
adrianhopebailie: I'd assume few humans ever read the ASN.1 files. Most specs that have ASN.1 use fragments of it throughout the spec and then have the full definition as an appendix.

I'd recommend using ASN.1 fragments in the spec and then following those with definitions.

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+  in the side protocol data (see below).
+
+- A **CLP Connection** is a websocket connection over which CLP Packets are
+  sent.
+
+- **CLP Packets** are the messages defined in the CLP ASN.1 spec.
+
+- **Peers** are the parties on a CLP connection. Your peer is the party on the
```
adrianhopebailie: Are you using the terms **CLP connection** and **CLP channel** interchangeably?

Earlier you talk about establishing a CLP channel, which I interpret as something logical (i.e. the physical connection could be anything, could go through proxies etc.)

I think of peers as logical entities on either side of a channel as opposed to some systems on either side of the connection. Which do you mean?

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+  other side of the CLP connection.
+
+- The **Bilateral Ledger** is the ledger which a CLP connection controls.  The
```
adrianhopebailie: This makes no sense to me. A connection can't control something. Is the bilateral ledger an actual stateful system or is it a logical thing that is "shared" by the two peers?

In reality a channel will have two ledgers, one maintained by each peers, that should always be in sync so it sounds like the **bilateral ledger** is just a logical construct?

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+  bilateral ledger is a persistent log of CLP packets, which can be used to
+deduce the current balance between two peers and the state of all transfers
+between them.
+
+- **Authoritative State** means that the entire Bilateral Ledger is being
```
adrianhopebailie: This paragraph is hard to understand. I'd change first sentence to something like:

> **Authoritative State** is the authoritative view of the state of the bilateral ledger maintained by either one or both peers. In the case where both peers keep ...

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+  | Side Protocol |
+  | Data          |
+  . (6)           |
+  .               |
+  .               |
+  |               |
+  +---------------+
+```
+
+1. **Type**: A 1-byte value describing what type of CLP packet this is.
```
adrianhopebailie: It may make things less confusing to not call this a CLP packet. `Message` or `Call` (common in other RPC protocols)?

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+`Prepare` is a request with side effects, because it creates a transfer with
+the given details. This transfer begins in the `prepared` state.
+
+`transferId` is a 128-bit unique ID that references the transfer. `amount` is
+a 64-bit integer denominated in ledger base units. The ledger base units can be
+anything, so long as both parties on the bilateral ledger agree on their
+meaning. Examples of base units are "micro-XRP," "satoshi," or "nano-dollars
+settled over paypal." `executionCondition` is a 256-bit integer containing the
+SHA-256 hash of this conditional transfer's fulfillment.  `expiresAt` is a
+ASN.1 UTC GeneralizedTime containing the expiry of this transfer.
```
adrianhopebailie: Should also say that it is restricted to UTC (i.e. no timezone info)

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+matches an existing transfer but that transfer is in the `fulfilled` state or
+the `rejected` state, an `Error` should be returned.
+
+#### Expiry
+
+After the `expiresAt` date is reached, the transfer can no longer be fulfilled.
+If one party on the CLP connection is keeping authoritative state, they should
+send a `Reject` request to the other party. If both parties are keeping
+authoritative state, they can independently expire the transfer (set the state
+to `rejected` and roll it back) automatically.
```
adrianhopebailie: Uses same `transferId` as original `Prepare`?

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+bilateral liquidity relationships. It is similar in function to [Plugin RPC],
+but has been rewritten to use OER instead of JSON, and extensibility features
+have been added.
+
+## Introduction
+
+### Motivation
+
+There are many different types of ledgers, and Interledger aims to interoperate
+all of them. For fast ledgers that support hashlocks, it is enough to wrap the
```
michielbdejong: s/hashlocks/hash time locks, memos, and messaging.

You need the timeout for expiresAt, the memo for the ILP packet, and the messaging for ILQP.

They also need to support losslessly transferring an integer amount from any account to any other account and checking your integer balance but we can say if they don't support those then the thing is not a ledger.

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+ledger's API in a [Ledger Plugin]. These allow you to send interledger payments
+through anyone else on that ledger who is running your plugin.
+
+In lots of scenarios, we don't have an underlying ledger that's fast enough to
+do every ILP payment on-ledger. There are also many situations when two parties
+trust one another up to an amount, and don't need any third-party ledger to
```
michielbdejong: These can't be disjoint, CLP is useless unless you have **both** a ledger that's slower than direct CLP **and** at least a little bit of off-ledger trust.

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+book-keep for them. Currently, this need is filled by [ILP Plugin Virtual],
+which uses [Plugin RPC] as its ledger protocol. JSON messages are passed back
+and forth to perform transfers and send quote requests. Custom JSON messages
+can be added to extend the Plugin RPC messages, but they aren't well defined.
```
michielbdejong: They are not better defined in CLP than in RPC; you still need to define siteProtocols in the same way you need to define the contents of `custom` for RPC. There is a difference though: CLP offers `protocolName` and `contentType` so you can more easily multiplex your side protocols without clashes.

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+
+Because the use case for a bilateral ledger protocols is so ubiquitous, the CLP
+has been invented to make the protocol extensible, efficient, and friendly to
```
michielbdejong: s/invented to make the protocol/designed to be

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
michielbdejong: I agree with Evan; we discussed this in the research team meeting yesterday. Sure, other protocols, like the existing RPC for instance, can do the same as CLP, but we think CLP is the best one, and we want people to use it whenever they implement Interledger from scratch, right? Or are you saying we don't care about that?

We also don't say in the ILQP spec that it's "just another quoting protocol".

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+re-implementation.
+
+### Scope
+
+The CLP, despite its name, is just another ledger protocol. It manages
+conditional transfers, messaging requests, and allows [side protocols] for
```
michielbdejong: s/messaging requests, and/messaging requests, error and result reporting, and

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
michielbdejong: Yes, both establishment and side protocols, see https://github.com/interledger/interledger/wiki/Interledger-over-CLP. We can copy-paste from there.

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
michielbdejong: In the asn.1 it's just called 'protocolData'. The term 'side protocol' was used in earlier versions, where there was a separation between ilp and the rest.

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
michielbdejong: Yeah, one option we discussed is to define "Request/Response over WebSockets (RRoW)", and then do "CLP over RRoW". It would mean the requestId disappears from CLP (but the transferId stays), and then it would make more sense to do CLP over http or over grpc.

But Stefan had some good arguments yesterday to say only webSockets, because ILP packets (ILQP, payments, errors) are already the common layer. So you either do ILP packets over [custom] or ILP packets over CLP, and in the latter case, let's just say you need to do it all-the-way including the WebSockets.

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+  +---------------+
+1 | Type (1)      |
+  +---------------+
+1 | Request ID    |
+2 | (2)           |
+3 |               |
+4 |               |
+  +---------------+
+1 | Length Prefix |
+2 | (3)           |
```
michielbdejong: I think this length prefix isn't there, because packet-specific data is either empty or a `SEQUENCE`, and sequences don't have lenght prefixes. See https://github.com/interledgerjs/clp-packet/blob/mj-testnet-1/test/indexSpec.js#L58

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+  +---------------+
+  | Packet-       |
+  | Specific      |
+  | Data          |
+  . (4)           |
+  .               |
+  .               |
+  |               |
+  +---------------+
+1 | Side Protocol |
```
michielbdejong: not 'Side'

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+2 | Count (5)     |
+  +---------------+
+  | Side Protocol |
```
michielbdejong: not 'Side'

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+```
+
+1. **Type**: A 1-byte value describing what type of CLP packet this is.
+The [CLP ASN.1 spec] describes what the different type IDs mean.
+
+2. **Request ID**: A 4-byte value used to correlate this requests and
+responses.
+
+3. **Length Prefix**: A 1-byte (if under 128) or 2-byte value, containing the
+combined length of the packet-specific data and side protocol data sections.
```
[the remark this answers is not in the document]
michielbdejong: See above, I think `Type (1) ` already defines how much bytes should be in `Packet-Specific Data (4)`.

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+
+4. **Packet-Specific Data:** Fields specific to the type of CLP packet. Variable
+length.
+
+5. **Side Protocol Count:** Variable-length integer containing the number of
+side protocols carried by this packet.
+
+5. **Side Protocol Data:** A list of key-value pairs, where the key is a
```
michielbdejong: not key-value pairs; it's a `SEQUENCE-OF`, so more like an array. Each element in the sequence has 3 parts, `protocolName`, `contentType`, and `data` (not two parts 'key' and 'value')

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+- **Authoritative State** means that the entire Bilateral Ledger is being
+  tracked by a party. Both peers on a CLP connection can keep authoritative
+state, which means they can get into dispute if they disagree on the state of a
+transfer. This usually happens when network latency causes the peers to
+disagree about expiries. If one party keeps authoritative state, the other
+party must trust them not to tamper with it.
+
+## Overview
+
+CLP is broken up into 4 different RPC requests, which can get 2 different
```
michielbdejong: You are combining the 'remove Ack' proposal into this PR, which I guess is not ideal, but if you do that, then at least also edit the asn.1 file.

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
michielbdejong: we should define what we mean with 'malformed CLP packet':
### unreadable packet (cannot answer):
* the first byte is outside the callset
* the total length is less than 5 bytes

### unexpected response (should not answer)
This is the one Evan refers to. The first byte indicates a response call (i.e., Ack, Response, or Error), and:
* the requestId is unknown to the recipient, or
* the requestId was used before in a well-formed request call, but after that the parser reached the end of the packet too early or too late.
* there is for instance an error code which is not listed in https://interledger.org/rfcs/0003-interledger-protocol/#errors

### unexpected request content (should answer with CLP-level error)
* the first byte indicates a request call (i.e., Message, Prepare, Fulfill, or Reject), and the next 4 bytes are a requestId that wasn't used before, but after that the parser reached the end of the packet too early or too late
* there is for instance an expiresAt that is in the past

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+
+Because it would be too slow to atomically save all requestIds that are
+processed, they are not idempotent. It is the responsibility of the requestor
+to make sure they don't duplicate requestIds. The implementation should ensure
+that no two in-flight requests are sent out with the same requestId. The
+responder should always send back a response to a request with the same
+requestId.
+
+The request types are `Message`, `Prepare`, `Fulfill` and `Reject`, and the
+response types are `Ack` (which may be [removed]), `Response`, and `Error`.
```
michielbdejong: if you don't then update above where it says '2 response calls'

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
michielbdejong: s/message data/packet-specific data
remove 'side'

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+
+### Message
+
+`Message` is used for sending information to the peer. It contains no message
+data, only side protocol data. ILQP packets are attached under the protocol
+`ilp`.
```
michielbdejong: protocolName 'ilp' and with contentType [octetString](https://github.com/interledger/rfcs/blob/master/asn1/CommonLedgerProtocol.asn#L39)

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+
+#### Response
+
+- `Response` is returned if the peer acknowledges the `Message`. If the peer
+  has data to send in reply (e.g. a quote response), it is carried in the side
+protocol data.
+
+- `Error` is returned if the peer is not able to process the message, or there
+  was an unexpected error and further messages should not be sent.
```
michielbdejong: clarify whether this includes errors on the protocol level, e.g. an ILP error saying 'no quote found'

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+`transferId` is a 128-bit unique ID that references the transfer. `amount` is
+a 64-bit integer denominated in ledger base units. The ledger base units can be
+anything, so long as both parties on the bilateral ledger agree on their
+meaning. Examples of base units are "micro-XRP," "satoshi," or "nano-dollars
+settled over paypal." `executionCondition` is a 256-bit integer containing the
+SHA-256 hash of this conditional transfer's fulfillment.  `expiresAt` is a
+ASN.1 UTC GeneralizedTime containing the expiry of this transfer.
+
+ILP payment packets are attached to the side protocol data under the protocol
+`ilp`.
```
michielbdejong: protocolName 'ilp' and with contentType octetString

michielbdejong: no 'side'

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+ASN.1 UTC GeneralizedTime containing the expiry of this transfer.
+
+ILP payment packets are attached to the side protocol data under the protocol
+`ilp`. 
+
+#### Response
+
+- `Response` is returned if the peer acknowledges the `Prepare`. This means the
+  transfer is now `prepared` and has been applied to the balance. It may carry
+side protocol data, for example a payment channel claim.
```
michielbdejong: in this case, 'side' is actually accurate, maybe we should say 'it may carry protocolData, but only for "side protocols", that is, protocols other than the one used in the prepare. Or just keep Ack, that would be simpler ;)

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+
+- `Error` is returned if the peer does not accept the transfer. This could be
+  because it is incompatible with an existing transfer with the same ID (see
+Idempotency below), or because it exceeds some constraint your peer has placed.
+The amount could be too high, or could exceeds the balance.
```
michielbdejong: Clarify that this should carry an error from https://interledger.org/rfcs/0003-interledger-protocol/#errors and clarify your stance on https://github.com/interledger/rfcs/issues/290

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+
+#### Idempotency
+
+If a transfer with the given ID already exists in any state, a new transfer
+should not be created. If the packet (including side protocol data) matches an
```
michielbdejong: no 'side'

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
michielbdejong: So you're proposing to make sending the 'Reject' optional, in this case? But only if you're sure the preparer is also keeping time? How can you know that? Wouldn't it be simpler to leave it as a Prepare that has had a Response (Ack) should always get either a Fulfill or a Reject?

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+
+`Fulfill` is used to change an existing transfer from the `prepared` state to
+the `fulfilled` state. The packet-specific data of `Fulfill` is made up of `transferId`
+and `fulfillment`. `Fulfill` is a request with side effects: it changes the
+state of the transfer and finalizes the transfer's balance update. If the
+transfer is already fulfilled, no changes are applied.
+
+`transferId` is a 128-bit unique ID, which references an existing transfer.
+`fulfillment` is a 256-bit integer, containing the preimage of the
+`executionCondition` of the referenced transfer.
```
michielbdejong: I think you added this by accident? If not, then please explain the motiviation and add it in  https://github.com/interledger/rfcs/blob/master/asn1/CommonLedgerProtocol.asn#L80 as well.

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+
+#### Response
```
michielbdejong: There already is a 'response' section above - maybe rename to 'response to fulfill'?

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+
+A `Fulfill` request must be sent from the receiver of the referenced transfer,
+not the sender.
+
+A `Fulfill` request is successful if `transferId` references an existing
+transfer in the `prepared` state, the `fulfillment` is the SHA-256 preimage of
+the referenced transfer's `executionCondition`, and the current time is before
+the referenced transfer's `expiresAt`.
```
michielbdejong: Current time when the last byte of the message is put on the wire? Or when its last byte reaches the preparer?

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+
+A `Fulfill` request is also successful if the transfer is in the `fulfilled`
+state, and this packet matches the one that fulfilled the transfer,
+including side protocols but not including requestId.
```
michielbdejong: i think it should always be successful if it successfully fulfills the transfer, so even if the protocolData is different or if the requestId is repeated.

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+
+- `Response` is returned if the request is successful.  data to be passed back.
+  This means the transfer's state is now `fulfilled`, and its balance change
+has been finalized.
```
michielbdejong: That depends on who is authoritative; if the slave fulfills, then this statement is correct, but if the master fulfills, the balance change is already finalized, even if the slave doesn't respond, or responds with an error. Note that this is independent of which party is the WebSocket server and which one is the WebSocket client.

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+
+### Reject
+
+`Reject` is used to change an existing transfer from the `prepared`
+state to the `rejected` state. The packet-specific data of `Reject` is made up
+of `transferId` and `rejectionReason`.
+
+`transferId` is a 128-bit unique ID, which references an existing transfer.
+`rejectionReason` is an [ILP Error], containing the reason that this transfer
+was rejected.
```
michielbdejong: mention if sending a reject is optional or not (above, you stated it can in some situations be ok to just let the transfer expire)

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+
+A `Reject` request must come from the receiver of the referenced transfer,
+not the sender.
+
+A `Reject` request is successful if `transferId` references an existing
+transfer in the `prepared` state.
+
+A `Reject` request is also successful if `transferId` references an existing
+transfer in the `rejected` state, and this packet matches the one that
+rejected this transfer, including side protocols but not including requestId.
```
michielbdejong: I would say excluding both requestId and protocolData.

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+
+- A `Response` is returned if the request is successful. This indicates that
+  the balance changes of the referenced transfer have been rolled back.
```
michielbdejong: I still think the current implementation with `Ack` is nicer; it feels weird to call it a "response" if you have nothing to say.

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
michielbdejong: important, yes, but  that list should be at https://interledger.org/rfcs/0003-interledger-protocol/#ilp-error-codes (even for ledger-level errors (whose codes that start with 'L'), probably).

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+This document describes the flow and messages that CLP uses. It does not cover
+establishment of CLP channels, nor does it cover any side protocols.  Side
+protocols include functionality like ledger metadata, balance, automated
+settlement, and dispute resolution. 
+
+The CLP Packet format is described exactly in the [CLP ASN.1 spec].
+
+## Terminology
+
+- **CLP** is the common ledger protocol, as described by this packet and the
```
dappelt: "... _as described by this packet_ ...". It is not clear from the context what _this packet_ is. Do you mean to write "... _as described by this document_ ..."?

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
dappelt: +1 sub-protocol

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+  ASN.1 spec.
+
+- A **Side Protocol** is a protocol which isn't defined by CLP and is carried
+  in the side protocol data (see below).
+
+- A **CLP Connection** is a websocket connection over which CLP Packets are
+  sent.
+
+- **CLP Packets** are the messages defined in the CLP ASN.1 spec.
```
dappelt: > **CLP Packets** are the messages ...

We should not use _packets_ and _messages_ synonymously in our CLP terminology. There is already a CLP packet type `Message`. Referring to the entire CLP packet as _the message_ will lead to ambiguities. In this sentence, I propose to use either _datagram_ or _unit of transfer_ instead of _messages_.

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+
+- **Peers** are the parties on a CLP connection. Your peer is the party on the
+  other side of the CLP connection.
+
+- The **Bilateral Ledger** is the ledger which a CLP connection controls.  The
+  bilateral ledger is a persistent log of CLP packets, which can be used to
```
dappelt: > ... is a persistent log of CLP packets ...

Rather than persisting a log of CLP packets, a ledger persists a state that is the result of a series of CLP packets.

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
dappelt: Yes, in a "*bilateral ledger*" both peers have a ledger that should be in sync, so it is a logical construct.

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
dappelt: Agreeing with @michielbdejong. In particular, there should be a distinction between handling unexpected response-type packets and handling unexpected/malformed request-type packets.

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
earizon: webSockets implicitly means one of the host is listening  for incoming webSocket connection establishment. Is there any preference in the CLP RFC about which host (ledger, connector, ...?) must be listening. The websocket can be setup for short periods of times, or "forever". Is the CLP protocol expected to be used in a particular way. (Looks to me that "wallets" will establish short connections, while connectors will be long-lasting).

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
earizon: Peer is confusing from my point of view. They are just peers at the websocket transport layer, but one is acting as server and the other one is a client, even if a given setup allows both CLP hosts to act as client and server for non-related messages.

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
earizon: A ledger can be anything. For example in the UTXO model of Bitcoin there is no real balance, just unexpent transactions outputs. Other ledgers can keep an internal database,  reset the state or forward to other ledgers, but the CLP packets will always be interpreted as external events arriving to the "internal ledger".

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
dappelt: Why is CLP packet confusing? 

Let's not use `Message` and CLP packet synonymously. (see my previous [comment](https://github.com/interledger/rfcs/pull/291/files/8e09dc9e044797097e6f5803c3abc4eb353cef1e?diff=unified#r137229950))

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+
+If a malformed CLP packet is sent, there should be no response. If a
+response-type CLP packet is sent without matching a previous request, it should
+be ignored.  If a request has already been responded to and another response is
+sent, the second response should be ignored.
+
+Because it would be too slow to atomically save all requestIds that are
+processed, they are not idempotent. It is the responsibility of the requestor
+to make sure they don't duplicate requestIds. The implementation should ensure
+that no two in-flight requests are sent out with the same requestId. The
```
dappelt: We should clarify that not all in-flight requests in a CLP communication channel need to have a unique request id, but only the requests send by the same peer need to have a unique request id. In other words, it is CLP-compliant to have both peers send each other a request with the same id (for example, this could happen if both sides coincidentally generate the same random request id).

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
dappelt: I don't think this is the right place to go into specifics about how ILP will be used in CLP. ILP is a higher level protocol and CLP should not have to care what kind of protocol it is transporting as payload.

All we should write is that the side-protocol `ilp` is reserved. Specifics about how ILP is transported over CLP should go into the documentation of the side-protocol `ilp`.

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
dappelt: As discussed [here](https://github.com/interledger/rfcs/issues/290), only CLP-level errors should result in an `Error` packet.

If my earlier arguments did not convince you, think about this analogy with TCP and HTTP: Sending ILP-level errors back as CLP `Error`s would be like your TCP socket is giving you connection errors on an HTTP 404 Page not found.

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
dappelt: Random.

With a sequential transfer id we might run into problems if the transfer id has to be unique across several CLP communication channels (think fb-ledger). A 3rd party could predict the "next" sequential transfer id and prevent you from creating a transfer.

Also, random transfer ids make it easier to map the transfer id from the LPI to the CLP transfer id.

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
sappenin: +1 for `Sub Protocol`

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,273 @@
...
+The amount could be too high, or could exceeds the balance.
+
+#### Idempotency
+
+If a transfer with the given ID already exists in any state, a new transfer
+should not be created. If the packet (including side protocol data) matches an
+existing transfer exactly and that transfer is in the `prepared` state, a
+`Response` should be returned. If the packet (including side protocol data) shares
+the ID of an existing transfer but other details do not match, an `Error`
+should be returned. If the packet (including side protocol data) exactly
```
sappenin: > If the packet (including side protocol data) shares the ID of an existing transfer but other details do not match, an `Error` should be returned.

What should this `Error` look like? Is there a RejectionReason?

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
michielbdejong: @earizon see https://github.com/interledger/interledger/wiki/Interledger-over-CLP#setup about picking who is server and who is client. If one of the two peers is not IPv4-addressable, then that will automatically become the client. If connections are short-lived, then you need to ask if they are "paused" or "ended", i.e. can the trustline later be resumed? Will the balance survive the downtime? If so, then you need something like the `TOKEN` described in that same section. We should probably include one or more setup protocols in this RFC (e.g. the testnet one plus one other people want to use, for instance a WebFinger-based one if someone wants to define it)

michielbdejong: @sappenin (paraphrasing from memory)
1. we already have ILP packets as a less strict compatibility layer; if you use some custom encoding of ILQP + ILPPayment + ILPError into gprc, then that's already good, and you are free to choose how you encode your expiresAt DateTime in gprc (no need to enforce OER on you for that).
2. if you use grpc-native data formats, you might have different max precisions and max string lengths than in OER, so that's tricky.
3. if we brand the grpc version as 'custom' rather than 'official' then not every implementation needs to implement it. Any situation with > 1 'official' transports therefore leads to bloaty software.
4. if we brand >1 transports as 'suggested' then you could avoid that, but then you need a double coincidence for e.g. an app and an online ILP service provider to be compatible.

Therefore, IMHO the best path, as I see it, seems to be something like:
* the official ILP node API is the secure WebSockets + OER + CLP + [ 'setup', 'vouch', 'ilp', 'info', 'balance' ], where ilp is a bag containing ILQP + ILPPayment + ILPError. With that, you can do the "core tasks": connect, and then send and receive from/to any ledger, discover your own ILP address, obtain various types of quotes, and check your trustline balance, all with a deterministic outcome.
* there is an official list of optional extensions, currently [ ccp, paychan ]. They are not needed for the "core tasks", but cpp allows you to announce routes (choose your own custom ledger name, as well as attract connector traffic), and paychan is a performance improvement that gives you a middle ground between escrow and trustline in terms of trustlessness vs speed.
* we would version the official ILP node API, the first version (used only by the testnet) is on https://github.com/interledger/interledger/wiki/Interledger-over-CLP, the second version (to be shared between testnet and ilp-kit reference stack) is scheduled for 18 september, the third version can hopefully be scheduled for a specific date, that hopefully lies in 2018. :)
* optional extensions can be versioned independently and more ad-hoc, as long as their `protocolName` doesn't become ambiguous

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
sharafian: whether one party is authoritative is already something that has to be decided out of band. If only one party is authoritative, they don't need to wait for RPC responses from their non-authoritative peer, for instance. If the peers know they're both authoritative, it's fine for them to eschew the useless Reject messages for timeout, because they both know that it's timing out

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
sharafian: the fifth byte in those test cases is this length prefix

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
sharafian: my concern was just that returning a success has the implication that your protocol data made it to the other side. If you re-fulfill with different protocol data, then a new fulfill event is not being emitted, so your new protocol data does not reach the other side. If you re-fulfill with the same protocol data, they already got your data the first time so the success accurately conveys that the data made it

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
sharafian: the exact storage format could be anything, but the data stored must match the data in the CLP packet in order for it to work properly. If something like protocolData isn't stored, for instance, idempotency doesn't work correctly. If amounts aren't stored in the same format, the balance cannot be derived correctly.

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
michielbdejong: That does simplify things, because that removes the only ledger-level `rejectionReason`. It means you can remove https://github.com/interledger/rfcs/blob/master/asn1/CommonLedgerProtocol.asn#L89 in favor of putting the rejectionReason (ILP-error) inside the 'ilp' entry in protocolData

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
michielbdejong: Oh wow, I think you mean the sixth byte, i.e. `buffer[5]`, and that indeed looks like one :) Also, https://github.com/interledger/rfcs/blob/master/asn1/CommonLedgerProtocol.asn#L118 states that packet.data should be length-prefixed. But from how I understand ASN.1/OER, that's not correct, is it? There is no reason to add a length-prefix to packet.data, because it's a sequence and you know which fields to expect.

michielbdejong: So I think that's a bug, then.

michielbdejong: It's not a bug, ASN.1 Open Types get a length prefix in OER (see section 30 of https://www.itu.int/rec/T-REC-X.696-201408-S/en)

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,348 @@
...
+
+`Message` is used for sending information to the peer. It contains no
+packet-specific data, only protocol data. [ILQP] packets are attached under the
+protocol name `ilp` with content-type `application/octet-stream`.
+
+- `Response` is returned if the peer acknowledges the `Message`. If the peer
+  has data to send in reply (e.g. a quote response), it is carried in the
+protocol data.
+
+- `Error` is returned if the peer is not able to process the message, or there
```
michielbdejong: s/message/packet

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,348 @@
...
+  was an unexpected error and further messages should not be sent. This does
```
michielbdejong: s/message/packet

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
```diff
@@ -0,0 +1,348 @@
...
+not include ILP errors such as "No quote found," only the cases in which the
+message cannot be sent/processed at all.
```
michielbdejong: s/message/packet

00XX-common-ledger-protocol/00XX-common-ledger-protocol.md
dappelt: The term message in this context refers to a CLP packet of type `Message`, so I think it is ok to keep it. I changed message to `Message` to be consistent with the other packet type descriptions.

michielbdejong:
Continued with Monday's last-minute changes in #300

PR #300, opened 2017-09-19
Apply changes from last night's meeting

michielbdejong:
@justmoon's notes, copied from https://ripplelabs.slack.com/files/U024N53C7/F74RUKDLK/notes.md: 

- BTP URIs: btp+wss://:token@hostname.example:port/path

- There is only one BTP version, but in the future we will likely add a subprotocol to the auth message that allows the client to advertise features and a subprotocol for the response that allows the server to acknowledge these features
- Subprotocols either use the feature 

- The first protocol in the packet is called the primary
- If the packet contains more than one protocol with the same name, reject the whole packet with a BTP error
- If there are zero elements in protocol data, this is syntactically valid, although in the case of Message may return a BTP error

- Nginx proxy statement for each peer
- Plugins listen with unencrypted WebSocket by default
- Make sure we use HSTS in ILP Kit

- Auth primary protocol 'auth' - no data
- Secondary protocol 'auth_token' - UTF-8: token
- Secondary protocol 'auth_username' - UTF-8: username
- If successful, server responds with empty response
- If unsuccessful, server responds with BTP error and closes the connection
- If client connects and doesn't send an auth packet within a timeout, the server closes the connection

- If we need multi-user support, we'll add a subprotocol that can be attached to each message to indicate which user it is for

- Clients should not have configure prefix and info

- btp-packet should parse messages as follows:
  - packetType: integer type
  - ... other packet type fields
  - primaryProtocol: string (primary protocol name)
  - protocolData: object (parsed protocol data, each value is json -> object, utf-8 -> string, binary -> Buffer)

- If you want the plugin to be a server, pass in a `listener` object
- If you want the plugin to be a client, pass in the `server` string which is a BTP uri
- If you want the plugin to be both, pass in both options
- Not urgent, but eventually, in case of redundant sockets, one should be dropped

- If you don't have an authenticated connection, the plugin will throw an error on any operation
- To ensure a connection, first `await connect()` before making any requests

- When designing configuration options, put what it is, not what format it's in (e.g. 'server' instead of 'uri')

- custom is a side protocol of type json, containing the entire custom object
- can be a side protocol to ILP or its own primary protocol (on messages)

### DIFF
--- a/0023-bilateral-transfer-protocol/0023-bilateral-transfer-protocol.md
+++ b/0023-bilateral-transfer-protocol/0023-bilateral-transfer-protocol.md
@@ -1,8 +1,7 @@
 ---
 title: Bilateral Transfer Protocol (BTP)
-draft: 1
+draft: 2
 ---
-
 # Bilateral Transfer Protocol (BTP)
 
 ## Preface
@@ -65,11 +64,12 @@ spec](https://github.com/interledger/rfcs/blob/master/asn1/BilateralTransferProt
   ASN.1 spec.
 
 - A **Sub-Protocol** is a protocol which isn't defined by BTP and is carried
-  in the protocol data (see below).
+  in the protocol data (see below). The first one is the primary sub-protocol,
+  subsequent entries are secondary sub-protocols.
 
 - A **BTP Connection** is a websocket connection over which BTP packets are
-  sent. Websockets are used because they provide message framing and allow BTP
-to use HTTP requests for authentication.
+  sent. Websockets (as opposed to raw TLS sockets) are used because they provide
+  message framing and can be used from the browser.
 
 - **BTP Packets** are the protocol data units described in this document. They are
   formally defined in the [BTP ASN.1 spec](https://github.com/interledger/rfcs/blob/master/asn1/BilateralTransferProtocol.asn).
@@ -179,6 +179,56 @@ ProtocolData ::= SEQUENCE OF SEQUENCE {
   data OCTET STRING
 }
 ```
+
+## Authentication
+
+Before anything else, when a client connects to a server, it sends a special
+`Message` request. Its primary `protocolData` entry MUST have name `'auth'`,
+content type `MIME_APPLICATION_OCTET_STREAM`,
+and empty data, and among the secondary entries, there MUST be a UTF-8
+`'auth_token'` entry, and a UTF-8 `'auth_username'` entry. The further secondary
+protocol data entries of this `Message` request MAY also be used to send
+additional information to the server. In situations where no authentication
+is needed, the `'auth_token'` and `'auth_username'` data can be set to the
+empty string, but they cannot be omitted.
+
+If the client sends any BTP call that is not a Message, or sends a Message call
+whose primary sub-protocol is not `auth`, the server should respond with an `Error`
+and close the connection.
+
+The server responds with a `Response` or `Error` as appropriate. Again, the
+`protocolData` field there MAY be used to send additional information to
+the client. To be clear, the server responds with an `Error` if:
+
+* any other packet is sent before the auth data
+* the provided authentication data is invalid or incorrect
+* any of the other protocol rules are violated (e.g. having two subprotos with the same name)
+* it takes too long before the authentication data is sent.
+
+If the server sent an `Error`, it subsequently closes the connection.
+If the server sent a `Response`, the BTP connection is open, until either
+one of the parties closes it. At the BTP level, the client and server play
+identical roles.
+
+If the client does not send an Auth packet within a reasonable time, the
+server optionally sends a Message informing the client that the authentication timed out,
+and then closes the connection. If the client did send an Auth packet, but
+got neither a `Response` nor an `Error` back from the server, the client
+closes the connection.
+
+If the connection is ever dropped and reconnected then it must be re-authenticated.
+
+## Sub-protocols
+
+In order to understand the different BTP calls, it is necessary to distinguish between the first ("primary") and subsequent ("secondary") sub-protocol entries. The primary sub-protocol
+entry defines what type of action or information is requested of the recipient of the message. The secondary sub-protocols should not request additional actions or information. If multiple actions or
+pieces of information are required, multiple separate Messages should be sent. The secondary sub-protocols should only modify the request made in the primary sub-protocol, or provide additional contextual data which can be consumed in a readonly way (without affecting the result).
+
+For example, the primary sub-protocol entry of a Message might represent a quote request, while one additional secondary sub-protocol entry may be present, indicating this request was forwarded by a proxy.
+
+Likewise, only the primary sub-protocol data in a Response indicates whether result of the request from the Message being responded to actually succeeded or not.
+
+In Error, Prepare, Fulfill, and Reject calls, the distinction between primary and secondary sub-protocol entries is less strict.
 
 ## Flow
 


michielbdejong:
Ha, RFC issue #300 :)

0023-bilateral-transfer-protocol/0023-bilateral-transfer-protocol.md
```diff
@@ -60,11 +64,12 @@ spec](https://github.com/interledger/rfcs/blob/master/asn1/BilateralTransferProt
   ASN.1 spec.
 
 - A **Sub-Protocol** is a protocol which isn't defined by BTP and is carried
-  in the protocol data (see below).
+  in the protocol data (see below). The first one is the primary sub-protocol,
+  subsequent entries are secondary sub-protocols.
```
dappelt: It should say something about the relationship between primary and secondary sub-proto. A secondary sub-protocol adds additional information to a primary-subprotocol.

Also, we should say that unrelated sub-protocols SHOULD NOT go into the same packet, but be sent in separate packets.

dappelt:
We only have to photoshop MARCH 9 to SEPTEMBER 18, the date BTP froze over 😝

0023-bilateral-transfer-protocol/0023-bilateral-transfer-protocol.md
dappelt: Also:
> If the packet contains more than one protocol with the same name, reject the whole packet with a BTP error

0023-bilateral-transfer-protocol/0023-bilateral-transfer-protocol.md
```diff
@@ -175,6 +180,31 @@ ProtocolData ::= SEQUENCE OF SEQUENCE {
...
+`protocolData` field there MAY be used to send additional information to
+the client.
+
+If the server sent an `Error`, it subsequently closes the connection.
+If the server sent a `Response`, the BTP connection is open, until either
+one of the parties closes it. At the BTP level, the client and server play
+identical roles.
+
+If the client does not send an Auth packet within a reasonable time, the
+server closes the connection. If the client did send an Auth packet, but
```
dappelt: If authentication times out and before closing the connection, should there be another message saying something like "Authentication timeout"? This is what I wrote in my version

> Before closing the connection, the server MAY send a Message informing the client that the authentication timed out. Such a message might be useful for debug purposes, but is not strictly necessary.

0023-bilateral-transfer-protocol/0023-bilateral-transfer-protocol.md
```diff
@@ -175,6 +180,31 @@ ProtocolData ::= SEQUENCE OF SEQUENCE {
 }
 ```
 
+## Authentication
+
+Before anything else, when a client connects to a server, it sends a special
```
dappelt: It should say that if the client sends any other packet than the auth data first, the server MUST reply with an `Error` and close the connection.

0023-bilateral-transfer-protocol/0023-bilateral-transfer-protocol.md
```diff
@@ -175,6 +180,31 @@ ProtocolData ::= SEQUENCE OF SEQUENCE {
...
+`Message` request. Its primary `protocolData` entry MUST have name `'auth'`
+and empty data, and among the secondary entries, there MUST be a UTF-8
+`'auth_token'` entry, and a UTF-8 `'auth_username'` entry. The further secondary
+protocol data entries of this `Message` request MAY also be used to send
+additional information to the server. In situations where no authentication
+is needed, the `'auth_token'` and `'auth_username'` data can be set to the
+empty string, but they cannot be omitted.
+
+The server responds with a `Response` or `Error` as appropriate. Again, the
```
dappelt: We should be more specific and list the situations when the server replies with a `Response` or an `Error`. A server replies with `Error` when:
- Any other packet is sent before the auth data.
- The provided authentication is invalid.
- Any of the other protocol rules are violated (e.g. having two subprotos with the same name)

The server sents a response if:
- the provided auth is correct.
- anything else?

0023-bilateral-transfer-protocol/0023-bilateral-transfer-protocol.md
sharafian: I think the exact semantics of sub-protocols belongs in the format section rather than the terminology section.

0023-bilateral-transfer-protocol/0023-bilateral-transfer-protocol.md
```diff
@@ -175,6 +180,31 @@ ProtocolData ::= SEQUENCE OF SEQUENCE {
 }
 ```
 
+## Authentication
+
+Before anything else, when a client connects to a server, it sends a special
+`Message` request. Its primary `protocolData` entry MUST have name `'auth'`
```
sharafian: and a content-type of application/octet-stream

0023-bilateral-transfer-protocol/0023-bilateral-transfer-protocol.md
```diff
@@ -175,6 +180,31 @@ ProtocolData ::= SEQUENCE OF SEQUENCE {
...
+and empty data, and among the secondary entries, there MUST be a UTF-8
+`'auth_token'` entry, and a UTF-8 `'auth_username'` entry. The further secondary
```
sharafian: `auth_username` may not be present in some cases (e.g. peering) where the username is implicit in the endpoint being connected to.

0023-bilateral-transfer-protocol/0023-bilateral-transfer-protocol.md
```diff
@@ -175,6 +180,31 @@ ProtocolData ::= SEQUENCE OF SEQUENCE {
...
+additional information to the server. In situations where no authentication
+is needed, the `'auth_token'` and `'auth_username'` data can be set to the
+empty string, but they cannot be omitted.
+
+The server responds with a `Response` or `Error` as appropriate. Again, the
+`protocolData` field there MAY be used to send additional information to
+the client.
+
+If the server sent an `Error`, it subsequently closes the connection.
+If the server sent a `Response`, the BTP connection is open, until either
```
sharafian: it's probably obvious, but I think we should specify that if the connection is ever dropped/reconnected then it must be re-authenticated.

0023-bilateral-transfer-protocol/0023-bilateral-transfer-protocol.md
michielbdejong: Hm, what is the type of empty content, though? 🤔 
I'm ok with picking octet-stream, I guess it feels nicer to pick one.

0023-bilateral-transfer-protocol/0023-bilateral-transfer-protocol.md
michielbdejong: That's asymmetric then: `auth_token` should be UTF-8 and either the token or the empty string, and `auth_username` should either be UTF-8 and the username, or missing? I don't see anything in Stefan's notes about that, but if you remember that's how we agreed it on Monday, then that's what I'll write.

michielbdejong: see also further comments on https://github.com/interledgerjs/ilp-plugin-payment-channel-framework/commit/f3f921a6240971127dd4b168d439adf7a5b8650a#commitcomment-24424484

michielbdejong: We discussed this in the #ripple-de-lux slack channel, I think the best option is:
* A server should give out BTP URIs of one form, `btp+wss://auth_username:auth_token@domain.com/path`
* If the server, for whatever reason, doesn't care about the `auth_username`, it can set it to `''` in the BTP URI.
* If the server, for whatever reason, doesn't care about the `auth_token`, it can set it to `''` in the BTP URI.
* a client should always send both `auth_username` and `auth_token` exactly as they appeared in the BTP URI (even if that means `''`)

0023-bilateral-transfer-protocol/0023-bilateral-transfer-protocol.md
emschwartz: Can we please put some examples of what we mean by this? The concept of a sub-protocol is already a bit confusing, and now it seems even more so because of this primary vs secondary distinction.

0023-bilateral-transfer-protocol/0023-bilateral-transfer-protocol.md
sharafian: we can always change the content type once we put things in there. I'm just using octet-stream in the current implementation

0023-bilateral-transfer-protocol/0023-bilateral-transfer-protocol.md
sharafian: sure, I implemented the first version without auth_username because it isn't an admin plugin, but I can just check to make sure its there and empty.

emschwartz:
What's the status on this PR?

michielbdejong:
I think we're in agreement about all discussion topics now, I'll rebase, update, and re-request review.

michielbdejong:
@sharafian bump

PR #304, opened 2017-09-26
docs(0004): noteToSelf, custom are optional

emschwartz:


michielbdejong:
I checked that both ilp-plugin-virtual and ilp-plugin-bells support these changes. Will continue to check other plugins. A bit late maybe now it's already merged, but I like to keep the rfcs repo as a representation of what is currently implemented.

michielbdejong:
For ripple/ilp-plugin-xrp-escrow, this is supported since https://github.com/ripple/ilp-plugin-xrp-escrow/commit/1b2f0636d9337d4c134b8e6d1f60ebaa51a8e1b1, which went into master 8 days ago.

PR #306, opened 2017-09-27
Updated HTTP-ILP (Continuation of #149)

michielbdejong:
* rebased on master
* filename corrected
* doc version header added
* Headers renamed from e.g. `X-Pay` to just `Pay`


0014-http-ilp/0014-http-ilp.md
```diff
@@ -0,0 +1,249 @@
...
+HTTP/1.1 200 OK
+Content-Type: application/json
+```
+``` js
+{
+  // TODO ...
+}
+```
+
+The HTTP-ILP server module now knows the ILP address and caches it in memory for a certain period (depending on the cache header of the above request.) The reason we had Ankita enter the SPSP address and not the ILP address is because the ILP address may change over time as ILP providers change their routing structure. Note that ILP may support multihoming (this is still being debated as of this writing), so there may also be more than one ILP address.
```
emschwartz: > Note that ILP may support multihoming (this is still being debated as of this writing), so there may also be more than one ILP address.

This can be removed

0014-http-ilp/0014-http-ilp.md
```diff
@@ -0,0 +1,249 @@
...
+```
+```
+[...]
+```
+
+Note that the client hasn't paid at this point and is only making the request to solicit a response from the server.
+
+### 2.2. Server rejects request, because token is new and unfunded
+
+The server returns an HTTP error of `402 Payment Required` and includes response headers showing the amount (how much the request would have cost), an ILP address and a condition seed.
```
emschwartz: It says it will show the amount the request would have cost but the example doesn't show that

0014-http-ilp/0014-http-ilp.md
```diff
@@ -0,0 +1,249 @@
...
+Pay-Balance: 0
+```
+
+The client may use the condition seed to create a condition to pay this host. The condition seed is generated by the server as follows:
+
+* Condition Seed: `HMAC(SHA256, receiver_secret, token)`
+
+The condition_seed is now a shared secret between the client and server, but will be unknown to any third-party connectors between them.
+
+That means that if the payment fails, the client can blame the server (assuming it considers itself honest and not compromised.) The client can't prove any wrongdoing to a third party, but since amounts are likely to be very low, this probably wouldn't be worth doing anyway. Also, proving wrongdoing would also involve proving that the server did not provide the agreed-upon service, which is probably near impossible in most cases anyway. However, since it can detect malicious behavior, the client can blacklist malicious hosts and not deal with them in the future.
```
emschwartz: I don't think we need this explanation. It's more confusing than helpful

0014-http-ilp/0014-http-ilp.md
```diff
@@ -0,0 +1,249 @@
...
+In order to refill its balance, the client now creates an ILP payment with the following properties:
+
+* Destination: `us.nexus.ankita.~recv.filepay`
+* Amount: `100`
+* Condition: `SHA256(fulfillment)`
+* Memo: `pay_token`
+
+The `fulfillment` is generated as:
+
+* Fulfillment: `HMAC(SHA256, condition_seed, destination_account || destination_amount)`
```
emschwartz: Shouldn't it also include the `pay_token`?

Is there a reason why this protocol doesn't use standard PSK?

0014-http-ilp/0014-http-ilp.md
```diff
@@ -0,0 +1,249 @@
...
+
+When the prepared payment reaches the receiving wallet, the webhook that is registered with the ankita+filepay@nexus.justmoon.com receiver is triggered. Note that the wallet does not fulfill the payment yet.
+
+### 2.4. Wallet POSTs a payment notification
```
emschwartz: Why not just have the server connect to the ILSP using something like BTP, rather than creating yet another encoding for ILP payment details? Websockets might not be ideal but I think it would be better if the server itself were a standard ILP receiver using some ledger protocol to talk to the ILSP.

sharafian:
This should show what https://github.com/justmoon/koa-ilp does now. A lot of the contents seem dated

michielbdejong:
In the updated version, the `Pay` header which [superagent-ilp expects](https://github.com/justmoon/superagent-ilp/blob/master/index.js#L38-L39) is:
`[ destinationAmount, destinationAccount, sharedSecret ].join(' ')`

And that's also the header that [koa-ilp produces](https://github.com/justmoon/koa-ilp/blob/master/index.js#L85).

So that's probably the one we want to document in this RFC and use in the [streaming payments tutorial](https://github.com/interledger/tutorials/pull/2)

michielbdejong:
Updated the documentation to remove the webhooks stuff (koa-ilp now just uses a plugin).
Also replaced the condition seed with standard PSK, and added a stub description of streaming http-ilp in section 3.

0014-http-ilp/0014-http-ilp.md
```diff
@@ -0,0 +1,149 @@
...
+
+## 1. Setup
+
+###### Scenario
+
+**Ankita** is a *server admin* who owns a file hosting service at `myservice.example` and would like to provide an API where her users can upload files without having to register first. In order to enable this in a standards-compliant way, she decides to use HTTP-ILP.
+
+### 1.1. Server admin installs an ILP payment module
+
+After deciding to use HTTP-ILP, Ankita searches the web on instructions on how to set this up. She finds an open-source HTTP-ILP server module which is compatible with the REST framework ([Koa](https://koajs.com)) that she is already using.
```
sharafian: This should include a link to the koa-ilp module too

0014-http-ilp/0014-http-ilp.md
```diff
@@ -0,0 +1,149 @@
...
+### 1.2. Server admin sets up a new receiver
+
+
+Ankita creates a dedicated subaccount under her account at her Interledger service provider, and takes a note of that subaccount's credentials.
+
+### 1.3. Server admin enters credentials into the HTTP-ILP server module config
+
+Next, Ankita logs back into her server and edits a config file of the HTTP-ILP server module to enter the plugin type and credentials she obtained:
+
+``` ini
```
sharafian: Right now this is configured through the environment, not an ini file.

0014-http-ilp/0014-http-ilp.md
```diff
@@ -0,0 +1,149 @@
...
+[plugin_creds]
+plugin_name=ilp-plugin-btp-client
+plugin_config={"server":"btp+wss://ankita+filepay:fxPERNaS4FGlC8H7eg6UfYVlglmFynFc8nh5la9PBGM@nexus.justmoon.com"}
+```
+
+Next, she restarts her server to load the new configuration.
+
+### 1.4. Paid HTTP server fetches receiver information
+
+When the HTTP-ILP server module loads up, it uses [plugin#getAccount](../0004-ledger-plugin-interface/0004-ledger-plugin-interface.md) to look up the Interledger address from the plugin.
```
sharafian: The plugin also gets a subscription to its ledger in order to listen for incoming payments

0014-http-ilp/0014-http-ilp.md
```diff
@@ -0,0 +1,149 @@
...
+
+###### Scenario
+
+**Marat** is a *graphic designer* who would like to upload a mockup image to share with a client. His friend recommended a tool which doesn't require any signup or configuration and uses his existing ILP account to pay for the storage and bandwidth fees.
+
+![Sequence Diagram](sequence.png)
+
+### 2.1. Client generates a server-specific token from the hostname and its local secret:
+
+The uploader tool contains an HTTP-ILP client module. This client module may for instance generate and locally save a `client_secret` using 256 bits of cryptographically secure randomness.
```
sharafian: Might be worth mentioning the existing client tools, `superagent-ilp` and `ilp-curl`

0014-http-ilp/0014-http-ilp.md
```diff
@@ -0,0 +1,149 @@
...
+```
+```
+[...]
+```
+
+Note that the client hasn't paid at this point and is only making the request to solicit a response from the server, that's why the OPTIONS verb is used..
+
+### 2.2. Server responds with payment details
+
+The server returns an HTTP error of `402 Payment Required` and includes response headers showing the amount, an ILP address and a shared secret.
```
sharafian: The server doesn't respond with `402` to an OPTIONS request; it uses a `204`. `402` is returned if the actual API call is attempted without payment.

0014-http-ilp/0014-http-ilp.md
```diff
@@ -0,0 +1,149 @@
...
+The amount expresses how much the request would have cost, as a decimal string and counted in the base unit of the ledger to which the ILP address belongs.
```
sharafian: Can we link to our description of what a base unit is?

0014-http-ilp/0014-http-ilp.md
```diff
@@ -0,0 +1,149 @@
...
+
+``` http
+HTTP/1.1 402 Payment Required
+Pay: 10 us.nexus.ankita.~recv.filepay SkTcFTZCBKgP6A6QOUVcwWCCgYIP4rJPHlIzreavHdU
+Pay-Balance: 0
+```
+
+The client can now use the shared secret to create a condition to pay this host. The shared secret may for instance be generated by the server as follows:
```
sharafian: Should explain that the destinationAccount and sharedSecret are used as the parameters to PSK

0014-http-ilp/0014-http-ilp.md
```diff
@@ -0,0 +1,149 @@
...
+* Destination: `us.nexus.ankita.~recv.filepay`
+* Amount: `100`
+* Condition: `SHA256(fulfillment)`
+* Memo: `pay_token`
+
+The `fulfillment` is generated from the shared secret using [PSK](../0016-pre-shared-key/0016-pre-shared-key.md).
+
+When the prepared payment reaches the server, it is fulfilled and the token's balance is increased.
+
+There is a chance that the HTTP-ILP server module will process the payment, but the fulfillment doesn't make it all te way back to the sender.
```
sharafian: te -> the

0014-http-ilp/sequence.mmd
```diff
@@ -0,0 +1,29 @@
+sequenceDiagram
+participant Payer
+participant Interledger
+participant PayeeWallet
+participant Payee
+activate Payer
+Payer->>+Payee: Attempt Request
+Payee-->>-Payer: 402 Payment Required
```
sharafian: If this is describing the `OPTIONS` flow, it should begin with an OPTIONS request that ends with a 204, rather than a regular request that ends with a 402.

0014-http-ilp/0014-http-ilp.md
```diff
@@ -0,0 +1,149 @@
...
+```
+```
+[...]
+```
+
+The request succeeds:
+
+``` http
+HTTP/1.1 200 OK
+Pay: 10 us.nexus.ankita.~recv.filepay SkTcFTZCBKgP6A6QOUVcwWCCgYIP4rJPHlIzreavHdU
```
sharafian: Unhash might make a better example application, because we know exactly what the API calls look like for it. For instance, an Unhash `OPTIONS` request includes the `Unhash-Content-Length` header to calculate the price.

0014-http-ilp/0014-http-ilp.md
```diff
@@ -0,0 +1,149 @@
...
+Pay-Balance: 90
+```
+
+Notice how the 100 units credit from the payment was added to the balance and the 10 unit cost for the current request was subtracted.
+
+### 3. Streaming flow
```
sharafian: I think the streaming flow is different enough that it should go in another proposal. The HTTP-ILP described here is pretty strictly request/response rather than streaming

0014-http-ilp/0014-http-ilp.md
michielbdejong: It feels weird that an rfc describes one specific implementation in such exact detail, but at the same time if the only current implementation uses env vars, then ok let's quote that example in the rfc too.

0014-http-ilp/0014-http-ilp.md
michielbdejong: ok, I'll move that to a separate PR.

0014-http-ilp/0014-http-ilp.md
michielbdejong: ok, I'll add `Unhash-Content-Length` to the OPTIONS call and explain that other services would have other product descriptors which the client can add to help determine the price for a request.

0014-http-ilp/0014-http-ilp.md
michielbdejong: I changed it to env vars, but not sure it now matches koa-ilp, because of https://github.com/justmoon/koa-ilp/issues/4

Issue #312, opened 2017-10-03
RFC-3: Remove amount field


justmoon:
Through [@emschwartz's work](#309) on the ILQPv2 protocol, it became apparent that there are use cases where we would like the connector to forward payments according to their offered rate. That is, we want connectors to ignore the `amount` field in the ILP packet.

The main reason there is an amount field in the ILP packet to begin with is for one specific use case: when the recipient wants to receive an exact amount.

With our modern view of ILP including micropayments and streaming payments, this seems like less of a core use case. But even if we still think this is an important use case, perhaps there is a better way to provide this functionality?

I believe there is. That is, by giving the recipient a way to reduce the amount received. The recipient is the party that cares about receiving an exact amount.

My suggestion to address this use case is that ledgers which serve merchants who have this requirement would allow merchants to pass back an amount together with the fulfillment. This amount must be lower or equal to the transfer amount and will be used to lower the actual amount transferred. The connector may also decide to fulfill their incoming payment with a lowered amount if they have a similar concern.

(One example where the connector would "pass on the savings" is a 1:1 connector that wants to keep the incoming and outgoing amounts exactly the same.)

@emschwartz suggested a slightly different method: The receiver could reject the payment with an "amount too high" error. This error could be addressed by the connector by re-sending the transfer with a lower amount, or it could pass the error on to the next connector or even all the way to the sender.

Either way is fine and either way would solve the "exact amount" use case in a much simpler way than the current solution.

#### Benefits of this change

* Simplify connector behavior

  The whole [forward vs deliver](https://github.com/interledger/rfcs/issues/77) goes away entirely. This is a currently a complex and awkward part of ILP.

* Allow more extensibility

  Forward vs deliver makes it harder to extend payment chains - once a connector in the chain thinks it is delivering (instead of forwarding) I can no longer have further currency exchanges later in the chain.

* Further simplify ILP packet format

#### Downsides of this change

* Changes ILP packet format

  The one thing we thought had finally stabilized. We could keep the amount field in the ILP packet and just permanently set it to zero or `0xffffffffffffffff`.

* Requires solving the "exact amount" use case in another way

justmoon:
If changing the ILP packet format is a major no-go, we could also keep it and just change who looks at it. I.e. connectors simply ignore it and it becomes a pure end-to-end field, like `data`. That would mean that the transport layer protocol defines how to interpret the `amount` field.

justmoon:
[...]

PR #315, opened 2017-10-05
docs: describe how competition ensures network health

michielbdejong:
By adding these two paragraphs, I tried to better describe how the current community network makes sure that connectors can't steal money.

I also wrote this in part as a strawman, to find out which parts people disagree with - both in terms of whether it's even desirable that connectors can't steal money (or of what behavior we even consider as stealing), and in terms of what Interledger's mechanism is to avoid stealing.

### DIFF
--- a/0001-interledger-architecture/0001-interledger-architecture.md
+++ b/0001-interledger-architecture/0001-interledger-architecture.md
@@ -1,6 +1,6 @@
 ---
 title: Interledger Architecture
-draft: 1
+draft: 2
 ---
 # Interledger Architecture
 
@@ -37,6 +37,12 @@ If the ledgers represent different assets, the connectors set the exchange rate 
 Connectors *peer* with one another to exchange information used to determine the best route for a payment.
 
 **Interledger ensures that senders' funds are safe throughout an multi-hop payment and cannot be stolen by faulty or malicious connectors (see [Interledger Security](#interledger-security)).**
+
+### Interledger Networks
+
+The Interledger protocol stack can be used to create a network that effectively offers the same functionality as a centralized ledger; namely, the ability to safely transfer value from one account to another. But since an Interledger network is made up of autonomously run ledgers and connectors, without any central authority, these ledgers and connectors can then compete, at each hop of a given payment, to offer a good service at a low price. This will lead, through market economics, to cheaper transaction costs than those that can be expected on a centrally controlled system.
+
+The system for ensuring that a payment either succeeds at a low price, or rolls back so that it can be re-routed within a short timeout, is described in the [pre-shared key](../0016-pre-shared-key/0016-pre-shared-key.md) specification.
 
 ### The Interledger
 
--- a/0016-pre-shared-key/0016-pre-shared-key.md
+++ b/0016-pre-shared-key/0016-pre-shared-key.md
@@ -1,6 +1,6 @@
 ---
 title: The Pre-Shared Key Transport Protocol (PSK)
-draft: 3
+draft: 4
 ---
 # Pre-Shared Key Transport Protocol (PSK)
 
@@ -13,6 +13,28 @@ As the name suggests, PSK relies on a pre-shared secret key. In application-laye
 An advantage of PSK is that the pre-shared secret key only needs to be shared once. As long as both parties possess this key and are listening for transfers, they can send payments between one another.
 
 A disadvantage of PSK is that it is repudiable. Although the sender does get cryptographic proof that the recipient received the payment, that proof cannot be used to convince 3rd parties that the sender did indeed send the funds, because the sender could have generated the fulfillment themselves. However, simple proof for the sender that the recipient got the funds is sufficient in many applications. The [Interledger Payment Request](https://github.com/interledger/rfcs/blob/master/0011-interledger-payment-request/0011-interledger-payment-request.md) transport protocol should be used instead in cases where non-repudiable proof is required.
+
+## Importance of PSK in ensuring network health
+
+When a sender or connector asks  the next connector to forward a payment to the receiver, they need to specify, among other things, the end-to-end data which may convince the receiver to fulfill the
+condition, and the source amount which will reward that next connector.
+
+PSK enables the safe delivery of the end-to-end data, which in turn allows the receiver to check whether the on-ledger destination amount is high enough.
+If the on-ledger destination amount is too low, that means either the sender, or one of the connectors along the path, used a transfer amount that was too low (i.e., is effectively trying to steal
+a little bit).
+
+If the destination amount is too low, even if just by a few percent, the receiver should reject the payment. If this happens, the sender should retry the payment, using a different route.
+This way, eventually, connectors that try to steal money will lose traffic, which is how the Interledger protocol stack ensures the network stays healthy.
+It is PSK's end-to-end data transportation that makes this continuous price competition possible, so that is one of the main advantages of PSK.
+
+The destination amount can either be public or secret. If the destination amount is public, each connector along the path can make sure they pay a fair but sufficient outgoing transfer amount.
+For the destination ledger, there will be a 1:1 rate between the transfer amount and the destination amount in the payment packet. For previous hops, this rate will be determined by their liquidity
+curve for the destination ledger. A rational connector will try to stay informed of the exact current liquidity curve for each destination ledger, so that they never overpay.
+Even if the destination amount is public and can be viewed by all connectors along the path, PSK ensures that it cannot be altered along the path.
+
+If the destination amount is secret, connectors have to guess how much they would need to pay to convince the rest of the hops along the path to get the payment fulfilled.
+Presumably the destination amount would be not much lower than the maximum achievable destination amount, so it will not be possible to hide the destination amount from the connectors entirely.
+Secret destination amounts are also not currently supported, but support for them may be added in the future.
 
 ## Flow
 


0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -38,6 +38,12 @@ Connectors *peer* with one another to exchange information used to determine the
 
 **Interledger ensures that senders' funds are safe throughout an multi-hop payment and cannot be stolen by faulty or malicious connectors (see [Interledger Security](#interledger-security)).**
 
+### Interledger Networks
+
+The Interledger protocol stack can be used to create a network that effectively offers the same functionality as a centralized ledger; namely, the ability to safely transfer value from one account to another. But since an Interledger network is made up of autonomously run ledgers and connectors, without any central authority, these ledgers and connectors can then compete, at each hop of a given payment, to offer a good service at a low price. This will lead, through market economics, to cheaper transaction costs than those that can be expected on a centrally controlled system.
+
+The system for ensuring that a payment either succeeds at a low price, or rolls back so that it can be re-routed within a short timeout, is described in the [pre-shared key](../0016-pre-shared-key/0016-pre-shared-key.md) specification.
```
emschwartz: That is the transport layer, not just PSK

0016-pre-shared-key/0016-pre-shared-key.md
```diff
@@ -14,6 +14,28 @@ An advantage of PSK is that the pre-shared secret key only needs to be shared on
 
 A disadvantage of PSK is that it is repudiable. Although the sender does get cryptographic proof that the recipient received the payment, that proof cannot be used to convince 3rd parties that the sender did indeed send the funds, because the sender could have generated the fulfillment themselves. However, simple proof for the sender that the recipient got the funds is sufficient in many applications. The [Interledger Payment Request](https://github.com/interledger/rfcs/blob/master/0011-interledger-payment-request/0011-interledger-payment-request.md) transport protocol should be used instead in cases where non-repudiable proof is required.
 
+## Importance of PSK in ensuring network health
```
emschwartz: This is the role of the Transport Layer, not just PSK. I don't think it belongs in this spec

0016-pre-shared-key/0016-pre-shared-key.md
```diff
@@ -14,6 +14,28 @@ An advantage of PSK is that the pre-shared secret key only needs to be shared on
...
+a little bit).
+
+If the destination amount is too low, even if just by a few percent, the receiver should reject the payment. If this happens, the sender should retry the payment, using a different route.
+This way, eventually, connectors that try to steal money will lose traffic, which is how the Interledger protocol stack ensures the network stays healthy.
+It is PSK's end-to-end data transportation that makes this continuous price competition possible, so that is one of the main advantages of PSK.
+
+The destination amount can either be public or secret. If the destination amount is public, each connector along the path can make sure they pay a fair but sufficient outgoing transfer amount.
+For the destination ledger, there will be a 1:1 rate between the transfer amount and the destination amount in the payment packet. For previous hops, this rate will be determined by their liquidity
+curve for the destination ledger. A rational connector will try to stay informed of the exact current liquidity curve for each destination ledger, so that they never overpay.
```
emschwartz: This has nothing to do with PSK

0016-pre-shared-key/0016-pre-shared-key.md
```diff
@@ -14,6 +14,28 @@ An advantage of PSK is that the pre-shared secret key only needs to be shared on
...
+Even if the destination amount is public and can be viewed by all connectors along the path, PSK ensures that it cannot be altered along the path.
+
+If the destination amount is secret, connectors have to guess how much they would need to pay to convince the rest of the hops along the path to get the payment fulfilled.
+Presumably the destination amount would be not much lower than the maximum achievable destination amount, so it will not be possible to hide the destination amount from the connectors entirely.
+Secret destination amounts are also not currently supported, but support for them may be added in the future.
```
emschwartz: I wouldn't add a paragraph like this until we decide what we're doing about the amount field in the packet. I also don't think this description belongs in this spec

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -38,6 +38,12 @@ Connectors *peer* with one another to exchange information used to determine the
 
 **Interledger ensures that senders' funds are safe throughout an multi-hop payment and cannot be stolen by faulty or malicious connectors (see [Interledger Security](#interledger-security)).**
 
+### Interledger Networks
```
emschwartz: I would call this something like "Interoperability and Competition"

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -38,6 +38,12 @@ Connectors *peer* with one another to exchange information used to determine the
...
+
+The Interledger protocol stack can be used to create a network that effectively offers the same functionality as a centralized ledger; namely, the ability to safely transfer value from one account to another. But since an Interledger network is made up of autonomously run ledgers and connectors, without any central authority, these ledgers and connectors can then compete, at each hop of a given payment, to offer a good service at a low price. This will lead, through market economics, to cheaper transaction costs than those that can be expected on a centrally controlled system.
```
emschwartz: Talking about how interoperability brings competition, which you would expect to drive down the rates, seems useful but I don't find the comparison to one centralized network very helpful. Right now there isn't one network. There are many networks. The key point is that all of those are disconnected and currently compete on reach.

0016-pre-shared-key/0016-pre-shared-key.md
```diff
@@ -14,6 +14,28 @@ An advantage of PSK is that the pre-shared secret key only needs to be shared on
...
+
+PSK enables the safe delivery of the end-to-end data, which in turn allows the receiver to check whether the on-ledger destination amount is high enough.
+If the on-ledger destination amount is too low, that means either the sender, or one of the connectors along the path, used a transfer amount that was too low (i.e., is effectively trying to steal
+a little bit).
+
+If the destination amount is too low, even if just by a few percent, the receiver should reject the payment. If this happens, the sender should retry the payment, using a different route.
+This way, eventually, connectors that try to steal money will lose traffic, which is how the Interledger protocol stack ensures the network stays healthy.
+It is PSK's end-to-end data transportation that makes this continuous price competition possible, so that is one of the main advantages of PSK.
+
+The destination amount can either be public or secret. If the destination amount is public, each connector along the path can make sure they pay a fair but sufficient outgoing transfer amount.
```
emschwartz: Right now the amount is always in the packet and public, and that isn't a decision on the part of PSK. The question in https://github.com/interledger/rfcs/issues/312 is whether the amount should be looked at by connectors or whether it should be in the packet at all. If we decided to take it out of the ILP packet, we would update PSK so that it always has the amount in the encrypted data.

0016-pre-shared-key/0016-pre-shared-key.md
```diff
@@ -14,6 +14,28 @@ An advantage of PSK is that the pre-shared secret key only needs to be shared on
...
+When a sender or connector asks  the next connector to forward a payment to the receiver, they need to specify, among other things, the end-to-end data which may convince the receiver to fulfill the
+condition, and the source amount which will reward that next connector.
+
+PSK enables the safe delivery of the end-to-end data, which in turn allows the receiver to check whether the on-ledger destination amount is high enough.
+If the on-ledger destination amount is too low, that means either the sender, or one of the connectors along the path, used a transfer amount that was too low (i.e., is effectively trying to steal
+a little bit).
+
+If the destination amount is too low, even if just by a few percent, the receiver should reject the payment. If this happens, the sender should retry the payment, using a different route.
+This way, eventually, connectors that try to steal money will lose traffic, which is how the Interledger protocol stack ensures the network stays healthy.
+It is PSK's end-to-end data transportation that makes this continuous price competition possible, so that is one of the main advantages of PSK.
```
emschwartz: Again, this is transport layer not just PSK that makes sure the right amount arrives. 

Also, ensuring that the correct amount arrives is different than creating competition. Having multiple options creates competition. Even if you didn't have a way to tell the receiver how much they were supposed to receive but they could tell you how much arrived, that would be enough for you to determine which connectors are charging you too much.

michielbdejong: > determine which connectors are charging you too much

But that means every connector can still steal an arbitrary amount from me whenever I decide to use them. I want the payment to fail instead.

0016-pre-shared-key/0016-pre-shared-key.md
michielbdejong: > update PSK so that it always has the amount in the encrypted data.

yes, that's what I was referring to with 'making the amount secret'

0001-interledger-architecture/0001-interledger-architecture.md
michielbdejong: OK, I see now how that's confusing. I'm comparing Interledger to an *imaginary* centralized system which doesn't really exist.

michielbdejong:
@emschwartz let's discuss the following points:
* do you agree that we should care about transaction fees, and that this affects our design choices?
* if so, which spec do we put this in? Obviously not the PSK one as you said, but then, where?
* I see a key point we don't agree on is whether it's a) enough to know, afterwards, how much was stolen (your standpoint, right?), or b) better to avoid stealing altogether. I think b), and that's why I think the receiver should only give out the fulfillment if *no less than 100%* of the destination amount was received.

0016-pre-shared-key/0016-pre-shared-key.md
emschwartz: Think about it this way. If you have a payment you need to make for some service, let's say it costs 10 units. Let's take the case where you can tell the receiver the exact amount. You tell them you're trying to send them 10 but only 8 arrive. So they reject. Now what do you do? You still want to make that payment. If you have another option, you'll try the other option. If you don't you'll keep using that same connector and just adjust the source amount upwards. It's the availability of other options that really keeps connectors in check.

emschwartz:
> do you agree that we should care about transaction fees, and that this affects our design choices?

Yes.

> if so, which spec do we put this in? Obviously not the PSK one as you said, but then, where?

Actually in thinking about the other streaming-focused option I'm coming to agree that this behavior is specific to PSK and IPR, as opposed to something you would _necessarily_ implement in other transport protocols. If other transport protocols would do something different (not what I thought before), then it would make sense to put it in the PSK (and maybe IPR) spec.

> I see a key point we don't agree on is whether it's a) enough to know, afterwards, how much was stolen (your standpoint, right?), or b) better to avoid stealing altogether. I think b), and that's why I think the receiver should only give out the fulfillment if no less than 100% of the destination amount was received.

PSK should ensure that receivers receive no less than the intended destination amount.

One difference in how I think about it though: if connectors deliver less it's not so much about them "stealing", it's more about them having a crappy rate. You only think of it as stealing if you have a quoting protocol where you're asking them a question beforehand and they give out a dishonest answer. If all you're ever asking them to do is forward payments, then you can have cheaper and more expensive connectors, you can have volatile and less volatile connectors, but them changing the rate isn't really stealing. There may be something to not wanting connectors to know at all if you're doing a quote or something else through them, because then it would be harder to implement trickier logic even if you wanted to.

emschwartz:
> do you agree that we should care about transaction fees, and that this affects our design choices?

Yes.

> if so, which spec do we put this in? Obviously not the PSK one as you said, but then, where?

Actually in thinking about the other streaming-focused option I'm coming to agree that this behavior is specific to PSK and IPR, as opposed to something you would _necessarily_ implement in other transport protocols. If other transport protocols would do something different (not what I thought before), then it would make sense to put it in the PSK (and maybe IPR) spec.

> I see a key point we don't agree on is whether it's a) enough to know, afterwards, how much was stolen (your standpoint, right?), or b) better to avoid stealing altogether. I think b), and that's why I think the receiver should only give out the fulfillment if no less than 100% of the destination amount was received.

PSK should ensure that receivers receive no less than the intended destination amount.

One difference in how I think about it though: if connectors deliver less it's not so much about them "stealing", it's more about them having a crappy rate. You only think of it as stealing if you have a quoting protocol where you're asking them a question beforehand and they give out a dishonest answer. If all you're ever asking them to do is forward payments, then you can have cheaper and more expensive connectors, you can have volatile and less volatile connectors, but them changing the rate isn't really stealing. There may be something to not wanting connectors to know at all if you're doing a quote or something else through them, because then it would be harder to implement trickier logic even if you wanted to.

michielbdejong:
I think these topics may be more fit for a tutorial than for rfcs; will continue this with https://github.com/interledger/tutorials/pull/5

michielbdejong:
> PSK should ensure that receivers receive no less than the intended destination amount.
[...]
> if connectors deliver less it's not so much about them "stealing"

If receivers are ensured to receive no less, then connectors may still *try* to deliver less, but they will never succeed in doing so because the receiver will put their foot down and refuse to give up the fulfillment.

Issue #314, opened 2017-10-05
Proposal: Include data with fulfillment


justmoon:
This has come up many times, but we previously lacked a strong, clear use case for it.

When doing streaming payments, the receiver needs to communicate information back to the sender - how much was received in the last payment, how much is left to pay, etc.

We could say that this needs to be communicated via the application layer. But this seems to add a lot of overhead. Imagine an SPSP payment of $10000 dollars with a max ILP payment size of $100. There isn't an obvious way for the SPSP server to communicate information back to the SPSP client. We could add a websocket connection, but that would complicate our "simple" HTTP based protocol quite a bit. The fulfillment is already being communicated back to the sender. And the prepare and reject state transitions already carry data, so the obvious solution is to add data to the fulfillment as well.

sharafian:
What's a connector's incentive to pass on the data with the fulfillment? And if a connector doesn't pass on the data with the fulfillment, is there any way for the sender to find that out?

It seems like connectors would have an interest in not passing on data with the fulfillment when an end-to-end quote is being performed, because they would have to retry the end to end quote.

emschwartz:
Repeat dealings.

If my streaming transport protocol relies on getting a message back and I don't, that's an error. I wouldn't send more payments through you and if this happens repeatedly I'd switch to using another connector that does work. That's another useful thing with streaming payments, is that it's never about a single payment. Would a connector really try to cheat me for a single $0.10 payment, or fail to pass on a couple kb of data, if that means they would miss out on the next $1000 that I'm about to send?

Also, this behavior would be part of the expectation of what all connectors implement. If you don't implement this functionality, you haven't implemented ILP properly.

adrianhopebailie:
> Repeat dealings.

If we consider this sufficient incentive for connectors to always behave correctly then we should be consistent in taking that position. 

That assumption makes the debate in #316 kind of meaningless. By extension of the assumption that connectors will always do what's most likely to earn them repeat business, we can assume they will always charge the least they viably can, assuming they have limited knowledge of what other connectors charge for the same route.

Starting to feel like the use case for crypto-conditions is rearing its head again

michielbdejong:
This would require a new (more generic) payment packet type, "payment with response", where each transfer is conditional not on the fulfillment of a sha256 condition, but on a signed message from the receiver. If we conclude sha256-based interledger is doomed without this, then we could switch to that, but otherwise, I think this would not be wise.

> how much was received in the last payment,

[...]

PR #317, opened 2017-10-20
LPI: Add fulfillment data, make rejection message binary

emschwartz:
Related to https://github.com/interledger/rfcs/issues/314. Also, when you reject payments you should be able to submit an ILP error (which is binary). I think the `RejectionMessage` object is a hold-over from before the ILP errors were binary.

Open questions:
- [ ] Should we keep the `*_cancel` events using the `RejectionMessage` class? (Are all of the plugins properly differentiating between `cancel` and `reject` events?)
- [ ] Should we make an object to hold the `fulfillment` and `ilp` in the `fulfillCondition` method or just have 3 parameters?
- [ ] Should we use `ilp` as the parameter name for `fulfillCondition` and `rejectIncomingTransfer` methods (to match the field in the `Transfer` object) or something more generic? Can this field hold any string or just ILP packets?
- [ ] Should we add an error to indicate if you try to send an ILP packet that's too big?

### DIFF
--- a/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
+++ b/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
@@ -1,6 +1,6 @@
 ---
 title: The Javascript Ledger Plugin Interface
-draft: 8
+draft: 9
 ---
 # Javascript Ledger Plugin Interface
 
@@ -28,8 +28,8 @@ This spec depends on the [ILP spec](../0003-interledger-protocol/).
 | | [**getFulfillment**](#getfulfillment) ( transferId ) <code>⇒ Promise.&lt;String></code> |
 | | [**sendTransfer**](#sendtransfer) ( transfer ) <code>⇒ Promise.&lt;null></code> |
 | | [**sendRequest**](#sendrequest) ( message ) <code>⇒ Promise.&lt;[Message](#class-message)></code> |
-| | [**fulfillCondition**](#fulfillcondition) ( transferId, fulfillment ) <code>⇒ Promise.&lt;null></code> |
-| | [**rejectIncomingTransfer**](#rejectincomingtransfer) ( transferId, reason ) <code>⇒ Promise.&lt;null></code> |
+| | [**fulfillCondition**](#fulfillcondition) ( transferId, fulfillment, ilp ) <code>⇒ Promise.&lt;null></code> |
+| | [**rejectIncomingTransfer**](#rejectincomingtransfer) ( transferId, ilp ) <code>⇒ Promise.&lt;null></code> |
 | | [**registerRequestHandler**](#registerrequesthandler) ( requestHandler ) <code>⇒ null</code> |
 | | [**deregisterRequestHandler**](#deregisterrequesthandler) ( ) <code>⇒ null</code> |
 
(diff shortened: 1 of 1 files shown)


0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
```diff
@@ -275,21 +275,21 @@ For a detailed description of these properties, please see [`Message`](#class-me
 
 
 #### fulfillCondition
-<code>ledgerPlugin.fulfillCondition( **transferId**:String, **fulfillment**:String ) ⇒ Promise.&lt;null></code>
+<code>ledgerPlugin.fulfillCondition( **transferId**:String, **fulfillment**:String, **ilp**:String ) ⇒ Promise.&lt;null></code>
 
 Submit a fulfillment to a ledger. Plugin must be connected, otherwise the promise should reject.
 
-The `fulfillment` is an arbitrary 32-byte buffer and is provided as a base64url-encoded string.
+The `fulfillment` is an arbitrary 32-byte buffer and is provided as a base64url-encoded string. `ilp` is an optional base64url-encoded ILP data packet that MAY be submitted alongside the fulfillment.
```
michielbdejong: Which [packet type](https://github.com/interledger/rfcs/blob/master/asn1/InterledgerPacket.asn#L30-L37) would this be?

michielbdejong:
If we would implement #314, then that would also require new ILP packet types, and a new form of `sendTransfer`, and a different type of conditional transfers in ledger implementations. I don't think such a fundamental change to the interledger design is necessary.

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
emschwartz: It would be a new one that would just be an `OCTET STRING` with data, but I don't think that the LPI needs to specify that

emschwartz:
@michielbdejong 

Wait, what? Why would `sendTransfer` change at all?

If by "different type of conditional transfers in ledger implementations", you're talking about special-casing the null condition, that isn't necessary to make the informational end-to-end quoting work. It'll work just fine if everyone just sends a normal transfer with that condition.

Also, after talking about it with @justmoon more, we concluded that it would be better to build another transport layer protocol that handles chunking payments, rather than implementing the end-to-end quoting as ILQPv2. In order to build transport layer protocols like this on top of ILP we need well-defined errors that will get relayed back and the ability for receivers to include data alongside the fulfillment. I think these are important building blocks that should be included in the core protocol.

michielbdejong:
>  Why would sendTransfer change at all?

OK, sorry, I now realize I misinterpreted #314, see [my update there](https://github.com/interledger/rfcs/issues/314#issuecomment-338600141).

So then indeed, you would implement it as proposed in this PR, without ledger-level checks.

But then my objection is that without ledger-level checks, this would not be secure.

This same objection has already been brought up in different words [by @sharafian](https://github.com/interledger/rfcs/issues/314#issuecomment-334541657), [by @adrianhopebailie](https://github.com/interledger/rfcs/issues/314#issuecomment-334605210), and [by @dappelt](https://github.com/interledger/rfcs/issues/314#issuecomment-338601674).

If people use this and rely on it, any connector along the path can annoy the senders: just pretend that there never was any data attached to the fulfillment, still receive their money, and annoy the sender. It would then be hard to determine which connector along the path removed the data.

How would you solve that problem?

adrianhopebailie:
> How would you solve that problem?

One option is to make the fulfillment a hash of the data. It's not going to force connectors to forward the data but it does alert the sender if the data is tampered with.

If we want to ensure the data is sent then we could make the fulfillment the data (likely encrypted).

This would require us to remind ourselves why we have standardized on a 32-byte fulfillment and and weigh up the pro's and con's of dropping that restriction.

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: The Javascript Ledger Plugin Interface
-draft: 8
+draft: 9
```
justmoon: Might be worth splitting the two changes. The fulfillment change is a non-breaking change (old implementations will ignore the extra parameter and simply not pass on the fulfillment data) while the rejection format change is breaking (new connectors will confuse old plugins by sending a string instead of an object.)

That said, I support both changes. :+1: - We could do the fulfillment data change right away and start adding support for accepting both the old and new rejection data format in plugins and connectors, but still emit the old format for now.

justmoon:
Split the fulfillment data change into a separate PR #338.

emschwartz:
Closing this as fulfillment data has now been added and https://github.com/interledger/rfcs/pull/347 proposes a more comprehensive set of changes to the LPI

PR #318, opened 2017-10-20
docs(3): add Payment Too Large error

emschwartz:
connectors may want to limit the maximum size of transfers going through them. this adds an ILP error code to indicate when such a limit is reached. it also recommends including the amount that arrived and the maximum amount supported to help senders adjust the size of their payment appropriately

### DIFF
--- a/0003-interledger-protocol/0003-interledger-protocol.md
+++ b/0003-interledger-protocol/0003-interledger-protocol.md
@@ -265,11 +265,9 @@ Date and time when the error was initially emitted. This MUST be expressed in th
 
     OCTET STRING (SIZE(0..8192))
 
-Error data provided primarily for debugging purposes. Systems that emit errors SHOULD include additional explanation or context about the issue.
+Error data either for additional fields relevant for handling a given error or for debugging information.
 
-Protocols built on top of ILP that define behavior for certain errors SHOULD specify the encoding format of error `data`.
-
-Unless otherwise specified, `data` SHOULD be encoded as UTF-8.
+For errors that specify additional data fields (defined below), the data SHOULD be encoded using OER. Other protocols built on top of ILP that define behavior for certain errors SHOULD specify the encoding format of error `data`. Debugging information SHOULD be encoded as UTF-8.
 
 ### ILP Error Codes
 
@@ -279,25 +277,25 @@ Inspired by [HTTP Status Codes](https://tools.ietf.org/html/rfc2616#section-10),
 
 Final errors indicate that the payment is invalid and should not be retried unless the details are changed.
 
-| Code | Name | Description |
-|---|---|---|
-| **F00** | **Bad Request** | Generic sender error. |
-| **F01** | **Invalid Packet** | The ILP packet was syntactically invalid. |
-| **F02** | **Unreachable** | There was no way to forward the payment, because the destination ILP address was wrong or the connector does not have a route to the destination. |
-| **F03** | **Invalid Amount** | The amount is invalid, for example it contains more digits of precision than are available on the destination ledger or the amount is greater than the total amount of the given asset in existence. |
-| **F04** | **Insufficient Destination Amount** | The receiver deemed the amount insufficient, for example you tried to pay a $100 invoice with $10. |
-| **F05** | **Wrong Condition** | The receiver generated a different condition and cannot fulfill the payment. |
-| **F06** | **Unexpected Payment** | The receiver was not expecting a payment like this (the memo and destination address don't make sense in that combination, for example if the receiver does not understand the transport protocol used) |
-| **F07** | **Cannot Receive** | The receiver is unable to accept this payment due to a constraint. For example, the payment would put the receiver above its maximum account balance. |
-| **F08** | **Amount Too Large** | The amount is larger than the maximum size supported by a connector, ledger or the receiver. The producer of this error SHOULD set the error `data` to the concatenation of the following two numbers, each encoded as a 64-bit unsigned, big-endian integer: the amount that arrived and the maximum amount supported by the party producing the error. |
-| **F99** | **Application Error** | Reserved for application layer protocols. Applications MAY use names other than `Application Error`. |
+| Code | Name | Description | Data Fields |
+|---|---|---|---|
+| **F00** | **Bad Request** | Generic sender error. | |
+| **F01** | **Invalid Packet** | The ILP packet was syntactically invalid. | |
+| **F02** | **Unreachable** | There was no way to forward the payment, because the destination ILP address was wrong or the connector does not have a route to the destination. | |
+| **F03** | **Invalid Amount** | The amount is invalid, for example it contains more digits of precision than are available on the destination ledger or the amount is greater than the total amount of the given asset in existence. | |
+| **F04** | **Insufficient Destination Amount** | The receiver deemed the amount insufficient, for example you tried to pay a $100 invoice with $10. | |
+| **F05** | **Wrong Condition** | The receiver generated a different condition and cannot fulfill the payment. | |
+| **F06** | **Unexpected Payment** | The receiver was not expecting a payment like this (the memo and destination address don't make sense in that combination, for example if the receiver does not understand the transport protocol used) | |
+| **F07** | **Cannot Receive** | The receiver is unable to accept this payment due to a constraint. For example, the payment would put the receiver above its maximum account balance. | |
+| **F08** | **Payment Too Large** | The payment is larger than the maximum size supported by a connector, ledger or the receiver. | `SourceAmount` (UInt64), `SourceAmountLimit` (UInt64) |
+| **F99** | **Application Error** | Reserved for application layer protocols. Applications MAY use names other than `Application Error`. | |
 
 #### T__ - Temporary Error
 
 Temporary errors indicate a failure on the part of the receiver or an intermediary system that is unexpected or likely to be resolved soon. Senders SHOULD retry the same payment again, possibly after a short delay.
 
-| Code | Name | Description |
-|---|---|---|
+| Code | Name | Description | Data Fields |
+|---|---|---|---|
 | **T00** | **Internal Error** | A generic unexpected exception. This usually indicates a bug or unhandled error case. |
 | **T01** | **Ledger Unreachable** | The connector has a route or partial route to the destination but was unable to reach the next ledger. Try again later. |
 | **T02** | **Ledger Busy** | The ledger is rejecting requests due to overloading. Try again later. |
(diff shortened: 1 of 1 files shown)


michielbdejong:
Actually, I now realize that the "amount" in the phrase "amount too large" would refer to the amount that arrived, not to the amount in the packet. In that case, I would rename it to something like "Superfluous Source Amount", to avoid confusion with the packet amount. I don't see why anybody would ever complain about a superfluous source amount (so that would need clarifying in the text), but having an error type for it is in itself harmless.

0003-interledger-protocol/0003-interledger-protocol.md
```diff
@@ -289,6 +289,7 @@ Final errors indicate that the payment is invalid and should not be retried unle
 | **F05** | **Wrong Condition** | The receiver generated a different condition and cannot fulfill the payment. |
 | **F06** | **Unexpected Payment** | The receiver was not expecting a payment like this (the memo and destination address don't make sense in that combination, for example if the receiver does not understand the transport protocol used) |
 | **F07** | **Cannot Receive** | The receiver is unable to accept this payment due to a constraint. For example, the payment would put the receiver above its maximum account balance. |
+| **F08** | **Amount Too Large** | The amount is larger than the maximum size supported by a connector, ledger or the receiver. The producer of this error SHOULD set the error `data` to the concatenation of the following two numbers, each encoded as a 64-bit unsigned, big-endian integer: the amount that arrived and the maximum amount supported by the party producing the error. |
```
sappenin: Couple of questions:

1. It's perhaps a distinction without a difference, but specifying the encoding of the data field here feels wrong to me. Shouldn't this be left to the ASN.1 definition, and be more loosely specified here?

1. Should those amounts be defined at all here in the spec? In the case of `F03` and `F04`, we don't specify the amounts that were too low. Seems like we should have consistency, and either mandate amounts in those errors too, or remove the amounts from this (F08) error.

1. Does this use-case (i.e., wanting to know specific and additional information about amounts in an ILP error) indicate that the `data` field, as defined, is insufficient for ILP errors? Instead, should we remove the `data` field and move to something like the `Subprotocol` concept? In that world, a subprotocol could define what information to provide (e.g., the amount that arrived and the max amount supported by the party) _and_ how to encode it.

1. Outside of sub-protocol, there are certain errors that require additional properties. Would it make sense to define actual error packets instead? Or use a collection of key-value pairs that could be easily extensible?

Overall, it feels like a bolt-on to just stuff two numbers into the `data` packet in such a specific manner - my feeling is that we're going to need the ability to include extra information about various errors in the future, and we should have a well-defined way to do such a thing in a consistent manner.

Not sure what that "consistent manner" should be just yet though...

emschwartz:
> the "amount" in the phrase "amount too large" would refer to the amount that arrived, not to the amount in the packet

Correct

> "Superfluous Source Amount", to avoid confusion with the packet amount. I don't see why anybody would ever complain about a superfluous source amount

I see your point about trying to avoid confusion with the amount in the packet but I think the term "superfluous" isn't quite right. It's not that you've sent more than you need to deliver a certain amount, it's that the payment you're trying to send is just too large. Would `Payment Too Large` be clearer?

0003-interledger-protocol/0003-interledger-protocol.md
emschwartz: Fair points.

**Should errors carry structured data?** Probably. There are at least a couple of cases where it will make the handling of those errors significantly more efficient.

**Is the extra data part of an extension or subprotocol, or is it linked to the error?** I think the more common case will be extra data that is intimately linked to the error. End-to-end protocols can include whatever data they want in the error data field already, so the case we're really talking about is errors that are produced and/or understood by senders and connectors.

**Should the error format include a place to put key-value pairs?** If the extra data that needs to be included is linked to the error itself, I don't think it needs to have the keys and could be encoded as just the values. Keys are most useful to include for extensibility mechanisms where you don't know beforehand which values will be included. However, if the data is thought of as part of an individual error, we do know the fields that will be there up front. If we use OER encoding, a `SEQUENCE` would just be the values of the individual fields concatenated together (I was thinking about that when I suggested that the two numbers be concatenated in the data field).

**Should the general error format include a place to put additional error-specific fields or should that be left to the definition of individual errors?** If there are fields that will be common across many different errors it would make sense to add those to the general error format itself. If the extra data is error-specific, I think it would make sense to have a generic field like `data` in the general format, and then to have specific errors dictate what should go in that field.

So my thoughts are that:
- we should keep the general error format as it is
- individual errors can specify what additional information SHOULD be put in the `data` field
- additional data SHOULD be encoded using OER
- the definitions of the additional data should be wherever the specific errors are defined (right now they're all in RFC-3), and we could additionally put them in an ASN.1 file

michielbdejong:
> not that you've sent more than you need to deliver a certain amount,

I don't understand. This is about standard interledger payments as they exist today, right? So the packet contains an amount, expressed in base units of the destination ledger. But you said *that* amount (the amount in the packet) is *not* too high. So then the only other amount involved is the incoming amount reaching the connector that throws the error. That implies that the source amount that was sent to the first connector was too high, right?

Assume all exchange rates are 1:1 and all fees are 1. The amount in the packet is 10, and there are 4 transfers making up the payment. Say the second connector has incorrect data, and overestimates the amount it needs to send by 100, and this causes the third connector to throw Payment Too Large. Then the transfer amounts will be:

```
HAPPY: sender -[13]-> conn1 -[12]-> conn2 -[11]-> conn3 -[10]-> receiver
ERROR: sender -[113]-> conn1 -[112]-> conn2 -[111]-> conn3 X
```

In both cases, the amount in the packet would be 10. `conn3` *could* just pocket the difference, but in an attempt to be nice to `conn2`, rejects their obvious overpayment. Is that what you mean?

michielbdejong:
It feels to me like you probably intend this error code to be used for best-effort payments, where the packet contains no `amount` field? If that's the case, then we should wait with adding this error type, until we added that ILP packet type.

dappelt:
@michielbdejong From what I understand, this is not necessarily related to if there is an amount field or not.  All this error indicates is that a connector is not willing to forward a given amount because he considers the risk/reward ratio not sufficient (i.e. the payment size is correlated with risk, keeping payment sizes small is a way to reduce risk). This error indicates to the sender that he should split up his payment in smaller chunks.

michielbdejong:
@dappelt 
> All this error indicates is that a connector is not willing to forward a given amount 

But there are three very different situations:
* the payment amount (as indicated in the packet) is too high
* the payment amount (as indicated in the packet) is low enough, but the incoming amount is too high, and the connector has the courtesy to tell their peer "I think you gave me too much money, something must be wrong here"
* (if we decide to implement this type of best-effort payments): there was no amount in the packet, and the payment amount (as, in that case, indicated by the incoming amount), is too high

This new error code, if we introduce it, should only describe one of these three situations. I now think that @emschwartz wants it to describe the third one, hence my remark.

dappelt:
As I understand it, it is neither of the three situations you mention. This error is not related to the `amount` in the ILP packet, but to the amount of the incoming transfer.

michielbdejong:
> This error is not related to the amount in the ILP packet, but to the amount of an incoming transfer.

That means it can still be either the second or the third.

@emschwartz  which one is it?

emschwartz:
@michielbdejong why do you think the error code needs to correspond to only one of those circumstances? As @dappelt said, no matter why the error is thrown it would communicate the same instructions to the sender: send a smaller payment (and it includes a suggestion as to how the sender should adjust their payment in order for it to be accepted).

(This also made me realize this error could be thrown by the receiver as well as connectors)

0003-interledger-protocol/0003-interledger-protocol.md
```diff
@@ -279,24 +277,25 @@ Inspired by [HTTP Status Codes](https://tools.ietf.org/html/rfc2616#section-10),
...
+| **F03** | **Invalid Amount** | The amount is invalid, for example it contains more digits of precision than are available on the destination ledger or the amount is greater than the total amount of the given asset in existence. | |
+| **F04** | **Insufficient Destination Amount** | The receiver deemed the amount insufficient, for example you tried to pay a $100 invoice with $10. | |
+| **F05** | **Wrong Condition** | The receiver generated a different condition and cannot fulfill the payment. | |
+| **F06** | **Unexpected Payment** | The receiver was not expecting a payment like this (the memo and destination address don't make sense in that combination, for example if the receiver does not understand the transport protocol used) | |
+| **F07** | **Cannot Receive** | The receiver is unable to accept this payment due to a constraint. For example, the payment would put the receiver above its maximum account balance. | |
+| **F08** | **Payment Too Large** | The payment is larger than the maximum size supported by a connector, ledger or the receiver. | `SourceAmount` (UInt64), `SourceAmountLimit` (UInt64) |
```
dappelt: According to OP, the data fields `SourceAmount` and `SourceAmountLimit` should help the sender adjust the payment amount. So the procedure the sender would adjust the payment size is:

1. the sender calculates `OverpayRatio` = `SourceAmount` / `SourceAmountLimit`
2. the sender divides the amount of the failed payment by `OverpayRatio` to obtain the new payment amount, which should be small enough to avoid a `Payment Too Large` error.

Is that correct? If so, how about replacing the data fields `SourceAmount` and `SourceAmountLimit` directly with `OverpayRatio`? The sender is likely not familiar with the unit of the connector's `SourceAmount`, so absolute integers are meaningless to the sender. Furthermore, it would save same bytes (one 32 bit float as opposed to two 64 bit integer).

Alternatively, if we stick with `SourceAmount` and `SourceAmountLimit`, each connector that passes on a `Payment Too Large` error could adjust the `SourceAmount` and `SourceAmountLimit` under consideration of it's own exchange rate (like with a "Quote by Destination Amount"). By doing so, the sender would obtain amounts in his own ledger units and adjusting the payment size would be trivial.

emschwartz: That's correct.

> how about replacing the data fields SourceAmount and SourceAmountLimit directly with OverpayRatio?

Fair question. I guess it mostly comes down to whether we want to include a float (losing some precision -- also it should be a 64 bit float at least) or the two integers. I don't feel too strongly about this.

Either way, I feel strongly that the connectors along the way should not try to modify the error. That would add another feature that every connector must implement (it would get seriously messed up if some connectors adjusted the values but not all), rather than limiting the feature to the party that has the payment size limit.

michielbdejong:
I don't agree at all that we should use one error code for these three very different situations.

If an amount was included in the packet, then we cannot distinguish between the first and second situation. That needs to be defined explicitly in the definition of this error type. I understand now, that the error is only intended for the second and third situation, not the first.

If the packet contained an amount, then it's the second situation, and when a connector receives this error, they should get a new quote and adjust their transfer amount. Maybe this can even be done within the current payment's timeout.

If the packet contained no amount, then the connector should reverse-apply its own liquidity curve to the limit amount, and that way, translate the error from its outgoing ledger to its incoming ledger.

It would be possible to overload one error class for both the second and the third situation, but I feel it would add more confusion than clarity?

0003-interledger-protocol/0003-interledger-protocol.md
dappelt: > I guess it mostly comes down to whether we want to include a float (losing some precision -- also it should be a 64 bit float at least) or the two integers.

Since exchange rates fluctuate (and liquidity curves don't have to be linear, at least for the moment), `OverpayRatio`/`SourceAmount` is anyway only an estimate. Therefore, I am not too much concerned with the loss of precision due to the floating point operation. Most likely a sender would subtract a bit more from the payment amount than what is indicated by `OverpayRatio` to avoid another `Payment Too Large` error.

justmoon: The main arguments for using two integers from my perspective:

* Floats allow special values like infinity, negative zero, NaN and subnormal numbers which could be used to trigger bugs or strange behavior in implementations.
* We already have 64-bit unsigned ints used for amounts in various places whereas using a float would introduce a new type that people have to implement in their parsers.
* Floats introduce some amount of precision error whereas the actual amount and max amount values are (usually) going to be actual values that tripped some *if* statement somewhere. If nothing else, that seems good for debugging.

Taken together I think we should use two integers.

One thing I don't understand is that it's called `SourceAmount` and `SourceAmountLimit`. That seems to indicate that it's the connector receiving the overpayment that is complaining. But wouldn't limits usually be on trustlines/channels? And if so, wouldn't the connector sending that transfer notice the fact that the payment would be over the limit (and be the one to trigger the error?)

I also don't like `SourceAmount` since we usually mean the source amount of the overall ILP path, not the *locally* incoming amount.

So I'd propose to use instead: `LocalAmount` and `LocalAmountLimit` which is neutral in terms of incoming vs outgoing and gives you a hint that this amount is in the complainant's local unit which may be different than the units of the source and destination ledgers.

Another problem is that the amount may be larger than UINT64_MAX. This could be an argument for using floats, but a) I think that this would be very rare and 2) all that would happen is that you would cap the `LocalAmount` at UINT64_MAX which would in the worst case trigger an extra overpayment or two. Plus, if there is a connector where this happens a lot and they don't mind the extra complexity, they could also adjust `LocalAmountLimit` downwards to ensure faster convergence.

emschwartz:
This may not actually be needed for chunked payments so I'm going to close it for now

PR #319, opened 2017-10-23
feat: add simplified flow for http-ilp

michielbdejong:


michielbdejong:
OK, but we're already using this to achieve streaming payments in the tutorials, so what should we do about that, then?

adrianhopebailie:
> into the actual spec until we're using it in implementation.

Not sure I agree. These specs are supposed to be drafts of ideas looking for feedback. We shouldn't hold anything back.

The best way to get feedback is make the ideas public and gather experience from implementors.

Right now this whole IL RFC is "invisible" because no version has ever been merged into master. I think that is a mistake.

sharafian:
I'm also not sure what the advantage of this is as compared to the existing HTTP ILP flow. Are you proposing that implementations should include both or should there be two incompatible HTTP ILP flows described here?

michielbdejong:
> I'm also not sure what the advantage of this is as compared to the existing HTTP ILP flow

Try out https://interledger.org/tutorials/trustlines and you'll see that it streams  hundreds of letters per second, as the response body of one single (streamed) http GET request.

Doing that with the existing HTTP ILP flow would require a lot of individual http GET requests, one for each letter, right? That would be a lot of overhead, so avoiding that overhead is the advantage.

michielbdejong:
I guess this is now no longer an Interledger-only question, maybe we should discuss the `Pay` header on https://lists.w3.org/Archives/Public/ietf-http-wg/2017OctDec/0118.html. Will ask @adrianhopebailie for advice about this.

michielbdejong:
@sharafian in the meantime, do you now at least see the advantage of this is as compared to the existing HTTP ILP flow? So can we merge this? If not, then let's schedule a 1:1 chat about this.

michielbdejong:
Merging this in to #329, will reopen as a PR once discussion reaches a consensus.

PR #320, opened 2017-10-24
feat: R03 Superfluous Source Amount

michielbdejong:
Inspired by discussions yesterday about New Interledger, it occurred to me that this error type is missing from Current Interledger. A sender, who doesn't receive route broadcasts, has no way to know when a quote expires unexpectedly (a new curve is broadcast). If the rate got worse, their next payment will fail with InsufficientSourceAmount. But if the rate got better, they would be paying too much.

Customers who want to take advantage of multiple ILSPs would still have to send quote requests to each of them each time, but if a customer has only one ILSP, this is a neat solution; a lot simpler than setting up subscriptions to route broadcasts.

An alternative would be to consider the amount in the transfer as a 'maximum I am willing to pay you', and then the ILSP would only charge the actual cost at the best current rate. But that would be a bigger change to how ledgers work.


0003-interledger-protocol/0003-interledger-protocol.md
```diff
@@ -314,6 +314,7 @@ Relative errors indicate that the payment did not have enough of a margin in ter
...
 | **R01** | **Insufficient Source Amount** | Either the sender did not send enough money or the exchange rate changed before the payment was prepared. The sender MAY try again with a higher amount, but they SHOULD NOT do this indefinitely or a malicious connector could steal money from them. |
 | **R02** | **Insufficient Timeout** | The connector could not forward the payment, because the timeout was too low to subtract its safety margin. The sender MAY try again with a higher expiry, but they SHOULD NOT do this indefinitely or a malicious connector could cause them to tie up their money for an unreasonably long time. |
+| **R03** | **Superfluous Source Amount** | The transfer amount is sufficient to cover the costs of the payment, but this error is thrown as a courtesy, to warn that a lower transfer amount would also suffice. The sender MAY obtain a new quote and try again with an adjusted amount, or retry with the same amount. If they retry with the same amount, the same error SHOULD NOT be thrown a second time. |
```
sharafian: In order to implement this logic, receivers would have to keep state about which packets they've seen and rejected with this error. Furthermore, a receiver that doesn't want to accept overpayment would be violating this spec by continuing to reject overpayment.

michielbdejong: good point, I'll remove the phrase

0003-interledger-protocol/0003-interledger-protocol.md
```diff
@@ -314,6 +314,7 @@ Relative errors indicate that the payment did not have enough of a margin in ter
...
 | **R01** | **Insufficient Source Amount** | Either the sender did not send enough money or the exchange rate changed before the payment was prepared. The sender MAY try again with a higher amount, but they SHOULD NOT do this indefinitely or a malicious connector could steal money from them. |
 | **R02** | **Insufficient Timeout** | The connector could not forward the payment, because the timeout was too low to subtract its safety margin. The sender MAY try again with a higher expiry, but they SHOULD NOT do this indefinitely or a malicious connector could cause them to tie up their money for an unreasonably long time. |
+| **R03** | **Superfluous Source Amount** | The transfer amount is sufficient to cover the costs of the payment, but this error is thrown as a courtesy, to warn that a lower transfer amount would also suffice. The sender SHOULD obtain a new quote and retry the Interledger payment with that adjusted transfer amount. |
```
emschwartz: How much higher would it have to be for you to throw this error? Seems like it could be kind of annoying if you intentionally send a high buffer because you really want the payment to go through on the first try and some connector who is trying to do you a favor gives you this error.

michielbdejong: Any amount that does not match the amount from a recent quote will be wrong. It's up to the ILSP to decide what is "recent".

Examples:
1. As an end-user, I do a quote-by-destination for destination amount 123. It tells me source amount 15.
1. As an end-user, I do a quote-by-source for source amount 15. It tells me destination amount 123. I then do the payment with packet amount 123 and transfer amount 15.
1. As an end-user, I do a curve-quote. My ILSP give me a curve: `[14, 122], [16, 124]`.

In all cases, I wait for say 20 seconds (e.g. waiting for user to click 'OK') and then do the payment with packet amount 123 and transfer amount 15.

Say the market moves a little bit during these 20 seconds, but not too much. The ILSP will then e.g. have a grace period of say 60 seconds during which it will serve both old and new quote.

Say the market moves a lot during these 20 seconds. The ILSP will then reject the payment with either `InsufficientSourceAmount` or `SuperfluousSourceAmount`, to allow the end-user to re-quote. The ILSP should keep enough margin so that such re-quotes are only necessary in relatively rare circumstances.

adrianhopebailie:
I think we need to distinguish between:

1. You sent more than is required, given the final destination amount so I'm going to do the courtesy of rejecting this quickly so you can try again with a lower amount (Only valid while we still use ILP v1 payment packet)
2. I don't have enough liquidity right now to service this request. Already dealt with by T04.

So, I assume this is only for case 1?

A counter-proposal is to change F04 to be "Incorrect Destination Amount". The error will then carry end-to-end data from the receiver back to the sender indicating the amount received and the amount actually expected. This shouldn't be visible to the intermediaries.

All that intermediaries know is why the transaction failed but since they don't know the desired destination amount they can't do anything about it anyway so that is enough.

michielbdejong:
>  I assume this is only for case 1?

yes

> change F04 to be "Incorrect Destination Amount"

I don't see any relationship between F04 (which is used when e.g. you pay 100 USD for a 110 USD invoice) and this proposal. As mentioned, this error is meant for what you called case 1, when you used the source amount based on a quote that is not, or no longer, correct. This is not related to the purpose of your payment to the receiver.

adrianhopebailie:
> This is not related to the purpose of your payment to the receiver.

The purpose of a payment is not relevant in this context. What is relevant is whether or not the node (be it an intermediary connector or the receiver) wants to accept the amount that is being transferred to them.

This could be because the amount is too low or too high. It's only worth differentiating between these if the way the transfer-sending-node reacts will be different. 

In ILP v2 it definitely won't be because no intermediary node will know the destination amount so they can't do anything but pass back the error.

You could argue that in ILP v1 the transfer-sending-node could retry with a different amount but I'm not sure that it's worth designing for that edge case if we expect it to be deprecated.

If we do then I'd expect both F04 and this new code to be deprecated in future in favour of a new permanent error that is for "Incorrect Destination Amount"

Also, I think you are assuming only receivers will ever raise an F04 error? I think it's valid for an intermediary node to raise an F04 error if they consider the amount they were sent in a transfer to be insufficient (or too much) for them to forward the payment. This could be due to a fixed per tx fee that creates a minimum payment amount or a compliance rule that creates a maximum amount they can send.

michielbdejong:
> You could argue that in ILP v1 the transfer-sending-node could retry with a different amount

Yes, that's what I'm proposing.

> but I'm not sure that it's worth designing for that edge case if we expect it to be deprecated.

Does that mean that you are rejecting my PR? Or can it be merged with that side-note?

0003-interledger-protocol/0003-interledger-protocol.md
emschwartz: So if the market doesn't move and the user submits a source amount of 16, is that invalid? What if the curve the user got is a bit imprecise and they are giving a higher number because of a rounding error? What if the user wants to add an extra buffer to give the payment a higher chance of going through? 

What if there are more connectors between the sender and the connector in question and due to rounding, accident, or other connectors' rates fluctuating more money ends up at this connector than they wanted? Would every connector need to remember not only their own rates but also the exact rate they should use when passing a payment to a specific peer?

michielbdejong:
> Would every connector need to remember not only their own rates but also the exact rate they should use when passing a payment to a specific peer?

yes

michielbdejong:
@emschwartz It's true that if you initiate the payment immediately after getting the quote response, then this error code would never be thrown.

But the advantage of defining it is that you could cache quotes indefinitely, and rely on this error to be thrown by your ILSP when a re-quote is necessary. It would therefore be a powerful tool to implement ILQP caching without the need to defining the expiry upfront.

michielbdejong:
It would be mathematically impossible to get a different source amount if both you and your ILSP use the same liquidity curve.

True, it would not combine well with overpaying, but I think its benefits (efficient ilqp caching) would be much greater than the potential benefits of overpaying.

michielbdejong:
I'll give up on this battle; it's indeed not ideal that sender and connector would need to coordinate the desired threshold for requoting; if the connector throws this error too often in the eyes of the sender, then it would be just annoying instead of helping.

So senders will just still need to re-quote periodically in order to get informed about rate improvements, as is the situation now.

PR #326, opened 2017-10-31
Add quoting to LPI for ILP v2

adrianhopebailie:
ILP v2 proposes to deprecate the ILP Payment Packet in favor of a generic ILP Data Packet containing just the ILP Address and data.

I propose we also deprecate ILQP in favor of ledger layer quoting using the new ILP Data Packet.

# Proposal

 1. Deprecate ILQP
 2. Add quoting functionality to the bilateral communications between nodes

NOTE: Here I have just proposed changes to the LPI but similar changes could be made to BTP.

### Fixed Receive Amount

1. A sender that wishes to know how much to send on the first transfer will use `sendQuote` and will attach the same ILP Packet that they intend to use for the transfer. They will not provide an amount or expiry.

2. The next connector will receive an `incoming_quoterequest` event and handle this by responding with a `Quote` response containing the minimum amount and maximum expiry times to use.

**NB:** It is not important how the connector determines the response. If the connector is able to inspect the ILP Packet and determine the destination amount it may choose not to forward the quote on but calculate the required transfer amount itself. Critically, the protocol does not force connectors to implement specific business rules or behaviour, their onkly obligation is to respond to the quote.

3. Assuming the connector chooses to simply forward the quote request on to the next connector it will do so with minimal processing required (purely routing based on the ILP Address in the packet).

The connector may even choose to forward the quote down multiple routes, collate the responses, and forward the best response down to the requestor.

4. This will continue until either an intermediary determines that it is able to respond or the quote gets to the receiver. If the quote gets to the receiver they will parse the ILP packet including any higher-layer protocol data that may have been protected from intermediaries. They will determine the desired amount that should be sent in an incoming transfer carrying this packet and will respond with the amount and the maximum time that this quote is valid in the response `Quote` object.

5. Each intermediary connector will then process an `incoming_quoteresponse` with attached amount and expiry, will calculate the desired incoming transfer they must receive and a safe timeout for that rate to still be valid. They will then forward the quote response on to the next connector with these values.

6. When the sender gets the response back they can construct a `sendTransfer` call using the values they received in the Quote from the first connector.

### Fixed Send Amount

1. A sender that wishes to know how much to the receiver will get based on a fixed amount in the first transfer will use `sendQuote` and provide an amount and an expiry that they would use if sending the transfer now (this will test if any connectors along the route reject the transfer because the telescoping timeouts get too low). The ILP Packet would contain the higher-layer protocol data. 

2. The first connector will process the quote request in the same way as it would a transfer for the same amount and expiry however knowing that it is a quote will not put any funds on hold. It will forward the quote on using an appropriate amount and expiry for the outgoing route.

3. When the receiver get's the quote request it will populate the higher-layer protocol data of the response with the amount that it received (and possibly the expiry that was specified) on the last quote request.

4. The response is then relayed back to the sender who is able to see how much will be received.

## Why not just use `sendRequest`?

The LPI methods used to do this need to exchange, at a minimum, a transfer amount and expiry. These should mirror their use in a `sendTransfer` call on the LPI.

This data should not be carried in the ILP packet. As we've determined with the ILP Payment Packet, it is best for data such as the destination amount to be carried in the higher-layer protocol data.

## Why not use ILQP's `QuoteBySourceAmount` and `QuoteByDestinationAmount` packet-pairs?

As explained above, the only data that intermediaries need to see is the ILP Address of the destination. The amount data that quoted between connectors should be carried at the same layer as the amount data in transfers.

## Why not used failed transfers?

Intermediaries that process failed transfers incur a cost to put funds on hold. This is unnecessary and can be easily avoided by simply offering the ability to get a quote based on the same data and flow but where connectors are aware they do not need to put the funds on reserve.

### DIFF
--- a/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
+++ b/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
@@ -1,6 +1,6 @@
 ---
 title: The Javascript Ledger Plugin Interface
-draft: 8
+draft: 9
 ---
 # Javascript Ledger Plugin Interface
 
@@ -26,6 +26,7 @@ This spec depends on the [ILP spec](../0003-interledger-protocol/).
 | | [**getAccount**](#getaccount) ( ) `⇒ String` |
 | | [**getBalance**](#getbalance) ( ) <code>⇒ Promise.&lt;String></code> |
 | | [**getFulfillment**](#getfulfillment) ( transferId ) <code>⇒ Promise.&lt;String></code> |
+| | [**sendQuoteRequest**](#sendquoterequest) ( quoteRequest ) <code>⇒ Promise.&lt;[Quote](#class-quote)></code> |
 | | [**sendTransfer**](#sendtransfer) ( transfer ) <code>⇒ Promise.&lt;null></code> |
 | | [**sendRequest**](#sendrequest) ( message ) <code>⇒ Promise.&lt;[Message](#class-message)></code> |
 | | [**fulfillCondition**](#fulfillcondition) ( transferId, fulfillment ) <code>⇒ Promise.&lt;null></code> |
@@ -39,6 +40,8 @@ This spec depends on the [ILP spec](../0003-interledger-protocol/).
 | [**connect**](#event-connect) | `( ) ⇒` |
 | [**disconnect**](#event-disconnect) | `( ) ⇒` |
 | [**error**](#event-error) | `( ) ⇒` |
+| [**incoming_quoterequest**](#event-_quoterequest) | <code>( quoteRequest:[IncomingQuoteRequest](#class-quote) ) ⇒</code> |
+| [**incoming_quoteresponse**](#event-_quoteresponse) | <code>( quoteResponse:[IncomingQuoteResponse](#class-quote) ) ⇒</code> |
 | [**incoming_transfer**](#event-_transfer) | <code>( transfer:[IncomingTransfer](#class-transfer) ) ⇒</code> |
 | [**incoming_prepare**](#event-_prepare) | <code>( transfer:[IncomingTransfer](#class-transfer) ) ⇒</code> |
 | [**incoming_fulfill**](#event-_fulfill) | <code>( transfer:[IncomingTransfer](#class-transfer), fulfillment:String ) ⇒</code> |
@@ -46,6 +49,8 @@ This spec depends on the [ILP spec](../0003-interledger-protocol/).
 | [**incoming_cancel**](#event-_cancel) | <code>( transfer:[IncomingTransfer](#class-transfer), cancellationReason:[RejectionMessage](#class-rejectionmessage) ) ⇒</code> |
 | [**incoming_request**](#event-_request) | <code>( message:[Message](#class-message) ) ⇒</code> |
 | [**incoming_response**](#event-_response) | <code>( message:[Message](#class-message) ) ⇒</code> |
+| [**outgoing_quoterequest**](#event-_quoterequest) | <code>( quoteRequest:[OutgoingQuoteRequest](#class-quote) ) ⇒</code> |
+| [**outgoing_quoteresponse**](#event-_quoteresponse) | <code>( quoteResponse:[OutgoingQuoteResponse](#class-quote) ) ⇒</code> |
 | [**outgoing_transfer**](#event-_transfer) | <code>( transfer:[outgoingTransfer](#class-transfer) ) ⇒</code> |
 | [**outgoing_prepare**](#event-_prepare) | <code>( transfer:[outgoingTransfer](#class-transfer) ) ⇒</code> |
 | [**outgoing_fulfill**](#event-_fulfill) | <code>( transfer:[outgoingTransfer](#class-transfer), fulfillment:String ) ⇒</code> |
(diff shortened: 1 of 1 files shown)


adrianhopebailie:
> I think sendQuote is not a correct name, it should be more like getRecommendedChunkSize.

This makes no sense to me. You'll have to explain. I am not looking for chunk sizes I am looking for a quote, either: 
- How much must I transfer to you for the attached ILP packet to be delivered correctly? OR
- How much will the destination account receive if I send the attached ILP packet in a transfer with the following amount and expiry?

Let's not get into the goals of the new transport layer stuff @emschwartz is working on, the goal here is a simple base layer that facilitates those goals. At this layer there is no such thing as chunked payments.

> I also think it should be wrapped in sendRequest, to make clear that it's the next connector, not the ledger, that processes this request.

I also disagree with this entirely. We can't define an interface for a ledger and then say _"Actually, some of the methods are not really for the ledger, they're for the next connector?"_

Everything on the LPI is for the ledger first and then expected to be forwarded to the connector, including `sendRequest`. Each method has its own data that may be processed by the ledger and an ILP Packet that is only processed by the next connector.

Because 'sendRequest' has no additional parameters it's effectively a tunnel through the ledger because we provide no data to the ledger to process. (Although, even that is not entirely true because a ledger can process anything in the `custom` field and behave differently based on what it finds). 

'sendQuote' is explicitly not just a tunnel. It contains data that the ledger is explicitly allowed to process and respond to. (Many ledgers won't but it is allowed since the logic will be very similar to that for `sendTransfer`)

Eg: The ledger could immediately raise an F04 error in response to a `sendQuote` if the account of the next node is short of liquidity. 
Eg: The ledger may use the custom data to supplement the quote as it passes back and forth.

Any argument you can make for why the ledger will not process the `sendQuote` can then be made against `sendTransfer`.

michielbdejong:
Wait, this is about type-9 payments, with no amount in the packet, right? Or about type-1? Or both?

> How much must I transfer to you for the attached ILP packet to be delivered correctly?

For type-1 that would be QuoteByDestination.
For type-9 that would be the minimal chunk size, so that at least zero (or one? - depends on how we define type-9 payments) ledger base units arrive at the destination account.

> How much will the destination account receive if I send the attached ILP packet in a transfer with the following amount and expiry?

There's no way of predicting that with certainty, right? So then I don't see the advantage over already setting a first chunk size of e.g. 1 USD; if the payment makes it all the way, the receiver can then see what destination amount results, and decide to fulfill or reject.

> some of the methods are not really for the ledger, they're for the next connector

I see your point, but isn't that the situation we already have? If you want to change this for quote requests, then shouldn't we also change it for transfers?

adrianhopebailie:
@michielbdejong where is all this chunking stuff coming from? Why are you assuming that all payments using the new packet will be for chunked payments?

adrianhopebailie:
> Overall, I think this would be a breaking change without much substantive difference

We're about to deprecate the primary ILP packet, I don't think introducing a breaking change is a strong argument against this.

> Either the amount is hidden or it is not hidden.

That is a false assumption. The new packet format assumes the amount is hidden by default but there is nothing stopping new protocols being defined on top of ILP that make some data about the payment visible to intermediaries (including the destination amount).

In fact, there may be compliance reasons that such a protocol will need to be developed in the future.

The key point I am trying to make is that we should not decide where data goes based on how we think connectors and ledgers will be implemented. We only have our limited experience to call on in order to check our assumptions.

We should put the data where it is most appropriate. The data in the parameters of methods to the LPI are bilateral exchanges. The data from Alice to Chloe is different to the data from Chloe to Bob. The data in the ILP packets is the data that is passed from the sender to the receiver. The connector behaviour should not change or be impacted if that data is suddenly encrypted by a new higher-layer protocol.

Passing the quote data in an ILQP packet is the wrong place to pass that data because it requires the connectors to parse it and change it at each hop.

> How is this different from ILQP? With ILQP every connector can do whatever they want to respond to the quote, they just need to respond to the quote.

That aspect of the new proposal is not different to ILQP but ILQP does not support carrying the same ILP packet end-to-end that would be used for the payment. In fact, it doesn't support carrying an ILP packet at all so there is no way to design higher-layer protocols that can exchange important end-to-end data  between the sender and receiver as part of the quoting process.

We could modify ILQP to address some of these issues but you'd still be carrying the quote data (amount and expiry) at the wrong layer of the stack.

> What is the substantive difference between where the fields go (whether in a quote packet or in the body of the request to the ledger)?

Why don't we put the transfer amount and expiry in the ILP packet and just let the connectors change that value for each onward hop?

The data in the body of the request to the ledger is bilateral data specific to the interactions that need to happen between two peers. Getting a quote is an interaction we have pushed up the stack into ILQP even though it should be a direct exchange between peers (in exactly the same way a transfer is)

> This may not be true if connectors allocate a certain amount of their liquidity (bandwidth) to each peer. If I have $10k split amongst 10 peers, would I care if one of them wastes their bandwidth with payments that aren't going to go through? I think connectors might have to reserve liquidity for each peer, because you wouldn't want one person's preparing tons of payments to interfere with another's experience.

As discussed this is a simplistic view of the business of being a connector. Failed payments cost money and any economically incentivized connector is going to maximize profit by reducing the ratio of failed payments they forward.

emschwartz:
Can this be closed now?

PR #334, opened 2017-11-11
feat: use interledger-psk as the payment method identifier for http-ilp

michielbdejong:


justmoon:
Can we do this change in a non-breaking way? E.g. by using `Pay-Acme` instead of `Pay` for a hypothetical `Acme` transport layer protocol.

michielbdejong:
That would still break https://tools.ietf.org/html/draft-hope-bailie-http-payments-00#section-4.1

adrianhopebailie:
@justmoon the motivation here is that there is support from a variety of groups to take the Web Payments work and apply it to other protocols.

In that sense we need a payment method identifier as the first parameter. Open to suggestions though on how we could achieve this differently.

michielbdejong:
There will never be one machine-readable payment request header.

screen shot 2017-11-27 at 11 33 30

Even between @adrianhopebailie, @sharafian and me, we can't agree on one favorite we all like. And other people will add other wishes and requirements. Some versions of the `Pay` header will not even be related to Interledger. Therefore, we can either let these competing versions proliferate so that software becomes incompatible (like what happened with WebDAV, basically), or say the `Pay` header name is now used up and other implementations will have to pick another header name, or try to unify all machine-readable payment request headers by adding a version identifier as the first part of the header value.

sharafian:
We've already got support for the `Pay: <payment method>` syntax in `superagent-ilp` so I'm ok with changing to that. I do think that using a different header name works too, though. I can't think of any practical reason why it's important for us to call it `Pay`.

michielbdejong:
OK, so can this be merged?

justmoon:
If the code already supports it, go for it. :+1:

michielbdejong:
OK, @sharafian can you review?

PR #337, opened 2017-11-14
docs(0003): add fulfillment data ilp packet

justmoon:
Implements #314.

### DIFF
--- a/0003-interledger-protocol/0003-interledger-protocol.md
+++ b/0003-interledger-protocol/0003-interledger-protocol.md
@@ -1,6 +1,6 @@
 ---
 title: The Interledger Protocol (ILP)
-draft: 5
+draft: 6
 ---
 # Interledger Protocol (ILP)
 
@@ -316,6 +316,32 @@ Relative errors indicate that the payment did not have enough of a margin in ter
 | **R02** | **Insufficient Timeout** | The connector could not forward the payment, because the timeout was too low to subtract its safety margin. The sender MAY try again with a higher expiry, but they SHOULD NOT do this indefinitely or a malicious connector could cause them to tie up their money for an unreasonably long time. |
 | **R99** | **Application Error** | Reserved for application layer protocols. Applications MAY use names other than `Application Error`. |
 
+### ILP Fulfillment Data Format
+
+This type of ILP packet is attached to a fulfillment and carries additional transport layer information that the sender may use to make further payments. Here is a summary of the fields in the ILP fulfillment data format:
+
+| Field | Type | Short Description |
+|:--|:--|:--|
+| data | OCTET STRING | Transport layer fulfillment data |
+
+#### Example
+
+Consider a payload of `4 16 65` (payload length is `3`).
+The length of the envelope content would then be `1+3+1 = 5`.
+The ILP packet type is always `9` for fulfillment data, and extensions is always `0`, so on the wire
+(and when passed between software modules like for instance ledger plugins),
+the resulting ILP packet would look like this:
+
+| type | content length | payload length | data | extensions |
+|:--|:--|:--|:--|:--|
+| 9 | 5 | 3 | 4 16 65 | 0 |
+
+#### data
+
+    OCTET STRING (SIZE(0..32767))
+
+Arbitrary data that is attached to the fulfillment. The contents are defined by the transport layer protocol.
+
 ## Appendix A: ASN.1 Module
 
 See [ASN.1 Definitions](../asn1/InterledgerProtocol.asn).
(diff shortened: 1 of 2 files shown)


0003-interledger-protocol/0003-interledger-protocol.md
```diff
@@ -316,6 +316,20 @@ Relative errors indicate that the payment did not have enough of a margin in ter
 | **R02** | **Insufficient Timeout** | The connector could not forward the payment, because the timeout was too low to subtract its safety margin. The sender MAY try again with a higher expiry, but they SHOULD NOT do this indefinitely or a malicious connector could cause them to tie up their money for an unreasonably long time. |
 | **R99** | **Application Error** | Reserved for application layer protocols. Applications MAY use names other than `Application Error`. |
 
+### ILP Fulfillment Format
+
+Here is a summary of the fields in the ILP fulfillment format:
```
michielbdejong: Where is the fulfillment itself? When you say "ILP fulfillment format", I expect to see 32 bytes

0003-interledger-protocol/0003-interledger-protocol.md
```diff
@@ -316,6 +316,20 @@ Relative errors indicate that the payment did not have enough of a margin in ter
...
+
+| Field | Type | Short Description |
+|:--|:--|:--|
+| data | OCTET STRING | Transport layer fulfillment data |
```
michielbdejong: This doesn't describe what the packet really looks like, it would be clearer to do it like https://github.com/interledger/rfcs/blob/master/0003-interledger-protocol/0003-interledger-protocol.md#ilp-payment-packet-format

0003-interledger-protocol/0003-interledger-protocol.md
```diff
@@ -316,6 +316,20 @@ Relative errors indicate that the payment did not have enough of a margin in ter
...
+
+#### data
+
+    OCTET STRING (SIZE(0..32767))
+
+Arbitrary data that is attached to the fulfillment. The contents are defined by the transport layer protocol.
```
michielbdejong: Right, the phrase 'attached to the fulfillment' makes more sense to me than the phrase 'the ILP fulfillment format' which was used above

0003-interledger-protocol/0003-interledger-protocol.md
justmoon: I didn't include the fulfillment since this is the ILP packet that travels *with* the fulfillment. Similar to how the payment packet doesn't contain the condition, because it travels *with* the condition.

I actually now agree with @adrianhopebailie that the condition probably should have been part of the ILP packet, but hindsight is 20/20 and I'd rather keep things consistent.

If we include the fulfillment in the ILP packet and we do this as a non-breaking change (add new `ilp` parameter in LPI `fulfillCondition`), then we'd be sending the fulfillment twice.

Agree, that the above rationale is not obvious, so I can add some verbiage to explain why this is the way it is.

0003-interledger-protocol/0003-interledger-protocol.md
justmoon: Good idea, I'll give that a shot. :+1:

justmoon:
@michielbdejong Addressed your comments

0003-interledger-protocol/0003-interledger-protocol.md
```diff
@@ -318,12 +318,18 @@ Relative errors indicate that the payment did not have enough of a margin in ter
...
 
 | Field | Type | Short Description |
 |:--|:--|:--|
 | data | OCTET STRING | Transport layer fulfillment data |
 
+#### Example
+
+| length | data | extensions |
+|:--|:--|:--|:--|
+| 3 | 4 16 65 | 0 |
```
michielbdejong: You're still missing the packet type and the envelope length, and would be good to explain that `4 16 65` is the actual payload. Also, number of `:--|` should match number of columns in the markdown table. So more like:

#### Example

Consider a payload of `4 16 65` (payload length is `3`).
The length of the envelope content would then be `1+3+1 = 5`.
The ILP packet type is always `10` for fulfillment data, and extensions is always `0`, so on the wire
(and when passed between software modules like for instance ledger plugins),
the resulting ILP packet would look like this:

| type | content length | payload length | data | extensions |
|:--|:--|:--|:--|:--|
| 10 | 5 | 3 | 4 16 65 | 0 |

michielbdejong:
Made this optional in IToT 17q4 and required in IToT 18q1: https://github.com/interledger/interledger/wiki/Interledger-over-BTP/3673b3f90fcdeb25061197a73394d02addd88f9f

justmoon:
@michielbdejong Ready for re-review!

0003-interledger-protocol/0003-interledger-protocol.md
justmoon: Nice! I added this example. Thanks!

PR #338, opened 2017-11-14
docs(0004): add fulfillment data to lpi

justmoon:
Implements #314 in LPI.

### DIFF
--- a/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
+++ b/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
@@ -1,6 +1,6 @@
 ---
 title: The Javascript Ledger Plugin Interface
-draft: 8
+draft: 9
 ---
 # Javascript Ledger Plugin Interface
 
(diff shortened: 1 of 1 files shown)


michielbdejong:
I think we should create a new payment type for when response data is required. In fact, you could even argue that if you call it a "payment", then proof-of-payment (the fulfillment) should be enough. Otherwise, you should call it the "sale" of content (the fulfillment data) which then may or may not be delivered.

justmoon:
Fulfillment data is not for selling content - it's for providing feedback about how the payment went. For example:

* When making a probing (quote) payment, it could tell you how much money arrived at the recipient.
* When making a streaming payment, it could tell you when the next iteration is due.
* When making a chunked payment, it could tell you how many more chunks you need to send and how you're doing against the total.
* When paying into an HTTP-ILP token, it could tell you what your new token balance is.

> I think we should create a new payment type for when response data is required.

If @emschwartz and I are correct and streaming payments are superior to larger discrete payments in almost every way, then basically all payments will use response data. At a minimum, we need to be able to test this hypothesis quickly. Since adding optional fulfillment data to the existing payments is a simple, non-breaking change, I prefer it over the more complicated change of adding a second type of payment.

michielbdejong:
Fair enough, although it does feel like a breaking change in disguise: passing back data is optional for each connector in the network, but as a sender you don't pick which connectors are on the path, so you can't control whether the data will actually make it back to you, unless you demand to be in a network where every single connector supports it. So I expect there would be two things that can happen: either everybody supports this optional feature, or failure rates will be so significant that nobody will want to rely on it in practice.

michielbdejong:
Ah, side note:

> When making a probing (quote) payment, it could tell you how much money arrived at the recipient.

I though the idea of a probing (quote) payment was that it gets rejected, not fulfilled?

justmoon:
> I though the idea of a probing (quote) payment was that it gets rejected, not fulfilled?

Could be either. If you've already gone through the trouble of putting all that money on hold and taking the exchange risk - may as well fulfill it at that point. So I think it's up to the transport layer protocol whether it wants to use rejected or fulfilled quotes.

Generally, I think that streaming/chunked transport layer protocols would fulfill it (if your payment is already in chunks, may as well use the quote as the first chunk) and locally atomic (receiver either gets all the money or none of it) transport layer protocols would reject it.

justmoon:
Alright, let's do it!

* [ ] interledgerjs/ilp-plugin-bells
* [x] interledgerjs/five-bells-ledger#436
* [x] interledgerjs/five-bells-shared#197
* [x] interledgerjs/ilp-plugin-payment-channel-framework#55
* [ ] ripple/ilp-plugin-xrp-escrow
* [ ] ripple/ilp-plugin-xrp-paychan
* [x] interledgerjs/ilp-connector#396
* [x] interledgerjs/ilp-connector-shard#3
* [ ] interledgerjs/ilp-plugin-lightning
* [ ] interledgerjs/ilp-plugin-mini-accounts
* [ ] interledgerjs/ilp-plugin-bitcoin-paychan
* [ ] interledgerjs/ilp-plugin-ethereum
* [ ] interledgerjs/ilp-plugin-chain

PR #341, opened 2017-11-15
fix #328: Transfer#to excludes interactions

### LINKED ISSUE #328: clarify relation between ILP addresses and on-ledger address in LPI
The [description of `Transfer#to` in the LPI](https://github.com/interledger/rfcs/blob/master/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md#transferto) states that it needs to be "the ILP address" of the credit account. But that can be interpreted in two ways: If you pay to `g.ledger.account.some.other.identifiers`, then should this field contain that full string or just `g.ledger.account`?

I would argue for the latter, since the sender already needs to know what the ledger prefix is (it's one of the fields as well, and also you need to know which level the `amount` applies to). The transfer object instructs the ledger about which on-ledger account it goes to, and the other identifiers are only needed in the packet that describe the payment to the receiver.

michielbdejong:


sharafian:
I think that adding interactions after the `to` field is permitted, but they will not be included in any `to`/`from` fields emitted by LPI events

justmoon:
How would you know where the interactions start? The addresses are supposed to be opaque to everyone except their owner (for a `destination` that would be the receiver.)

michielbdejong:
> How would you know where the interactions start?

Right after the account identifier, which starts right after the ledger prefix, which is determined by the (longest) prefix for which you have a liquidity curve (whether obtained from a broadcast or from a quote `response.appliesToPrefix`, that doesn't matter).

Suppose you have a liquidity curve for `g.one.` and a destination address of `g.one.two.three.four`. As I always understood it, that means `g.one.` is the ledger prefix, `two` is the account identifier and `.three.four` are the interactions.

justmoon:
2016 me would respond: What if `g.one.two.three.` is a subledger? Users are supposed to be able to subdivide their accounts if they like.

2017 me thinks: Post-enlightenment I think we should get rid of the concept of a ledger prefix entirely. Connectors should just forward based on the longest matching prefix in their routing table. They should not make a futile attempt to guess whether they are the last ones to forward the payment or not. They should not assume anything about the structure of the ILP address.

I think that my distinction between [forward and deliver](https://github.com/interledger/rfcs/issues/77) was a big mistake. Without it, connectors are simpler (no more `isLocal`) and user accounts are more powerful (anyone can easily create subledgers in any currency.)

michielbdejong:
>  What if `g.one.two.three.` is a subledger?

Yeah, in the LPI, subledgers are exactly the type of thing (apparently called "interactions" in the ilp-addresses rfc) that will be allowed in the `to` address in a `sendTransfer` or `sendRequest` call, but that will be omitted in the events that get emitted.

One thing to be aware of is that that subledger needs to use the same unit of value as the ledger for which the liquidity curve exists, otherwise its unit of value would be over-defined.

> I think we should get rid of the concept of a ledger prefix

I don't think the interledger enlightenment and the recent resolution of the forward-or-deliver confusion are related to this, but I do see your point in relation to Evan's work on amount-less packets; if there is no amount in the packet, then nobody has to wonder what the unit of value is of the destination ledger. Is that what you mean by 'get rid of the concept of a ledger prefix'? That would make total sense to me, then.

But ILP addresses are used in two ways in the LPI: for referring to the the destination, and for referring to `to` and `from` accounts on the *current* ledger (the ledger to which this LPI instance gives access). Do you think ledgers should be abstracted as having nested accounts as well? Wouldn't that make the abstraction unnecessarily complex to think about, if you're just talking about a specific ledger?

justmoon:
> But ILP addresses are used in two ways in the LPI: for referring to the destination, and for referring to to and from accounts on the current ledger 

`to` and `from` don't exist post-enlightenment. If the ledger is a connector, then it routes based on `ilp.destination`, it doesn't make sense to have `to` or `from` fields if ledgers are connectors.

Most plugins would work like trust lines or payment channels - these already represent a relationship between connectors.

In some cases, you may have a ledger that is fast enough to do ILP transactions, but does not have an ILP connector built-in, e.g. something like five-bells-ledger. In that case, you can simulate a connector in the plugin. If the ILP destination address starts with a ledger prefix that the plugin is configured with, it takes the first segment after the plugin and forwards the payment to that account. Otherwise, it forwards the payment to the default connector on that ledger. (`getInfo().connectors[0]`)

> One thing to be aware of is that that subledger needs to use the same unit of value of the ledger to which this LPI instance gives access, otherwise its unit of value would be over-defined.

True with Current Interledger, however with Best-effort Interledger subledgers can have different units.

michielbdejong:
> to and from don't exist post-enlightenment.

That's irrelevant here, this PR is just trying to clarify the use and meaning of the current LPI's `to` field, which does exist, right? So what should we do with that, leave it ambiguous as it is now? Or fast-track its deprecation?

michielbdejong:
@justmoon What should we do with this, leave it ambiguous as it is now? Or leave it broken until lpi1 is fully deprecated?

justmoon:
Would be good to leave this in a consistent, non-broken state.

I agree with @sharafian's comment above. We could say that `to` and `from` MUST start with the ledger prefix and that the plugin may decide how many segments after that it will consider, usually one. When emitting events, the plugin MUST include one or more segments after the ledger prefix in the `to` and `from` fields, usually one.

Note that a later draft of the ILP address spec may be changed to better match a post-enlightenment viewpoint, so I'd avoid referring to specific sections in that spec.

michielbdejong:
Deprecated

Issue #340, opened 2017-11-15
forwarded payments and the diner's dilemma


michielbdejong:
When deciding on what transaction fee to charge, a connector will take into consideration how much money it wants to earn from each transaction (an argument for charging higher fees), as well as how much business it expects to attract (an argument for charging lower fees).

Attracting business can work in two ways:
* in the short run, give a competitive quote and have your peer believe that you generally succeed at delivering payments at whatever price you quote
* in the long run, have your peer believe that not only do you generally succeed in delivering a payment, you also generally charge a low fee

In cases where the sender send a little bit more than strictly necessary, both are more complicated when using forwarded payments (without explicit destination amount in the packet) than when using exact payments (with explicit destination amount in the packet). This is due to the [diner's dilemma](https://en.wikipedia.org/wiki/Unscrupulous_diner%27s_dilemma).

The analogy of ordering more expensive food if the bill is shared in the diner's dilemma is that of charging a higher fee if not you as a connector, but all connectors as a group, are made responsible for how much arrives at the destination for a given source amount.

The analogy is not 100% because in the diner's dilemma, everybody can see what the others are ordering, and in forwarded payments, all you see is that one person ordered the lobster, not who it was.

emschwartz:
Do you still think it would play out this way if:

1. The game is played _many_ times
2. The value of each one is low
3. It's not a group of friends going out to dinner but instead each person is choosing to bring along one other friend, who chooses one of their friends, and so on (so it's not about blaming the person who ordered the lobster but just about blaming the friend that you brought)
4. There are multiple paths the payment could take (if we don't have options connectors don't need to do anything tricky to raise their prices)
5. We adjust who we bring to each successive meal based on the results of all the previous ones?

michielbdejong:
Each time I forward a payment, I can choose who I forward it onto, and how much commission I charge myself. The information I get back is whether that payment succeeded or failed. If a payment succeeded, I think, I was too careful, I could have charged more. If it failed, I can seek one of four explanations:
1. someone before me charged too much
2. I charged too much
3. the person directly after me charged too much, but will correct this next time
4. the person directly after me charged too much, and will not correct this next time
5. someone after that next person charged too much
6. the receiver was never going to fulfill this payment anyway

In cases 1., 3., and 6., there was nothing I could do anyway.
In case 5., someone after me should adjust their routing.
In case 2. I should adjust my pricing. 
Only in case 4. I have to change where I forward payments to next time.

[...]

Issue #333, opened 2017-11-10
when selling apples, embrace rejection


michielbdejong:
One day, I found myself selling 1000 apples at the market. I deposited the apples at the auctioneer's desk, asked around to find buyers and get quotes, and when I got the best quote, I told the auctioneer to give the apples to that buyer at the agreed price. That night, I dreamed about ILQP.

The next day, it occurs to me that I had better sell the apples one by one. So I decide a minimum price per apple, tell this (in secret) to the auctioneer, and advertise the quality of my apples to potential buyers. They put in their bids, and if their bid is high enough, they can buy one apple. The buyers soon use binary search to find out which minimum price I told the auctioneer, and the rest of my apples all sell at that price within a few minutes. That night, I dreamed about ilp3.

The third day, it occurs to me that telling the auctioneer a minimum price was not ideal, because my apples sold quickly at the minimum price I had set. So this day, I tell the auctioneer to sell my apples to the highest bidders throughout the day. So the auctioneer now sells one apple per minute, only to the highest bidder that presented themselves during that minute. Each minute, buyers adjust their bid to what they think the others are bidding. That night, I dreamed about [FPSBAs](https://en.wikipedia.org/wiki/First-price_sealed-bid_auction).

The fourth day, I decide I don't want to stay at the market all day, if that's not necessary, so I tell the auctioneer to offer bulk discounts for people who are willing to buy more than one apple per minute (e.g. 1% off if you buy 10 at a time, 2% off if you buy 100 at a time, 3% off if you buy the whole lot). That night, in my dreams I turned into an sender; my apples converted into prepared source transfers, the auctioneer became a receiver, the buyers turned out to be connectors, and the price per apple was the destination amount. The Interledger network in my dream had a lot of rejected transfers; in fact, just like the auctioneer only accepted winning bids from the apple buyers, the receiver only accepted very few of the transfers that reached it - the vast majority of transfers were rejected. Everybody embraced the rejection of transfers as just a way to do auctions, and lived happily ever after.

sappenin:
Amazing!

emschwartz:
Can this issue be closed?

PR #343, opened 2017-11-17
ILP packet type for payment chunks

michielbdejong:
Obsoletes the amount-less ("best effort") payments proposed in #323.

Solves the problems identified in #340, using the ideas from #333.

By adding a 'chunk size', which the receiver can use to compare the chunks as they arrive. The receiver sees with what transfer amount chunks come in, and only accepts the most profitable ones; the rest is rejected. The ratio between chunks that are accepted and rejected by the receiver is prescribed by the sender in `targetChunksAcceptedPerHour`.

### DIFF
--- a/asn1/InterledgerPacket.asn
+++ b/asn1/InterledgerPacket.asn
@@ -9,6 +9,7 @@ IMPORTS
     FROM GenericTypes
 
     InterledgerProtocolPaymentData,
+    PaymentChunkData,
     InterledgerProtocolErrorData
     FROM InterledgerProtocol
 
@@ -34,7 +35,8 @@ PacketSet PACKET ::= {
     {5 QuoteBySourceAmountResponseData} |
     {6 QuoteByDestinationAmountRequestData} |
     {7 QuoteByDestinationAmountResponseData} |
-    {8 InterledgerProtocolErrorData}
+    {8 InterledgerProtocolErrorData} |
+    {10 PaymentChunkData}
 }
 
 InterledgerPacket ::= SEQUENCE {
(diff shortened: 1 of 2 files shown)


adrianhopebailie:
> Solves the problems identified in #340, using the ideas from #333.

This assumes that we all agree that the problems in #340 are actually problems that will occur in the real world and not just theoretical.

Until this is proven I don't think there is any justification in adding this complexity to the core protocol. Chunking etc are not core to ILP, they can be implemented at the transport layer.

michielbdejong:
> This assumes that we all agree that the problems in #340 are actually problems

Sure, so far I seem to be the only person who thinks there is anything wrong with the proposal from #312 which it tries to compete with. :)

michielbdejong:
Right, the `targetChunksAcceptedPerHour` and `paymentId` can go inside `packet.data`. But do you see why the chunk size *is* needed? Or did I really fail miserably at explaining the concept of the cigar box size in https://github.com/interledger/rfcs/issues/340#issuecomment-345220420?

adrianhopebailie:
> Right, the targetChunksAcceptedPerHour can go inside packet.data. But do you see why the chunkSize and paymentId is needed? Or did I really fail miserably at explaining the concept of the cigar box size in #340 (comment)?

You are solving this at the wrong layer IMO.

The assumptions you make in your cigar box example don't hold for the majority of real-world payments systems today.

ILP should be dead-simple. It should not concern itself with anything that is influenced by economic incentives as these are inherently use-case specific. Rather those should be dealt with in the transport or application layer.

emschwartz:
Can this be closed? It's possible that something like this could go in a transport layer protocol, but it doesn't belong in the interoperability layer

PR #347, opened 2017-12-02
Ledger Plugin Interface v2

### LINKED ISSUE #302: Clarify how to recognize different error types in from LPI
The way ilp-connector knows that a plugin throws one of https://github.com/interledger/rfcs/blob/master/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md#errors is not by using `instanceof` because that would require shared JavaScript code between the plugin and the code that uses the plugin, or it would require the plugin to expose these error classes as constants. Instead, it [looks at `err.name`](https://github.com/interledgerjs/ilp-connector/blob/0697d37b65d2fcaa6aac42eb93871d96ae173530/src/lib/executeSourceTransfer.js#L75-L79).

I think it may be clear from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/name that an error object should have a `.name` field with its name, but if you look at for instance https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch and also how it's usually done in Java, C++, C#, etc. I think it doesn't hurt to add a sentence 'when catching an error, look for the `.name` field, which will then be the string from the left column in the table (case-sensitive)'

### LINKED ISSUE #332: RFC-4: Change RejectionMessage to match ILP (RFC-3)
There are a couple of issues with the RFC-4 [RejectionMessage](https://github.com/interledger/rfcs/blob/75d819d5947c6cee94510b7b1cd009e1627fc748/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md#class-rejectionmessage):

* `forwarded_by` should be an array
* `message` does not exist in the ILP packet and should presumably move to `additional_info`

justmoon:
*Please note the changes below are in roughly chronological order, later changes may revise earlier ones.*

### Ledgers are a type of connector (Interledger enlightenment)
@sharafian and I realized a while ago that ILP could be simplified by treating ledgers as another type of connector. This later came to be known as "Interledger enlightenment".

Plugins that represent ledgers have to do a simple form of routing. If the ILP destination address starts with the ledger prefix, they should forward the transfer to the account mentioned in the destination address (usually the first segment after the ledger prefix). If it doesn't, they should forward the payment to the default connector. This new behavior no longer allows a connector to reach multiple other connectors through one plugin. If that is desired, the connector should create a separate plugin per peer connector.

- Remove `connectors`: All payments to a plugin go through the default connector if the ILP destination is not on the ledger
- Remove `account`/`from`/`to`: It's the plugin's responsibility to determine the destination
- Remove `prefix`: Ledger prefixes are now handled internally by the plugin

### Ephemeral transfers
Fulfillments and rejections are simply returned by sendTransfer calls. Now that we envision Interledger transfers to always be relatively small, we can slightly relax some of the guarantees around the individual transfer. Based on work by @michielbdejong and @emschwartz.
- Change `sendTransfer` to return `Promise<FulfillmentInfo>` or reject with `InterledgerRejectionError`
- Add `registerTransferHandler` and `deregisterTransferHandler` (if no handler is registered, plugin should reject all incoming transfers)
- Remove `noteToSelf`: We can now use local state in the sendTransfer caller function
- Remove `fulfillCondition`: Instead, return FulfillmentInfo
- Remove `rejectIncomingTransfer`: Instead, throw an InterledgerRejectionError
- Remove `getFulfillment`: With short-lived transfers, the need to retrieve a fulfillment for the purpose of retrying later is no longer relevant

### LPI is not a different layer than ILP
The ILP packet has become so simple that it no longer makes sense to encapsulate it with the transfer object. Based on #270 by @adrianhopebailie and further work by @emschwartz.
- Remove `ilp` from Transfer, FulfillmentInfo
- Add `destination:String`, `data:Buffer` to Transfer
- Add `data:Buffer` to FulfillmentInfo

### No optimistic mode
As suggested by @emschwartz, optimistic mode can be simulated using Optimistic-over-Universal (OoU).
- Remove `*_transfer` events
- Make `executionCondition` and `expiresAt` mandatory for transfers

### Middleware provides better extensibility
As discovered by @emschwartz, many use cases of plugins can be better represented using middleware.
- Remove `*_{prepare,fulfill,reject,cancel}` events: These can be provided by a middleware that emits such events, if desired
- Remove `*_{request,response}` events

### Quote by probing
Instead of a dedicated quoting protocol, we quote by sending a probe payment. See #309.
- Remove `sendRequest`: We no longer need a dedicated request/response feature

### Bandwidth instead of balances
@sharafian made the point that ILP participants only care about their balances insofar as they affect their ability to send payments. In fact, balances are just a special case of a function that takes a payment and returns whether the payment is allowed or not. We can model this more generically by treating it similarly to IP bandwidth. IP doesn't track how much bandwidth is available but instead returns an error when packets are coming through too quickly.
- Remove `getBalance`
- Remove `minBalance`/`maxBalance`

### Cleanup LPI errors
ILP/LPI errors are a bit of a mess right now.

- Define an `InterledgerRejectionError` to be a proper JavaScript error object
- Change `forwardedBy` to be an array to match the ILP packet
- Make `message` a local property since it does not exist in the ILP error packet and can therefore not be forwarded

Fixes #332.

### ILDCP replaces LedgerInfo
It's not a plugin concern to configure the client. This can be done via messages sent on the Interledger layer.

- Remove `getInfo`, `getAccount` and `LedgerInfo`
- Remove `info_change` event

### Conditional transfers are an Interledger concern
See #359. We don't need to enforce conditions on the ledger layer. This can be done on the Interledger layer.

- Remove all mentions of `executionCondition`, `expiresAt`
- Replace `transfer: Transfer` with `amount: string`
- Replace `sendTransfer(transfer: Transfer): Promise<FulfillmentInfo>` with `sendMoney(amount: string): Promise<void>`
- Replace `sendRequest(message: Message): Promise<Message>` with `sendData(data: Buffer): Promise<Buffer>`
- Replace `[de]registerRequestHandler` and `[de]registerTransferHandler` with `[de]registerDataHandler` and `[de]registerMoneyHandler`

### Interledger rejections should be emitted on the Interledger layer only
The plugin previously emitted some Interledger rejections. The plugin shouldn't even know what Interledger is. All errors emitted by the plugin are now plain JavaScript exceptions.

- Remove `InterledgerRejectionError`

### Other changes
- Add `lpiVersion = 2` static variable
- Change most optional fields to required fields to reduce ambiguity
- Clarify that errors should be distinguished by `name` (fixes #302)

### DIFF
--- a/0000-ledger-plugin-interface-2/0000-ledger-plugin-interface-2.md
+++ b/0000-ledger-plugin-interface-2/0000-ledger-plugin-interface-2.md
@@ -31,7 +31,7 @@ This spec depends on the [ILP spec](../0003-interledger-protocol/).
 
 | | Name |
 |:--|:--|
-| `static` | [**lpiVersion**](#version) = 2 |
+| `static` | [**version**](#ledgerpluginversion) = 2 |
 
 ###### Events
 | Name | Handler |
@@ -39,7 +39,6 @@ This spec depends on the [ILP spec](../0003-interledger-protocol/).
 | [**connect**](#event-connect) | `( ) ⇒` |
 | [**disconnect**](#event-disconnect) | `( ) ⇒` |
 | [**error**](#event-error) | `( ) ⇒` |
-| [**info_change**](#event-info_change) | <code>( info:[LedgerInfo](#class-ledgerinfo) ) ⇒</code> |
 
 ###### Errors
 | Name | Description |
@@ -89,6 +88,11 @@ const ledgerPlugin = new LedgerPlugin({
 ```
 
 For a detailed description of these properties, please see [`PluginOptions`](#class-pluginoptions).
+
+#### LedgerPlugin.version
+<code>LedgerPlugin.**version**:Number</code>
+
+Always `2` for this version of the Ledger Plugin Interface.
 
 ### Connection Management
 
@@ -181,7 +185,7 @@ p.sendTransfer({
 For a detailed description of these properties, please see [`Class: Transfer`](#class-transfer).
 
 #### LedgerPlugin#registerTransferHandler
-<code>ledgerPlugin.registerTransferHandler( **transferHandler**: ( transfer: [Transfer](#class-transfer) ) ⇒ Promise&lt;[Message](#class-message)> ) ⇒ null</code>
+<code>ledgerPlugin.registerTransferHandler( **transferHandler**: ( transfer: [Transfer](#class-transfer) ) ⇒ Promise&lt;[FulfillmentInfo](#class-fulfillmentinfo)> ) ⇒ null</code>
 
 Set the callback which is used to handle incoming prepared transfers. The callback should expect one parameter (the [Transfer](#class-transfer)) and return a promise for the resulting [FulfillmentInfo](#class-fulfillmentinfo). If the transfer is rejected or an error occurs, the callback should reject the transfer. In general, the callback should behave as [`sendTransfer`](#ledgerpluginsendtransfer) does.
 
@@ -195,15 +199,6 @@ If an incoming transfer is received by the plugin, but no handler is registered,
 Removes the currently used transfer handler. This has the same effect as if [`registerTransferHandler`](#ledgerpluginregistertransferhandler) had never been called.
 
 If no transfer handler is currently set, this method does nothing.
-
-### Event: `info_change`
-<code style="">ledgerPlugin.on('info_change',
-  (
-    **info**:[LedgerInfo](#class-ledgerinfo)
-  ) ⇒
-)</code>
-
-Emitted any time the plugin's `LedgerInfo` cache changes.
 
 ## Class: Transfer
 <code>class Transfer</code>
@@ -319,8 +314,6 @@ A cryptographic fulfillment that is the SHA-256 preimage of the hash provided as
 
 Fulfillments are base64url-encoded values with a length of exactly 32 bytes.
 
-Ledger plugins that do not support holds MUST reject with an `HoldsNotSupportedError` if this parameter is provided.
-
 #### FulfillmentInfo#ilp
 <code>**ilp**:Buffer</code>
 
@@ -331,7 +324,9 @@ An [ILP packet](https://interledger.org/rfcs/0003-interledger-protocol/draft-4.h
 
 Ledger plugins MAY use this object to accept and/or set additional fields for other features they support. The object MUST be serializable, i.e. only plain JSON types are allowed anywhere in the object or sub-objects.
 
-Note that connectors MAY forward some fields of `custom` data from plugin to plugin, but generally are not expected to. All `custom` fields that were passed to `sendTransfer` MUST be passed to the transfer handler by the plugin on the receiving side. The only exception are properties which start with the underscore character (`_`), which MAY be consumed by the plugin and not passed on.
+Note that connectors MAY forward some fields of `custom` data from plugin to plugin, but generally are not expected to. All `custom` fields that were set in `FulfillmentInfo#custom` on the receiving side MUST be set in `FulfillmentInfo#custom` when it is resolved by the promise on the sending side. The only exception are properties which start with the underscore character (`_`), which MAY be consumed by the plugin and not passed on.
+
+All custom fields that were set in FulfillmentInfo#custom on the receiving side MUST be set in FulfillmentInfo#custom when it is resolved by the promise on the sending side.
 
 ###### Example
 ``` js
--- a/0000-ledger-plugin-interface-2/0000-ledger-plugin-interface-2.md
+++ b/0000-ledger-plugin-interface-2/0000-ledger-plugin-interface-2.md
@@ -175,7 +175,7 @@ This method MAY reject with any arbitrary JavaScript error.
 ```js
 p.sendTransfer({
   amount: '10',
-  executionCondition: '47DEQpj8HBSa-_TImW-5JCeuQeRkm5NMpJWZG3hSuFU',
+  executionCondition: Buffer.from('47DEQpj8HBSa-_TImW-5JCeuQeRkm5NMpJWZG3hSuFU', 'base64'),
   expiresAt: '2016-05-18T12:00:00.000Z',
   ilp: Buffer.alloc(0),
   custom: {}
@@ -210,7 +210,7 @@ The `Transfer` class is used to describe transfers from the originator of the se
 |:--|:--|:--|
 | `String` | [amount](#transferamount) | Integer transfer amount, in the ledger's base unit |
 | `Buffer` | [ilp](#transferilp) | Attached data (ILP packet) |
-| `String` | [executionCondition](#transferexecutioncondition) | Cryptographic hold condition |
+| `Buffer` | [executionCondition](#transferexecutioncondition) | Cryptographic hold condition |
 | `String` | [expiresAt](#transferexpiresat) | Expiry time of the cryptographic hold |
 | `Object` | [custom](#transfercustom) | Object containing ledger plugin specific options |
 
@@ -218,7 +218,7 @@ The `Transfer` class is used to describe transfers from the originator of the se
 ``` js
 {
   amount: '100',
-  executionCondition: 'I3TZF5S3n0-07JWH0s8ArsxPmVP6s-0d0SqxR6C3Ifk',
+  executionCondition: Buffer.from('I3TZF5S3n0-07JWH0s8ArsxPmVP6s-0d0SqxR6C3Ifk', 'base64'),
   expiresAt: '2017-12-02T11:51:26.627Z',
   ilp: Buffer.alloc(0),
   custom: {
@@ -249,16 +249,16 @@ An [ILP packet](https://interledger.org/rfcs/0003-interledger-protocol/draft-4.h
 If the `ilp` data is too large, the ledger plugin MUST reject with a `MaximumIlpDataSizeExceededError`.
 
 #### Transfer#executionCondition
-<code>**executionCondition**:String</code>
+<code>**executionCondition**:Buffer</code>
 
 A cryptographic challenge used for implementing holds. The underlying ledger MUST hold the transfer until the condition has been fulfilled or the `expiresAt` time has been reached.
 
-Conditions are the base64url-encoded SHA-256 hash of a random or pseudo-random 32-byte preimage called the fulfillment.
+The `executionCondition` is a Node.js `Buffer` containing the 32-byte SHA-256 hash of a random or pseudo-random 32-byte preimage called the `fulfillment`.
 
 ###### Example
 
 ``` js
-'I3TZF5S3n0-07JWH0s8ArsxPmVP6s-0d0SqxR6C3Ifk'
+Buffer.from('I3TZF5S3n0-07JWH0s8ArsxPmVP6s-0d0SqxR6C3Ifk', 'base64')
 ```
 
 #### Transfer#expiresAt
@@ -308,11 +308,11 @@ The `FulfillmentInfo` class is used to describe the fulfillment and associated d
 ### Fields
 
 #### FulfillmentInfo#fulfillment
-<code>**fulfillment**:String</code>
+<code>**fulfillment**:Buffer</code>
 
 A cryptographic fulfillment that is the SHA-256 preimage of the hash provided as the [`executionCondition`](#transferexecutioncondition) when the transfer was first prepared.
 
-Fulfillments are base64url-encoded values with a length of exactly 32 bytes.
+The `fulfillment` is a Node.js `Buffer` with a length of exactly 32 bytes.
 
 #### FulfillmentInfo#ilp
 <code>**ilp**:Buffer</code>
--- a/0000-ledger-plugin-interface-2/0000-ledger-plugin-interface-2.md
+++ b/0000-ledger-plugin-interface-2/0000-ledger-plugin-interface-2.md
@@ -19,15 +19,15 @@ To send ILP payments through a new ledger, one must implement a ledger plugin th
 | | Name |
 |:--|:--|
 | `new` | [**LedgerPlugin**](#new-ledgerplugin) ( opts ) |
-| | [**connect**](#ledgerpluginconnect) ( options ) `⇒ Promise.<null>` |
-| | [**disconnect**](#ledgerplugindisconnect) ( ) `⇒ Promise.<null>` |
+| | [**connect**](#ledgerpluginconnect) ( options ) `⇒ Promise.<undefined>` |
+| | [**disconnect**](#ledgerplugindisconnect) ( ) `⇒ Promise.<undefined>` |
 | | [**isConnected**](#ledgerpluginisconnected) ( ) `⇒ Boolean` |
 | | [**sendData**](#ledgerpluginsenddata) ( data ) <code>⇒ Promise.&lt;Buffer></code> |
 | | [**sendMoney**](#ledgerpluginsendmoney) ( amount ) <code>⇒ Promise.&lt;Buffer></code> |
-| | [**registerDataHandler**](#ledgerpluginregisterdatahandler) ( dataHandler ) <code>⇒ null</code> |
-| | [**deregisterDataHandler**](#ledgerpluginderegisterdatahandler) ( ) <code>⇒ null</code> |
-| | [**registerMoneyHandler**](#ledgerpluginregistermoneyhandler) ( moneyHandler ) <code>⇒ null</code> |
-| | [**deregisterMoneyHandler**](#ledgerpluginderegistermoneyhandler) ( ) <code>⇒ null</code> |
+| | [**registerDataHandler**](#ledgerpluginregisterdatahandler) ( dataHandler ) <code>⇒ undefined</code> |
+| | [**deregisterDataHandler**](#ledgerpluginderegisterdatahandler) ( ) <code>⇒ undefined</code> |
+| | [**registerMoneyHandler**](#ledgerpluginregistermoneyhandler) ( moneyHandler ) <code>⇒ undefined</code> |
+| | [**deregisterMoneyHandler**](#ledgerpluginderegistermoneyhandler) ( ) <code>⇒ undefined</code> |
 
 ###### Constants
 
@@ -67,7 +67,7 @@ const ledgerPlugin = new LedgerPlugin({
     // (e.g. when the ledger has reduced memo capability and we can only put an ID in the memo)
     // Store a value under a key
     put: (key, value) => {
-      // Returns Promise.<null>
+      // Returns Promise.<undefined>
     },
     // Fetch a value by key
     get: (key) => {
@@ -75,7 +75,7 @@ const ledgerPlugin = new LedgerPlugin({
     },
     // Delete a value by key
     del: (key) => {
-      // Returns Promise.<null>
+      // Returns Promise.<undefined>
     }
   }
 })
@@ -91,7 +91,7 @@ Always `2` for this version of the Ledger Plugin Interface.
 ### Connection Management
 
 #### LedgerPlugin#connect
-<code>ledgerPlugin.connect( options:[ConnectOptions](#class-connectoptions ) ⇒ Promise.&lt;null></code>
+<code>ledgerPlugin.connect( options:[ConnectOptions](#class-connectoptions ) ⇒ Promise.&lt;undefined></code>
 
 `options` is optional.
 
@@ -101,7 +101,7 @@ Rejects with `InvalidFieldsError` if credentials are missing, and `NotAcceptedEr
 Rejects with `TypeError` if `options.timeout` is passed but is not a `Number`.
 
 #### LedgerPlugin#disconnect
-<code>ledgerPlugin.disconnect() ⇒ Promise.&lt;null></code>
+<code>ledgerPlugin.disconnect() ⇒ Promise.&lt;undefined></code>
 
 Unsubscribe from ledger events.
 
@@ -148,7 +148,7 @@ const responseBuffer = await p.sendData(requestBuffer)
 ```
 
 #### LedgerPlugin#sendMoney
-<code>ledgerPlugin.sendMoney( **amount**:string ) ⇒ Promise.&lt;null></code>
+<code>ledgerPlugin.sendMoney( **amount**:string ) ⇒ Promise.&lt;undefined></code>
 
 Transfer `amount` units of money from the caller to the counterparty of the account.
 
@@ -157,7 +157,7 @@ All plugins MUST support amounts in a range from one to some maximum.
 ### Receiving
 
 #### LedgerPlugin#registerDataHandler
-<code>ledgerPlugin.registerDataHandler( **dataHandler**: ( data: Buffer ) ⇒ Promise&lt;Buffer> ) ⇒ null</code>
+<code>ledgerPlugin.registerDataHandler( **dataHandler**: ( data: Buffer ) ⇒ Promise&lt;Buffer> ) ⇒ undefined</code>
 
 Set the callback which is used to handle incoming prepared data packets. The callback should expect one parameter (the data as a Buffer)) and return a promise for the resulting response data packet (as a Buffer.) If an error occurs, the callback MAY throw an exception. In general, the callback should behave as [`sendData`](#ledgerpluginsenddata) does.
 
@@ -166,14 +166,14 @@ If a data handler is already set, this method throws a `DataHandlerAlreadyRegist
 If an incoming packet is received by the plugin, but no handler is registered, the plugin SHOULD respond with an error.
 
 #### LedgerPlugin#deregisterDataHandler
-<code>ledgerPlugin.deregisterDataHandler( ) ⇒ null</code>
+<code>ledgerPlugin.deregisterDataHandler( ) ⇒ undefined</code>
 
 Removes the currently used data handler. This has the same effect as if [`registerDataHandler`](#ledgerpluginregisterdatahandler) had never been called.
 
 If no data handler is currently set, this method does nothing.
 
 #### LedgerPlugin#registerMoneyHandler
-<code>ledgerPlugin.registerMoneyHandler( **moneyHandler**: ( amount: string ) ⇒ Promise&lt;null> ) ⇒ null</code>
+<code>ledgerPlugin.registerMoneyHandler( **moneyHandler**: ( amount: string ) ⇒ Promise&lt;undefined> ) ⇒ undefined</code>
 
 Set the callback which is used to handle incoming money. The callback should expect one parameter (the amount) and return a promise. If an error occurs, the callback MAY throw an exception. In general, the callback should behave as [`sendMoney`](#ledgerpluginsendmoney) does.
 
@@ -182,7 +182,7 @@ If a money handler is already set, this method throws a `MoneyHandlerAlreadyRegi
 If incoming money is received by the plugin, but no handler is registered, the plugin SHOULD return an error (and MAY return the money.)
 
 #### LedgerPlugin#deregisterMoneyHandler
-<code>ledgerPlugin.deregisterMoneyHandler( ) ⇒ null</code>
+<code>ledgerPlugin.deregisterMoneyHandler( ) ⇒ undefined</code>
 
 Removes the currently used money handler. This has the same effect as if [`registerMoneyHandler`](#ledgerpluginregistermoneyhandler) had never been called.
 
@@ -217,7 +217,7 @@ Method names are based on the popular LevelUP/LevelDOWN packages.
 {
   // Store a value under a key
   put: (key, value) => {
-    // Returns Promise.<null>
+    // Returns Promise.<undefined>
   },
   // Fetch a value by key
   get: (key) => {
@@ -225,7 +225,7 @@ Method names are based on the popular LevelUP/LevelDOWN packages.
   },
   // Delete a value by key
   del: (key) => {
-    // Returns Promise.<null>
+    // Returns Promise.<undefined>
   }
 }
 ```
--- a/0000-ledger-plugin-interface-2/0000-ledger-plugin-interface-2.md
+++ b/0000-ledger-plugin-interface-2/0000-ledger-plugin-interface-2.md
@@ -147,7 +147,7 @@ General event for fatal exceptions. Emitted when the plugin experienced an unexp
 #### LedgerPlugin#sendTransfer
 <code>ledgerPlugin.sendTransfer( **transfer**:[Transfer](#class-transfer) ) ⇒ Promise.&lt;[FulfillmentInfo](#class-fulfillmentinfo)></code>
 
-Initiates an account-local transfer. A transfer MUST contain an `amount` of zero or more and MAY have attached `data`. See the description of the [Transfer](#class-transfer) class below.
+Initiates an account-local transfer. A transfer MUST contain an `amount` of zero or more and MAY have attached `ilp` data. See the description of the [Transfer](#class-transfer) class below.
 
 All plugins MUST support amounts in a range from zero to some maximum. Plugins MAY implement zero-amount transfers differently than other transfers.
 
@@ -156,7 +156,7 @@ All plugins MUST support amounts in a range from zero to some maximum. Plugins M
 |:--|:--|:--|
 | transfer | <code>[Transfer](#class-transfer)</code> | Properties of the transfer to be created |
 
-When sending transfers, the [amount](#transferamount), [destination](#transferdestination), [executionCondition](#transferexecutioncondition) and [expiresAt](#transferexpiresat) fields are required. [data](#transferdata) and [custom](#transfercusto) MAY be left undefined.
+When sending transfers, all fields are required and MUST be provided, even if they contain an empty value, such as an empty object in `custom`.
 
 ###### Returns
 **`Promise.<FulfillmentInfo>`** A promise which resolves when the transfer has been fulfilled/executed.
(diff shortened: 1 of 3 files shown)


justmoon:
Created a module that can turn an LPI1 plugin into an LPI2 plugin, see [interledgerjs/ilp-compat-plugin](https://github.com/interledgerjs/ilp-compat-plugin).

justmoon:
> LPI is not a different layer than ILP

Having some second thoughts about this one. It seems like we're losing extensibility in a pretty major way. In ILPv1, we can add things to the ILP packet which some connectors understand in such a way that it is ignored by the receiver and other connectors that don't support it.

For example, in ILPv1 we could have added a hop limit in a non-breaking way. With this, we can't because there is no way to attach ILP layer information beyond the two natively supported fields (`destination` and `data`).

That's fine if we think we never need to extend ILP. But that's a hell of a thing to say.

michielbdejong:
Why not use http calls instead of JavaScript function calls?
I think previously the answer was "because fulfillments travel in the opposite direction and so we need to listen for events".

justmoon:
> Why not use http calls instead of JavaScript function calls?

What do you mean by HTTP calls? You mean the plugin should expose a local HTTP server and the connector would call it? That would add a lot of overhead for no (?) benefit?

xxxx-ledger-plugin-interface-2/xxxx-ledger-plugin-interface-2.md
```diff
@@ -0,0 +1,502 @@
...
+  expiresAt: '2016-05-18T12:00:00.000Z',
+  data: Buffer.alloc(0),
+  custom: {}
+})
+```
+
+For a detailed description of these properties, please see [`Class: Transfer`](#class-transfer).
+
+#### LedgerPlugin#registerTransferHandler
+<code>ledgerPlugin.registerTransferHandler( **transferHandler**: ( transfer: [Transfer](#class-transfer) ) ⇒ Promise&lt;[Message](#class-message)> ) ⇒ null</code>
```
adrianhopebailie: `Message` -> `FulfillmentInfo`

xxxx-ledger-plugin-interface-2/xxxx-ledger-plugin-interface-2.md
```diff
@@ -0,0 +1,502 @@
...
+
+This date MUST be in the future, otherwise the plugin MUST reject with a `UnacceptableExpiryError`.
+
+###### Example
+
+``` js
+'2017-12-02T11:51:26.627Z'
+```
+
+#### Transfer#custom
```
adrianhopebailie: It's not clear to me how this differs from `data` if the fields MUST be passed to the receiver.

My assumption is that the plugin will know what needs to go to the receiver and what doesn't based on the underlying connection type.

xxxx-ledger-plugin-interface-2/xxxx-ledger-plugin-interface-2.md
```diff
@@ -0,0 +1,502 @@
...
+Transport-layer data that travels with the fulfillment.
+
+#### FulfillmentInfo#custom
+<code>**custom**:Object</code>
+
+Ledger plugins MAY use this object to accept and/or set additional fields for other features they support. The object MUST be serializable, i.e. only plain JSON types are allowed anywhere in the object or sub-objects.
+
+If the `custom` data is too large, the ledger plugin MUST reject with a `MaximumCustomDataSizeExceededError`.
+
+Note that connectors MAY forward some fields of `custom` data from plugin to plugin, but generally are not expected to. All `custom` fields that were passed to `sendTransfer` MUST be passed to the transfer handler by the plugin on the receiving side. The only exception are properties which start with the underscore character (`_`), which MAY be consumed by the plugin and not passed on.
```
adrianhopebailie: Replace: "All `custom` fields that were passed to `sendTransfer` MUST be passed to the transfer handler by the plugin on the receiving side"

With: "All `custom` fields that were set in `FulfillmentInfo#custom` on the receiving side MUST be set in `FulfillmentInfo#custom` when it is resolved by the promise on the sending side."

adrianhopebailie:
> If the ILP destination address starts with the ledger prefix, they should forward the transfer to the account mentioned in the destination address (usually the first segment after the ledger prefix). If it doesn't, they should forward the payment to the default connector.

I think this could be restated as:

If the ILP destination address is the address of the receiver, they should forward the transfer to the receiver's account. If it isn't, the connector should attempt to route the payment via a new outgoing transfer in the direction of the destination ILP address.

adrianhopebailie:
p.s. I think the build fail is because you have used folder `xxxx-` and it's matching on `(\d\d\d\d)`

michielbdejong:
> the plugin should expose a local HTTP server and the connector would call it?

you can short-circuit that if it's within the same process. Instead of calling `plugin.listen(LOCALHOST_PORT)` and the connector then using `node-fetch` or superagent, the connector can call `plugin.handleRequest` directly

xxxx-ledger-plugin-interface-2/xxxx-ledger-plugin-interface-2.md
```diff
@@ -0,0 +1,502 @@
...
+
+Persistence MAY be required for internal use by some ledger plugins. For this purpose hosts MAY be configured with a persistence layer.
+
+Method names are based on the popular LevelUP/LevelDOWN packages.
+
+###### Example
+```js
+{
+  // Store a value under a key
+  put: (key, value) => {
```
emschwartz: Is the `value` assumed to be JSON? Could you store a `Buffer` instead?

xxxx-ledger-plugin-interface-2/xxxx-ledger-plugin-interface-2.md
```diff
@@ -0,0 +1,502 @@
...
+| Type | Name | Description |
+|:--|:--|:--|
+| `Number` | [timeout](#connectoptions-timeout) | milliseconds |
+
+### Fields
+
+#### ConnectOptions#timeout
+<code>**timeout**:Number</code>
+
+The number of milliseconds that the plugin should spend trying to connect before giving up.
```
emschwartz: Couldn't this just be handled by the caller (with a `Promise.race`)?

xxxx-ledger-plugin-interface-2/xxxx-ledger-plugin-interface-2.md
justmoon: > It's not clear to me how this differs from data

@emschwartz is correct.

`data` is passed on by connectors and plugins. `custom` is only passed on by plugins (generally speaking), but not by connectors. Although there are exceptions. (For example, I'm using `custom.ilpLegacyAmount` to pass on the ILP amount in my connector refactor.)

I'll see if I can improve the wording.

justmoon:
> p.s. I think the build fail is because you have used folder xxxx- and it's matching on (\d\d\d\d)

Yes, I'm aware. I'll assign an RFC number when we're closer to being ready to merge.

xxxx-ledger-plugin-interface-2/xxxx-ledger-plugin-interface-2.md
justmoon: Hmm, how would you tell the plugin to stop trying to connect? Call `disconnect`?

justmoon: Ironically, I just ran into an issue in `ilp` that's very similar to this. `ilp` uses `Promise.race` to timeout connecting, but `Promise.race` doesn't do anything to stop the thing that "lost" the race.

In my case what this did is that the connection timeout was keeping the tests from exiting. (Mocha 4 no longer force-exits unlike Mocha 3, so you have to clean everything up properly.)

I did a quick fix by setting a shorter timeout. But imo, the proper fix is to remove `Promise.race` and replace it with `setTimeout` and call `clearTimeout` once the connection completes successfully. If you really wanted to use promises, you could use bluebird and it's idea of cancellation, but I think that's actually less clean.

General lesson: When using `Promise.race`, be aware that the thing that loses the race may keep doing stuff that's no longer necessary and/or that you don't want it to be doing.

adrianhopebailie:
> Yes, I'm aware. I'll assign an RFC number when we're closer to being ready to merge.

What I meant is that you could use 0000- until you take a number

xxxx-ledger-plugin-interface-2/xxxx-ledger-plugin-interface-2.md
adrianhopebailie: I worry about having something that is intended for the next connector (local receiver) mixed in with something for the plugin. 

This implies the plugin must know how to differentiate between the two (and only send the data intended for the next connector) and that the caller must have some knowledge of the underlying plugin (i.e. does it support X in `custom`).

This also suggests that we've simply moved the implementation challenges of `sendMessage` into `sendTransfer`. If I have a ledger that doesn't support messaging I need to establish a communications channel specifically for sending this data in parallel to the transfers.

Could we split `custom` into `pluginData` and `peerData`?

As a caller, if I have no idea what plugin I am talking to, I can still safely put anything serializable into `peerData` and know it will get to my peer.

I will only use `pluginData` if I know what plugin I am speaking to or if some of the data in there becomes standard and I expect all plugins to understand it.

xxxx-ledger-plugin-interface-2/xxxx-ledger-plugin-interface-2.md
```diff
@@ -0,0 +1,502 @@
...
+```json
+{
+  "currencyCode": "USD",
+  "currencyScale": 4
+}
+```
+
+For a detailed description of these properties, please see [`LedgerInfo`](#class-ledgerinfo).
+
+#### Event: `connect`
```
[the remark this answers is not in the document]
adrianhopebailie: It's useful for monitoring to have events even if they are just pub-sub type notifications. i.e. You likely won't react to them other than to log them.

I was tempted to suggest we should still keep the transfer and fulfillment events for this reason

xxxx-ledger-plugin-interface-2/xxxx-ledger-plugin-interface-2.md
```diff
@@ -0,0 +1,502 @@
...
+<code>ledgerPlugin.on('connect', () ⇒ )</code>
+
+Emitted whenever a connection is successfully established.
+
+#### Event: `disconnect`
```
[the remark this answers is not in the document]
adrianhopebailie: > Should ledger plugins try to reconnect themselves or should that be up to the user of the plugin?

-1 - That seems like functionality that should be written once in whatever consumes the plugin.

Also, a plugin CAN try to reconnect silently if it is able to buffer requests in the interim. That would be a feature of a good plugin but seems like a high bar for all plugins.

Finally, that makes some assumptions about the underlying transport that will be used which I think is a mistake.

xxxx-ledger-plugin-interface-2/xxxx-ledger-plugin-interface-2.md
```diff
@@ -0,0 +1,502 @@
...
+| `Object` | [custom](#fulfillmentinfocustom) | Object containing ledger plugin specific options |
+
+### Fields
+
+#### FulfillmentInfo#fulfillment
+<code>**fulfillment**:String</code>
+
+A cryptographic fulfillment that is the SHA-256 preimage of the hash provided as the [`executionCondition`](#transferexecutioncondition) when the transfer was first prepared.
+
+Fulfillments are base64url-encoded values with a length of exactly 32 bytes.
```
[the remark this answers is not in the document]
adrianhopebailie: They are fixed length. Isn't a Buffer uneccessary over-head.

xxxx-ledger-plugin-interface-2/xxxx-ledger-plugin-interface-2.md
```diff
@@ -0,0 +1,502 @@
...
+If an incoming transfer is received by the plugin, but no handler is registered, the plugin should reject the transfer.
+
+#### LedgerPlugin#deregisterTransferHandler
+<code>ledgerPlugin.deregisterTransferHandler( ) ⇒ null</code>
+
+Removes the currently used transfer handler. This has the same effect as if [`registerTransferHandler`](#ledgerpluginregistertransferhandler) had never been called.
+
+If no transfer handler is currently set, this method does nothing.
+
+### Event: `info_change`
```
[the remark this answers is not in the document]
adrianhopebailie: It also occurred to me that the connect promise should also resolve to a `LedgerInfo` object.

xxxx-ledger-plugin-interface-2/xxxx-ledger-plugin-interface-2.md
```diff
@@ -0,0 +1,502 @@
...
+|:--|:--|:--|
+| `String` | [name](#interledgerrejectionerrorname) | `'InterledgerRejectionError'` |
+| `String` | [message](#interledgerrejectionerrormessage) | Error message for local use |
+| `Object` | [ilpRejection](#interledgerrejectionerrorilprejection) | Information about the ILP rejection |
+| `String` | [ilpRejection.code](#interledgerrejectionerrorilprejection) | Machine-readable error code |
+| `String` | [ilpRejection.name](#interledgerrejectionerrorilprejection) | Human-readable description of the error code |
+| `String` | [ilpRejection.triggeredBy](#interledgerrejectionerrorilprejection) | ILP address from which the rejection originates |
+| `String[]` | [ilpRejection.forwardedBy](#interledgerrejectionerrorilprejection) | The ILP addresses of zero or more connectors who forwarded the rejection |
+| `String` | [ilpRejection.triggeredAt](#interledgerrejectionerrorilprejection) | ISO 8601 datetime. Describes when the rejection originally occurred. |
+| `Object` | [ilpRejection.additionalInfo](#interledgerrejectionerrorilprejection) | Additional details about the error |
```
emschwartz: I think we need more details on what value types are allowed in here and how you're supposed to use it. Do we only support JavaScript JSON values? What's the range on supported numbers?

Also, is every implementation free to use its own key names? Can connectors modify this to add more details? What if they want to add a property but the key is already used by another application?

Overall, I'm not sure that having an object here is much more helpful than having a single `data` property that transport protocols can define the encoding for, or is otherwise assumed to be a string.

emschwartz: How about adding a `data` property that is specifically for Transport Layer protocols and then specifying that `additionalInfo` can contain key-value pairs as UTF-8 strings?

xxxx-ledger-plugin-interface-2/xxxx-ledger-plugin-interface-2.md
```diff
@@ -0,0 +1,502 @@
...
+)</code>
+
+Emitted any time the plugin's `LedgerInfo` cache changes.
+
+## Class: Transfer
+<code>class Transfer</code>
+
+The `Transfer` class is used to describe transfers from the originator of the sendTransfer call towards some `destination`. All fields are required and MUST NOT be undefined. However, `amount` MAY be the value `'0'`, `data` MAY be an empty buffer and `custom` MAY be an empty object.
+
+###### Fields
```
michielbdejong: In the light of https://github.com/interledger/rfcs/pull/349 I would change this to:
```js
{
  headers: {
    'ILP-Destination': 'g.crypto.bitcoin.1XPTgDRhN8RFnzniWCddobD9iKZatrvH4.~asdf1234',
    'ILP-Condition': 'x73kz0AGyqYqhw_c5LqMhSgpcOLF3rBS8GdR52hLpB8',
    'ILP-Expiry': '2017-12-07T18:47:59.015Z',
    'ILP-Amount': '1000'
 },
 body: <body>
}
```

michielbdejong: Or maybe
```js
{
  destination: 'g.crypto.bitcoin.1XPTgDRhN8RFnzniWCddobD9iKZatrvH4.~asdf1234',
  condition: 'x73kz0AGyqYqhw_c5LqMhSgpcOLF3rBS8GdR52hLpB8',
  expiry: '2017-12-07T18:47:59.015Z',
  amount: '1000'
  body: <body>
}
```

xxxx-ledger-plugin-interface-2/xxxx-ledger-plugin-interface-2.md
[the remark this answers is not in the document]
sappenin: @emschwartz Great point - I've been wondering the same thing (i.e., how to handle an event saying the currency_code changed). It seems like that would be a big problem.

xxxx-ledger-plugin-interface-2/xxxx-ledger-plugin-interface-2.md
```diff
@@ -0,0 +1,502 @@
...
+| Type | Name | Description |
+|:--|:--|:--|
+| `String` | [amount](#transferamount) | Integer transfer amount, in the ledger's base unit |
+| `String` | [destination](#transferdestination) | ILP destination address |
+| `Buffer` | [data](#transferdata) | Transport-layer data |
+| `String` | [executionCondition](#transferexecutioncondition) | Cryptographic hold condition |
```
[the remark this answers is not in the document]
sappenin: Minor nit, but I prefer `expiresAt`. 

Also, I like `executionCondition` because, in our implementation at least, we have other types of conditions floating around, and the naming is nice to be able to distinguish (I know this spec won't exactly match Atomic-mode implementations, but FWIW).

xxxx-ledger-plugin-interface-2/xxxx-ledger-plugin-interface-2.md
sappenin: -1 for the first `transfer` proposal, but +1 for something like the 2nd `transfer` proposal.

* The prefixes (i.e., `ILP-...` in #349 are there as an artifact of using HTTP headers to carry the data (to distinguish between other headers that might be arbitrarily added to the request as part of a normal and valid HTTP request). Conversely, In this interface, the prefixes are not needed because there's less of a risk of "other data fields" colliding with these fields names -- mainly because we're defining the field names here.
* `body` is an artifact of HTTP. We should call it what it really is here (I like Adrian's suggestion above of `peerData` and `pluginData`), and just make new attributes for the data we need.

xxxx-ledger-plugin-interface-2/xxxx-ledger-plugin-interface-2.md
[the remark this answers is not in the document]
adrianhopebailie: > I like executionCondition because, in our implementation at least, we have other types of conditions floating around

Fair point

justmoon:
Regarding whether to keep the ILP packet as an opaque blob or not:

Randomly reread chapters 1 and 2 of [ASN.1 Complete](http://www.oss.com/asn1/resources/books-whitepapers-pubs/larmouth-asn1-book.pdf) recently, which talk about the exact problem we've run into. Should you separate your protocol into layers (as TCP/IP does) or should you merge them into one (as most telecommunication protocols did at the time)? It's a trade-off of simplicity vs extensibility. If you layer your protocol, each layer can evolve independently. If you merge them, they become simpler (in the short run.)

It's a trade-off and therefore: Too many layers are bad. As are too few.

So the question is: Do we need the ledger layer to be able to evolve independently from the Interledger layer? Do we want to be able to think about them independently?

I would say, yes, absolutely. We already have two actively used, competing, ledger layer protocols in BTP and ILP-over-HTTP (not to be confused with HTTP-ILP) just based on preferences without our own small team. Others may wish to use protocol buffers. In the future, we may not want to have a dependency on the TCP/IP stack at all and come up with a ledger protocol that layers straight on a secure communication link.

Merging the Interledger and ledger layers *seems* simple, because you're getting rid of a layer, but it isn't because now it's harder to evolve your ledger layer or your Interledger layer independently since they're tied together.

If we agree that we want to have separation between the ledger/account layer and the Interledger layer, then we need to decide whether the interface is:

1. You need to transfer this small set of fields (`Destination`, `Data`, `Error-Code`, `Error-Triggered-By`, ...) or
2. You need to transfer an opaque blob (ILP packet)

The second provides a lot less opportunity for ledger protocols to get it wrong and allows new ILP packet types to be created in the future, without having to modify each ledger protocol. (Note that you wouldn't just want to update the implementations, you'd want to update the *protocols*.)

We (at least some of us) have collectively come to this conclusion before. That's where the ILP packet came from. So perhaps it's not surprising that thinking about it again would lead back to the same place.

So, considering this, I would strongly advocate we make the LPIv2 interface such that preparation, fulfillment and rejection each take/emit an ILP `Buffer`, which is treated as an opaque blob by the plugin.

adrianhopebailie:
Personally I am in favor of us leaning more toward simplicity in this case.

I agree that the "correct" way to layer these protocols is to have distinct layers between the ledger and ILP layers but I think the cost is higher than we estimate and that the extensibility requirement is overstated.

The interface between the Ledger and ILP layers is unique (compared to other inter-layer interfaces) because the majority of the data elements are standardized by the Interledger standard yet in the current model they are passed both at the ledger and ILP layers.

Ledger layer:
 - Condition
 - Expiry
 - Fulfillment

ILP layer:
 - ILP Address
 - ILP Error Code/Name
 - ILP Error Source (and intermediaries)
 - ILP Error Timestamp

I think there is a strong argument that these should all be at the same layer.

I have made the case previously that the Condition and Fulfillment should be in the ILP layer and I think we have agreed that this is theoretically correct but practically makes no sense because they will be mostly used at the ledger layer. We compromised for simplicity.

In my opinion the same argument could be made for moving the other fields to the ledger layer.

There are a lot of benefits:

1. No need to define a new packet format and encoding.
2. The end-to-end data is just transported as the payload of whatever Application Communication Protocol is used for the ledger layer protocol (eg: HTTP, CoAP, JSON-RPC, custom).
3. We can leverage existing methods for securing and integrity-checking this data end-to-end (otherwise they are only usable at the ledger layer)
4. No need for a fulfillment packet that is just an envelope around a BLOB
5. No need for a payment packet that simply adds an address to the data

In summary, re-use a lot of what already exists and invent less ourselves making implementation and understanding simpler.

I find it very difficult to justify the complexity we have currently to cater for extensibility that seems very unlikely to ever be required.

sappenin:
I concur 100% with @adrianhopebailie -- the benefits he outlined are really compelling (I had a bunch of feedback, but I think Adrian captured nearly all of it).

Only thing I would underscore is developer adoption -- implementing a packet with codecs in different languages might seem trivial, but it is a significant barrier to adoption, IMHO, and it's not clear that we couldn't accomplish the same things using HTTP and a well defined spec.

xxxx-ledger-plugin-interface-2/xxxx-ledger-plugin-interface-2.md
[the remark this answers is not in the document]
justmoon: I think we can use @emschwartz' idea of middleware to serve the transfer events use case. That's what convinced me to take them out.

``` js
class EventMiddleware extends EventEmitter
{
  constructor (next) {
    super()
    this.next = next
  }

  getTransferHandler() {
    return async (transfer) => {
      this.emit('prepare', transfer)

      try {
        const fulfillInfo = await this.next(transfer)
        this.emit('fulfill', transfer, fulfillInfo)
      } catch (err) {
        this.emit('reject', transfer, err)
      }
    }
  }
}
```

And you could hook that into `sendTransfer` to get events for outgoing transfers and info `registerTransferHandler` for incoming transfers.

``` js
const incomingEvents = new EventMiddleware(myActualTransferHandler)
plugin.registerTransferHandler(incomingEvents.getTransferHandler())

incomingEvents.on('prepare', (transfer) => {
  console.log('An incoming transfer for %s was prepared', transfer.amount)
})
```

xxxx-ledger-plugin-interface-2/xxxx-ledger-plugin-interface-2.md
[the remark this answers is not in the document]
justmoon: Currently, the connector listens to the `disconnect` event to know when it should stop advertising a certain route. In fact, it actively informs peers that it can no longer service this route.

For this use case alone, I think this is an important feature.

xxxx-ledger-plugin-interface-2/xxxx-ledger-plugin-interface-2.md
```diff
@@ -0,0 +1,502 @@
...
+<code>ledgerPlugin.on('connect', () ⇒ )</code>
+
+Emitted whenever a connection is successfully established.
+
+#### Event: `disconnect`
+<code>ledgerPlugin.on('disconnect', () ⇒ )</code>
+
+Emitted when the connection has been terminated or lost.
+
+#### Event: `error`
```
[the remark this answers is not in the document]
justmoon: Some fatal errors happen asynchronously. This is to make sure those errors don't just happen silently. If a plugin triggers this event, the connector should probably shut down or at least destroy that plugin.

xxxx-ledger-plugin-interface-2/xxxx-ledger-plugin-interface-2.md
```diff
@@ -0,0 +1,502 @@
...
+General event for fatal exceptions. Emitted when the plugin experienced an unexpected unrecoverable condition. Once triggered, this instance of the plugin MUST NOT be used anymore.
+
+### Ledger Transfers
+
+#### LedgerPlugin#sendTransfer
+<code>ledgerPlugin.sendTransfer( **transfer**:[Transfer](#class-transfer) ) ⇒ Promise.&lt;[FulfillmentInfo](#class-fulfillmentinfo)></code>
+
+Initiates an account-local transfer. A transfer MUST contain an `amount` of zero or more and MAY have attached `data`. See the description of the [Transfer](#class-transfer) class below.
+
+All plugins MUST support amounts in a range from zero to some maximum. Plugins MAY implement zero-amount transfers differently than other transfers.
```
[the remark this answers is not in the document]
justmoon: Don't seem like we need a maximum. We just need an error that tells you what the maximum is.

A connector should:
``` js
try {
   plugin.sendTransfer(...)
} catch (err) {
  if (err.name === 'MaxAmountExceeded') {
    // Get maximum amount from err.maxAmount and trigger a
    // "payment too large" ILP error to tell the sender to send a
    // smaller amount.
  }
}
```

xxxx-ledger-plugin-interface-2/xxxx-ledger-plugin-interface-2.md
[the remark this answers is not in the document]
justmoon: Yeah for `executionCondition` I ran into the same issue - ilp-connector supports a cancellationCondition. For `expiresAt` - it's more consistent to say `expiresAt`, because we have other timestamps using that `...At` format, e.g. `triggeredAt`.

Don't think there is a strong reason to rename these fields. I was ready to do it, but @sharafian vetoed it and I'm kind of in the mode of, "let's just get this done," so if I can avoid some extra renaming, I'm happy.

dappelt:
On a different note: If `getBalance` is removed, how does a user know what his balance is? Or is the idea that the user is not aware of his balance at all?

Knowing how much funds a user has available is essential for many purchase decisions. For example, let's say a user has 5€ left in his wallet. He would like to buy a coffee and a newspaper, both cost 3€. If the user is aware of his balance, he would know that he can afford either the coffee or the newspaper, but not both, and make a conscious purchase decision. On the other hand, if the balance is unknown to the user, he would buy one item and only when he tries to pay for the second item he would figure out that he has not enough funds for both.

0000-ledger-plugin-interface-2/0000-ledger-plugin-interface-2.md
```diff
@@ -0,0 +1,510 @@
...
+  custom: {}
+})
+```
+
+For a detailed description of these properties, please see [`Class: Transfer`](#class-transfer).
+
+#### LedgerPlugin#registerTransferHandler
+<code>ledgerPlugin.registerTransferHandler( **transferHandler**: ( transfer: [Transfer](#class-transfer) ) ⇒ Promise&lt;[FulfillmentInfo](#class-fulfillmentinfo)> ) ⇒ null</code>
+
+Set the callback which is used to handle incoming prepared transfers. The callback should expect one parameter (the [Transfer](#class-transfer)) and return a promise for the resulting [FulfillmentInfo](#class-fulfillmentinfo). If the transfer is rejected or an error occurs, the callback should reject the transfer. In general, the callback should behave as [`sendTransfer`](#ledgerpluginsendtransfer) does.
```
emschwartz: @justmoon While working on the PSKv2 implementation using the LPI2 I realized we also want to have a way to ignore an incoming transfer. This is necessary if you have two receivers listening on one account and want to make sure they don't interfere with one another.

sharafian: actually @justmoon and I were briefly discussing this yesterday. It might be a best practice to have each listener get its own sub-account. This means your ILP address would change every time you reconnect (which is OK because they're ephemeral anyways), but all your listeners would still have a shared balance. The ILP-DCP protocol would just have to take into account which listener is asking, and then send out to that same listener when notifications come in.

michielbdejong:
I think we will eventually want to add the `to` field back as an optional parameter on `sendData` and `sendMoney` so that you can address multiple peers on one ledger. If you would split it into one plugin per peer, then their balances would [go up and down in unison](https://github.com/interledger/rfcs/issues/356#issuecomment-352532810), which would be less helpful than if we just have one plugin per ledger, with a `to` field. After all, this is still called the *ledger* plugin interface, not the Peer Interface. But for now I guess it's fine to remove `to`, because for the forseeable future, we will mainly be working with one-to-one relationships, not one-to-many.

0000-ledger-plugin-interface-2/0000-ledger-plugin-interface-2.md
```diff
@@ -22,9 +22,12 @@ This spec depends on the [ILP spec](../0003-interledger-protocol/).
 | | [**connect**](#ledgerpluginconnect) ( options ) `⇒ Promise.<null>` |
 | | [**disconnect**](#ledgerplugindisconnect) ( ) `⇒ Promise.<null>` |
 | | [**isConnected**](#ledgerpluginisconnected) ( ) `⇒ Boolean` |
-| | [**sendTransfer**](#ledgerpluginsendtransfer) ( transfer ) <code>⇒ Promise.&lt;[FulfillmentInfo](#class-fulfillmentinfo)></code> |
-| | [**registerTransferHandler**](#ledgerpluginregistertransferhandler) ( transferHandler ) <code>⇒ null</code> |
-| | [**deregisterTransferHandler**](#ledgerpluginderegistertransferhandler) ( ) <code>⇒ null</code> |
+| | [**sendData**](#ledgerpluginsenddata) ( data ) <code>⇒ Promise.&lt;Buffer></code> |
+| | [**sendMoney**](#ledgerpluginsendmoney) ( amount ) <code>⇒ Promise.&lt;Buffer></code> |
```
emschwartz: Nitpick: `sendMoney` or `sendValue`?

sharafian: sendValue would be a confusing name, because it sounds like it takes a value and communicates that value to your counterparty. sendMoney is technically not correct in all circumstances, but I think it makes more sense

sappenin: sendAsset?

0000-ledger-plugin-interface-2/0000-ledger-plugin-interface-2.md
```diff
@@ -0,0 +1,249 @@
...
+Emitted when the connection has been terminated or lost.
+
+#### Event: `error`
+<code>ledgerPlugin.on('error', ( **err**:Error ) ⇒ )</code>
+
+General event for fatal exceptions. Emitted when the plugin experienced an unexpected unrecoverable condition. Once triggered, this instance of the plugin MUST NOT be used anymore.
+
+### Sending
+
+#### LedgerPlugin#sendData
```
emschwartz: If you specifically wanted to implement a ledger plugin that did send ILP packets over conditional ledger transfers, would you just implement one that looks at the data being sent and handles it in a different way?

michielbdejong:
I think it would be good to clarify two things:
* how do sendData and sendMoney affect the account's balance
* how can you use on-ledger transfers or paychan claim updates to rebalance?

For example, think about how this would work in ilp-plugin-xrp-paychan:

Both parties open a payment channel to the other party, but don't send any claims yet. Balance starts at zero.

Using `plugin.sendData`, the Sender sends a Prepare packet for 10 drops, and the Receiver fulfills it, so now the Receiver's balance is 10, and the Sender's balance is -10. Right? This is the unsecured balance.

Now in order not to hit their peer's maxUnsecuredBalance, the sender sends a paychan claim for 10 drops. How can they send this claim to their peer, and make the their own balance go back up to 0?

Option A) the Sender calls `plugin.sendMoney` with a negative amount, and attaches the claim
Option B) the Sender sends the claim out-of-band (it used to be in the Fulfill Response), and then the Receiver calls `plugin.sendMoney` with a positive amount
Option C) instead of `plugin.sendMoney`, we define `plugin.claimMoney`, such that `plugin.claimMoney` with a positive amount is equivalent to `plugin.sendMoney` with a negative amount.

sharafian:
```
Using plugin.sendData, the Sender sends a Prepare packet for 10 drops, and the Receiver fulfills it, so now the Receiver's balance is 10, and the Sender's balance is -10. Right? This is the unsecured balance.

Now in order not to hit their peer's maxUnsecuredBalance, the sender signs a paychan claim for 10 drops. How can they send this claim to their peer, and make their own balance go back up to 0?
```

I would look at this a different way. When the sender calls `sendData`, they are giving nothing to their peer (a sendData call carries no money), but the peer is forwarding an interledger payment on their behalf, which they will later have to settle. This means the peer is out 10 units. This is what brings the sender's balance to `-10`. They've effectively fronted you the money.

Now when the sender wants to pay off their debt (i.e. bring the balance back up), they call `sendMoney`. This call actually transmits 10 units of money to the peer, and the peer keeps this settlement without forwarding anything out. This means the peer has gained 10 units. This brings the sender's balance back to `0`.

TL;DR for other people following this: sendData increases debt because your peer gets nothing in and sends a payment out for you. sendMoney decreases debt because the peer gets money in and sends nothing out for you.

michielbdejong:
Right, I see your point. So then I agree, for the LPI the naming does makes sense; we just say that plugin.sendData decreases the sender's trustline balance, and plugin.sendMoney decreases the sender's on-ledger or paychan-secured balance, so that both can have positive amounts and it still makes sense. So then my only issue is with the way the word "transfer" is used in #360.

PR #350, opened 2017-12-09
docs(0019): add definitions for streaming and chunked payments

justmoon:
We're starting to use these terms a bunch, perhaps it's time to add them to the glossary.


0019-glossary/0019-glossary.md
```diff
@@ -236,6 +239,11 @@ See [IL-RFC 9](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-p
 ## Source (transfer, ledger, amount)
 The transfer/ledger/amount directly adjacent to the sender.
 
+## Streaming Payment
```
sappenin: I like the term `Streaming`, but still feel like it still collides with meaning of a "chunked" payment (and can think of some examples that seem to fit both definitions).

Is a "Streaming Payment" actually a type of Chunked Payment, or are these meant to be distinct?

emschwartz: It's possible that you would use the same protocols for both, but the idea was to separate when you're talking about paying for an ongoing or streaming service vs when you're trying to send a single larger amount and break it up for other reasons.

0019-glossary/0019-glossary.md
```diff
@@ -30,6 +30,9 @@ When part of an account's balance is not readily usable, it is up to the ledger
 ## Centralized Ledger
 A ledger which is operated by a single entity.
 
+## Chunked Payment
+A large payment which is split into smaller chunks that are then sent sequentially. Chunked payments reduce the total amount of money in-flight at a given time, which can reduce certain risks and costs.
```
michielbdejong: A large application-level payment which is split into smaller interledger-level chunks that are then sent sequentially. Chunked payments reduce the total amount of money in-flight at a given time, which can reduce certain risks and costs for intermediate connectors, while increasing certain risks for the sender and receiver.

0019-glossary/0019-glossary.md
```diff
@@ -236,6 +239,11 @@ See [IL-RFC 9](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-p
 ## Source (transfer, ledger, amount)
 The transfer/ledger/amount directly adjacent to the sender.
 
+## Streaming Payment
+An ongoing payment where small amounts of money are sent over time to pay for some ongoing service or rent. For instance, a user might pay for a streaming video with a streaming payment. Or a tenant may pay rent as a streaming payment.
+
+The amount sent in each iteration and the interval between iterations are inversely proportional. When the interval is long, e.g. a month, this would usually be called a subscription or recurring payment. The term "streaming payment" is commonly reserved for cases where the interval is short, from one second up to about a day.
```
michielbdejong: A more specific meaning of the term "streaming payments" in the context of http-ilp is the practice of streaming one single http resource in exchange for more than one Interledger payment.

PR #351, opened 2017-12-11
PSKv2

emschwartz:
Transport protocol that handles encryption, fulfillment and condition generation, end-to-end quoting, and chunked payments.

There are still a number of TODOs scattered throughout the spec but I figured we might as well get the discussion going.

Curious to hear if you think all the description up front is helpful or makes the protocol seem more complicated than necessary. Should we cut that down, move more to the appendix, or add more explanation?

Please leave comments inline (as opposed to in the conversation view) so we can keep discussion threads together.

### DIFF
--- a/0025-pre-shared-key-2/0025-pre-shared-key-2.md
+++ b/0025-pre-shared-key-2/0025-pre-shared-key-2.md
@@ -1,0 +1,287 @@
+---
+title: Pre-Shared Key V2 (PSKv2) Transport Protocol
+draft: 1
+---
+# Pre-Shared Key V2 (PSKv2) Transport Protocol
+
+This spec defines version 2 of the Pre-Shared Key Transport Protocol. PSKv2 is intended to be used for micropayments and larger [Chunked Payments](#chunked-payments). It includes end-to-end data encryption, condition and fulfillment generation, [End-to-End Quoting](#end-to-end-quoting) and an approach to Chunked Payments inspired by the Internet's Transmission Control Protocol (TCP).
+
+**Table of Contents**
+
+* [Overview](#overview)
+  + [Setup](#setup)
+  + [Encryption](#encryption)
+  + [Condition and Fulfillment Generation](#condition-and-fulfillment-generation)
+  + [End-To-End Quoting](#end-to-end-quoting)
+  + [Chunked Payments](#chunked-payments)
+* [Model of Operation](#model-of-operation)
+  + [Informational Quote](#informational-quote)
+  + [Single Payment](#single-payment)
+  + [Chunked Payment](#chunked-payment)
+* [Specification](#specification)
+  + [Encryption](#encryption-1)
+  + [Data Packet](#data-packet)
+* [Appendix A: Advanced Sender Features](#appendix-a-advanced-sender-features)
+
+## Overview
+
+Like [PSKv1](./0016-pre-shared-key/0016-pre-shared-key.md), PSKv2 uses a secret shared between the sender and receiver to generate the conditions and fulfillments for payments without repeated end-to-end communication. PSKv2 can be used for individual or streaming micropayments and also includes support Chunked Payments.
+
+This specificiation aims to standardize the behavior of receivers while allowing for sophisticated sender behavior to be developed in the future. [Appendix A: Advanced Sender Features](#appendix-a-advanced-sender-features) outlines areas for potential sender features to be implemented in the future.
+
+### Setup
+
+Before using PSKv2, the sender and receiver use an authenticated, encrypted communication channel (generally provided as part of an [Application Layer Protocol](./0001-interledger-architecture/0001-interledger-architecture.md#application-layer)) to exchange:
+
+- PSK version (such as `2.0` for this version)
+- A 32-byte random or pseudorandom shared secret
+- The receiver's ILP address
+
+The receiver MAY generate the shared secret from a longer-term secret and a nonce, as described in [PSKv1's Appendix A: Recommended Algorithm for Generating Shared Secrets](./0016-pre-shared-key/0016-pre-shared-key.md#appendix-a-recommended-algorithm-for-generating-shared-secrets). In this case, the receiver would append the nonce (**not** the secret) to their ILP address to create the destination address they communicate to the sender.
+
+### Encryption
+
+In PSKv2.0, request and response data is encrypted using [AES-256-GCM](https://en.wikipedia.org/wiki/Galois/Counter_Mode) with a 12-byte Initialization Vector (IV) and a 16-Byte Authentication Tag.
+
+If subsequent versions support additional encryption algorithms, those details should be exchanged between the sender and receiver when they establish the pre-shared key. If a receiver attempts to decrypt an incoming packet but is unable to (perhaps because the sender is using an unsupported cipher), they SHOULD reject the incoming transfer with an `F06: Unexpected Payment` error.
+
+### Condition and Fulfillment Generation
+
+The sender and receiver generate the condition and fulfillment using the pre-shared key that gives this protocol its name. The fulfillment is computed from the pre-shared key and the encrypted PSK data. The condition is the hash of the fulfillment.
+
+As in PSKv1, this method for generating the condition and fulfillment enables many payments to be sent without end-to-end communication for each one. However, because the sender also knows the fulfillment up-front, neither version of PSK allows the sender to use the fulfillment as a non-repudiable proof of payment. For use cases where this is desired, a Transport Protocol like the [Interledger Payment Request (IPR)](./0011-interledger-payment-request/0011-interledger-payment-request.md) should be used instead.
+
+### End-To-End Quoting
+
+In contrast to PSKv1, which depends on the [Interledger Quoting Protocol (ILQP)](./0008-interledger-quoting-protocol/0008-interledger-quoting-protocol.md), PSKv2 has two built-in mechanisms for discovering the exchange rate of a given path: informational quotes and dynamic price discovery.
+
+For purely informational quotes, PSKv2 uses unfulfillable test payments to the receiver to determine the exchange rate for a given path. The condition is set to a random 32-byte condition (with no known fulfillment). When the receiver gets the incoming payment, they reject the incoming transfer and include in their response how much arrived in that transfer. The sender can use the source amount they set and the receiver's response to determine the rate.
+
+When a sender is sending many payments to the receiver, they need a dynamic view of the exchange rate because the rate may change during the course of the interaction. To handle these cases, every PSKv2 payment response includes the amount that arrived in the final transfer to the receiver. Similar to the informational quotes, the sender can use the chunk source amount and the receiver's response to determine and monitor the exchange rate.
+
+**Judging Prices by Source Units:** With any type of Interledger quoting, whether end-to-end or using ILQP, senders SHOULD judge prices in their source units. Receivers can lie about how much arrives in a test payment for a quote or they could run their own ledger and connector that happen to have a bad exchange rate. Therefore, senders must assume that whatever amount the receiver says has arrived represents the "real" rate of the path.
+
+**Linear Exchange Rates and Fixed Destination Amount Quotes:** End-to-End Quoting works best with linear exchange rates. If the rate for a path is linear, it can be determined by sending a single test payment for an arbitrary amount. The source amount required to deliver a fixed destination amount can be easily determined by dividing the source amount by the rate. If exchange rates are non-linear, getting an informational quote for a fixed destination amount would require a binary search, which is not ideal. However, in an Interledger network primarily or exclusively suited to small payments, there should be little reason for rates to vary by payment size.
+
+### Chunked Payments
+
+PSKv2 is designed to support both "chunked" and "streaming" payments. Both involve sending many small payment chunks to pay for some interaction. "Chunked Payments" refers to splitting one larger transaction into smaller chunks, for example sending $1000 as one thousand $1 payments. Chunked payments are useful for dealing with paths that have low Maximum Payment Sizes (MPS). "Streaming Payments" refers to sending payment chunks in exchange for an ongoing service, for example paying for a movie stream 1 Mb at a time.
+
+Interledger connectors set their Maximum Payment Size (MPS) based on their available liquidity, the "bandwidth" they allocate to each of their peers, and the risk they are willing to take on each payment. Analogous to the Internet's Path Maximum Transmission Unit (MTU), an Interledger path MPS is the minimum MPS of the connectors involved.
+
+If senders want to use a path with an MPS lower than their payment amount, they need to split their payment into chunks. PSKv2 includes a Payment ID and Sequence number to enable receivers to associate incoming payment chunks properly.
+
+**Possibility for Incomplete Payments:** If the path between the sender and receiver runs out of liquidity while a chunked payment is in progress, it could be impossible for the payment to complete fully. This should be a rare case, analogous to Internet Service Providers getting cut off because they fail to pay their electricity bills. However, in such cases, the receiver may need to refund the partial payment by sending another chunked payment back and either the sender or receiver would incur the transaction costs.
+
+**Fluctuating Exchange Rates:** The rate from the sender to the receiver may change due to legitimate market movements during the course of a payment. Senders SHOULD monitor the rate changes and they may not want to send payments that are significantly larger than their bandwidth (this could be compared to how downloading an ultra high-definition movie over a dial-up connection is possible but would not make for a good experience).
+
+**Malicious Connectors:** Exchange rates can also change for illegitimate reasons, such as connectors driving up the price mid-payment. Ultimately, the main force keeping connectors' rates low is competition. If there is only one path, the connectors can raise the rate without resorting to any sneaky tactics and the sender would need to decide if they are willing to pay that rate. If a sender has multiple options for the connector they send through, they may want to use [Multi-Path Chunked Payments](#multi-path-chunked-payments) to automatically switch paths if one becomes more expensive.
+
+**Trusting the Receiver:** The sender of a payment must always trust the receiver, except in the rare case that they are doing an atomic exchange of money for goods or services. Whether or not the sender is sending a single- or multiple-chunk payment, a malicious receiver could simply refuse to deliver the goods or services once the payment arrives. If senders must trust the receivers anyway, it is no worse to trust receivers to accurately report how much money arrived and not collude with connectors to raise the rates mid-payment. Since senders will judge merchants' prices by the total source amount they pay, receivers that pull such tricks should develop negative reputations and be avoided.
+
+## Model of Operation
+
+### Informational Quote
+
+1. Sender creates a packet with a type of 1, a random Payment ID, a Sequence number of 0 (or the next Sequence if the quote is part of a multi-chunk payment), and both the Payment Amount and Chunk Amount set to 2^64 - 1 (the maximum unsigned 64-bit integer, or 18446744073709551615). Sender [encrypts](#encryption-envelope) the packet.
+2. Sender generates a random 32-byte condition.
+3. Sender sends a test payment with the condition, encrypted data and either the chosen source amount (in the case of a fixed source amount payment) or an arbitrary amount (in the case of a fixed destination amount).
+4. Receiver is notified of the incoming transfer and decrypts the data.
+5. Receiver compares the Chunk Amount from the packet with the incoming transfer amount and notes that the incoming transfer amount is too low.
+6. Receiver creates a response packet, including the Payment ID and Sequence from the request, a Payment Amount of 0 (or the total they have fulfilled with this Payment ID), and the Chunk Amount set to the amount that arrived on the incoming transfer. Receiver [encrypts](#encryption-envelope) the packet.
+7. Receiver rejects incoming transfer, setting the rejection message's data to the encrypted quote response.
+8. Sender decrypts the quote response and divides the receiver's Chunk Amount by the source amount to determine the path exchange rate.
+
+### Single Payment
+
+1. In the case of a fixed source amount payment, sender multiplies the source amount by the rate to determine the Payment Amount. In the case of a fixed destination amount payment, sender divides the destination amount by the rate to determine the amount for their outgoing transfer.
+2. Sender creates a packet with a Type of 1, a random Payment ID, a Sequence number of 0, and the Payment Amount and Chunk Amount set to the destination amount determined in step 2. Sender [encrypts](#encryption-envelope) the request.
+3. Sender generates the payment condition as described [below](#fulfillable-condition).
+4. Sender sends a payment using the source amount determined in step 2, the condition from step 3, and the encrypted request as the data.
+5. Receiver is notified of the incoming transfer and decrypts the data.
+6. Receiver checks to ensure that the incoming transfer amount is greater than or equal to the Chunk Amount specified in the request. If the Payment Amount is not equal to 0, the receiver checks that the transfer amount is less than or equal to the Payment Amount specified in the request.
+7. Receiver creates a [Payment Chunk Response](#payment-chunk-response) with the same Payment ID and Sequence number. The Payment Amount and Chunk Amount are set to the amount that arrived on the incoming transfer. The receiver then [encrypts](#encryption-envelope) the data.
+8. Receiver generates fulfillment as described [below](#fulfillment-generation) and fulfills the incoming transfer with the fulfillment and the encrypted response data.
+9. Sender is notified of the outgoing transfer being fulfilled and can use the decrypted response to determine the amount that arrived at the destination.
+
+### Chunked Payment
+
+1. Sender creates a packet with Type 0, a random Payment ID, a Sequence number of 0 and a Chunk Amount of 0. In the case of a fixed destination amount, sender sets the Payment Amount to the destination amount. In the case of a fixed source amount, sender sets the Payment Amount to 2^64 - 1 (the maximum unsigned 64-bit integer, or 18446744073709551615). Sender [encrypts the request](#encryption-envelope).
+2. Sender generates the payment condition as described [below](#fulfillable-condition).
+3. Sender sends a payment with the condition from step 2., the encrypted request from step 1. as the data, and an arbitrary source amount.
+4. Receiver is notified of the incoming transfer and decrypts the data.
+5. Receiver checks that the incoming transfer amount is greater than or equal to the Chunk Amount and that the total that has arrived thus far does not exceed the Payment Amount.
+6. Receiver creates a response packet with the Type 2, the same Payment ID, Sequence. The Payment Amount is set to the total the receiver has received and fulfilled thus far (including the current chunk). Receiver sets the Chunk Amount to the amount that arrived on the incoming transfer. Receiver [encrypts the response](#encryption-envelope).
+7. Receiver generates fulfillment as described [below](#fulfillment-generation) and fulfills the incoming transfer.
+8. Sender is notified of the outgoing transfer being fulfilled and decrypts the attached data.
+9. Sender determines the exchange rate by dividing the Chunk Amount from the receiver by the chunk's source amount.
+10. Sender repeats step 1-8 with the following modifications, until the fixed source or destination amount has been sent or delivered, respectively:
+  - Sender MUST create a new PSK data packet for each chunk with the same Payment ID and an incremented Sequence number (the Sequence enables the sender to reliably correlate outgoing requests with their responses)
+  - Sender MAY specify the minimum Chunk Amount the receiver should accept for each transfer based on the calculated path exchange rate
+  - Sender MAY adjust the transfer amount upwards or downwards in response to successful and unsuccessful payments, or to deliver a fixed destination amount exactly
+  - Sender SHOULD set the Type to 1 on the last chunk. Receivers SHOULD NOT fulfill additional chunks after a Type 1 chunk has been received. Note that receivers SHOULD stop fulfilling chunks after the specified Payment Amount has been reached, even if the sender did not send a Type 1 packet.
+
+## Specification
+
+### Encryption
+
+Request and response data are encrypted using [AES-256-GCM](https://en.wikipedia.org/wiki/Galois/Counter_Mode) with a random 12-byte Initialization Vector (IV) and a 16-byte Authentication Tag.
+
+#### Encryption Envelope
+
+```
+(Numbers represent bytes)
+
+       4       8       12      16      20      24      28      32
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|       Random IV       |       Authentication Tag      | Ciph..|
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                          Ciphertext                           |
+\                                                               \
+\                                                               \
+|                          Ciphertext                           |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+```
+
+| Field | Type | Description |
+|---|---|---|
+| Random IV  | 12-Byte UInt | Nonce used as input to the AES-GCM algorithm. Also ensures conditions are random. Senders MUST NOT encrypt two packets with the same nonce |
+| Authentication Tag   | 16-Byte UInt | Authentication tag produced by AES-GCM encryption that ensures data integrity |
+| Ciphertext | 0-32739 Bytes | Encrypted data (see below for contents) |
+
+#### Encryption Pseudocode
+
+Note that the encryption key is derived from the shared secret using an HMAC.
+
+```
+iv = random_bytes(16)
+encryption_key = hmac_sha256(shared_secret, "ilp_psk_encryption")
+{ ciphertext, auth_tag } = aes_256_gcm(encryption_key, iv, data)
+```
+
+### Data Packet
+
+All PSKv2.0 requests and responses encode their data in the same byte format, though the meanings of the fields differ depending on whether it is a request or a response.
+
+The following is encrypted, as defined [above](#encryption), and set as the ILP payment data on a payment from the sender to the receiver.
+
+```
+(Numbers represent bytes)
+
+       4       8       12      16      20      24      28      32
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|T|           Payment ID          |  Seq  | Payment Amount| Chu-|
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|nk Amount|             Application Data                        |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                       Application Data                        |
+\                                                               \
+\                                                               \
+|                       Application Data                      |E|
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|                          Junk Data                            |
+\                                                               \
+\                                                               \
+|                          Junk Data                            |
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+```
+
+#### Packet Types
+
+| Type ID | Description |
+|---|---|
+| `0` | Payment chunk. May be carried on a fulfillable transfer (in the case of a payment) or an unfulfillable transfer (in the case of a quote). |
+| `1` | Last payment chunk. Same as `0` but indicates that there will be no more chunks after this one is fulfilled. |
+| `2` | Success response. Must be carried on the data accompanying the fulfillment. |
+| `3` | Error response. Must be carried in the error with which the receiver rejects the incoming transfer. |
+
+#### Request Fields
+
+| Field | Type | Description |
+|---|---|---|
+| Type (T) | 1-Byte UInt (UInt8) | `0` to indicate a payment chunk, `1` to indicate it is the last chunk for the payment. Receivers SHOULD NOT fulfill additional chunks for the given payment after the last chunk has been received |
+| Payment ID | 16-Byte UInt | Unique ID of this payment |
+| Sequence | 4-Byte UInt (UInt32) | Zero-indexed chunk sequence number. Receivers SHOULD NOT fulfill two chunks with the same sequence for a single payment. Receivers SHOULD use the sequence to order chunks in case application data is split across multiple chunks, but chunks do not necessarily need to be delivered in order |
+| Payment Amount (Total)* | 8-Byte UInt (UInt64) | The total amount the payment is intended to deliver. A value of `0` in a request indicates that the sender will determine when to stop sending chunks |
+| Chunk Amount (Minimum) | 8-Byte UInt (UInt64) | Minimum amount the receiver should receive for this chunk in order to fulfill it. A value of `0` in a request indicates that any amount should be accepted |
+| Application Data | [OER Variable-Length Octet String](http://www.oss.com/asn1/resources/books-whitepapers-pubs/Overview%20of%20OER.pdf) | User data carried along with the payment |
+| Extensions (E) | 1-Byte UInt (UInt8) | Currently set to `0`. Reserved for future use to indicate the presence of extension fields. Parsers that do not understand the extensions MUST ignore them |
+| Junk Data | 0-? Bytes | _(Optional)_ Extra data included to obscure what the ILP payment is for. Receivers MUST ignore this data. Senders SHOULD include either the maximum or a random number of empty (0) bytes |
+
+\* **Note:** The largest Payment Amount (the sum of all of the chunks) supported by PSKv2.0 is the maximum value for a UInt64 or 18446744073709551615.
+
+#### Response Fields
+
+The following apply to both fulfillment and error responses:
+
+| Field | Type | Description |
+|---|---|---|
+| Type (T) | 1-Byte UInt (UInt8) | `3` to indicate it is a fulfillment response, `4` to indicate it is an error |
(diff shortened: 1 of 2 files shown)


0000-pre-shared-key-v2.md
```diff
@@ -0,0 +1,291 @@
+---
+title: Pre-Shared Key V2 (PSKv2) Transport Protocol
+draft: 1
+---
+# Pre-Shared Key V2 (PSKv2) Transport Protocol
+
+This spec defines version 2 of the Pre-Shared Key Transport Protocol. PSKv2 implements end-to-end data encryption, condition and fulfillment generation, [End-to-End Quoting](#end-to-end-quoting) and [Chunked Payments](#chunked-payments). The approach to chunked payments is inspired by the Internet's Transmission Control Protocol (TCP).
```
michielbdejong: Can it also be used for test payments and/or stats collection?

0000-pre-shared-key-v2.md
```diff
@@ -0,0 +1,291 @@
...
+
+When a sender is sending many payments to the receiver, they need a dynamic view of the exchange rate because the rate may change during the course of the interaction. To handle these cases, every PSKv2 payment response includes the amount that arrived in the final transfer to the receiver. Similar to the informational quotes, the sender can use the chunk source amount and the receiver's response to determine and monitor the exchange rate.
+
+**Trusting the Receiver:** With any type of Interledger quoting, whether end-to-end or using ILQP, senders SHOULD judge prices in their source units. Receivers can lie about how much arrives in a test payment for a quote or they could run a connector that happens to have a bad exchange rate to their destination units. Therefore, senders must assume that whatever amount the receiver says has arrived represents the "real" rate of the path.
```
michielbdejong: Strictly speaking, in ILQP the receiver cannot lie about how much arrives, unless they run the destination ledger

emschwartz: "or they could run a connector that happens to have a bad exchange rate to their destination units"

0000-pre-shared-key-v2.md
emschwartz: Sure it can be used for test payments, but would those be any different from normal payments or quotes? If not, I'm not sure it needs to be documented here

0000-pre-shared-key-v2.md
michielbdejong: running one such connector wouldn't make a difference, the network would route around it, using other connectors to reach the destination ledger

0000-pre-shared-key-v2.md
```diff
@@ -0,0 +1,291 @@
...
+9. Sender repeats step 1-8 until the fixed source or destination amount has been sent or delivered, respectively. Sender MAY adjust the chunk amount upwards or downwards in response to successful and unsuccessful payments.
+
+## Specification
+
+### Encryption
+
+Request and response data are encrypted using [AES-GCM](https://en.wikipedia.org/wiki/Galois/Counter_Mode) with a 16-byte Initialization Vector (IV).
+
+#### Encryption Envelope
+
```
adrianhopebailie: Strongly recommend using the format from RFC8188 rather than making up our own.
See: http://httpwg.org/specs/rfc8188.html#header

This would make PKS over ILP over HTTP compliant with RFC8188.

emschwartz: > This would make PKS over ILP over HTTP compliant with RFC8188.

This wouldn't be true if we [bring back the ILP packet](https://github.com/interledger/rfcs/pull/347#issuecomment-350759502), because the HTTP body will be the ILP packet and this envelope will be inside there. I think it's more important to be consistent with the rest of our stack (using OER)

adrianhopebailie: I agree. This makes me sad 😢 

1. Put ILP Address in HTTP headers use RFC 8188 as is. Profit
OR
2. Put ILP Address in packet in body. Invent everything from scratch.

0000-pre-shared-key-v2.md
```diff
@@ -0,0 +1,324 @@
...
+```
+iv = random_bytes(16)
+encryption_key = hmac_sha256(shared_secret, "ilp_psk_encryption")
+{ ciphertext, auth_tag } = aes_256_gcm(encryption_key, iv, data)
+```
+
+### Quotes
+
+#### Quote Request
```
emschwartz: @justmoon brought up the point: do we need purely informational quotes or should these be removed?

If on a normal payment the receiver will tell you how much arrived, the sender could just use an actual payment to determine the rate for the path. It seems like the main problem would be that the rounding error could get quite large if you send a super tiny payment first. That would also mean you need to pay a small amount for quotes, but that might not be the worst thing.

Thoughts?

sentientwaffle: If a quote is a "Payment Chunk Request" with the "Chunk Minimum" set to the max uint64, then it won't actually cost anything.

emschwartz: That's a great point. I'm going to update it to remove the quote request and response types. It's very cool that that means we can use the same request/response/error packets to handle quotes, one-off payments, and chunked payments.

0000-pre-shared-key-v2.md
```diff
@@ -234,6 +231,42 @@ The following is encrypted, as defined [above](#encryption). In the case of a re
...
+
+| Field | Type | Description |
+|---|---|---|
+| Type (T) | 1-Byte UInt (UInt8) | Always `4` to indicate it is a response |
+| Payment ID | 16-Byte UInt | Unique ID from the request. Senders SHOULD ensure that the response ID matches the request |
+| Sequence | 4-Byte UInt (UInt32) | Chunk sequence number from the request. Senders SHOULD ensure that the sequence matches the request |
+| Payment Received | 8-Byte UInt (UInt64) | The amount the receiver has received thus far for the payment (the sum of fulfilled chunks) |
+| Chunk Received | 8-Byte UInt (UInt64) | The amount the receiver received for the given chunk |
+| Application Data | [OER Variable-Length Octet String](http://www.oss.com/asn1/resources/books-whitepapers-pubs/Overview%20of%20OER.pdf) | User data carried along with the payment |
+| Extensions (E) | 1-Byte UInt (UInt8) | Always `0`. Reserved for future use to indicate the presence of extension fields |
```
sharafian: Extensions must be ignored if you don't implement the logic to handle OER extensions. "Always `0`" implies that a parser should verify that the extensions byte is zero, which would actually break extensibility.

0000-pre-shared-key-v2.md
```diff
@@ -234,6 +231,42 @@ The following is encrypted, as defined [above](#encryption). In the case of a re
...
+The following is encrypted, as defined [above](#encryption), and set as the data accompanying the payment chunk fulfillment.
+
+```
+(Numbers represent bytes)
+
+ 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|T|           Payment ID          |  Seq  | Payment Rcved | Chu-|
++-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+|nk Rcved |             Application Data                        |
```
sharafian: Maybe include the length prefix for the application data in this diagram

0000-pre-shared-key-v2.md
```diff
@@ -0,0 +1,324 @@
...
+| Payment Total | 8-Byte UInt (UInt64) | The total amount the payment is intended to deliver. A value of `0` in a request indicates that the sender will determine when to stop sending chunks |
+| Chunk Minimum | 8-Byte UInt (UInt64) | Minimum amount the receiver should receive for this chunk in order to fulfill it. A value of `0` in a request indicates that any amount should be accepted |
+| Application Data | [OER Variable-Length Octet String](http://www.oss.com/asn1/resources/books-whitepapers-pubs/Overview%20of%20OER.pdf) | User data carried along with the payment |
+| Extensions (E) | 1-Byte UInt (UInt8) | Always `0`. Reserved for future use to indicate the presence of extension fields |
+| Junk Data | 0-? Bytes | _(Optional)_ Extra data included to obscure what the ILP payment is for. Receivers MUST ignore this data. Senders SHOULD include either the maximum or a random number of empty (0) bytes |
+
+**TODO:** should we add a source address for refunds? (that should only be necessary in extreme cases)
```
sentientwaffle: I think that refunds are an uncommon enough case that psk shouldn't account for them.

0000-pre-shared-key-v2.md
emschwartz: I'd like to but I'm not sure how to illustrate a variable length prefix...

0000-pre-shared-key-v2.md
```diff
@@ -122,74 +137,9 @@ encryption_key = hmac_sha256(shared_secret, "ilp_psk_encryption")
...
-Senders SHOULD use a random 32-byte value as the payment condition.
-
-#### Quote Fulfillment
-
-Not applicable. The quote is intentionally unfulfillable.
-
-### Payment Chunks
-
-#### Payment Chunk Request
+All PSKv2.0 requests and responses encode their data in the same byte format, though the meanings of the fields differ depending on whether it is a request or a response.
```
adrianhopebailie: What's the motivation for defining a binary packet format?

emschwartz: 1. We need some encoding format
2. There's no reason to have the overhead of including the keys as strings, since we know what they are up front
3. Might as well make it consistent with the rest of our stack by using OER

0000-pre-shared-key-v2.md
```diff
@@ -0,0 +1,273 @@
...
+
+**Fluctuating Exchange Rates:** The rate from the sender to the receiver may change due to legitimate market movements during the course of a payment. Senders SHOULD monitor the rate changes and they may not want to send payments that are significantly larger than their bandwidth (this could be compared to how downloading a 4K movie over a dial-up connection is possible but would not make for a good experience).
+
+**Malicious Connectors:** Exchange rates can also change for illegitimate reasons, such as connectors driving up the price mid-payment. Ultimately, the main force keeping connectors' rates low is competition. If there is only one path, the connectors do not need to do anything sneaky to drive up the rate. If a sender has multiple options for the connector they send through, they may want to use [Multi-Path Chunked Payments](#multi-path-chunked-payments) to automatically switch paths if one becomes more expensive.
```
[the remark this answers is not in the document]
emschwartz: The point is that they don't need to do anything sneaky. They can simply charge a higher rate up front, since there's nowhere else to go. Sending a chunked payment is no worse than sending a single payment in this case.

0024-pre-shared-key-v2/0024-pre-shared-key-v2.md
```diff
@@ -0,0 +1,266 @@
...
+
+**Trusting the Receiver:** When doing any Interledger payment aside from an atomic exchange of money for goods or services, the sender must trust the receiver to deliver the goods once they receive the payment. With chunked payments, the sender must trust the receiver to deliver the goods and accurately report how much money has arrived on their incoming transfer. A malicious receiver could keep reporting less money having arrived on every transaction in an attempt to extract more money from the sender, but this is no worse a threat than a receiver demanding more money after a non-chunked payment arrives but before they release the goods or services. Since senders will judge merchants' prices by the total source amount they pay, receivers that pull such tricks should develop negative reputations and be avoided.
+
+## Model of Operation
+
+### Informational Quote
+
+1. Sender creates a packet with a type of 1, a random Payment ID, a Sequence number of 0, a Payment Amount of 0, a Chunk Amount set to 18446744073709551615 (the maximum value) and [encrypts it](#encryption-envelope).
```
michielbdejong: By payment amount and chunk amount, do you mean transfer amount and packet amount? Why would anyone forward such a payment if the transfer amount is zero?

0024-pre-shared-key-v2/0024-pre-shared-key-v2.md
```diff
@@ -0,0 +1,266 @@
...
+4. Receiver is notified of the incoming transfer and decrypts the data.
+5. Receiver compares the Chunk Amount from the packet with the incoming transfer amount and notes that the incoming transfer amount is too low.
+6. Receiver creates a response packet, including the Payment ID and Sequence from the request, a Payment Amount of 0 (or the total they have fulfilled with this Payment ID), and the Chunk Amount set to the amount that arrived on the incoming transfer. Receiver [encrypts](#encryption-envelope) the packet.
+7. Receiver rejects incoming transfer, setting the rejection message's data to the encrypted quote response.
+8. Sender decrypts the quote response and divides the receiver's Chunk Amount by the source amount to determine the path exchange rate.
+
+### Single Payment
+
+1. In the case of a fixed source amount payment, sender multiplies the source amount by the rate to determine the Payment Amount. In the case of a fixed destination amount payment, sender divides the destination amount by the rate to determine the amount for their outgoing transfer.
```
michielbdejong: I would prefer to use ilqp in this case

0024-pre-shared-key-v2/0024-pre-shared-key-v2.md
emschwartz: No, it means the fields called Payment Amount and Chunk Amount in the PSK data packet defined below

0024-pre-shared-key-v2/0024-pre-shared-key-v2.md
emschwartz: I don't think that will realistically be an option. As I've said before, ILQP doesn't work unless every connector supports it and/or every connector pushes up to date price information to one another very frequently. This is a high burden to place on the connectors that isn't necessary if senders can get the same information in a different way that doesn't require a specific feature from the connectors.

emschwartz:
I do think we should deprecate the ILP packet type with the amount in it and ILQP. 

We are trying to create interoperability. Having multiple different protocols at the core interoperability layer means more that everyone needs to agree on and implement and I don't think they provide enough substantive benefit to merit the cost. Someone could build another set of protocols using the same chaining idea as ILP, but it's not our concern because it wouldn't actually be interoperable.

Forwarded payments are a much simpler and more generic building block, as evidenced by the fact that you can build the three different flows described in this spec using only that one connector feature. I would strongly argue that this is the right building block to standardize on at the Inyerledger layer.

0000-pre-shared-key-v2.md
adrianhopebailie: I think pushing OER (or any binary encoding) up to this layer is a mistake because it is passed opaquely through all intermediaries so the interoperability requirement is significantly diminished.

Only the sender and receiver need to interpret this so it's more likely to be implemented by clients than connectors and something more common at this layer (like JSON) just seems like a better natural fit.

michielbdejong:
> more that everyone needs to agree on and implement

I think that's an illusion. For ledgers and ledger plugins, it doesn't matter because they treat the packet as opaque. The question is how connectors can be interoperable with each other.

Of course connectors need to talk each other's complex language to do complex things like path finding, liquidity management, and dynamic pricing. So yes, interoperability between connectors will be hard, regardless of whether we make the interledger layer a thick layer or a thin layer.

The beauty is that the ledger requirements are low: to become interoperable, a ledger just needs to support timed sha256 hashlock escrow + messaging, and if the ledger supports both (or the messaging is added in a way that makes sense for that ledger) then someone needs to write a plugin for that ledger, done. That doesn't change by moving quoting and path finding up or down in the stack.

adrianhopebailie:
> I do think we should deprecate the ILP packet type with the amount in it and ILQP.

I agree that we should deprecate the ILP packet with the amount (we need one or the other in the ILP layer).

I'm not sure about ILQP. It is not a fundamental part of ILP but that doesn't mean it may not be used between connectors. i.e. It's not part of ILP but it may still be used so it's probably out of place in the ILP packet definitions but not entirely useless

michielbdejong:
The destination amount can easily be considered to be implied by the transfer amount, so I wouldn't be so opposed to moving it into the packet data. We could also say ILQP is part of the connector-to-connector (and/or sender-to-connector) layer, the fact that we have ilp-packet types for ILQP just makes it easy to send such messages as a payload to some other account on the same ledger

emschwartz:
Ok I actually don't care if we deprecate ILQP, but it is important to agree that it is not required and you cannot expect all other connectors to have implemented it (as was the assumption before). 

I'm coming at this from the perspective of someone who wants to reimplement ILP in another language. My question is what must I write to make my implementation interoperable. For a minimal connector (with configured routes) you need one or two ledger protocols and ILP forwarded payments/fulfillments/errors. For a sender or receiver you need at least one ledger layer protocol, ILP forwarded payments/fulfillments/errors, a transport protocol and an application layer protocol.

0000-pre-shared-key-v2.md
emschwartz: It is only the sender and receiver that need to implement this, but the sender and receiver also need to support OER to send and receive ILP packets.

emschwartz: Agreed, I think many application layer protocols will use JSON or some other "friendlier" encoding. If that's the case though, that's all the more reason to have as minimal byte overhead as possible in the transport layer so the application layer can send more data.

In most cases I would expect the ILP and PSK2 modules to be implemented roughly at the same time, by the same people. If you're building a sender or receiver right now you'll need both, and I think that is a good place to get started for an implementation.

0000-pre-shared-key-v2.md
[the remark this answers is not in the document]
dappelt: If such a connector sets a higher rate up front, a customer could simply not send the transfer at all. By providing a cheaper rate initially and increasing the rate half-way through the sender is trapped: Either he finishes the payment at the higher rate or he asks the receiver for a refund, for which the connector would also apply the higher rate. Either way, the sender is trapped.

I propose to drop the sentence _[...] the connectors do not need to do anything sneaky to drive up the rate [...]_.

0024-pre-shared-key-v2/0024-pre-shared-key-v2.md
```diff
@@ -0,0 +1,266 @@
...
+
+## Setup
+
+Before using PSKv2, the sender and receiver use an authenticated, encrypted communication channel (generally provided as part of an [Application Layer Protocol](./0001-interledger-architecture/0001-interledger-architecture.md#application-layer)) to exchange:
+
+- PSK version (such as `2.0` for this version)
+- A 32-byte random or pseudorandom shared secret
+- The receiver's ILP address
+
+The receiver MAY generate the shared secret from a longer-term secret and a nonce that they put into the ILP address they give to the sender, as described in [PSKv1's Appendix A: Recommended Algorithm for Generating Shared Secrets](./0016-pre-shared-key/0016-pre-shared-key.md#appendix-a-recommended-algorithm-for-generating-shared-secrets).
```
dappelt: > [...] longer-term secret and a nonce that they put into the ILP address [...]

Consider rewording to make it clear that only the nonce goes into the ILP address, not the long-term secret.

0024-pre-shared-key-v2/0024-pre-shared-key-v2.md
```diff
@@ -0,0 +1,266 @@
...
+
+### Encryption
+
+In PSKv2.0, request and response data is encrypted using [AES-256-GCM](https://en.wikipedia.org/wiki/Galois/Counter_Mode) with a 16-byte Initialization Vector (IV) and a 16-Byte Authentication Tag.
```
dappelt: Regarding the size of the Authentication Tag. This is an excerpt from the [NIST publication](http://nvlpubs.nist.gov/nistpubs/Legacy/SP/nistspecialpublication800-38d.pdf) on GCM:
> The  bit  length  of  the  tag,  denoted  t,  is  a  security  parameter,  as  discussed  in  Appendix  B. In general, t may  be  any  one  of  the  following  five values:  128,  120,  112,  104,  or  96.    For  certain applications, t  may  be  64  or  32;  guidance  for  the  use  of  these  two  tag  lengths, including requirements on the length of the input data and the lifetime of the key in these cases, is given in Appendix C.

So, a 16 byte Authentication Tag can be considered secure.

Another source recommending a 16 byte authentication tag is _Cryptography Engineering: Design Principles and Practical Applications_ (p. 309):
> “Therefore, although it is possible to reduce the size of the authenticator for GCM from 128 bits to something less, we recommend not doing so. Our recommendation is to only use GCM with the full 128-bit authentication tag.”

emschwartz: Ok, and I think the recommendation is also to use a 16-byte IV, right?

dappelt: The NIST publication recommends a bit length of 96 for the IV:
> For  IVs,  it  is  recommended  that  implementations restrict  support  to  the length  of  96  bits,  to promote interoperability, efficiency, and simplicity of design.

0024-pre-shared-key-v2/0024-pre-shared-key-v2.md
```diff
@@ -0,0 +1,266 @@
...
+
+If subsequent versions support additional encryption algorithms, those details should be exchanged between the sender and receiver when they establish the pre-shared key.
+
+### Condition and Fulfillment Generation
+
+The sender and receiver generate the condition and fulfillment from the pre-shared key that gives this protocol its name and the data that is sent with each payment. The fulfillment is an HMAC (using SHA256) of a key derived from the shared secret and the encrypted data.
```
dappelt: > The sender and receiver generate the condition and fulfillment from the pre-shared key that gives this protocol its name and the data that is sent with each payment.

If the reader is not already familiar with PSK, this sentence is confusing. Consider splitting this sentence up. The fulfillment is computed from the pre-shared key and payment data. The condition is the SHA256 hash of the fulfillment.

0024-pre-shared-key-v2/0024-pre-shared-key-v2.md
```diff
@@ -0,0 +1,266 @@
...
+
+**Trusting the Receiver:** When doing any Interledger payment aside from an atomic exchange of money for goods or services, the sender must trust the receiver to deliver the goods once they receive the payment. With chunked payments, the sender must trust the receiver to deliver the goods and accurately report how much money has arrived on their incoming transfer. A malicious receiver could keep reporting less money having arrived on every transaction in an attempt to extract more money from the sender, but this is no worse a threat than a receiver demanding more money after a non-chunked payment arrives but before they release the goods or services. Since senders will judge merchants' prices by the total source amount they pay, receivers that pull such tricks should develop negative reputations and be avoided.
+
+## Model of Operation
+
+### Informational Quote
+
+1. Sender creates a packet with a type of 1, a random Payment ID, a Sequence number of 0, a Payment Amount of 0, a Chunk Amount set to 18446744073709551615 (the maximum value) and [encrypts it](#encryption-envelope).
```
dappelt: 18446744073709551615 -> 2<sup>64</sup>-1

dappelt: (the maximum value) -> (max. unsigned 64 bit integer)

0024-pre-shared-key-v2/0024-pre-shared-key-v2.md
```diff
@@ -0,0 +1,266 @@
...
+5. Receiver checks that the incoming transfer amount is greater than or equal to the Chunk Amount and that the total that has arrived thus far does not exceed the Payment Amount.
+6. Receiver creates a response packet with the Type 2, the same Payment ID, Sequence. The Payment Amount is set to the total the receiver has received and fulfilled thus far (including the current chunk). Receiver sets the Chunk Amount to the amount that arrived on the incoming transfer. Receiver [encrypts the response](#encryption-envelope).
+7. Receiver generates fulfillment as described [below](#fulfillment-generation) and fulfills the incoming transfer.
+8. Sender is notified of the outgoing transfer being fulfilled and decrypts the attached data.
+9. Sender determines the exchange rate by dividing the Chunk Amount from the receiver by the chunk's source amount.
+10. Sender repeats step 1-8 until the fixed source or destination amount has been sent or delivered, respectively. Sender MAY adjust the chunk amount upwards or downwards in response to successful and unsuccessful payments. The sender SHOULD set the Type to 1 on the last chunk.
```
dappelt: The sender does not exactly repeat steps 1 to 8. At 1), the sender creates a new packet with the same `Payment ID` as before, increments `Sequence number` etc.

0024-pre-shared-key-v2/0024-pre-shared-key-v2.md
```diff
@@ -0,0 +1,266 @@
...
+9. Sender is notified of the outgoing transfer being fulfilled and can use the decrypted response to determine the amount that arrived at the destination.
+
+### Chunked Payment
+
+1. Sender creates a packet with Type 0, a random Payment ID, a Sequence number of 0 and a Chunk Amount of 0. In the case of a fixed destination amount, sender sets the Payment Amount to the destination amount. In the case of a fixed source amount, sender sets the Payment Amount to 18446744073709551615 (the maximum possible value). Sender [encrypts the request](#encryption-envelope).
+2. Sender generates the payment condition as described [below](#fulfillable-condition).
+3. Sender sends a payment with the condition from step 2., the encrypted request from step 1. as the data, and an arbitrary source amount.
+4. Receiver is notified of the incoming transfer and decrypts the data.
+5. Receiver checks that the incoming transfer amount is greater than or equal to the Chunk Amount and that the total that has arrived thus far does not exceed the Payment Amount.
```
dappelt: At step 1) a packet with Chunk Amount of 0 is created. So how can this constraint ever be violated:
> Receiver checks that the incoming transfer amount is greater than or equal to the Chunk Amount [...]

 The incoming transfer would need to have an amount < 0 to violate this constraint.

dappelt: I assume for consecutive packets the chunk amount would be greater than 0.

0024-pre-shared-key-v2/0024-pre-shared-key-v2.md
```diff
@@ -0,0 +1,266 @@
...
+4. Sender sends a payment using the source amount determined in step 2, the condition from step 3, and the encrypted request as the data.
+5. Receiver is notified of the incoming transfer and decrypts the data.
+6. Receiver checks to ensure that the incoming transfer amount is greater than or equal to the Chunk Amount specified in the request. If the Payment Amount is not equal to 0, the receiver checks that the transfer amount is less than or equal to the Payment Amount specified in the request.
+7. Receiver creates a [Payment Chunk Response](#payment-chunk-response) with the same Payment ID and Sequence number. The Payment Amount and Chunk Amount are set to the amount that arrived on the incoming transfer. The receiver then [encrypts](#encryption-envelope) the data.
+8. Receiver generates fulfillment as described [below](#fulfillment-generation) and fulfills the incoming transfer with the fulfillment and the encrypted response data.
+9. Sender is notified of the outgoing transfer being fulfilled and can use the decrypted response to determine the amount that arrived at the destination.
+
+### Chunked Payment
```
dappelt: Adding pseudo-code would make this more clear.

0024-pre-shared-key-v2/0024-pre-shared-key-v2.md
```diff
@@ -0,0 +1,266 @@
...
+### Encryption
+
+Request and response data are encrypted using [AES-256-GCM](https://en.wikipedia.org/wiki/Galois/Counter_Mode) with a 16-byte Initialization Vector (IV).
+
+#### Encryption Envelope
+
+```
+(Numbers represent bytes)
+
+ 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1 2 3 4 5 6 7 8 9 0 1
```
dappelt: If numbers represent bytes, this should start counting at 1. Also, it should not start over at 0 but should keep going 10, 11, etc.

Maybe better to use powers of 2 instead.

0024-pre-shared-key-v2/0024-pre-shared-key-v2.md
emschwartz: I could add that but I think it would be fairly long. It might be better to just read an actual implementation if you want to reimplement it

michielbdejong:
> My question is what must I write to make my implementation interoperable. For a minimal connector (with configured routes) you need [...]

This simplification makes sense for an initial network where all connectors trust all other connectors through configured routes,  but I would like to be more ambitious and hope that we can later open up that network, and allow random unvetted connectors to join. Then, we'll need to add back things like quote caching, maybe not on the interledger layer, but on the connector-to-connector layer.

Some of the decisions we are proposing now (e.g. relaxing the definition of ledger prefix and not including a `targetPrefix` in end-to-end quote responses) affect not only the initial closed network, but also the ability to open that network up later. So it's better to build in "trustlessness" from the start, even if it's not an MVP prerequisite.

Especially since we already have an implementation of it, into which years of work were invested and which does already support quote caching, liquidity routing, and route broadcasts with exchange rate information. And it would be easy to add this new experiment without throwing away all that existing work, right?

sharafian:
> ...and allow random unvetted connectors to join. Then, we'll need to add back things like quote caching

Can you explain why allowing unvetted connectors to join the network requires quote caching?

0024-pre-shared-key-v2/0024-pre-shared-key-v2.md
```diff
@@ -0,0 +1,287 @@
...
+
+- PSK version (such as `2.0` for this version)
+- A 32-byte random or pseudorandom shared secret
+- The receiver's ILP address
+
+The receiver MAY generate the shared secret from a longer-term secret and a nonce, as described in [PSKv1's Appendix A: Recommended Algorithm for Generating Shared Secrets](./0016-pre-shared-key/0016-pre-shared-key.md#appendix-a-recommended-algorithm-for-generating-shared-secrets). In this case, the receiver would append the nonce (**not** the secret) to their ILP address to create the destination address they communicate to the sender.
+
+### Encryption
+
+In PSKv2.0, request and response data is encrypted using [AES-256-GCM](https://en.wikipedia.org/wiki/Galois/Counter_Mode) with a 12-byte Initialization Vector (IV) and a 16-Byte Authentication Tag.
```
sharafian: From what source did you determine that those were the correct lengths to use? (I think PSK v1 uses a 16-byte IV)

0024-pre-shared-key-v2/0024-pre-shared-key-v2.md
```diff
@@ -0,0 +1,287 @@
...
+
+If subsequent versions support additional encryption algorithms, those details should be exchanged between the sender and receiver when they establish the pre-shared key.
```
sharafian: Is there any part of the PSK details that says what the encryption algorithm is, or is the expectation just that if you failed to exchange details at a higher level the data will appear to be corrupt

0024-pre-shared-key-v2/0024-pre-shared-key-v2.md
```diff
@@ -0,0 +1,287 @@
...
+
+When a sender is sending many payments to the receiver, they need a dynamic view of the exchange rate because the rate may change during the course of the interaction. To handle these cases, every PSKv2 payment response includes the amount that arrived in the final transfer to the receiver. Similar to the informational quotes, the sender can use the chunk source amount and the receiver's response to determine and monitor the exchange rate.
+
+**Trusting the Receiver:** With any type of Interledger quoting, whether end-to-end or using ILQP, senders SHOULD judge prices in their source units. Receivers can lie about how much arrives in a test payment for a quote or they could run their own ledger and connector that happen to have a bad exchange rate. Therefore, senders must assume that whatever amount the receiver says has arrived represents the "real" rate of the path.
```
sharafian: This is a really important and subtle point to grasp in order to understand Interledger. It might make sense to put this as a whole section in some kind of document describing how the ILP world differs from the ordinary world of payments, and explain why not attributing any meaning to any foreign units is A) necessary, and B) a good thing.

"Trusting the Receiver" seems like a strange header for this paragraph, because it describes why you can't trust the receiver.

0024-pre-shared-key-v2/0024-pre-shared-key-v2.md
```diff
@@ -0,0 +1,287 @@
...
+
+**Fluctuating Exchange Rates:** The rate from the sender to the receiver may change due to legitimate market movements during the course of a payment. Senders SHOULD monitor the rate changes and they may not want to send payments that are significantly larger than their bandwidth (this could be compared to how downloading a 4K movie over a dial-up connection is possible but would not make for a good experience).
+
+**Malicious Connectors:** Exchange rates can also change for illegitimate reasons, such as connectors driving up the price mid-payment. Ultimately, the main force keeping connectors' rates low is competition. If there is only one path, the connectors do not need to do anything sneaky to drive up the rate. If a sender has multiple options for the connector they send through, they may want to use [Multi-Path Chunked Payments](#multi-path-chunked-payments) to automatically switch paths if one becomes more expensive.
```
sharafian: What should the sender do if the only path available exceeds the rate that they're willing to spend? Should they just stand by until the rate comes back down?

0024-pre-shared-key-v2/0024-pre-shared-key-v2.md
```diff
@@ -0,0 +1,287 @@
...
+
+**Trusting the Receiver:** When doing any Interledger payment aside from an atomic exchange of money for goods or services, the sender must trust the receiver to deliver the goods once they receive the payment. With chunked payments, the sender must trust the receiver to deliver the goods and accurately report how much money has arrived on their incoming transfer. A malicious receiver could keep reporting less money having arrived on every transaction in an attempt to extract more money from the sender, but this is no worse a threat than a receiver demanding more money after a non-chunked payment arrives but before they release the goods or services. Since senders will judge merchants' prices by the total source amount they pay, receivers that pull such tricks should develop negative reputations and be avoided.
```
sharafian: It's worth pointing out that if the receiver colludes with a connector to have them raise their rates mid-payment, it's no different from the receiver deciding to raise the prices for their goods mid payment. If we don't point that out, then the Malicious Connector attack sounds a lot worse than it is.

michielbdejong:
When information about alternative routes arrives, you want to compare it to information you received earlier. That comparison cannot happen if you didn't cache the information.

0024-pre-shared-key-v2/0024-pre-shared-key-v2.md
```diff
@@ -0,0 +1,287 @@
...
+  + [Single Payment](#single-payment)
+  + [Chunked Payment](#chunked-payment)
+* [Specification](#specification)
+  + [Encryption](#encryption-1)
+  + [Data Packet](#data-packet)
+* [Appendix A: Advanced Sender Features](#appendix-a-advanced-sender-features)
+
+## Overview
+
+Like [PSKv1](./0016-pre-shared-key/0016-pre-shared-key.md), PSKv2 uses a secret shared between the sender and receiver to generate the conditions and fulfillments for payments without repeated end-to-end communication. PSKv2 additionally includes support for End-to-End Quoting and Chunked Payments.
```
sharafian: What kinds of payments will PSK2 be used for? Will we use it for everything, or will micropayments (which we expect to be under the MTU) use a different protocol? A lot of the points in this document are very different depending on whether we're thinking about a large payment or a small payment. Most of the time, it seems like this spec is describing retail-sized payments that require lots of chunking.

If it's also being used for micropayments, I think that deserves more mentioning. For instance, many of the attacks described really don't work against micropayments. And many of the quoting mechanisms aren't needed when you're just sending a 1-chunk micropayment.

0024-pre-shared-key-v2/0024-pre-shared-key-v2.md
```diff
@@ -0,0 +1,287 @@
...
+4. Receiver is notified of the incoming transfer and decrypts the data.
+5. Receiver checks that the incoming transfer amount is greater than or equal to the Chunk Amount and that the total that has arrived thus far does not exceed the Payment Amount.
+6. Receiver creates a response packet with the Type 2, the same Payment ID, Sequence. The Payment Amount is set to the total the receiver has received and fulfilled thus far (including the current chunk). Receiver sets the Chunk Amount to the amount that arrived on the incoming transfer. Receiver [encrypts the response](#encryption-envelope).
+7. Receiver generates fulfillment as described [below](#fulfillment-generation) and fulfills the incoming transfer.
+8. Sender is notified of the outgoing transfer being fulfilled and decrypts the attached data.
+9. Sender determines the exchange rate by dividing the Chunk Amount from the receiver by the chunk's source amount.
+10. Sender repeats step 1-8 with the following modifications, until the fixed source or destination amount has been sent or delivered, respectively:
+  - Sender MUST create a new PSK data packet for each chunk with the same Payment ID and an incremented Sequence number
```
sharafian: What's the purpose of the sequence number? What should the receiver do if they miss a sequence number? If one of the payments fails, should the sender increment the sequence number, or retry it with the same sequence number? Will an informational quote always have a sequence number of 0, even if it occurs partway through a payment (for example to refresh the exchange rate)?

0024-pre-shared-key-v2/0024-pre-shared-key-v2.md
```diff
@@ -0,0 +1,287 @@
...
+  - Sender MAY specify the minimum Chunk Amount the receiver should accept for each transfer based on the calculated path exchange rate
+  - Sender MAY adjust the transfer amount upwards or downwards in response to successful and unsuccessful payments, or to deliver a fixed destination amount exactly
+  - Sender SHOULD set the Type to 1 on the last chunk
```
sharafian: What if the sender didn't set the type to 1, but the chunk they sent completes the payment's destination amount. Would the receiver reject subsequent payments? Do they have a way of telling the sender that they've completed the payment already?

What if the sender sets the type to 1, but the destination amount isn't complete yet? Does the receiver just take it to mean that the sender won't pay the rest of what they owe?

0024-pre-shared-key-v2/0024-pre-shared-key-v2.md
```diff
@@ -0,0 +1,287 @@
...
+
+## Specification
+
+### Encryption
+
+Request and response data are encrypted using [AES-256-GCM](https://en.wikipedia.org/wiki/Galois/Counter_Mode) with a random 12-byte Initialization Vector (IV).
```
sharafian: and 16 byte auth tag

0024-pre-shared-key-v2/0024-pre-shared-key-v2.md
dappelt: https://github.com/interledger/rfcs/pull/351#discussion_r156799912

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -119,8 +103,9 @@ The response body is a JSON object that includes basic account details necessary
 
 | Field | Type | Description |
 |---|---|---|
+| `psk_versions` | Array of Strings | PSK versions supported by the server (written as `2.0`, indicating the major and minor version) |
 | `destination_account` | [ILP Address](../0015-ilp-addresses/0015-ilp-addresses.md) | ILP Address of the receiver's account |
-| `shared_secret` | 32 bytes, [base64-url encoded](https://en.wikipedia.org/wiki/Base64#URL_applications) | The shared secret to be used by this specific http client in the [Pre-Shared Key protocol](../0016-pre-shared-key/0016-pre-shared-key.md). Should be shared only by the server and this specific http client, and should therefore be different in each query response. |
+| `shared_secret` | 32 bytes, Base64-Encoded | The shared secret to be used by this specific http client in PSK. Should be shared only by the server and this specific http client, and should therefore be different in each query response. |
```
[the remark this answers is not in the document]
emschwartz: Yes padding, even though it's kind of silly

0024-pre-shared-key-v2/0024-pre-shared-key-v2.md
emschwartz: Correct, you should exchange that out of band and you won't be able to make sense of the data if you haven't agreed on that.

0024-pre-shared-key-v2/0024-pre-shared-key-v2.md
emschwartz: The main purpose of the sequence number right now is to allow the sender to ensure that responses match the requests they sent out. Without the sequence, a connector could attach an old response (that would say a lower amount had arrived) to a later fulfillment and the sender wouldn't know the difference.

Originally I thought that receivers should keep track of the sequence numbers to avoid accepting two with the same sequence. @justmoon argued that wasn't necessary because unlike with data, money is fungible so it doesn't matter if the chunks arrive out of order as long as you get the total amount.

I don't think informational quotes need to have a sequence of 0. From the receiver's perspective, there is no such thing as an informational quote, it's just a chunk that arrived with too little money (and potentially an unfulfillable condition).

0024-pre-shared-key-v2/0024-pre-shared-key-v2.md
emschwartz: > What if the sender didn't set the type to 1, but the chunk they sent completes the payment's destination amount. Would the receiver reject subsequent payments?

Yes, the receiver should reject further chunks once the destination amount specified has arrived

> What if the sender sets the type to 1, but the destination amount isn't complete yet? 

The receiver should assume that means the sender won't send any more. I think the PSK implementation would need to turn it over to the Application Layer protocol at that point to figure out what to do

emschwartz:
> When information about alternative routes arrives, you want to compare it to information you received earlier. That comparison cannot happen if you didn't cache the information.

@michielbdejong this seems orthogonal to PSKv2, no?

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -119,8 +103,9 @@ The response body is a JSON object that includes basic account details necessary
 
 | Field | Type | Description |
 |---|---|---|
+| `psk_versions` | Array of Strings | PSK versions supported by the server (written as `2.0`, indicating the major and minor version) |
```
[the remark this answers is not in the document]
emschwartz: We could. The main argument I see against that would be: what if we add a PSKv2.1 that supports ciphers other than AES-GCM? Would we really need to do a major version bump for SPSP just to support that?

0024-pre-shared-key-v2/0024-pre-shared-key-v2.md
emschwartz: Note that the [Rust](https://briansmith.org/rustdoc/ring/aead/static.AES_256_GCM.html), [Go](https://golang.org/pkg/crypto/cipher/#AEAD), and [Java](https://stackoverflow.com/questions/31851612/java-aes-gcm-nopadding-what-is-cipher-getiv-giving-me) implementations either require or recommend using a 96-bit IV as well.

emschwartz:
(@justmoon PSKv2 is currently slated to use IL-RFC number 24 but I can change that to 25)

PR #355, opened 2017-12-14
docs(0003): add rejection packet

justmoon:


michielbdejong:
This looks like a breaking change compared to the ilp-kit version which was released earlier this week. Should we maybe keep this PR open until it's implemented? Or should we just decouple the rfcs repo from the reference implementation again, to allow more rapid iteration?

justmoon:
I'm fine holding off on merging this until the implementation is there.

michielbdejong:
Actually, it's not as critical as I first thought, we do need a major version for ilp-kit, but we don't need to break the network for it, because a v5 ilp-kit can know that it's talking to a v4 peer, and then still send (legacy) type 8 instead of type 11.

0003-interledger-protocol/0003-interledger-protocol.md
```diff
@@ -245,8 +245,49 @@ Amount in discrete units of the receiving ledger's asset type. Note that the amo
...
+Invalid destination. destination=example.us.bob
+```
+
+This field MUST be encoded as UTF-8.
+
+#### data
+
+    OCTET STRING (SIZE(0..32767))
+
+Machine-readable data. The format is defined for each error code. Implementations MUST follow the correct format for the code given in the `code` field.
```
sentientwaffle: Where are each code's data formats defined? I don't see any mention of it at https://github.com/interledger/rfcs/blob/cecf8d2f6b50556bbb28427caac7ee29c6ed97f1/0003-interledger-protocol/0003-interledger-protocol.md#ilp-error-codes.

justmoon: I remembered that @emschwartz had already added a column for that in #318, so I'm planning to copy it from there. Will add that before merging. Note that currently all errors have empty data.

michielbdejong:
> decouple the rfcs repo from the reference implementation again, to allow more rapid iteration?

This ended the May-December 2017 period in which the rfcs repo described what ilp-kit implements.

Issue #359, opened 2017-12-22
Conditions are NOT a ledger layer concern


justmoon:
Brought to you from our sponsors at holy ****, there are still ways to simplify ILP...

This is a follow-on to @adrianhopebailie's famous ticket #270. He was the first to recognize that executionConditions awkwardly straddle two layers: Interledger and ledger. This dissonance stuck with me until it lead me to the realization described in this ticket, which finally resolves it.

---

We recently started seriously implementing an earlier realization we dubbed the Interledger enlightenment. Before the enlightenment, we thought that ledgers were the nodes and connectors were the edges of the financial graph. But we realized that a simpler and more general model is that the nodes in the Interledger graph are senders, connectors and receivers and the edges are accounts they have with one another. A ledger then is just another type of connector.

Interestingly, there aren't a lot of different types of account you can have. You can have a credit relationship, which we sometimes call a trustline. Or you can have a payment channel, which essentially the crypto-equivalent of a check: One party signs something called a claim that the other party can take to someone else (such as a blockchain) to get paid. Checks reduce the trust needed between the parties because the settlement is enforceable.

In ILP, we use unsecured payment channels. What we do is we transfer a claim after receiving a fulfillment. In other words, for us, a payment channel is just a very efficient way to send some money. Note that there is no condition involved.

We believe that such unsecured payment channels are fine, because each payment is very small, so it is not worth worrying about losing a payment, as long as the cumulative loss is limited.

If we use unconditional transfers between each pair of nodes, then how come we haven't gotten rid of conditions entirely? (We call this "optimistic mode" in ILP speak.)

The reason we still have conditions is to isolate risk to a specific hop. I.e. if someone along the chain fails, we want to be able to assign blame correctly. That's why the two phase execution still makes sense.

**But if conditions are only evaluated by connectors and ledger plugins only do unconditional transfers, then conditions are an Interledger concern, not a ledger layer concern.**

In other words, the LPI `sendTransfer` should be changed from:

``` ts
interface Transfer {
  amount: string,
  executionCondition: Buffer,
  expiresAt: string,
  ilp: Buffer
}

interface FulfillmentInfo {
  fulfillment: Buffer,
  ilp: Buffer
}

class Plugin {
  sendTransfer(transfer: Transfer): FulfillmentInfo
}
```

``` js
const { fulfillment, data: ilpFulfillmentPacket } = await plugin.sendTransfer({
  amount: '10',
[...]

PR #360, opened 2017-12-22
docs(0023): unconditional transfer type for ilp4

sharafian:
Adds unconditional transfer type to accommodate https://github.com/interledger/rfcs/issues/359 . Related to https://github.com/interledgerjs/btp-packet/pull/18 .


michielbdejong:
Wouldn't it be more common for the party who does the on-ledger or on-paychan payment, to send a BTP message saying "My BTP balance just went up because I paid you on-ledger, look here are the details" instead of what currently seems to be a call type to notify your peer that "Your BTP balance just went up because you paid me on-ledger, look here are the details"

michielbdejong:
@sharafian see also https://github.com/interledger/rfcs/pull/347#issuecomment-355498032

sharafian:
sendMoney makes your balance go down, because you're sending money to your peer. A BTP connection/relationship no longer has a balance, but the `Transfer` call is used to carry information about an on-ledger (in this context payment channels count as "on ledger" because they're secured by the ledger) transfer. I believe it's pretty similar to your vouch protocol in that way

michielbdejong:
> A BTP connection/relationship no longer has a balance, but the Transfer call is used to carry information

If you have one call registry where `Prepare` sends a conditional transfer, and `Transfer` sends an unconditional one, then it would be super-confusing, if their amounts work in opposite directions (`Prepare` conditionally increases the sender's debt, `Transfer` unconditionally lowers that same debt)

michielbdejong:
How about 'Settle' instead of 'Transfer'?

sharafian:
Transfer and Prepare aren't in the same call registry though. From the plugin's point of view, sendData just sends data. The `Transfer` call is the only one that involves money.

sharafian:
I wouldn't be hugely opposed to Settle instead of Transfer, but having to refactor around naming is always a little annoying.

michielbdejong:
> Transfer and Prepare aren't in the same call registry

They are, call types 3 and 7, see https://github.com/interledger/rfcs/pull/360/files#diff-2c87a4d4e57dd9430f40e0fe2b4d295aR114

michielbdejong:
sendData is the LPI method which causes a Prepare to be sent, it increases the sender's debt.
sendMoney is the LPI method which causes a Transfer to be sent, it decreases the sender's debt.

emschwartz:
> sendData is the LPI method which causes a Prepare to be sent, it increases the sender's debt.

Wouldn't it be an `IlpPrepare` sent over a BTP Type 6 `Message`? I don't think we'll be using the BTP `Prepare` anymore

michielbdejong:
Ah ok, then `Transfer` is a correct name. But if we go in that direction, then we should put `Fulfill` and `Reject` into `Response` packets, and we need to mark the three existing calls as deprecated, otherwise it will be very confusing. And this would be a breaking change, of which this PR by itself seems to describe just a small part, right?

> A BTP connection/relationship no longer has a balance

I think a good way to explain the new situation is to avoid the term 'balance', and say the `Message` calls can increase your `unsecuredAmount`, while `Transfer` calls decrease it.

If you use BTP as just a trustline, then the `unsecuredAmount` plays the role of a "balance that shall not be named"; the reason we should not call it a balance is that it works in the opposite direction: unsecured amount goes up when you send Interledger payments, and it goes down when you receive them.

michielbdejong:
I would propose we keep putting the payment channel claim updates in Transfer calls for the 18Q1 version, since that's what we've already been implementing, and I want to put this live on the testnet today.

Then we can always move them to Message calls in the 18Q2 version.

emschwartz:
Maybe the real benefit of the Transfer type is so that we can clearly distinguish between the `sendData` and `sendMoney` methods in LPI2

michielbdejong:
> clearly distinguish between the sendData and sendMoney methods in LPI2

The protocolName would be 'ilp' for sendData, and something else (e.g. 'paychan') for sendMoney.

emschwartz:
So the only difference when using the `Transfer` is that there is an amount field that you need to check against the claim that you've gotten to make sure they match? That sounds like it adds a potential footgun without any benefit, no?

sharafian:
Currently, it's used as an extra check to see if the balance is in dispute (the `TRANSFER` amount says how much the peer thinks this claim increases the balance). We could switch over to using `MESSAGE` without much trouble, though.

emschwartz:
What happens if they are out of sync? Would it just log a warning to the console? Doesn't seem like there's much you would actually do about that

sharafian:
Yeah, it just logs a warning to make you aware of it

michielbdejong:
Sorry, I realize my earlier remark was confusing. I wrote:
> we can always move them to Message calls in the 18Q2 version

As @emschwartz remarked on Slack, deprecating the version that uses the `Transfer` call can be done on 1 April 2018, but adding the newer version can already be done now! :)

@sharafian wrote:
> We could switch over to using MESSAGE without much trouble, though.

Sounds good, let's do it! :)

michielbdejong:
@sharafian I'm merging this so I can continue https://github.com/interledger/rfcs/pull/383 on top of it, see further discussion there.

PR #362, opened 2017-12-24
Add RFC to explain OER choice and how to implement it

justmoon:
As requested by @emschwartz, I wrote up an RFC draft on why we chose OER. The idea is that other specs could reference this document when talking about things like the length determinant.

This document will likely go through several revisions over time to make it more complete and more readable.

---

While working on this I noticed that it would be good not to require a non-zero length for all ILP addresses, because we may want to use OER in contexts where zero-length addresses make sense, e.g. the default route (empty prefix). I patched the ILP payment packets to retain their previous definition.


0000-notes-on-oer-encoding/0000-notes-on-oer-encoding.md
```diff
@@ -0,0 +1,190 @@
...
+
+## Basics
+
+### Octet strings
+
+All Interledger formats are byte-aligned messages and should be passed as octet strings, i.e. sequences of bytes. The correct type for this would be a `Buffer` in JavaScript, `byte[]` in Java, `bytes` in Python, etc.
+
+### Extensibility
+
+For most Interledger formats, we require that implementations MUST ignore extra bytes at the end of the message. This behavior allows additional fields to be added in the future that are automatically ignored by existing implementations.
```
sharafian: end of the message == end of the message inside the envelope. Extra bytes after the ILP packet would be considered the next message, right?

0000-notes-on-oer-encoding/0000-notes-on-oer-encoding.md
```diff
@@ -0,0 +1,190 @@
...
+| UInt64 | 8 bytes | 0 .. 18446744073709551615 (2^64 - 1) |
+| UInt128 | 16 bytes | 0 .. 2^128 - 1 |
+| UInt160 | 20 bytes | 0 .. 2^160 - 1 |
+| UInt192 | 24 bytes | 0 .. 2^192 - 1 |
+| UInt224 | 28 bytes | 0 .. 2^224 - 1 |
+| UInt256 | 32 bytes | 0 .. 2^256 - 1 |
+| UInt384 | 48 bytes | 0 .. 2^384 - 1 |
+| UInt512 | 64 bytes | 0 .. 2^512 - 1 |
+
+### Fixed-length signed integers
```
sharafian: Do the Interledger protocols actually use any signed integers?

0000-notes-on-oer-encoding/0000-notes-on-oer-encoding.md
```diff
@@ -0,0 +1,190 @@
...
+All fixed-length signed integers are encoded in big-endian, least-significant bit last. Negative values are encoded using 2's complement binary encoding.
+
+| Type | Size | Range |
+| :--- | :--- | :--- |
+| Int8 | 1 byte | -128 .. 127 |
+| Int16 | 2 bytes | -32768 .. 32767 |
+| Int32 | 4 bytes | -2147483648 .. 2147483647 |
+| Int64 | 8 bytes | -9223372036854775808 .. 9223372036854775807 |
+
+### Fixed-length floating point numbers
```
sharafian: We don't use these either

0000-notes-on-oer-encoding/0000-notes-on-oer-encoding.md
```diff
@@ -0,0 +1,190 @@
...
+
+All variable-length fields are prefixed with a length determinant. The length determinant can be either short form or long form.
+
+#### Short form
+
+A single byte in the range 0 .. 127, denoting the length.
+
+#### Long form
+
+A single byte in the range 128 .. 255, denoting the value *128 + n*, where *n* is the length-of-length. This is followed by an unsigned integer of *n* bytes denoting the length. The length MUST NOT contain leading zeros and MUST be greater than 127.
```
sharafian: Might be a little confusing here; the length-of-length needn't be greater than 127, but the length-of-length long integer containing the length of the data must be greater than 127

0000-notes-on-oer-encoding/0000-notes-on-oer-encoding.md
```diff
@@ -0,0 +1,190 @@
...
+
+### Variable-length octet string
+
+A variable length octet string consists of a length determinant, followed by that many bytes.
+
+### Variable-length unsigned integer
```
sharafian: Not used in ILP

0000-notes-on-oer-encoding/0000-notes-on-oer-encoding.md
```diff
@@ -0,0 +1,190 @@
...
+
+A variable length unsigned integer consists of a length determinant, followed by a big-endian integer of that many bytes. Leading zeros MUST NOT be used.
+
+### Variable-length signed integer
```
sharafian: Not used in ILP

0000-notes-on-oer-encoding/0000-notes-on-oer-encoding.md
```diff
@@ -0,0 +1,190 @@
...
+
+All Interledger formats are byte-aligned messages and should be passed as octet strings, i.e. sequences of bytes. The correct type for this would be a `Buffer` in JavaScript, `byte[]` in Java, `bytes` in Python, etc.
+
+### Extensibility
+
+For most Interledger formats, we require that implementations MUST ignore extra bytes at the end of the message. This behavior allows additional fields to be added in the future that are automatically ignored by existing implementations.
+
+## Types
+
+### Fixed-length unsigned integers
```
sharafian: Only a few of these integer formats are required. We should just specify the minimum formats in order to implement ILP. I believe we only need UInt8, UInt64, and UInt256.

0000-notes-on-oer-encoding/0000-notes-on-oer-encoding.md
```diff
@@ -0,0 +1,190 @@
...
+
+### Why CANONICAL-OER?
+
+For Interledger, we wanted to use the power of ASN.1 in order to design and validate our protocols and to avoid any ambiguity for implementors. However, we also wanted it to be easy to implement Interledger encodings, without having to first learn a bunch of ASN.1 and related technology. OER gives us the best of both worlds, because its encodings are extremely simple and can easily be described and implemented without referencing ASN.1 or OER. Yet, we can use standard tools to generate test data and validate our implementations. And we can refer to the ASN.1/OER specs to resolve any ambiguities that are discovered later.
+
+OER is designed to strike a better compromise between simplicity, speed and size than previous ASN.1 encoding rules. All OER types are byte-aligned, which makes them easier to serialize and parse in most programming languages than Packed Encoding Rules (PER) and also faster on most platforms.
+
+We chose CANONICAL-OER instead of BASIC-OER because we wanted a canonical encoding format so that Interledger encodings could be used for cryptographic operations like hashing and signing.
+
+## Basics
```
sharafian: I think each of these should include an example of what the type's ASN.1 syntax looks like. That way someone reading the ASN.1 docs and making a parser for the ILP types will better understand what's going on.

0000-notes-on-oer-encoding/0000-notes-on-oer-encoding.md
justmoon: I was considering talking about ILP packets here, but this is the general OER guide, so I opted not to. I think anything that's specific to one protocol should go into that protocol's spec.

0000-notes-on-oer-encoding/0000-notes-on-oer-encoding.md
justmoon: Disagree. This isn't just meant for ILP, it's also meant for BTP, PSKv2 and any future protocols in the Interledger ecosystem.

Here's my idea for a compromise: Why don't I add a "used in" table for each type, so if you're writing a serdes just for ILP packets you know which types to ignore.

0000-notes-on-oer-encoding/0000-notes-on-oer-encoding.md
```diff
@@ -0,0 +1,285 @@
+---
+title: Notes on OER Encoding
+draft: 1
+---
+
+# Introduction
+
+The Interledger stack consists of a collection of different protocols. In order to keep some consistency between these separate, but related protocols, the Interledger team has decided to use some common choices for encoding formats across at least those protocols we control.
+
+These choices will not be applicable for all protocols in the Interledger ecosystem. For example, we used a binary encoding while a text-based encoding may be more suitable. We used an encoding with a fixed schema, while a schema-less encoding may be easier to extend by third-parties.
```
emschwartz: "For example, we used a binary encoding while a text-based encoding may be more suitable" -> "...suitable for all use cases"

0000-notes-on-oer-encoding/0000-notes-on-oer-encoding.md
```diff
@@ -0,0 +1,285 @@
...
+
+## CANONICAL-OER
+
+### What is OER/CANONICAL-OER?
+
+The Octet Encoding Rules (OER) are a set of simple byte-aligned encoding rules for ASN.1 types.
+
+CANONICAL-OER narrows BASIC-OER to a single canonical encoding.
+
+### Why CANONICAL-OER?
```
emschwartz: Could you also add a section at the beginning or end describing the other encodings we considered and why we decided against each of them?

0000-notes-on-oer-encoding/0000-notes-on-oer-encoding.md
```diff
@@ -0,0 +1,285 @@
+---
+title: Notes on OER Encoding
+draft: 1
+---
+
+# Introduction
```
emschwartz: I think it would be useful to structure this article like a Wirecutter review. The top should say very quickly what we picked and why it's great, then it should explain why you should trust this decision and go into the details.

0000-notes-on-oer-encoding/0000-notes-on-oer-encoding.md
```diff
@@ -0,0 +1,285 @@
...
+
+Using a formalized language like ASN.1 to describe one's protocol means that one can use automated tools to verify the consistency and completeness of one's specification. This is a very valuable tool. It also means that one can refer to existing specifications which already cover many edge cases. Otherwise, each protocol would have to answer all possible edge cases itself, which quickly becomes tedious.
+
+ASN.1 is by far the most widely used formalized language for defining protocols. As an ISO and ITU standard, it is used in telecommunications protocols, cryptography standards and many more.
+
+The Interledger team chose ASN.1 because of it's long history and robust ecosystem. It may not be the most modern of standards, but it is extremely powerful and comes with huge amounts of implementation experience. Also, there exist high-quality tools for protocol designers.
+
+## CANONICAL-OER
+
+### What is OER/CANONICAL-OER?
```
emschwartz: What I'm missing from the intro up to here is how awesome it is that the only special thing you really need to implement to support this format is for variable-length fields. Everything else is encoded in a way that most programming languages support out of the box. I'm not sure you could say that for any other binary format

0000-notes-on-oer-encoding/0000-notes-on-oer-encoding.md
emschwartz: Also worth explaining somewhere why we used a binary format

0000-notes-on-oer-encoding/0000-notes-on-oer-encoding.md
```diff
@@ -0,0 +1,285 @@
+---
+title: Notes on OER Encoding
```
emschwartz: "Why Interledger Uses OER and How To Implement It"? Or is that too long?

0000-notes-on-oer-encoding/0000-notes-on-oer-encoding.md
```diff
@@ -0,0 +1,285 @@
...
+
+Abstract Syntax Notation One (ASN.1) is an international standard for describing abstract data types. ITU gives the following definition:
+
+> Abstract Syntax Notation number One
+> is a standard that defines a formalism
+> for the specification of abstract data types.
+
+In plain English, ASN.1 describes what your data can be, without making any statements about how it could be encoded in bits and bytes.
+
+### Why ASN.1?
```
emschwartz: A lot of developers have bad associations with ASN.1 and I remember you ascribing most of that to BER and DER. It might be worth mentioning your argument that OER basically fixes most of the things people hated before about the ASN.1 family

0000-notes-on-oer-encoding/0000-notes-on-oer-encoding.md
emschwartz: I wonder if it would be worth including a pseudocode parser at the end of this RFC to show how easy it is to implement and how the only thing that isn't a standard language function is the variable-length strings and integers

0000-notes-on-oer-encoding/0000-notes-on-oer-encoding.md
adrianhopebailie: I think this doc should talk about SEQUENCES and how these are used to define most compound types in Interledger protocols. 

For extensibility these will often contain an extensibility byte as the last element in the sequence. This will be `00` (indicating a length of zero) for most data sequences but where a non-zero value is encountered this should be interpreted as a length indicator and the subsequent `n` bytes (where `n` is the length specified by the length indicator) can be ignored.

0000-notes-on-oer-encoding/0000-notes-on-oer-encoding.md
```diff
@@ -0,0 +1,285 @@
+---
+title: Notes on OER Encoding
+draft: 1
+---
+
+# Introduction
+
+The Interledger stack consists of a collection of different protocols. In order to keep some consistency between these separate, but related protocols, the Interledger team has decided to use some common choices for encoding formats across at least those protocols we control.
```
michielbdejong: I would describe the "Interledger community" as made up of potentially several "teams"?

0000-notes-on-oer-encoding/0000-notes-on-oer-encoding.md
```diff
@@ -0,0 +1,285 @@
...
+| Float32 | `3F8FCD36` | `1.12345` (approximately) |
+| Float64 | `3FF1F9A6B50B0F28` | `1.12345` (approximately) |
+
+### Fixed-length octet string
+
+> Used in: ILP, BTP
+
+A fixed length octet string is encoded as itself with no added prefix or suffix.
+
+### Length determinant
```
emschwartz: Based on https://github.com/interledger/rfcs/pull/387#discussion_r167688168, this doc should also include a mention of open types and where they're used in ILP

justmoon:
Note to self: Incorporate this note on open types: https://github.com/interledger/rfcs/pull/387#discussion_r167688168

michielbdejong:
I think https://github.com/interledgerjs/ilp-packet/issues/13 is an issue @adrianhopebailie now ran into independently, and more people probably will in the future. Maybe it's worth mentioning this slightly unintuitive issue in this RFC?

justmoon:
Yeah, I left a comment here to remind myself to add a note about open types, but then I saw that @emschwartz beat me to it. :point_up:

0000-notes-on-oer-encoding/0000-notes-on-oer-encoding.md
```diff
@@ -0,0 +1,285 @@
...
+
+# Foundation
+
+Interledger encodings are based on and compliant with Abstract Syntax Notation One (ASN.1) using Canonical Octet Encoding Rules (CANONICAL-OER).
+
+## ASN.1
+
+### What is ASN.1?
+
+Abstract Syntax Notation One (ASN.1) is an international standard for describing abstract data types. ITU gives the following definition:
```
dora-gt: What is ITU? (International Telecommunication Union?)

0000-notes-on-oer-encoding/0000-notes-on-oer-encoding.md
```diff
@@ -0,0 +1,285 @@
...
+
+The Interledger team chose ASN.1 because of it's long history and robust ecosystem. It may not be the most modern of standards, but it is extremely powerful and comes with huge amounts of implementation experience. Also, there exist high-quality tools for protocol designers.
+
+## CANONICAL-OER
+
+### What is OER/CANONICAL-OER?
+
+The Octet Encoding Rules (OER) are a set of simple byte-aligned encoding rules for ASN.1 types.
+
+CANONICAL-OER narrows BASIC-OER to a single canonical encoding.
```
dora-gt: I could not understand what "BASIC-OER" means.
It will be so helpful to give a reference to "BASIC-OER".

0000-notes-on-oer-encoding/0000-notes-on-oer-encoding.md
dora-gt: It will definitely help to show sample of entire stream.
e.g. sample ASN.1 code and how it is encoded by OER.

adrianhopebailie:
As this is an informational document I think we should merge it and make it available as soon as possible then address the many comments provided as improvements.

I don't think any of the comments point out errors in the doc so there is no risk that publishing will result in spreading incorrect info.

@emschwartz @sharafian if you agree can you update your reviews so we can merge and then we can log the outstanding changes requested as an issue.

sharafian:
Everything looks like it's been addressed; is there anything keeping us from merging this?

PR #364, opened 2017-12-30
Enlighten RFC 1

justmoon:
Update RFC 1 to reflect latest thinking. I.e. ledgers do not implement escrow functionality. Rather, it is used by Interledger to localize risk.


0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -1,82 +1,91 @@
 ---
 title: Interledger Architecture
-draft: 1
+draft: 2
 ---
 # Interledger Architecture
 
 Interledger provides for secure payments across multiple assets on different ledgers. The architecture consists of a conceptual model for interledger payments, a mechanism for securing payments, and a suite of protocols that implement this design.
```
michielbdejong: Maybe find a different phrasing for 'secure payments' / 'mechanism for securing payments'

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -1,82 +1,91 @@
...
+Senders typically choose a connector and send all their traffic through this connector. If they are at any point dissatisfied with their connector, they may switch to a different one. Sophisticated senders may operate a connector of their own and maintain relationships with multiple connectors to minimize costs and maximize availability.
 
-Connectors *peer* with one another to exchange information used to determine the best route for a payment.
+### Receivers
+
+A *receiver* is a system which accepts incoming Interledger packets and provides a cryptographic receipt for each packet which is returned to the sender along the same path. For more information on the cryptographic exchange, see [Payment Flow](#payment-flow).
 
 **Interledger ensures that senders' funds are safe throughout an multi-hop payment and cannot be stolen by faulty or malicious connectors (see [Interledger Security](#interledger-security)).**
```
michielbdejong: This safety now comes from the small chunk size, right? Since transfers now go over accounts, not ledger, a faulty or malicious connector *can* steal a chunk, but only one, and when it does steal a chunk, it would immediately lose traffic to competitors.

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -1,82 +1,91 @@
...
 
 ### The Interledger
 
-The Interledger protocol suite may be used to transact across any ledgers and connectors, whether they are public or private. There is no single network that all parties must connect to to use these protocols.
+The "Interledger protocol suite" can be used among any peered group of connectors, be it are public or private. There is no single network that all parties must connect to to use these protocols.
```
michielbdejong: "be it are"

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -1,82 +1,91 @@
...
 
-"The Interledger" is a conceptual network made up of independent and diverse ledgers linked by connectors. Each account "on the Interledger" is part of a particular ledger, but they may transact with others by sending interledger payments through different ledgers and connectors. Like "the Internet", the Interledger is not a single network but is comprised of multiple interconnected networks.
+"The Interledger" is the name of a public instance of the Interledger protocol suite which seeks to provide a coherent global financial instructure. Ideally, anyone connected to the Interledger should be able to transact with anyone else, each using their currency and connector of choice.
```
michielbdejong: "instructure"

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -1,82 +1,91 @@
...
 
 <span class="show alert alert-info">**Hint:** Conditional transfers or *authorization holds* are the financial equivalent of a [two-phase commit](http://foldoc.org/two-phase%20commit).</span>
 
-### Conditional Transfers
+### Interledger Protocol Flow
 
-Each local transfer is first *prepared* and then either *executed* or *rejected*. When a transfer is prepared, the ledger puts the funds of the source account on hold with a *cryptographic condition* and *timeout*. If the condition is fulfilled before the timeout, the transfer is executed and the funds are transferred. If the timeout is reached, the transfer expires and the ledger returns the funds to the source account automatically.
+The sender first sends an Interledger `prepare` packet. Each connector may either forward the `prepare` packet, or, in the case of an error, return a `reject` packet. If the `prepare` packet receiver, they may respond with either a `fulfill` or `reject` packet, depending on whether they wish to accept the Interledger packet. In either case, the resulting packet is forwarded along the same path back to the sender.
```
michielbdejong: If the `prepare` packet [?] receiver -> missing "reaches the"?
accept the Interledger packet -> "accept the Interledger payment"?
forwarded along the same path back -> "passed along the same path back"?

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -1,82 +1,91 @@
...
 
-Inspired by the [Lightning Network](http://lightning.network), Interledger uses the digest of the [SHA-256](https://en.wikipedia.org/wiki/SHA-2) hash function as the condition for transfers. The fulfillment is a valid 32-byte preimage for the hash specified when the transfer was prepared. Ledgers are responsible for validating fulfillments. [Transport Layer](#transport-layer) protocols are used by the sender and receiver to generate the condition for a particular payment.
+ Each `prepare` packet contains a *cryptographic condition* and *timeout*. If a connector receives and forwards the `fulfill` packet before the timeout, the `prepare` packet is considered fulfilled, creating an obligation for the peer that sent the packet to settle with the connector. If the timeout is reached before the `fulfill` packet is received, the connector will consider the `prepare` packet expireds and no obligation is created.
```
michielbdejong: "expireds"

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -1,82 +1,91 @@
...
 
-### Payment Flow
+Settlement may either be based solely on trust between the connector and its peers, or be supported with functionality on the settlement ledger. See [IL-RFC 17](../0017-ledger-requirements/0017-ledger-requirements.md) for the full description and tiers of ledger requirements.
 
-In Interledger payments, all component transfers are prepared before any are executed. No funds are transferred, so none can be lost if a connector fails or attempts to redirect the payment.
+As the `prepare` packet makes its way to the receiver, connectors only reserve the appropriate amount of funds, they do not settle it yet. No funds are transferred, so none can be lost if a connector fails or attempts to redirect the packet.
```
michielbdejong: Instead of "No funds are transferred" we could say "No obligation is agreed yet between the peers" since this is now about accounts and peer obligations, not on-ledger transfers.

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -1,82 +1,91 @@
...
 
-When the *receiver* is notified of funds on hold for them, they submit the *fulfillment* of the cryptographic condition to claim their funds. Each connector uses the same fulfillment to claim their incoming transfer.
+When the *receiver* receives the `prepare` packet, they return the `fulfill` packet to claim their funds. Each connector uses the same `fulfill` packet to claim their funds respectively from the party (sender or connector) that sent them the `prepare` packet.
```
michielbdejong: Not sure if the phrase "claim their funds" is accurate here - the receiver does establish a claim, but the last connector does not atomically pay out on-ledger.

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -1,82 +1,91 @@
...
 
-The timeout of each successive transfer is shorter than the previous one, giving each connector a window of time to deliver the fulfillment even if their outgoing transfer was executed at the last possible moment.
+When forwarding a `prepare` packet from the sender-side account to the receiver-side account, each connector will reduce the timeout by a fixed amount. This provides with a fixed time window to pass back the `fulfill` packet from the receiver-side account to the sender-side account later. A connector must have a fair chance to pass the `fulfill` packet back before the timeout on the sender-side account is reached, even if the `fulfill` packet was received on the receiver-side account at the last possible moment. Otherwise, it would be obligated to pay out on the receiver-side account with no obligation on their counterparty on the sender-side account to pay them back.
```
michielbdejong: "provides [with] a"

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -1,82 +1,91 @@
...
 
 For more details on the flow, see the [Interledger Protocol specification](../0003-interledger-protocol/0003-interledger-protocol.md) and the [Interledger whitepaper](https://interledger.org/interledger.pdf).
 
 <span class="show alert alert-info">**Note:** Interledger only supports Universal mode as described in the whitepaper. Atomic mode can be used by adjacent subsets of participants in an Interledger payment if desired, but this is not part of the standard.</span>
```
michielbdejong: maybe it's time for a new whitepaper?

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -1,82 +1,91 @@
...
 
 ### Connector Risk and Mitigation
 
-Interledger connectors accept some risk in exchange for the revenue they generate from facilitating payments. In the Interledger payment flow, connectors' outgoing transfers are executed before their incoming transfers. After each connector is notified that the outgoing transfer has been executed, they have a window of time to deliver the fulfillment and execute the incoming transfer. Connectors that fail to deliver the fulfillment in time may lose money.
+Interledger connectors accept some risk in exchange for the revenue they generate from facilitating payments. In the Interledger payment flow, connectors incur outgoing obligations on the receiver-side account, before they become entitled to incoming obligations on the sender-side account. After each connector receives a `fulfill` packet, they have a window of time to deliver the `fulfill` packet to their counterparty on the sender-side account. Connectors that fail to deliver the `fulfill` packet in time may lose money.
```
michielbdejong: May be necessary here to describe who keeps time, now that there's no longer a third-party ledger for that. I think it should always be the receiving connector, so that you don't need a 'fulfillment too late' error. The sending connector can then decide to unpeer if in practice it's often too late to pass back fulfillments from that peer.

0001-interledger-architecture/0001-interledger-architecture.md
justmoon: Hmm, because it's too abstract? What would you say instead?

0001-interledger-architecture/0001-interledger-architecture.md
justmoon: No, the small chunk size quantitatively reduces the risk, it does not change the nature of the risk.

Before: **Everyone has to trust their direct neighbors, Interledger ensures you do not have to trust any other intermediaries.** Theoretically, these were meant to always be ledgers you already have to trust. But, in practice, we started using trustlines which didn't really fit the ledger model at all.

Now: **Everyone has to trust their direct neighbors, Interledger ensures you do not have to trust any other intermediaries.** (The same.) The only difference is that an escrowing connector (formerly known as a ledger) is now only one of several possible constellations. At each hop (account), one side needs to take risk.

You might say: But before, I only had to trust the ledger, which could be a decentralized blockchain. Well, connectors could be decentralized blockchains, too. The reduction in risk is just likely not worth the increase in complexity/cost, so I don't think decentralized connectors will play a big role anytime soon.

michielbdejong: So then changing it to "cannot be stolen by faulty or malicious *intermediate* connectors" would make the statement true. Maybe we want to give a special name to the first connector, e.g. "entry node", to distinguish its special role in the trust model from that of intermediate nodes? cc @dappelt

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -1,82 +1,91 @@
...
 ## Interledger Model
 
-![Interledger Model: Sender, Connectors, Ledgers, Receiver](../shared/graphs/interledger-model.svg)
+![Interledger Model: Sender, Connectors, Receiver](../shared/graphs/interledger-model.svg)
 
-### Ledgers
+### Connectors
 
-*Ledgers* are stateful systems that track the ownership of *assets*. Ledgers contain buckets of assets known as *accounts* and record *transfers* between them. Each account has a *balance*, which is the amount of the ledger's assets the account holds. Account balances may be positive or negative, representing assets or liabilities.
+A *connector* is a system that provides the service of forwarding Interledger packets towards the destination provided in the packet. Packets are forwarded across connections between senders, connectors and receivers, called *accounts*.
```
emschwartz: What are "called accounts"? It's hard to understand what in this sentence that is referring to.

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -1,82 +1,91 @@
...
 
-**In the Interledger model, a ledger only tracks a single asset, which may be a currency, stock, commodity, etc.** One entity that maintains accounts denominated in multiple assets is described as having multiple ledgers.
+Interledger packets describe tiny amounts of money, which may be settled individually or in aggregate through various kinds of external payment systems collectively referred to as *ledgers*. Ledgers can include blockchains, but also banks, peer-to-peer payment schemes, automated clearing house (ACH), mobile money institutions and even central-bank operated real-time gross settlement (RTGS) systems among many other types. Some ledgers support very efficient settlement, while others support only slower forms. In general, the faster and cheaper the ledger, the more often an account is settled. See [IL-RFC 22](../0022-hashed-timelock-agreements) for more information on different ledger types and settlement strategies. Settlement does not occur between the sender and receiver directly. Each connector settles only with its direct neighbors.
```
emschwartz: > Interledger packets describe tiny amounts of money

I don't think this is necessarily the case. The whole small packet idea comes from us thinking the early Interledger connectors will have small Maximum Payment Sizes. Some paths or connectors in the future may support larger MPSs.

> See [IL-RFC 22](../0022-hashed-timelock-agreements)

Should link to the markdown file

0001-interledger-architecture/0001-interledger-architecture.md
emschwartz: > can be used among any peered group of connectors

This makes it sound like all of the connectors must be connected with one another, rather than there being a chain of peering or customer relationships between them

0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -1,82 +1,91 @@
...
 
-To be fully Interledger-compatible, ledgers MUST support conditional transfers, though it is possible to send Interledger payments over a ledger that does not natively support the recommended features. See [IL-RFC 17](../0017-ledger-requirements/0017-ledger-requirements.md) for the full description and tiers of ledger requirements.
+Inspired by the [Lightning Network](http://lightning.network), Interledger uses the digest of the [SHA-256](https://en.wikipedia.org/wiki/SHA-2) hash function as the condition for `prepare` packets. The `fulfill` packet contains a valid 32-byte preimage for the hash specified in the `prepare` packet. Connectors are responsible for validating fulfillments. [Transport Layer](#transport-layer) protocols are used by the sender and receiver to generate the condition for a particular packet.
 
-### Payment Flow
+Settlement may either be based solely on trust between the connector and its peers, or be supported with functionality on the settlement ledger. See [IL-RFC 17](../0017-ledger-requirements/0017-ledger-requirements.md) for the full description and tiers of ledger requirements.
```
emschwartz: I'm not sure we still want to link to this document, because it was intended to describe requirements for ledgers to directly participate in the Interledger payment.

michielbdejong:
> senders' funds are safe throughout an (sic) multi-hop payment

> [the ledger] is used by Interledger to localize risk.

I thought about this again and think it's no longer accurate to think of funds traveling across the network (as is implied a bit by the word "throughout"), or about ledgers being involved at all in the sender's (or connector's) risk model. Instead, we can say the beauty of Interledger is that there's a local "action" (first transfer) that is conditional on a remote "action" (last transfer), irrespective of how many (if any) intermediate transfers exist. The first "action" increases the unsecuredAmount of the sender and decreases that of the receiver.

After that, the sender will have to "pay in" to their connector in order to lower their unsecuredAmount again, and the receiver may request a "pay out" from their connector, and that's where funds move. Before that, it's only data that moves, so connectors don't really transmit the sender's money.

Data can only be called Money if the Medium is a Ledger! :)

michielbdejong:
I'll see if I can resume this work; I think one of the main questions is how to deal with IPR, since it relies on type-1 packets and still points to IL-RFC-3, which was superseded by IL-RFC-27.

0001-interledger-architecture/0001-interledger-architecture.md
michielbdejong: It was because I thought 'secure' referred specifically to 'semi-atomic', but I guess we can still call it secure in a more abstract way, so let's leave it as-is.

0001-interledger-architecture/0001-interledger-architecture.md
michielbdejong: i'll propose to use the word 'intermediaries' in my continuation of this PR. we should probably revise the glossary and add a few clarifying notes there about what a ledger is, what a connector is, and what a 'node'/ILSP/neighbor is, once this is merged.

0001-interledger-architecture/0001-interledger-architecture.md
michielbdejong: Leaving this as it is, becuase I don't really know how to improve it.

0001-interledger-architecture/0001-interledger-architecture.md
michielbdejong: We discussed this in this week's team meeting - I'll remove the link to the old whitepaper, and then we can add a link to the new one once it's ready.

michielbdejong: Also updating this ILP link from IL-RFC-3 to IL-RFC-27

0001-interledger-architecture/0001-interledger-architecture.md
michielbdejong: let's leave those details out of scope here, as it would probably get too complex for an architecture overview.

0001-interledger-architecture/0001-interledger-architecture.md
michielbdejong: changing this to 'across connections (*accounts*) between ...'

0001-interledger-architecture/0001-interledger-architecture.md
michielbdejong: * removed 'tiny'
* fixed the hyperlink

0001-interledger-architecture/0001-interledger-architecture.md
michielbdejong: changed to 'network of peered connectors'

0001-interledger-architecture/0001-interledger-architecture.md
michielbdejong: removed the 'See ...' sentence

michielbdejong:
> it's no longer accurate to think of funds traveling across the network, or about ledgers being involved at all in the sender's (or connector's) risk model.

I guess both issues are just a matter of how you look at it, so I think we can leave these idioms as they are in this doc.

PR #369, opened 2018-01-19
Document plugin helpers

michielbdejong:
In the reference implementation of LPI v2, the plugin constructor receives a second argument, called 'api'. It contains two helpers, a PluginStore, and a Logger. We should document both! :)

See https://github.com/interledgerjs/ilp-connector/blob/refactor/st-lpi2/src/services/accounts.ts#L138-L149


michielbdejong:
This was merged, see https://github.com/interledger/rfcs/blob/master/0024-ledger-plugin-interface-2/0024-ledger-plugin-interface-2.md#new-ledgerplugin

PR #372, opened 2018-01-22
feat: allow empty auth_username if auth_token is a bearer token

michielbdejong:


michielbdejong:
Included this proposal in https://github.com/interledger/interledger/wiki/Interledger-over-BTP#changes-between-versions-17q4-and-18q1

0023-bilateral-transfer-protocol/0023-bilateral-transfer-protocol.md
```diff
@@ -186,11 +186,13 @@ Before anything else, when a client connects to a server, it sends a special
 `Message` request. Its primary `protocolData` entry MUST have name `'auth'`,
 content type `MIME_APPLICATION_OCTET_STREAM`,
 and empty data, and among the secondary entries, there MUST be a UTF-8
-`'auth_token'` entry, and a UTF-8 `'auth_username'` entry. The further secondary
+`'auth_token'` entry, and there MAY be a UTF-8 `'auth_username'` entry. The further secondary
```
dappelt: These two statements contradict each other: 
> [...] there MAY be a UTF-8 `'auth_username'` entry.

and later 

>[...] `'auth_username'` data can be set to the empty string, but they cannot be omitted.

michielbdejong: ah, good catch! fixed.

sharafian:
Even in `ilp-plugin-btp` where we ordinarily don't specify an `auth_username`, the `auth_username` protocol is still sent to the server. To reflect the current behavior you could add a note saying that the `auth_username` may be an empty string

michielbdejong:
Will merge once https://circleci.com/gh/interledger/rfcs/tree/master is fixed

PR #373, opened 2018-01-23
Update IL-RFC-14 to use PSK2

michielbdejong:


michielbdejong:
Importantly, compared to the `'interledger-psk'` payment method, for `'interledger-psk2'` the amount moved to the end of the `Pay` header.

PR #374, opened 2018-01-24
Fix size constraints to be whole KiB

adrianhopebailie:
ASN.1 size constraints expressed as ranges define a max and min amount. The max amounts previously used were 1 byte short of complete Kibibytes.

8 KiB = 8192 Bytes (Octets or chars when encoded using OER)
32 KiB = 32768 Bytes (Octets)

### DIFF
--- a/asn1/InterledgerProtocol.asn
+++ b/asn1/InterledgerProtocol.asn
@@ -35,7 +35,7 @@ InterledgerFulfill ::= SEQUENCE {
     fulfillment UInt256,
 
     -- Information for sender (transport layer information)
-    data OCTET STRING (SIZE (0..32767))
+    data OCTET STRING (SIZE (0..32768))
 }
 
 InterledgerReject ::= SEQUENCE {
@@ -46,10 +46,10 @@ InterledgerReject ::= SEQUENCE {
     triggeredBy Address,
 
     -- User-readable error message
-    message UTF8String (SIZE (0..8191)),
+    message UTF8String (SIZE (0..8192)),
 
     -- Machine-readable error data, dependent on code
-    data OCTET STRING (SIZE (0..32767))
+    data OCTET STRING (SIZE (0..32768))
 }
 
 END


emschwartz:
I think @justmoon did that intentionally. If I remember correctly I think it was so that the length prefix wouldn't need a whole extra byte

adrianhopebailie:
> If I remember correctly I think it was so that the length prefix wouldn't need a whole extra byte

  1. I think it is a mistake to define something in ASN.1 based on encoding rules as the ASN.1 definition is independent of the encoding.
  2. I think you are incorrect with regards to how a length prefix is encoded using OER anyway:

The length prefix in OER (for both Octet and Character strings) is a single byte unless the length is > 127.

If the length is > 127 then the first byte has it's most significant bit set and the remaining bits indicate the number of subsequent bytes that encode the length.

If the length  <= 255 then the first byte is 0x81 (or 1000 0001) and the second is the actual length of the field. (Length prefix = 2 bytes)

If the length <= 65535 then the first byte is 0x82 (or 1000 0010) and the the actual length of the field is encoded as a UInt16 in the subsequent 2 bytes. (Length prefix = 3 bytes)

If the length <= 16777215 then the first byte is 0x83 (or 1000 0011) and the the actual length of the field is encoded as a UInt24 in the subsequent 3 bytes. (Length prefix = 4 bytes)

Etc... up until you are encoding the length as an unsigned 1016 bit integer at which point the field value itself is...

So if you are picking max field sizes for variable length fields that will be at the limit of adding a byte for the length indicator then you should have fields that are:

  - <= 127 (length prefix = 1 byte)
  - <= 255 (length prefix <= 2 bytes)
  - <= 65535 (length prefix <= 3 bytes)
  - <= 16777215 (length prefix <= 4 bytes)

Beyond this it seems odd to be worrying about a single byte if the field can be so huge.

justmoon:
@adrianhopebailie Do you want to discuss this further or can this be closed?

adrianhopebailie:
> The size of ~32kb was chosen so that a typical ILP packet would fit comfortably inside a UDP packet.

Interesting! Isn't that closer to 64k though? Well, a bit less for the headers etc.

> It's easier to remember that all limits are a power of two minus one.

At sizes like 32k it seems odd to be haggling over wasting a single bit but I guess that's not the case for smaller limits so I think just being consistent is reason enough to leave this as is.

PR #376, opened 2018-01-25
feat: add test. as the prefix for the testnet

michielbdejong:


michielbdejong:
This was originally proposed in https://github.com/interledger/rfcs/pull/154#discussion_r101749528 
Now, the counter-argument no longer holds, because `test.` happens to be what we use on the Interledger testnet (specifically, `test.amundsen.`, `test.crypto.xrp.`, and `test.crypto.eth.rinkeby.`)

michielbdejong:
@adrianhopebailie ok, merged those two table rows into one.

Issue #375, opened 2018-01-24
Payment pointers and internationalized domain names


dappelt:
I created this issue to discuss potential problems and solution of internationalized domain names in payment pointers. [Payment pointers](https://github.com/interledger/rfcs/blob/5641d91e806a8c3e27d97b91c76cacd13a87444b/0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md#appendix-a-payment-pointer) are introduced as part of SPSPv2. They support [internationalized domain names](https://en.wikipedia.org/wiki/Internationalized_domain_name) the same way URLs do, which makes payment pointers susceptible to [homograph attacks](https://en.wikipedia.org/wiki/IDN_homograph_attack).

To give a concrete example, let's assume alice is regularly paying to `$bob.example.com`. The invoices she is paying look something like `$bob.example.com/invoice/12345/whatever`
Mallory is aware of this and sends a forged payment pointer to Alice that looks like `$bob.example.com∕invoice∕12345∕.evil.com`. Note that there are no backslashes in that payment pointer, but the unicode char \u2215 (`∕ `), so the host is `evil.com`. Since both payment pointers look identical when displayed, Alice might be tricked in querying evil.com instead of example.com.

michielbdejong:
I vote for restricting to ASCII. That way, you can still phish with `$exampIe.com/bob` which will look like $exampIe.com/bob, but at least that's easier for mainstream users to comprehend than weird unicode characters.

Also, as I said on the call, I have never seen anybody use international characters in their email address either, and even for instance Japanese native speakers will probably be able to use just ASCII characters without much inconvenience, when picking a payment pointer.

adrianhopebailie:
> I vote for restricting to ASCII.

I vote for leaving international support but mentioning the risks in the "Security Considerations" section of the specification.

1. It is much simpler and easier for implementors who simply re-use whatever libraries and logic they already have for URL parsing to parse these. There is no character manipulation required, simply drop the leading `$`, split the pointer on the first `/` and construct a URL with the pieces.
2. How does a service that already hosts other services at a domain that uses non-latin characters create payment pointers for their users? Do we expect them to register a new domain just to host payment services.
[...]

PR #377, opened 2018-01-26
Payment Pointers

adrianhopebailie:
Initial draft of spec for payment pointers as discussed in #367


0000-payment-pointers/0000-payment-pointers.md
```diff
@@ -0,0 +1,126 @@
...
+
+### Widely Usable
+
+It should be simple for individuals or small companies to host their own payment setup protocol receiver endpoints at a URL that can be resolved via a simple and easily recognizable payment pointer. Likewise, it should be possible for a payment services provider to host payment setup protocol receiver endpoints for multiple entities without the risk of hosting them at endpoint URLs that conflict with their other services. To that end the provider should have the option of hosting different receivier endpoints under the same domain and a different path or at a different subdomain for each receiver.
+
+## Payment Setup Protocols
+
+A payment setup protocol is defined very generically as a protocol for exchanging payment information between a sender and receiver before executing the payment.
+
+An example of a payment setup protocol is the [Simple Payment Setup Protocol](https://interledger.org/rfcs/0009-simple-payment-setup-protocol/) used to setup an Interledger payment.
```
emschwartz: This should just be a relative link to that RFC

0000-payment-pointers/0000-payment-pointers.md
```diff
@@ -0,0 +1,126 @@
...
+
+Payment pointers MUST resolve to an endpoint hosted by a payment receiver at which a payment setup protocol can be initiated. This endpoint is known as the payment setup protocol receiver endpoint.
+
+A payment setup protocol receiver endpoint is identified by an https URL as defined in [RFC7230](https://tools.ietf.org/html/rfc7230#section-2.7.2). It MUST accept HTTP GET requests to initiate the protocol.
```
emschwartz: Wouldn't whether it responds to GET requests depend on the setup protocol used?

0000-payment-pointers/0000-payment-pointers.md
```diff
@@ -0,0 +1,126 @@
...
+
+Payment setup protocols SHOULD define a custom MIME type for their message definitions and use HTTP content negotitation between the client and endpoint server to determine the payment setup protocol and protocol versions supported by the requesting client and the endpoint server.
```
emschwartz: Add an example for this

0000-payment-pointers/0000-payment-pointers.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Payment Pointer Syntax
+
+This specification uses the Augmented Backus-Naur Form (ABNF) notation of [RFC2234](https://tools.ietf.org/html/rfc2234) and imports the following rules from [RFC3986](https://tools.ietf.org/html/rfc3986#section-3.3)): _host_ and _path-abempty_.
+
+The syntax of a payment pointer is:
+
+```
+  "$" host path-abempty
```
emschwartz: Add an example

0000-payment-pointers/0000-payment-pointers.md
```diff
@@ -0,0 +1,126 @@
...
+
+```
+  "https://" host path-abempty
+```
+
+If _path-abempty_ is empty then it is assigned the value `/.well-known/pay`.
+
+Note that this is a restrcited profile of the data allowed in a full https URL. Where the URL syntax supports an _authority_ the payment pointer syntax only supports a _host_ which excludes the _userinfo_ and _port_. The payment pointer syntax also excludes the _query_ and _fragment_ parts that are allowed in the URL syntax.
+
+### Examples
```
emschwartz: I think it would be more helpful for examples to be spread throughout the explanation

0000-payment-pointers/0000-payment-pointers.md
```diff
@@ -0,0 +1,126 @@
...
+$example.com/bob ->            https://example.com/bob
+```
+
+# Payment Initiation
+
+Given a payment pointer for a receiver, a sending client should resolve the https URL of the payment setup protocol receiver endpoint according to the rules defined in this specification.
+
+The client should then initiate an https connection with the endpoint server and perform an http GET request to the endpoint URL.
+
+The cleint should express the protocols it supports by including the media types of the protocol messages in the `Accept` header of the request.
```
emschwartz: "protocols" is plural, but I think it should be even more explicit that clients should send multiple types in the Accept header if they support multiple protocols (and we should have an example of what that looks like)

Also, typo: "cleint"

0000-payment-pointers/0000-payment-pointers.md
```diff
@@ -0,0 +1,126 @@
+---
+title: Payment Pointers
+draft: 1
+---
+# Payment Pointers and Payment Setup Protocols
+
+## Abstract
+
+Payment pointers are a standardized identifier for accounts that are able to recieve payments. In the same way that an email address provides an identifier for a mailbox in the email ecosystem a payment pointer is used by an account holder to share the details of their account with anyone that wishes to make a payment to them.
```
michielbdejong: "recieve"

0000-payment-pointers/0000-payment-pointers.md
```diff
@@ -0,0 +1,126 @@
...
+
+A payment pointer can be normalized to an "https" URL that provides the location of a payment setup service endpoint at which a sender can initiate a payment to the receiver.
```
michielbdejong: I think "normalize" means change from one of the variations to the canonical one, but the https URL is not itself a payment pointer, so I would rather use the verb "dereference" instead.

0000-payment-pointers/0000-payment-pointers.md
```diff
@@ -0,0 +1,126 @@
...
+
+As a result the pointer is constructed that is intended to mislead a human user by appearing to identify one (trusted) naming authority while actually identifying a different authority hidden behind the noise. 
+
+As detailed in [RFC7230], the "https" scheme (Section 2.7.2) is intended to prevent (or at least reveal) many of these potential attacks on establishing the authority behind a pointer, provided that the negotiated TLS connection is secured and the client properly verifies that the communicating server's identity matches the target URI's authority component (see [RFC2818]).Correctly implementing such verification can be difficult (see [Georgiev]).
+
+Payment Setup Protocols should take heed of this risk in their design and provide a mechainsim for the sender to reliably verify the receiver's identity. This may be as simple as relying on the certificate of the receving endpoint server to identify the receiver, or may be handled explicitly within the protocol.
```
emschwartz: This seems to downplay the risks. If you use SPSP with the wrong endpoint, you're sending money to the wrong person. 

The client should check that the cert matches the endpoint you query, but that doesn't help you if you've just queried the wrong endpoint. Furthermore, if the payment pointer resolves to a URL with a redirect, the endpoint you're talking to will have a cert that has no relationship to the payment pointer at all.

0000-payment-pointers/0000-payment-pointers.md
```diff
@@ -0,0 +1,126 @@
...
+
+### Examples
+
+The following payment pointers resolve to the specified endpoint URLS:
+
+```
+$example.com ->                https://example.com/.well-known/pay
+$example.com/invoices/12345 -> https://example.com/invoices/12345
+$bob.example.com ->            https://bob.example.com/.well-known/pay
+$example.com/bob ->            https://example.com/bob
```
michielbdejong: maybe for completeness, add:
```sh
$example.com/ ->            https://example.com/
```

0000-payment-pointers/0000-payment-pointers.md
```diff
@@ -0,0 +1,126 @@
...
+Given a payment pointer of the form `"$" host path-abempty` the payment setup protocol receiver endpoint URL is:
+
+```
+  "https://" host path-abempty
+```
+
+If _path-abempty_ is empty then it is assigned the value `/.well-known/pay`.
+
+Note that this is a restrcited profile of the data allowed in a full https URL. Where the URL syntax supports an _authority_ the payment pointer syntax only supports a _host_ which excludes the _userinfo_ and _port_. The payment pointer syntax also excludes the _query_ and _fragment_ parts that are allowed in the URL syntax.
+
```
emschwartz: Mention that you can have a redirect, but maybe clients shouldn't follow more than 1 (or 2?) redirects

0000-payment-pointers/0000-payment-pointers.md
emschwartz: That's kind of confusing, so we may want to explicitly disallow that

0000-payment-pointers/0000-payment-pointers.md
```diff
@@ -0,0 +1,126 @@
...
+Given a payment pointer of the form `"$" host path-abempty` the payment setup protocol receiver endpoint URL is:
+
+```
+  "https://" host path-abempty
+```
+
+If _path-abempty_ is empty then it is assigned the value `/.well-known/pay`.
+
+Note that this is a restrcited profile of the data allowed in a full https URL. Where the URL syntax supports an _authority_ the payment pointer syntax only supports a _host_ which excludes the _userinfo_ and _port_. The payment pointer syntax also excludes the _query_ and _fragment_ parts that are allowed in the URL syntax.
```
dappelt: Mention that payment pointers that do not abide by these restrictions must not be resolved into URLs.

0000-payment-pointers/0000-payment-pointers.md
dappelt: Can you remind me why we need `$bob.example.com` and `$example.com/bob`?

0000-payment-pointers/0000-payment-pointers.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Reliability and Consistency
+
+There is no guarantee that once a payment pointer has been used resolve a payment setup protocol receiver endpoint, the same service will be hosted at that URI in the future.
+
+## Semantic Attacks and Phishing
+
+Because payment pointers only support a limited profile of the data in a URL, not all of the attacks described in RFC3986 apply to payment pointers.
+
+However it is possible for a malicious actor to construct a payment pointer that appears to point to a trusted receiver but in fact poimnts to a malicious actor. This is especially so when using internationalized domain names in the payment pointer as these can be constcuted to appear as if there is a path separator (`/`) but this may in fact be a non-latin character in the _host_ portion of the pointer.
```
dappelt: poimnts -> points

dappelt: constcuted -> constructed

0000-payment-pointers/0000-payment-pointers.md
adrianhopebailie: No, that needs to be a minimal requirement because that kicks of protocol negotiation

0000-payment-pointers/0000-payment-pointers.md
```diff
@@ -0,0 +1,126 @@
...
+
+As a result the pointer is constructed that is intended to mislead a human user by appearing to identify one (trusted) naming authority while actually identifying a different authority hidden behind the noise. 
+
+As detailed in [RFC7230], the "https" scheme (Section 2.7.2) is intended to prevent (or at least reveal) many of these potential attacks on establishing the authority behind a pointer, provided that the negotiated TLS connection is secured and the client properly verifies that the communicating server's identity matches the target URI's authority component (see [RFC2818]).Correctly implementing such verification can be difficult (see [Georgiev]).
```
dappelt: HTTPS does not help with homograph attacks. When spoofing gmail.com, an attacker can get a perfectly valid cert for, say, gmai1.com.

adrianhopebailie: Correct but it does help reveal such an attack.
e.g. By using EV certs you can verify the entity behind the cert

0000-payment-pointers/0000-payment-pointers.md
adrianhopebailie: We could add DNS SRV based discovery but I think that would be good as a fall-back we can add later if there are setup protocols people want to introduce that are not HTTP based

0000-payment-pointers/0000-payment-pointers.md
adrianhopebailie: Used resolved as that's what I have used consistently

0000-payment-pointers/0000-payment-pointers.md
adrianhopebailie: Have amended the rules to say that if path is empty or equals `/` then use the default

adrianhopebailie: Options for the hosting company. Different infrastructure challenges

0000-payment-pointers/0000-payment-pointers.md
emschwartz: I was thinking more about protocols that might use a POST or another HTTP call instead of a GET

0000-payment-pointers/0000-payment-pointers.md
sharafian: I think we should just disallow trailing characters

adrianhopebailie: I think it's easier to catch this one ambiguous case and have a rule to handle it

dappelt:
I added a new response to the [issue](https://github.com/interledger/rfcs/issues/375#issuecomment-364134132) about internationalized paypointers.

adrianhopebailie:
@dappelt I think the following text in the spec deals with your concerns in #375 

> Payment Setup Protocols should take heed of this risk in their design and provide a mechanism for the sender to reliably verify the receiver's identity. This may be as simple as relying on the certificate of the receiving endpoint server to identify the receiver, or may be handled explicitly within the protocol.

> Sending clients MUST provide a way for a sender, after providing a pointer to the client, to verify that the entity they are sending to (or at least the entity hosting the payment setup protocol receiver endpoint) is who they intended otherwise senders can easily be fooled into sending money to the wrong receiver.

I also added explicit mention of the ASCII limitation and a warning about attempting to resolve "Internationalized" pointers

PR #378, opened 2018-01-28
Deprecate RFC 4

adrianhopebailie:
Update web scripts and template for deprecated specs

@bsteinlo I have updated the spec template to insert a new `<p class="deprecated">...</p>` tag at the top of the page if the spec is deprecated. Right now it has no CSS defined for `.deprecated` but the plan will be to make it banner or similar eye-catching element.

Is this HTML okay to work with?

### DIFF
--- a/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
+++ b/0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
@@ -1,7 +1,9 @@
 ---
 title: The Javascript Ledger Plugin Interface
-draft: 10
+deprecated: See IL RFC 24 for the latest version
+draft: FINAL
 ---
+
 # Javascript Ledger Plugin Interface
 
 The Interledger Protocol is a protocol suite for connecting blockchains and other ledgers.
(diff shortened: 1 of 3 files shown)


bsteinlo:
@adrianhopebailie HTML is fine for me. Let's just take a look at the styling before publishing something with the deprecated class.

adrianhopebailie:
@bsteinlo great. Can you push another commit that updates the CSS? The HTML for this spec will be:

```html
<p class="deprecated">
    This specification has been deprecated.
    <br/>See IL RFC 24 for the latest version.
</p>
```

bsteinlo:
@adrianhopebailie  No problem. Do you have a way for me to look at this in context?

bsteinlo:
@adrianhopebailie Okay cool. How about instead of the `deprecated` class, you adjust it to have `alert alert-warning` classes. So the markup would look like 
`<div class="col-sm-8 markdown-body">
<p class="alert alert-warning">
    This specification has been deprecated.
    <br>See IL RFC 24 for the latest version.
  </p>
 <h1 id="javascript-ledger-plugin-interface">Javascript Ledger Plugin Interface</h1>`

And then just give it a `margin-top: 35px;`

Does that work?

bsteinlo:
And then just give it a `margin-top: 35px;`

Does that work?

0004-ledger-plugin-interface/0004-ledger-plugin-interface.md
```diff
@@ -1,7 +1,9 @@
 ---
 title: The Javascript Ledger Plugin Interface
-draft: 10
+deprecated: See IL RFC 24 for the latest version
```
emschwartz: Is it possible to link to it?

PR #384, opened 2018-02-02
Remove chunked payments from PSKv2

emschwartz:
TL;DR: PSKv2 is more like UDP than TCP.

Draft 1 attempted to have PSKv2 handle one-off payments, streaming payments, and chunked payments. However, @sharafian made good points (summarized [here](https://gist.github.com/emschwartz/235ac5bdfcd0a6b07315a248c0cd84ee)) that some of the features in the protocol and implementation of PSKv2 were only useful for chunked payments and not for streaming. When streaming payments, it makes more sense to have a completely stateless receiver that has a single callback to handle all incoming chunks, and applications will want more direct control over how they determine whether to accept or reject individual packets. Also, we wanted to [support application data](https://github.com/interledgerjs/ilp-protocol-psk2/issues/5) but PSKv2 did not go all the way to implementing segmentation and reassembly like TCP. (@sentientwaffle also asked the first time he read the spec why we didn't separate the chunked payments part from the rest of PSKv2)

This change simplifies the protocol and includes only the features necessary for:
1. Generating the conditions and fulfillments from a shared secret
2. Encrypting and authenticating request and response data
3. Providing the sender a way to indicate to the receiver the minimum amount they should get to accept a given packet.

This is implemented in https://github.com/interledgerjs/ilp-protocol-psk2/pull/19


0025-pre-shared-key-2/0025-pre-shared-key-2.md
```diff
@@ -186,58 +156,59 @@ The following is encrypted, as defined [above](#encryption), and set as the ILP
...
+| Type (T) | Always `4` (UInt8) | Indicates that this is a PSK2 request. Request packets SHOULD be rejected or ignored if they are not in an ILP Prepare packet. |
+| Request ID (ReqID) | UInt32 | ID used to correlate requests and responses. Senders CANNOT rely on receivers enforcing uniqueness, but they SHOULD ensure that responses carry the same Request ID as the outgoing request |
+| Request Amount | UInt64 | Minimum amount that should arrive in the ILP Prepare packet for the receiver to fulfill it. Receivers SHOULD NOT accept packets where the Request Amount is greater than the ILP Prepare amount |
+| Data | [OER Variable-Length Octet String](http://www.oss.com/asn1/resources/books-whitepapers-pubs/Overview%20of%20OER.pdf) | User data carried along with the payment. Note this data is encrypted and authenticated |
+| Extensions | N/A | Addtional fields may be added after the `Application Data` later. Implementations that do not support any extensions SHOULD ignore all additional bytes |
```
sentientwaffle: typo: "Addtional"

0025-pre-shared-key-2/0025-pre-shared-key-2.md
```diff
@@ -186,58 +156,59 @@ The following is encrypted, as defined [above](#encryption), and set as the ILP
...
 
-The following apply to both fulfillment and error responses:
+| Field | Type | Description |
+|---|---|---|
+| Type (T) | Always `5` (UInt8) | Indicates that this is a PSK2 response. Response packets SHOULD be rejected or ignored if they are not in an ILP Fulfill packet. |
+| Request ID (ReqID) | UInt32 | ID used to correlate requests and responses. Receivers MUST use the same Request ID from the Request in their Response. Senders SHOULD ignore responses whose Request ID does not match their original request. |
+| Request Amount | UInt64 | Amount that arrived in the ILP Prepare packet the receiver got. This is used to help senders determine the path exchange rate and how much the receiver has gotten. Note the sender must trust the receiver to report this honestly. |
+| Data | [OER Variable-Length Octet String](http://www.oss.com/asn1/resources/books-whitepapers-pubs/Overview%20of%20OER.pdf) | User data carried along with the payment. Note this data is encrypted and authenticated |
+| Extensions | N/A | Addtional fields may be added after the `Application Data` later. Implementations that do not support any extensions SHOULD ignore all additional bytes |
```
sentientwaffle: typo: "Addtional"

0025-pre-shared-key-2/0025-pre-shared-key-2.md
```diff
@@ -186,58 +156,59 @@ The following is encrypted, as defined [above](#encryption), and set as the ILP
...
+| Type (T) | Always `6` (UInt8) | Indicates that this is a PSK2 error. Response packets SHOULD be rejected or ignored if they are not in an ILP Reject packet. |
+| Request ID (ReqID) | UInt32 | ID used to correlate requests and responses. Receivers MUST use the same Request ID from the Request in the Error. Senders SHOULD ignore responses whose Request ID does not match their original request. |
+| Request Amount | UInt64 | Amount that arrived in the ILP Prepare packet the receiver got. This is used to help senders determine the path exchange rate and how much the receiver has gotten. Note the sender must trust the receiver to report this honestly. |
+| Data | [OER Variable-Length Octet String](http://www.oss.com/asn1/resources/books-whitepapers-pubs/Overview%20of%20OER.pdf) | User data carried along with the payment. Note this data is encrypted and authenticated |
+| Extensions | N/A | Addtional fields may be added after the `Application Data` later. Implementations that do not support any extensions SHOULD ignore all additional bytes |
```
sentientwaffle: typo: "Addtional"

0025-pre-shared-key-2/0025-pre-shared-key-2.md
```diff
@@ -186,58 +156,59 @@ The following is encrypted, as defined [above](#encryption), and set as the ILP
...
+| Type (T) | Always `4` (UInt8) | Indicates that this is a PSK2 request. Request packets SHOULD be rejected or ignored if they are not in an ILP Prepare packet. |
+| Request ID (ReqID) | UInt32 | ID used to correlate requests and responses. Senders CANNOT rely on receivers enforcing uniqueness, but they SHOULD ensure that responses carry the same Request ID as the outgoing request |
+| Request Amount | UInt64 | Minimum amount that should arrive in the ILP Prepare packet for the receiver to fulfill it. Receivers SHOULD NOT accept packets where the Request Amount is greater than the ILP Prepare amount |
+| Data | [OER Variable-Length Octet String](http://www.oss.com/asn1/resources/books-whitepapers-pubs/Overview%20of%20OER.pdf) | User data carried along with the payment. Note this data is encrypted and authenticated |
+| Extensions | N/A | Addtional fields may be added after the `Application Data` later. Implementations that do not support any extensions SHOULD ignore all additional bytes |
+| Junk Data | 0-? Bytes | _(Optional)_ Extra data included to obscure what the ILP payment is for. Receivers SHOULD ignore this data |
```
sentientwaffle: Receivers should ignore Extensions/Junk, but connectors are supposed to relay them, right (referencing https://github.com/interledgerjs/ilp-connector/issues/405).

0025-pre-shared-key-2/0025-pre-shared-key-2.md
```diff
@@ -253,7 +224,7 @@ condition = sha256(fulfillment)
 
 The following pseudocode details how the receiver regenerates the fulfillment from the data.
 
-**Note:** Senders MAY use unfulfillable conditions to get an informational quote. Receivers MUST return an [Error Response](#response-fields) packet that includes the amount that arrived on the incoming transfer, even if they are unable to regenerate the fulfillment from the request.
+**Note:** Senders MAY use unfulfillable conditions to get an informational quote. Receivers MUST return an [Error Response](#error) packet that includes the amount that arrived on the incoming transfer, even if they are unable to regenerate the fulfillment from the request.
 
 The `shared_secret` is the pre-shared key that gives this protocol its name. The `data` is the **encrypted** payment chunk request.
```
adrianhopebailie: The word 'chunk' seems weird in this iteration of the doc.

0025-pre-shared-key-2/0025-pre-shared-key-2.md
emschwartz: > Receivers should ignore Extensions/Junk

Yes

> but connectors are supposed to relay them, right?

This is part of the encrypted PSK packet so connectors won't know whether or not it's there. This doesn't use the extensions mentioned in https://github.com/interledgerjs/ilp-connector/issues/405

0025-pre-shared-key-2/0025-pre-shared-key-2.md
```diff
@@ -63,63 +68,30 @@ When a sender is sending many payments to the receiver, they need a dynamic view
...
+7. Receiver creates a PSK Response with the same Request ID and the Request Amount set to the amount that arrived in the ILP Prepare packet. The receiver then [encrypts](#encryption-envelope) the data using the same encryption key.
+8. Receiver generates fulfillment as described [below](#fulfillment-generation) and fulfills the incoming transfer with the fulfillment and the encrypted response data. The receiver SHOULD return an ILP Reject with an `F05: Wrong Condition` code if they are unable to regenerate the correct condition.
+9. Sender gets the ILP Fulfill packet and decrypts the PSK Response from the data. They SHOULD ignore the Response if the Request ID does not match their original request or if the PSK packet is not a Response type. They may use the Request Amount from the Response to estimate the path exchange rate. They may also use the data in the PSK Response from the receiver.
 
 ## Specification
```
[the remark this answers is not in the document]
emschwartz: 12, that was a mistake left from an earlier draft

0025-pre-shared-key-2/0025-pre-shared-key-2.md
```diff
@@ -45,7 +50,7 @@ In PSKv2.0, request and response data is encrypted using [AES-256-GCM](https://e
 
 If subsequent versions support additional encryption algorithms, those details should be exchanged between the sender and receiver when they establish the pre-shared key. If a receiver attempts to decrypt an incoming packet but is unable to (perhaps because the sender is using an unsupported cipher), they SHOULD reject the incoming transfer with an `F06: Unexpected Payment` error.
```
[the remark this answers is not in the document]
emschwartz: Maybe we could add another that's even more specific but the definition does say (emphasis added):

> The receiver was not expecting a payment like this (the memo and destination address don't make sense in that combination, for example if the **receiver does not understand the transport protocol used**)

FWIW, the implementation uses an `F06` to determine if it should try sending a legacy packet to support implementations of Draft 1 of this spec: https://github.com/interledgerjs/ilp-protocol-psk2/blob/master/src/sender.ts#L200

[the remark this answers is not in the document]
adrianhopebailie: I would vote for a new error code because I would consider this "invalid" as opposed to "unknown"

asn1/PreSharedKeyV2.asn
```diff
@@ -0,0 +1,46 @@
...
+Psk2Packet ::= SEQUENCE {
+    -- Type ID
+    type UInt8,
+
+    -- ID used to correlate requests and responses
+    requestId UInt32,
+
+    -- In PSK2 Requests, this indicates the minimum the receiver should accept
+    -- In PSK2 Responses and Errors, it is the amount that the receiver got
+    requestAmount UInt64,
```
adrianhopebailie: Suggest calling this field `amount` since it occurs in both requests and responses and the type of amount is contextual.

0025-pre-shared-key-2/0025-pre-shared-key-2.md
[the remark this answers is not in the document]
emschwartz: Actually when you try to decrypt the packet and it doesn't work, you don't know if it was modified along the way or using an encryption algorithm you don't understand. All you know is that you weren't able to decrypt it.

asn1/PreSharedKeyV2.asn
emschwartz: I think the term `requestAmount` still fits, because in one it's saying how much should be delivered in the request and the other is saying how much the request delivered

0025-pre-shared-key-2/0025-pre-shared-key-2.md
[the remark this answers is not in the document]
adrianhopebailie: I would interpret "Unknown" to mean I got a payment that I wasn't expecting. I.e. I was able to parse it but it doesn't match up to anything I am expecting.

I would interpret "Invalid" to mean I got a payment that I can't even parse.

asn1/PreSharedKeyV2.asn
```diff
@@ -0,0 +1,46 @@
...
+
+    -- Application data carried in the encrypted PSK Request/Response/Error
+    data VarBytes
+}
```
adrianhopebailie: You are missing the Extensions byte and the junk data.
A sequence has no overall length indicator so you need these to be able to parse this.

The envelope is no use because it wraps encrypted data (unless there is a guarantee that the encrypted data and decrypted data will be the same length). Even if you did figure out the total length from the envelope this would still mean a lot of custom decoding logic and it would be impossible to differentiate between the extensions and the junk data.

emschwartz: Those are just the rest of the bytes, which you should ignore anyway. If you don't understand the extensions there's no need to parse them, because unlike in ILP packets you don't need to pass them on

adrianhopebailie: In that case I'd remove the mention of extensions completely. We don't actually support extensions if you're expected to just throw away any data you don't expect.

In future we can add new fields and retain the current packet numbers without breaking existing implementations because they will simply throw away the new fields along with the junk data.

The exception would be if we tried to add optional fields which cause a bitmap to be added at the beginning so we'd simply be prevented from adding new optional fields.

emschwartz: Removed in https://github.com/interledger/rfcs/pull/384/commits/fd7b42dcf022c0e1be9b34de8f90906799a5dd3b

0025-pre-shared-key-2/0025-pre-shared-key-2.md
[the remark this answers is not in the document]
emschwartz: Don't know what to say. That error code has meant what it says there since it was defined. The description pretty clearly indicates that it should be used for saying you don't understand the transport protocol.

If the user rejects the packet, for example because they weren't expecting the payment, that's treated as an `F99: Application Error` and the user can include some response in the PSK Response data to tell the sender what's up.

Small point but to me "invalid" sounds like "I parsed it and you sent something bad" like an HTTP 400 Bad Request error.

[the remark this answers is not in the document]
adrianhopebailie: Yeh, not worth making a big deal over.

Issue #388, opened 2018-02-11
Proposal: Duplex PSK


sharafian:
This is something that @emschwartz and @justmoon have thought about extensively, but I was pretty skeptical until the last conversation that we had about it.

The current version of PSK can be treated as a unidirectional stream. You can send data and money one way (and there's a method of acknowledging these), but the receiver cannot actively tell the sender what to do. So:

The sender can:

- Send money w/ attached data

The receiver can:

- Respond to money w/ attached data

If the receiver wants the sender to stop sending money, they need to communicate that in the response of the sender's packet. The sender needs to send continuous "polling" packets in order for the receiver to tell them when to send again.

This is workable in some use cases, but a little clunky. Furthermore, it completely fails to solve some cases like refunds, where the sender actually wants the receiver to start sending money back.

We can solve this issue by making PSK a duplex stream. The sender and receiver have each others' ILP addresses in this model, so packets can be sent actively both ways.

We can get rid of polling now, and say that instead of periodically asking when to send money, the sender and receiver (who are now identical in their capabilities) can actively request money by sending a packet. It goes from the sender being able to "push", to both parties being able to "push" and "pull." So:

Both parties can:

- Send money w/ attached data
- Request money w/ attached data

You'll notice that this opens up a couple of different flows. One flow is that someone is requesting money (and the other party doesn't know how much to send). The other flow is that someone wants to send an amount of money (and the other party doesn't know how much they're going to receive).

We can reason about this by saying the PSK stream's balance starts at zero, and each party has a minimum and maximum balance. The rules are as follows:

- If you send money, your balance goes down.
- If you receive money, your balance goes up.
- If the current balance is below your peer's minimum, you'll send money.
- If the current balance is above your own maximum, you'll send money.
- Take care of your own min/max before your peer's, i.e. Don't send money if it puts you below your minimum.
- Either party can change their minimum/maximum at any time by sending a special PSK packet to the other party.

That's a little hard to grasp in all at once, so I'll give a couple of examples:

#### I want to send 1,000,000 units

I set my minimum to `-1,000,000` and my maximum to `-1,000,000`. The other party's maximum is `Infinity` (I can pay them as much as I like) and their minimum is `0` (They're not going to pay me).

The balance starts at zero, so I'm above my maximum. I begin sending money. Once the other party gets a payment they increase their minimum by the payment amount, because they're not willing to refund it. Eventually I send 1,000,000 units.

#### I am paying an invoice for 1,000,000 units

[...]

PR #394, opened 2018-02-15
First draft of tokens proposal [WIP]

adrianhopebailie:
This is the first draft of a proposal for an ILP Token standard to accommdate a number of use cases that are currently difficult to achieve.

### DIFF
--- a/0000-interledger-tokens/0000-interledger-tokens.md
+++ b/0000-interledger-tokens/0000-interledger-tokens.md
@@ -1,0 +1,242 @@
+---
+title: Interledger Tokens
+draft: 1
+---
+# Interledger Tokens
+
+An **Interledger Token** is a special form of JSON Web Token that can be used to initiate an Interledger payment. The token provides proof that the payment has been pre-authorized by the payer. By using these tokens new use cases for "debit pull" style payments such as recurring subscriptions or offline authorization are possible.
+
+## Background
+
+TODO
+
+## Motivation
+
+TODO
+
+## Terminology
+
+The following terms are used throughout this specification. Where they are used their meaning as follows:
+
+### Payer
+
+The entity whose funds are being spent when the token is used to initiate a payment.
+
+### Payer's Interledger Service Provider (ILSP)
+
+The entity that accepts and processes the token and payment request. This may be the payer or they may delegate it to some other entity that is able to make payments on their behalf.
+
+### Payee
+
+The entity that is requesting payment using the token.
+
+## The Interledger Token
+
+An **Interledger Token** is a special form of a JSON Web Token (JWT) [RFC7519](https://tools.ietf.org/html/rfc7519) that can be submitted to a participant on the Interledger network to request that they initiate a new payment via ILP. The content of the token describes how much to pay, who to pay and provides information for the entity processing the token to verify that the payment was authorized by the owner of the funds being sent.
+
+**Interledger Tokens** are limited profile of a standard JWT. This profile limits the allowed algorithms, allowed claims and defines specific rules for what constitutes valid claim values. A token is used by a payee to "claim" that a payment has been authorized by the payer in accordance with the terms of the token.
+
+After processing a token and making a payment the participant that has made the payment MAY return a new token to the payee to use again in the future. For example, a payee may posses a token that authorizes a payment to them of USD 100. They may then submit this to initiate a payment but request that only USD 25 is sent in the payment. Following successful execution of the payment they may be returned a new token that authorizes a payment to them of USD 75.
+
+An example token payload is shown below (with further details provided in later sections):
+
+```json
+{
+  "jti": "48359d89-e9ef-44dd-9a3d-68f991ed7556", //Unique Token ID
+  "iss": "872369652347412343", //Unique identifier for the payer (recognized by their ILSP)
+  "sub": "example.payee123", //ILP Address of the payee
+  "aud": "https://superwallet.example/ilp/", //Endpoint to which the token is submitted
+  "exp": "1518699103", //The Expiry of the token (Thursday, February 15, 2018 12:51:43 PM)
+  "payee": {
+    "max": "100", //The maximum amount that should be delivered to the payee when this token is presented 
+    "min": "0", //The minimum amount that should be delivered to the payee when this token is presented
+    "asset": "USD", //The asset that the min and max values are denominated in
+    "scale": "2" //The scale of the min and max (eg. if scale = 2 then a value of 1234 = USD 12.34)
+  },
+  "payer": {
+    "protected":"eyJhbGciOiJkaXIiLCJlbmMiOiJBMjU2R0NNIn0=", // BASE64URL({"alg":"dir","enc":"A256GCM"})
+    "iv":"48V1_ALb6US04U3b", //Random 96-bit IV
+    "ciphertext":"5eym8TW_c8SuK0ltJ3rpYIzOeDQz7TALvtu6UG9oMo4vpzs9tX_EFShS8iB7j6ji
+     SdiwkIr3ajwQzaBtQD_A", //Data for the payer's ILSP encrypted using AES GCM and pre-shared key
+    "tag":"XFBoMYUZodetZdvTiFvSkQ" //Authentication Tag
+  }
+}
+```
+The `payer` is a JSON Web Encryption (JWE) object (See [RFC7516]](https://tools.ietf.org/html/rfc7516)) containing data that is securely exchanged between the payer and the payer's ILSP. The plain-text representation of `payer.ciphertext` in this example is:
+
+```json
+{
+  "max": "100", //The maximum amount that must be sent from the payer's account when this token is presented.
+  "min": "0", //The minimum amount that must be sent from the payer's account when this token is presented.
+  "asset": "USD", //The asset that must be spent when making a payment from the payer's account. (In the event that the payer holds multiple accounts with the ILSP)
+  "scale": "2", //The scale of the max and min amounts above
+}
+```
+
+The payee can POST this token to `https://superwallet.example/ilp/` to initiate the process of making ILP payments to `example.payee123.tx567`.
+
+### Valid Headers
+
+Only the following header values are allowed in an **Interledger Token**:
+
+| Header | Name      | Description                               |
+|--------|-----------|-------------------------------------------|
+| alg    | Algorithm | The algorithm used to sign the token body |
+
+### Valid Header Values
+
+Only the following header values are allowed in an **Interledger Token**:
+
+| Header | Value | Description                                                                                     |
+|--------|-------|-------------------------------------------------------------------------------------------------|
+| alg    | HS256 | HMAC using SHA-256 (See [RFC7518 Section 3.2](https://tools.ietf.org/html/rfc7518#section-3.2)) |
+
+### Valid Claims
+
+The following Registered Claim Names (as defined in RFC7519) are allowed in an **Interledger Token**:
+
+| Claim | Name            | Description |
+|-------|-----------------|-------------|
+| jti   | Token ID        | The unique identifier of the token. This must unique per payer (issuer). |
+| iss   | Issuer          | An identifer for the payer that can will allow the Payer's ILSP to uniquely identify the payer. |
+| sub   | Subject         | The ILP address of the payee. |
+| aud   | Audience        | The URL of the **token submission endpoint** of the Payer's ILSP. Only a valid HTTPS URL is allowed with no query string or fragment portions. |
+| exp   | Expiration Time | The expiry time of the token (per [RFC7519](https://tools.ietf.org/html/rfc7519)). |
+| nbf   | Not Before      | The time before which the token may not be used (per [RFC7519](https://tools.ietf.org/html/rfc7519)). |
+| iat   | Issued At       | The time at which the JWT was issued (per [RFC7519](https://tools.ietf.org/html/rfc7519). |
+
+The following Public Claim Names (as defined in RFC7519) are allowed in an **Interledger Token**. These will be registered with IANA in accordance with the process defined in RFC7519:
+
+| Claim | Name       | Description |
+|-------|------------|-------------|
+| payee | Payee Data | Data about the payment that can be made with this token as relates to the account of the payee. |
+| payer | Payer Data | Data about the payment that can be made with this token as relates to the account of the payer. |
+
+#### Payer and Payee Data
+
+The Payer Data and Payee Data contain the same fields however the Payee Data is un-encrypted and visible to the token holder (likely the payee) whereas the Payer Data is encrypted and visible only to the Payer's ILSP.
+
+Both data sets define limits for the token, denominated in the asset and scale of the payee and payer accounts.
+
+| Field | Description |
+|-------|-------------|
+| asset | The asset of the account. If a payer has multiple accounts that the ILSP could send from this provides guidance as to which asset should be spent when this token is redeemed by the payee. |
+| scale | Provides the scale of the number used in `min` and `max` which are both expressed an unsigned 64-bit integers. |
+| min   | In the Payee Data, this is the minimum amount that will be delivered. The ILSP will not attempt the payment if the current rate suggests this will not be possible. In the Payer Data, this is the minimum amount that must be sent per payment. This may be set by the payer in the case that there are fixed fees to send that make sending payments smaller than a certain size un-viable. Assumed to be 0 if not provided. |
+| max   | In the Payee Data, this is the maximum amount that should be delivered. The ILSP will attempt to deliver no more than this amount. The payee may request less than this, in which case the ILSP should return a new token after a successful payment with this value reduced by the amount delivered. In the Payer Data, this is the maximum amount that must be delivered. This is the limit that the payer has defined for this token and the ILSP must not attempt to make a payment if the current rate suggests that this will need to be exceeded in order to deliver the `payee.minimum`.
+
+## Flow
+
+### Request
+
+A payee or other system may request a new token and specify the parameters of the token that they require including the ILP Address to send to, asset of the payee account and max and min amounts to set in the Payee Data.
+
+### Issuing
+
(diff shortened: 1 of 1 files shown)


0000-interledger-tokens/0000-interledger-tokens.md
```diff
@@ -0,0 +1,242 @@
...
+
+The entity that accepts and processes the token and payment request. This may be the payer or they may delegate it to some other entity that is able to make payments on their behalf.
+
+### Payee
+
+The entity that is requesting payment using the token.
+
+## The Interledger Token
+
+An **Interledger Token** is a special form of a JSON Web Token (JWT) [RFC7519](https://tools.ietf.org/html/rfc7519) that can be submitted to a participant on the Interledger network to request that they initiate a new payment via ILP. The content of the token describes how much to pay, who to pay and provides information for the entity processing the token to verify that the payment was authorized by the owner of the funds being sent.
```
emschwartz: Have you looked at macaroons at all? I think they'd be pretty cool for something like this because whoever you give it to could add further "caveats" and pass it on more

0000-interledger-tokens/0000-interledger-tokens.md
```diff
@@ -0,0 +1,242 @@
...
+  "jti": "48359d89-e9ef-44dd-9a3d-68f991ed7556", //Unique Token ID
+  "iss": "872369652347412343", //Unique identifier for the payer (recognized by their ILSP)
+  "sub": "example.payee123", //ILP Address of the payee
+  "aud": "https://superwallet.example/ilp/", //Endpoint to which the token is submitted
+  "exp": "1518699103", //The Expiry of the token (Thursday, February 15, 2018 12:51:43 PM)
+  "payee": {
+    "max": "100", //The maximum amount that should be delivered to the payee when this token is presented 
+    "min": "0", //The minimum amount that should be delivered to the payee when this token is presented
+    "asset": "USD", //The asset that the min and max values are denominated in
+    "scale": "2" //The scale of the min and max (eg. if scale = 2 then a value of 1234 = USD 12.34)
```
emschwartz: Relying on a fixed destination amount and currency is dangerous because there aren't really comparable assets in Interledger and the rate could change drastically between when you give out this token and when your ILSP goes to pay it. The payee could be running their own connector and mess with the rate to take all of your money.

0000-interledger-tokens/0000-interledger-tokens.md
```diff
@@ -0,0 +1,242 @@
...
+
+After processing a token and making a payment the participant that has made the payment MAY return a new token to the payee to use again in the future. For example, a payee may posses a token that authorizes a payment to them of USD 100. They may then submit this to initiate a payment but request that only USD 25 is sent in the payment. Following successful execution of the payment they may be returned a new token that authorizes a payment to them of USD 75.
+
+An example token payload is shown below (with further details provided in later sections):
+
+```json
+{
+  "jti": "48359d89-e9ef-44dd-9a3d-68f991ed7556", //Unique Token ID
+  "iss": "872369652347412343", //Unique identifier for the payer (recognized by their ILSP)
+  "sub": "example.payee123", //ILP Address of the payee
```
emschwartz: Putting the address without the secret isn't that useful because if someone is able to give you a bad secret to do PSK with, they could re-route the money to themselves and provide valid fulfillments even if it didn't get to the real ILP address.

0000-interledger-tokens/0000-interledger-tokens.md
```diff
@@ -0,0 +1,242 @@
...
+
+The `iat` value should be set to the current time as this will assist in identifying the `sender_secret` that was used (in combination with the `iss` value) assuming the issuer regularly changes the `sender_secret`.
+
+The token should then be signed using the `HMAC using SHA-256` algorithm as described in [RFC7515](https://tools.ietf.org/html/rfc7515) and serialized using the Flattened JWS JSON Serialization Syntax as defined in [RFC7515 Section 7.2.2](https://tools.ietf.org/html/rfc7515#section-7.2.2).
+
+No unprotected headers are allowed.
+
+### Redemption
+
+A token is redeemed by submitting it to the endpoint defined in the `aud` value. An ILSP that is processing a token should reject it if the value of `aud` does not match the URL of the endpoint at which the token was submitted.
```
emschwartz: I'm not a fan of using a completely different protocol just for this use case. I think we could enable this kind of flexible pull-payment with either a payment pointer and version of SPSP that supports [duplex payment streams](https://github.com/interledger/rfcs/issues/388) or something like a BTP endpoint and token that lets you connect and send payments to yourself through that using any higher-level protocol you like.

0000-interledger-tokens/0000-interledger-tokens.md
adrianhopebailie: These are just a "guide". The payee wants to know what value the token has to them. The ILSP that does the sending should always respect the limits set in the payer data first.

0000-interledger-tokens/0000-interledger-tokens.md
adrianhopebailie: It's a bearer token. That's okay.

emschwartz: Also, if ILP addresses are ephemeral, your receiver might be listening on a different address by the time the merchant gets around to paying you

0000-interledger-tokens/0000-interledger-tokens.md
adrianhopebailie: > something like a BTP endpoint and token that lets you connect and send payments to yourself through that using any higher-level protocol you like.

That's exactly what this is except that it's a more sophisticated token because it's a bearer token with the limits encoded in the token itself that can be generated by an entirely offline sender. That is a critical differentiator.

emschwartz: Okay, but then it shouldn't have a different way to redeem it. If you just use BTP with a bearer token, the fact that you're using a fancy token that encodes the rules will be completely opaque to the merchant and your BTP server can support whatever token formats it wants.

0000-interledger-tokens/0000-interledger-tokens.md
adrianhopebailie: I'd assume as a receiver you'd have some control over that or at least get short lived tokens if you expect your address to change often.

Bare in mind that the majority of payments are one off. The payee will get the token from the payer, redeem it and then move on all within probably a few seconds.

0000-interledger-tokens/0000-interledger-tokens.md
adrianhopebailie: Not in depth admittedly. JWT is a now a pretty well bedded down standard. It has had a lot of criticism but I believe this spec addresses all of them by using a very limited profile of JWT.

0000-interledger-tokens/0000-interledger-tokens.md
adrianhopebailie: The purpose of a standard is to not tie the token format to the protocol. It's doesn't have to be BTP and in many cases it will be desirable not to be. Such as when the entity redeeming the token has no idea what BTP is they simply have a token they want to submit to a REST API somewhere and get a response to say they got paid.

> BTP server can support whatever token formats it wants.

Sure, if my BTP server is the only thing that can issue the tokens for me. i.e. That doesn't support an offline sender.

> the fact that you're using a fancy token that encodes the rules will be completely opaque to the merchant

Not at all. The token is intentionally not encrypted but just signed. The merchant can look at the token and determine immediately if they are likely to be able to redeem it and for how much.

emschwartz: Unless you're using a payment pointer, which includes protocol negotiation in it, I don't see how you're not tied to a protocol for actually redeeming the money.

> they simply have a token they want to submit to a REST API somewhere and get a response to say they got paid.

The receiver also needs to be running a receiver that speaks the same application layer protocol as the payer's ILSP.

> Sure, if my BTP server is the only thing that can issue the tokens for me. i.e. That doesn't support an offline sender.

That's not true, and I implemented this in "ILP3" with macaroons. If the BTP server uses a token format that allows for offline generation or caveating, the sender can be offline and generate sub-tokens that the BTP server will honor.

The benefit of using bearer tokens that are opaque to the merchant is that ILSPs can support different formats and ever more complex rules without the merchant changing their flow at all.

sharafian: I think payment pointers are preferable to BTP because they make use of a layer on top of ILP instead of a layer underneath it. But yeah, a payment pointer is also capable of containing identifying information so that the ILSP can tell what merchant is pulling.

adrianhopebailie:
> once we have duplex PSK

I'm not convinced this will happen in a way that will actually work reliably for most normal retail payments use cases so I guess we'll just have to wait and see.

As I said in https://github.com/interledger/rfcs/issues/388#issuecomment-364860980 I'm not sure how you plan to handle connectors that charge fees in both directions. So far there doesn't seem to be an answer to that and that seems like a deal breaker for me.

That aside, and assuming duplex PSK works. How does a payer give a payment pointer to a payee, that allows them to only pull a specific amount of money from their account, if the payer is offline? (The most common retail payments use case)

Maybe there is a case for use payment pointers inside this token?

0000-interledger-tokens/0000-interledger-tokens.md
adrianhopebailie: > The receiver also needs to be running a receiver that speaks the same application layer protocol as the payer's ILSP.

The current proposal allows the payee to submit the application data so they are defining the application layer protocol on both ends.

> That's not true, and I implemented this in "ILP3" with macaroons. If the BTP server uses a token format that allows for offline generation or caveating, the sender can be offline and generate sub-tokens that the BTP server will honor.

Then that was just another format for exactly the same thing. Macaroons instead of JWTs.

> The benefit of using bearer tokens that are opaque to the merchant is that ILSPs can support different formats and ever more complex rules without the merchant changing their flow at all.

The merchant doesn't HAVE to change their flow. If the merchant has the ability to read some data in the token then they have the advantage of being able to change their flow if they choose to.

The benefit of using tokens that have some opaque and some visible data is that the bearer knows what they are actually carrying. If I get a token for an ILP Address I no longer own or that has expired I know that before trying to redeem it.

adrianhopebailie: > I think payment pointers are preferable to BTP because they make use of a layer on top of ILP instead of a layer underneath it.

I don't really follow the on top vs underneath thing.

We should be very wary of payment pointers as a way to pull payments. This is exactly the mistake that was made decades ago with cards. All you needed was someone's card number and you could start pulling money out of their account. The result is layers and layers of complexity to try and shield card numbers from being shared.

The purpose of payment pointers is to give someone a pointer to an account they can pay into not the other way around.

sharafian:
The requirements to be always online are no different in this proposal than the requirements for payment pointers. Someone needs to be running an endpoint that's actively listening, whether it's the human you're referring to as the "payer" or their ILSP.

In the benign case of a connector that charges a spread, there's just a small fee associated with a refund. The network's job is to keep this fee low through competition. If there's a malicious connector who is charging fees with the intention of holding payments hostage, that's something that applies equally to this proposal as it does to duplex PSK. If a connector in the network raised their rates during the pull payment, the payment may be forced to hang until the network routes around them.

0000-interledger-tokens/0000-interledger-tokens.md
sharafian: > I don't really follow the on top vs underneath thing.

using BTP for application layer concerns means everyone needs to use it as the communication under ILP. Using a protocol like SPSP that rides on top of ILP means that everyone only needs to have ILP in common. It's like how lightning uses the lnd RPC endpoint in their application level invoice flow, which makes it harder to become interoperable.

> This is exactly the mistake that was made decades ago with cards.

Not sure what the difference is between having a token get leaked vs. having a payment pointer get leaked. Either way, they're authorized for a specific amount of money/throughput and can be revoked.

adrianhopebailie:
>The requirements to be always online are no different in this proposal than the requirements for payment pointers.

I'm not sure how you come to that conclusion. To get any information from a payment pointer I must go online and query the SPSP endpoint. The token contains static data that can be inspected entirely offline.

I can deposit money in an account controlled by an ILSP, agree the sender-key with them and then issue tokens whenever I want that can be redeemed by whoever has them whenever they want. The only entity that needs to be online is the ILSP and then only at the time of redemption.

Further, I can issue a token that has no fixed receiver by setting `sub` to `g.` and I have the equivalent of a cash cheque.

> If there's a malicious connector who is charging fees with the intention of holding payments hostage, that's something that applies equally to this proposal as it does to duplex PSK.

Not true. This proposal is simply an authorization for a one way flow of funds. The sender is expected to not even start the payment if the current rate suggests they won't be able to deliver the required destination amount.

There is no attempt to recoup payments sent if the rate changes and risk losing that too.

emschwartz:
> The token contains static data that can be inspected entirely offline.

It shouldn't matter if the merchant can inspect data while being offline, because that data in no way guarantees their ability to get money from the sender. I thought the offline point you were bringing up was about the payer being offline (in which case they could have a token protocol with their ILSP that allows them to generate auth tokens offline).

> The only entity that needs to be online is the ILSP and then only at the time of redemption.

And whoever is receiving money on behalf of the payee

> This proposal is simply an authorization for a one way flow of funds. The sender is expected to not even start the payment if the current rate suggests they won't be able to deliver the required destination amount.

Your malicious connector will just let you start sending and then drive up the rate with each chunk you send. It doesn't make a big difference whether the flow of funds is one- or two-way

> There is no attempt to recoup payments sent if the rate changes and risk losing that too.

I don't think the duplex stream would necessarily try to recoup the money by default. Just because you have the ability to send in both directions doesn't mean you'll use it.

emschwartz:
The alternative to this proposal using duplex payment streams would be that the payer gives the merchant a payment pointer that looks like `$evan.cash/180ca4c3c0a2d662130cf88aeee8c5842b421bf0e27f53cd609a2403c10b91ca`. The merchant doesn't know or care what those random 32 bytes mean, but they can open a socket to the address that resolves to and request money.

0000-interledger-tokens/0000-interledger-tokens.md
emschwartz: I have to say that I never really regained respect for JWTs after the [`none` bug](https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/)...

adrianhopebailie:
> The alternative to this proposal using duplex payment streams would be that the payer gives the merchant a payment pointer that looks like $evan.cash/180ca4c3c0a2d662130cf88aeee8c5842b421bf0e27f53cd609a2403c10b91ca. The merchant doesn't know or care what those random 32 bytes mean, but they can open a socket to the address that resolves to and request money

That is just a level of indirection on the same thing.

That token (the path of the pointer) has to be accepted by the payer's ILPS as proof that the payer has authorized this payment. 

If that pointer is produced offline there is no way for the payer to indicate to their ILSP what limits they are putting on the payment (i.e amount, expiry or recipient).

I think you are focusing on the use case too much and missing the point of the spec: A standard way to encode authorization of a payment that can be produced by a payer and redeemed by the bearer to request a payment.

I hear you that this could also be done with Macaroons. I prefer this more widely used (and in my opinion more flexible) standard.

I will add some more flesh to the motivations to try and make this clearer and see if there is a way to weave in payment pointers that make sense.

0000-interledger-tokens/0000-interledger-tokens.md
adrianhopebailie: I think that's a little unfair. That was a foot-gun sure, but it also only materialized in bad implementations that ignored the advice **in the spec** to use different keys for different algorithms. 

For me, the whole value of using a standard like JWT is it has lots of eyes on it so the vuln was found and fixed. It is anti-fragile.

emschwartz:
> That token (the path of the pointer) has to be accepted by the payer's ILPS as proof that the payer has authorized this payment.
> If that pointer is produced offline there is no way for the payer to indicate to their ILSP what limits they are putting on the payment (i.e amount, expiry or recipient).

I don't think I'm getting my point across. I _like_ the idea of being able to generate tokens offline with limits that the payer's ILSP will respect. However:
1. The simple* token standard is a concern only of the payer, their client software, and their ILSP (not the merchant). It should be totally fine if I want to use macaroons with my ILSP, you want to use JWTs, and someone else wants to use pre-agreed random tokens that are all valid for specific amounts. The merchant should treat all of these as opaque bearer tokens and submit whatever they're given. Having some standards for this is mostly useful for client software to be able to add further caveats or limitations on the payer's behalf, so I'm not sure I'd even think of these as "ILP tokens".
2. The way of redeeming the tokens should be based on Payment Sockets and the flow for the merchant should be exactly the same as any other use of those (resolve the payment pointer, open a socket, and request money)

\* The reason to define a different standard that the merchant would be aware of would be if you want merchants to accept payment tokens as payments while the merchant is offline. However, then you need the payer's ILSP to vouch for the payer and their device (that it won't sign claims for more money than it has) and you need a system by which the merchant can trust the ILSP to pay for the token when the merchant goes online. Since all of that is considerably more complicated than this, I think that should be a separate standard (that we may or may not get around to implementing).

adrianhopebailie:
> The reason to define a different standard that the merchant would be aware of would be if you want merchants to accept payment tokens as payments while the merchant is offline.

I think I see where we are missing each other. Not having a standard is problematic because it fragments the ecosystem. 

This is analogous to the gateway tokens in use today for card payments. Each payment gateway has a different opaque token they expect merchants to submit for a payment. As a result, wallets that generate the tokens, like Android Pay need to have integrations for each token type they will support.

If we define a standard token then we decouple the wallet from the ILSP which I think is positive.

I also think there is value in being able to validate a token (as the payee) without redeeming it. i.e. See the signed content and validate the signature.

> The way of redeeming the tokens should be based on Payment Sockets and the flow for the merchant should be exactly the same as any other use of those (resolve the payment pointer, open a socket, and request money)

I like the idea of the socket for pull-payments in general but I think we should be conscious of who operates at what layer. At the application layer we should be using Web technology like JWTs.

I think we could actually come up with a more efficient (and appropriate) serialization for exchange at the ILP/Transport layers. I.e. The token used in the socket `connect()` from the payee to the payer is probably better OER encoded.

I think we should finish the socket and duplex payments work and then revisit this.

PR #398, opened 2018-02-23
Only check asn1 if it changed

michielbdejong:


michielbdejong:
Merged as part of #372, cannot merge this separately because now RFCs were changed.

PR #401, opened 2018-02-24
update three draft numbers on which generate_web script fails

michielbdejong:


michielbdejong:
My theory is that some dependency of scripts/generate_web.js changed, and therefore the generated html content is slightly different for these three documents, even if their markdown content did not change. [The error](https://circleci.com/gh/interledger/rfcs/1085) only seems to happen on circleci, when I run generate_web locally, it doesn't. After we merge this we can see how the html changed.

michielbdejong:
I thought of a slightly less strange fix, reducing those three draft version in the gh-pages branch - then tests should also pass, and the result should be noop, which would be better. will try if that works instead.

michielbdejong:
hm, no wait, that would mean draft versions would temporarily be incorrect, and old versions may be overwritten, i'll just merge this one.

emschwartz:
Damn, this still failed https://circleci.com/gh/interledger/rfcs/1091

michielbdejong:
Latest https://circleci.com/gh/interledger/rfcs/tree/es-ilp4-spec runs succeeded

Issue #405, opened 2018-02-26
Document Web Monetization and compare to IL-RFC-14


michielbdejong:
In https://github.com/sharafian/web-monetization-leaderboard and www.siren.sh, @sharafian is using a protocol that, IIUC, does not use 402 responses with `Pay` headers, and actually allows two flows, one in which payment is required, and one in which payment is optional. We should document the comparison, and decide if we want to document that newer approach (and if we want to support it on the testnet), and also if we want to continue using http-ilp and this newer approach side-by-side.

For now, on the testnet, we only have https://letter-shop.herokuapp.com/ which supports http-ilp as documented in IL-RFC-14, but we can add more experiments if desired!

@sharafian I understand your proposal is to use http-ilp for machine-readable http requests (paid API calls) and to use your newer approach for human-readable ones (humans viewing html using a browser), right?

sharafian:
Yeah, http-ilp is desirable for APIs because it charges a fixed price and pays once. The new protocol, which we're calling "Web Monetization" right now, is better for content because the amount paid is entirely up to the user. The longer the user stays on the page, the longer their extension will be sending money to the site, but their extension also has the option to not pay at all. Just like most sites don't care about an ad-blocker-blocker because they would rather leave an impression on the user than not show them anything at all, most sites probably won't turn away users if their traffic is not monetized.

michielbdejong:
Can you document your Web Monetization proposal somewhere?

michielbdejong:
Thanks for starting that! Discussion continued there.

PR #406, opened 2018-02-26
Docs: web monetization api

sharafian:
Closes https://github.com/interledger/rfcs/issues/405 . Further discussion can be done on this PR.


0000-web-monetization/0000-web-monetization.md
```diff
@@ -0,0 +1,69 @@
...
+- The **user** is the party who is accessing the site.
+
+### Design Goals
+
+- Should be extremely simple for webmasters to use in their site.
+- Backend infrastructure should be optional; should be usable on a static site.
+- Should give users a choice about how much to spend, and which sites to support.
+- Should give advanced webmasters a way to associate payments with their users, in order to unlock premium experiences.
+
+### Relation to Other Protocols
```
emschwartz: What's the relation between this and the W3C Web Payments API?

0000-web-monetization/0000-web-monetization.md
```diff
@@ -0,0 +1,69 @@
...
+
+Web Monetization makes use of [Payment Pointers](../0026-payment-pointers/0026-payment-pointers.md) and [SPSP](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md).
```
emschwartz: "makes use of" is a little too vague. What does it use them for?

0000-web-monetization/0000-web-monetization.md
```diff
@@ -0,0 +1,69 @@
...
+## Flow
+
+- The user visits a webpage. The page is loaded and the page's javascript is run.
+- The site checks whether `window.monetize` is defined. This function is injected into the page by the browser or browser extension which provides Web Monetization.
+- If `window.monetize` is not defined, the webpage may show the user a message asking them to install a Web Monetization extension in order to support this site.
+- If `window.monetize` is defined, the webpage calls `monetize({ receiver })`, where `receiver` is their payment pointer. This function returns a promise.
+- The user's browser or Web Monetization extension decides whether to pay the page.
+- If the user's browser decides not to pay, an error is thrown from the `monetize` promise.
+- If the user's browser pays, the `monetize` promise resolves.
+- The webpage may run some code in order to thank the user or offer them additional content.
```
emschwartz: Unclear from this description how the website knows if they've actually gotten paid and, if so, how much

sharafian:
> couldn't we just use the Web Payments API?

I think this is aimed at a different problem. Web Payments is for payments, while monetization is intended to replace what ads are doing right now.

- Web payments is more geared towards discrete payments, as you pointed out
- Web payments require user interaction (as far as I understand)
- We may require additional parameters that are specific to the problem of monetizing a page (i.e. the page could express that it's playing audio content and needs to be paid even when its in the background)
- We don't want one page to be able to monetize itself multiple times; the object based API of web payments makes this harder to constrain and more confusing to use
- The site doesn't know how much it wants to get paid (or they may give an amount but only as a guideline); it's up to the user agent.

emschwartz:
@adrianhopebailie Is this true? "Web payments require user interaction"

> The site doesn't know how much it wants to get paid (or they may give an amount but only as a guideline); it's up to the user agent.

Not sure that's true. If they're charging you for content, or for an ad-free experience, wouldn't there be a specific amount they expect to get (possibly per time) in order to keep giving that to you?

adrianhopebailie:
> @adrianhopebailie Is this true? "Web payments require user interaction"

Currently yes

sharafian:
> Not sure that's true. If they're charging you for content, or for an ad-free experience, wouldn't there be a specific amount they expect to get (possibly per time) in order to keep giving that to you?

Potentially for content, but that really just depends on their current business model. Youtube doesn't refuse to show you content if you don't view their ads, and most sites don't either. In general it's better to make an impression on a user even if you get no revenue because it increases the chance they'll share it.

Today, a site that charges a fixed price for content would sell it using an online payment method like paypal. That particular case might be better served by web payments.

emschwartz:
> Today, a site that charges a fixed price for content would sell it using an online payment method like paypal. That particular case might be better served by web payments.

Why? If you already have payments set up and ready to stream automatically, why shouldn't I be able to say "this is how much you need to send in order to get the improved experience"? If there's no way to express that, the extension will need to guess how much you need to be sending to see some different experience. If the Web Payments API isn't a good fit for the whole monetization use case because it requires user interaction, I don't see why it would be any more appropriate for the case of "you need to give me at least $0.25 (or $0.001 per second) in order to have an ad-free experience".

sharafian:
> If you already have payments set up and ready to stream automatically, why shouldn't I be able to say "this is how much you need to send in order to get the improved experience"?

If that were an option, it would lead to one of three scenarios:

1. The user is willing to pay more than you asked for, and so you're getting less money than you could.
2. The user is not willing to pay that much, so you need to bother them to turn up the amount of money they can pay. What's more likely is that they'll just leave your page and pay nothing.
3. The user's maximum is perfectly matched to the amount you ask for.

3 is the most desirable scenario, so why wouldn't the API always do that? It's not like there's a fixed cost associated with distributing your content that needs to be recouped. The user's agent can just decide how much to pay based on static settings and/or data that it builds up on the user's browsing habits.

sharafian:
Can we merge this so it can be linked to from other sources? We can make this the first draft but I'd like to be able to reference it from several different places.

0000-web-monetization/0000-web-monetization.md
```diff
@@ -0,0 +1,78 @@
...
+
+This function does not guarantee that the user has paid the server. The backend which runs the SPSP receiver
+will have to be queried in order to confirm how much has been paid to a specific receiver.
+
+##### Parameters
+
+| Name | Type | Description |
+|:---|:---|:---|
+| opts | `Object` | The options for monetization. |
+| opts.receiver | `String` | The payment pointer or SPSP endpoint to which ILP payments should be sent. |
```
michielbdejong: "SPSP" should link to [IL-RFC-9, draft 4](https://interledger.org/rfcs/0009-simple-payment-setup-protocol/draft-4.html)
"payment pointer" should link to [IL-RFC-26, draft 1](https://interledger.org/rfcs/0026-payment-pointers/draft-1.html)

Otherwise, LGTM

0000-web-monetization/0000-web-monetization.md
```diff
@@ -69,7 +69,7 @@ will have to be queried in order to confirm how much has been paid to a specific
 | Name | Type | Description |
 |:---|:---|:---|
 | opts | `Object` | The options for monetization. |
-| opts.receiver | `String` | The payment pointer or SPSP endpoint to which ILP payments should be sent. |
+| opts.receiver | `String` | The [payment pointer](../0009-simple-payment-setup-protocol/draft-4.html) or [SPSP endpoint](../0026-payment-pointers/draft-1.html) to which ILP payments should be sent. |
```
michielbdejong: looks like the payment pointer link points to the spsp spec and vice versa

Issue #397, opened 2018-02-22
Fix errors related to amounts, balances, and bandwidth


emschwartz:
We need to decide on the error codes that will be returned when a packet exceeds the balance or bandwidth so that the sender can react appropriately.

**Currently Implemented in the Connector**
- [Balance limit exceeded in connector middleware](https://github.com/interledgerjs/ilp-connector/blob/master/src/middlewares/balance.ts#L34): `F00: Bad Request`
- [Maximum packet amount exceeded](https://github.com/interledgerjs/ilp-connector/blob/master/src/errors/max-packet-amount-exceeded-error.ts): `F03: Invalid Amount`
- [Throughput exceeded](https://github.com/interledgerjs/ilp-connector/blob/master/src/errors/throughput-exceeded-error.ts): `F02: Unreachable`
- [Rate limit exceeded](https://github.com/interledgerjs/ilp-connector/blob/master/src/errors/rate-limit-exceeded-error.ts): `F02: Unreachable`

(The plugins are even worse because they don't return any ILP Reject packets at all: https://github.com/interledgerjs/ilp-plugin-xrp-asym-server/issues/20, https://github.com/interledgerjs/ilp-plugin-mini-accounts/issues/22)

**Proposal**
- Balance limit exceeded: `T04: Insufficient Liquidity` or a new error
- Max packet amount: new final error `F08: Maximum Payment Size Exceeded` (the amount isn't really _invalid_, it's just too high for this path)
- Throughput exceeded: `T05: Rate Limited` or a new error if we want to differentiate between "you're sending too many packets and you're sending too much money"
- Rate limit exceeded: `T05: Rate Limited` (just needs to use the right code)

justmoon:
Nice, thanks for working on this!

> Balance limit exceeded: T04: Insufficient Liquidity or a new error

Yeah, T04 sounds right.

> Max packet amount: new final error F08: Maximum Payment Size Exceeded (the amount isn't really invalid, it's just too high for this path)

:+1:

> Throughput exceeded: T05: Rate Limited or a new error if we want to differentiate between "you're sending too many packets and you're sending too much money"

I would have expected T04 here. A throughput limit is more like a liquidity limit to me than a rate limit imo. Throughput and balance limits prevent me from spending too much money. A rate limit prevents me from overloading the connector's CPU/memory/bandwidth.

If anyone disagrees strongly with this I would say another option would be to just create a separate T** code for throughput exceeded.

[...]

PR #407, opened 2018-02-28
docs(0003): add F08: Maximum Payment Size Exceeded

sentientwaffle:
See https://github.com/interledger/rfcs/issues/397


sentientwaffle:
Oops, missed that

PR #413, opened 2018-03-22
[WiP] add loopback transport in discovery methods

michielbdejong:
work in progress...

### DIFF
--- a/0029-loopback-transport/0029-loopback-transport.md
+++ b/0029-loopback-transport/0029-loopback-transport.md
@@ -1,6 +1,6 @@
 ---
 title: The Loopback Transport Protocol (LT)
-draft: 1
+draft: 2
 ---
 # Loopback Transport Protocol (LT)
 
@@ -10,7 +10,7 @@ The Loopback Transport protocol is a technique for transporting value over the I
 
 The following flow describes how the sender's loopback transfer module ("LT module") interacts with the application layer protocol, the receiver, and the ILPv4 network.
 
-1. The loopback connection is established out of band. For instance, the application layer protocol may allow the receiver to specify a [BTP/2.0](../0023-bilateral-transfer-protocol/0023-bilateral-transfer-protocol.md) URL for the sender to connect to.
+1. The loopback connection is established out of band (see also [discovery](#discovery)). For instance, the application layer protocol may allow the receiver to specify a [BTP/2.0](../0023-bilateral-transfer-protocol/0023-bilateral-transfer-protocol.md) URL for the sender to connect to.
 1. The sender's LT module is initialized with this open connection.
 1. The sender's LT module discovers the loopback address using [IL-DCP](https://github.com/interledgerjs/ilp-protocol-ildcp).
 1. The sender's application layer protocol tells the LT module to send a payment for a given source amount, and expiry time.
@@ -31,3 +31,14 @@ When designing an application layer protocol, it's advisable to support both LT 
 If LT is chosen, the loopback link needs to be established; this will usually require some data exchange (e.g., one of the two parties sends connection details to the other party).
 
 The flow described above assumes one party only acts as the sender, and the other party only acts as the receiver, but once the loopback link is established, it can easily be used in both directions (in that case, each participant would send an IL-DCP request to the other participant, to ask for a loopback address).
+
+## Discovery
+
+A loopback server can be defined in the Loopback Server Definition (LSD) format, whose mime-type is `application/lsd+json`:
+
+```js
+{
+  loopback: 'wss+btp://token@example.com/'
+  protocol: 'BTP/2.0'
+}
+```
--- a/0014-http-ilp/0014-http-ilp.md
+++ b/0014-http-ilp/0014-http-ilp.md
@@ -1,6 +1,6 @@
 ---
 title: HTTP-ILP
-draft: 3
+draft: 4
 ---
 # HTTP-ILP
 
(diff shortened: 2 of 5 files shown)


michielbdejong:
Had an interesting chat with @adrianhopebailie about this, some things that still need to be fleshed out would be how the application layer negotiates currency code/scale. Another idea that came up was that you could do receiver-initiated loopback, and it would be like IPR.

michielbdejong:
For payment setup, need to add a way to discover a loopback server in:
* [ ] (IL-RFC-14) http-ilp,
* [ ] (IL-RFC-28) web-monetization,
* [ ] (IL-RFC-26) payment-pointers, and with that, [web-payments](https://w3c.github.io/webpayments/proposals/interledger/)
* (maybe) ILP URL scheme
* (maybe) SPSP

michielbdejong:
Hm, maybe it's easier to extend SPSP, and catch  payment-pointers with that, and with that web-payments and web-monetization. And then instead of adding a http-ilp scheme, http-ilp could also just use a payment pointer.

0029-loopback-transport/0029-loopback-transport.md
```diff
@@ -0,0 +1,50 @@
...
+
+Once the loopback link is up, and each party who wants to act as a sender knows what their loopback address is, LT exposes two methods and one event to the application layer: one method for the one-shot flow, one for the chunked flow, one event that gets triggered whenever an incoming payment gets fulfilled, and one for when an outgoing payment gets fulfilled.
+
+It is up to the application layer protocol to decide how much to send, and whether to use a one-shot or a chunked flow.
+
+In future, hooks may be provided through which the application layer can also decide to reject incoming payments instead of looping them back to the appropriate sender.
+
+## Discovery
+
+A loopback server can be defined in the Loopback Server Definition (LSD) format, whose mime-type is `application/lsd+json`:
```
adrianhopebailie: Think this should be an extension of SPSP rather.
Example:

```json
{
  "destination_account": "example.ilpdemo.red.bob",
  "loopback": "wss+btp://token@example.com/",
  "maximum_destination_amount": "100000",
  "minimum_destination_amount": "10",
  "ledger_info": {
    "currency_code": "USD",
    "currency_scale": 2
  },
  "receiver_info": {
    "name": "Bob Dylan",
    "image_url": "https://red.ilpdemo.org/api/spsp/bob/profile_pic.jpg"
  }
}
```

I don't think it's necessary to put the `protocol` in as this should be determined from the loop back URI implicitly.

adrianhopebailie: The "destination_account" could be left out and established via ILDCP but having it in here is an optimization

michielbdejong: > Think this should be an extension of SPSP rather.

I could go either way with this, I kinda liked using the switch that we already have [at the payment-pointer level](https://github.com/interledger/rfcs/pull/413/files#diff-26073d3cbab98c2b76aeb396ea5ef212R103)

adrianhopebailie: I think it's too early to be fragmenting the system like that. SPSP v3 which supports more transport options seems better than a whole new setup protocol to me.

The ability to support other setup protocols via Payment Pointers is future proofing and loose coupling so that if someone has a better idea than SPSP we can still use Payment Pointers. 

That shouldn't be an excuse to not use SPSP when it actually already delivers everything we need for loopback IMO

adrianhopebailie: > it should be the client, not the server, who picks the transport

That is not how most protocol negotiation works. Servers usually provide options and clients pick from those. 

Payment Pointers is not a protocol for discovering a transport protocol, it's a protocol for discovering information about a receiver, including all the transport protocols it supports.

The protocol negotiation built into Payment Pointers is not for picking a transport protocol its for picking a setup protocol.

Unless loop-back is incompatible with SPSP I don't see why you'd invent a new setup protocol.

What prevents the SPSP response from providing multiple protocol options to the client?

```
{
  "destination_account": "example.ilpdemo.red.bob",
  "shared_secret": "vdytf65rdhgcioIYbuffGKItgRfg"
  "loopback": "wss+btp://token@example.com/",
  "balance": {
    "minimum": "0",
    "current": "-1000"
  },
  "asset_info": {
    "code": "USD",
    "scale": 2
  },
  "receiver_info": {
    "name": "Bob Dylan",
    "image_url": "https://red.ilpdemo.org/api/spsp/bob/profile_pic.jpg"
  }
}
```

michielbdejong: @adrianhopebailie you're making a valid point, so I'll split up this PR into one part which just describes the loopback transport (which I want to keep simple and finish quickly), and a second part (which can take longer and be a more fundamental cleanup) dealing with cleaning up our setup protocols so that they support multiple transports.

sharafian:
I personally think one of these should be the go-to standard, so most implementations don't need both. For now, though, I like the idea of doing this at the discovery level. Payment pointers already have protocol negotiation (via the Accept header) while SPSP doesn't. If you created a new accept format (like application/spsplt+json, or something), then existing applications wouldn't get confused by your SPSP responses.

michielbdejong:
> If you created a new accept format (like application/spsplt+json, or something), then existing applications wouldn't get confused by your SPSP responses.

That's what i did (see https://github.com/interledger/rfcs/pull/413/files#diff-26073d3cbab98c2b76aeb396ea5ef212R103), and i think Evan also preferred doing the switch there, but Adrian said https://github.com/interledger/rfcs/pull/413#discussion_r176721037

michielbdejong:
ah wait, i see the nuance of calling it SPSPLT instead of LSD. I think my name is more fun, though ;)

sharafian:
Groovy

sharafian:
Adrian has some points about how the LSD and SPSP fields are pretty much the same, but we'd still be creating a breaking SPSP version to add that feature (because it's breaking to not always use PSK). So I don't see much of a compelling reason why we should do SPSPv4 over LSDv1

michielbdejong:
I'm going to Backronym LSD from "Loopback Server Description" to "Loopback Server Discovery", because I keep getting it wrong myself :)

michielbdejong:
We decided not to add Loopback to the RFCs repo until at least one application uses [Loopback Server Discover](https://github.com/interledgerjs/ilp-protocol-lsd) to set up an Interledger loop, instead of using a payment-pointer/SPSP to exchange a STREAM shared secret.

PR #414, opened 2018-03-23
chore: fix merge error

michielbdejong:


michielbdejong:
Looks quite ugly in https://github.com/interledger/rfcs/blob/master/0023-bilateral-transfer-protocol/0023-bilateral-transfer-protocol.md#authentication

PR #415, opened 2018-03-23
Add IPRv3 spec

adrianhopebailie:
Updated IPR based on loop-back ILP as discussed with @michielbdejong


sharafian:
The ideas in here make sense as a part of the loopback transport (the case where the sender is acting as a connector rather than the receiver acting as a connector), but I don't see how they fit into IPR.

IPRv1 was just an encoding for condition and ILP packet (when they were different). Now that we already have a format for encoding all of that (InterledgerPrepare) I don't think we need to have anything called IPR.

michielbdejong:
Good point! But then we should still remove IPRv1 from the list of transport protocols, and replace it with that "Just Send A Prepare Packet" method, right?

0000-interledger-payment-request-3/0000-interledger-payment-request-3.md
```diff
@@ -0,0 +1,113 @@
...
+
+1. The receiver creates the payment request, an `InterledgerPrepare` packet, with:
+    1.  the `amount` that is being paid from the sender to the receiver (in the receiver's currency and scale),
+    2.  a `destination` ILP Address that routes to the receiver (and may contain a payment specific suffix to help the receiver identify it when it arrives),
+    3.  an `executionCondition` generated using one of algorithms outlined in [Condition Generation Recommendations](#condition-generation-recommendations),
+    4.  an appropriate `expiresAt` that will allow the sender to both, determine the appropriate amount to send (a process that could potentially take some time), and complete the payment, and
+    5.  optional application `data` that is opaque to the sender.
+2. The receiver sends the payment request (the `InterledgerPrepare` packet) to the sender.
+3. The sender routes this packet out via one of its ILP peers, behaving like a typical ILP connector, and MUST therefor:
+    1.  determine the appropriate `amount` to send on the next hop,
```
emschwartz: How does the sender determine the rate between their units and the receiver's units? As @michielbdejong mentioned, IPRv1 depended on ILQP, but since that was removed from the Interledger layer transport protocols need to specify a quoting mechanism.

Also, what happens if the amount the receiver requests is larger than the Maximum Packet Amount for the path from the sender to the receiver? The receiver has no idea what the limit might be, because they don't know where the sender is sending from

adrianhopebailie: > How does the sender determine the rate between their units and the receiver's units? As @michielbdejong mentioned, IPRv1 depended on ILQP, but since that was removed from the Interledger layer 

That's not in scope for this protocol. The sender can use whatever mechanism they wish, they could send a probing payment, query a third-party or send an amount hey consider "enough"

> transport protocols need to specify a quoting mechanism.

Why? How you determine the amount to send shouldn't be tightly coupled to the transport protocol.

adrianhopebailie:
> Now that we already have a format for encoding all of that (InterledgerPrepare) I don't think we need to have anything called IPR.

I disagree. This spec defines more than just an encoding it, defines a flow too and guidance for the condition and fulfillment generation. This needs to be documented somewhere to be useful.

emschwartz:
Can this be closed now?

PR #416, opened 2018-03-28
[WiP] Add loopback transport

michielbdejong:
This PR just adds the loopback as a transport; adding support for it in our setup protocols to deal is continued on https://github.com/interledgerjs/ilp/issues/133


0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -110,17 +110,22 @@ SPSP uses Webfinger ([RFC 7033](https://tools.ietf.org/html/rfc7033)), an HTTPS-
...
 
-    In the Interledger Payment Request (IPR) protocol, the receiver generates the payment details and condition. The receiver does not share the secret used to generate the condition and fulfillment with the sender or anyone else, but the sender must ask the recipient to generate and share a condition before sending each payment. IPR is primarily useful for building non-repudiable application layer protocols, in which the sender's posession of the fulfillment proves to third parties that the sender has paid the receiver for a specific obligation.
+    In the Interledger Payment Request (IPR) protocol, the receiver generates the payment details and condition. The receiver does not share the secret used to generate the condition and fulfillment with the sender or anyone else, but the sender must ask the receiver to generate and share a condition before sending each payment. IPR is primarily useful for building non-repudiable application layer protocols, in which the sender's possession of the fulfillment proves to third parties that the sender has paid the receiver for a specific obligation.
```
michielbdejong: changed "recipient" to "receiver" for consistency

michielbdejong:
Not sure about the constant use of the phrase "the sender's LT module", it's hard to read. But I like how it describes exactly what tasks/steps an LT module should implement, and what tasks/steps are done by the application layer protocol.

Also not entirely sure about how to compare loopback and psk in IL-RFC-1; apart from latency, @emschwartz identified the following reasons why people might still want to use psk even though we now have loopback:
* it's not really that much complexity to add (i guess this is hard to quantify)
* as a receiver, you may want to use psk without spsp and without payment pointer, to keep your server's IP address secret
* in the future, we may have multiple BTP versions in use, and although it will be easy enough for a sender to pick a software application that allows her to connect to her own connector, picking one that supports the BTP versions used by all receivers out there might be harder

I think those three arguments are valid, but too complex to include in IL-RFC-1, so current text only mentions that PSK is an advanced technique to reduce latency in cases where the remote accept/reject decision making is not too hard.

michielbdejong:
To do:

* [ ] clean up code structure
* [ ] add unit tests

0029-loopback-transport/0029-loopback-transport.md
```diff
@@ -0,0 +1,33 @@
...
+---
+# Loopback Transport Protocol (LT)
+
+The Loopback Transport protocol is a technique for transporting value over the Interledger. It relies on the establishment of a temporary, direct, ILPv4 link between sender and receiver (the loopback link). The flow is as follows.
+
+## Flow
+
+The following flow describes how the sender's loopback transfer module ("LT module") interacts with the application layer protocol, the receiver, and the ILPv4 network.
+
+1. The loopback connection is established out of band. For instance, the application layer protocol may allow the receiver to specify a [BTP/2.0](../0023-bilateral-transfer-protocol/0023-bilateral-transfer-protocol.md) URL for the sender to connect to.
```
emschwartz: This makes it sound like a BTP URL is just one of multiple options, is that true? Would clients / servers be expected to support other ways of connecting?

0029-loopback-transport/0029-loopback-transport.md
```diff
@@ -0,0 +1,33 @@
...
+1. The sender's LT module is initialized with this open connection.
+1. The sender's LT module discovers the loopback address using [IL-DCP](https://github.com/interledgerjs/ilp-protocol-ildcp).
```
emschwartz: the loopback address = receiver's address, right?

Could you describe somewhere what the requirements are for the receiver? What are they running? Sounds like a BTP server and an ILDCP server. Is there anything special about how they determine the loopback address for a given connection, or is that just done by the connector / multi-plugin?

0029-loopback-transport/0029-loopback-transport.md
```diff
@@ -0,0 +1,33 @@
...
+1. The sender's application layer protocol tells the LT module to send a payment for a given source amount, and expiry time.
+1. The sender's LT module chooses a fulfillment at random.
+1. The sender's LT module creates the condition of their payment by computing the SHA-256 hash of the fulfillment.
+1. The sender's LT module sets the destination address to the loopback address for this lookback link.
+1. Using this condition, address, and amount, the sender's LT module prepares an ILPv4 payment to a connector of her choice.
+1. The payment will find its way from this connector, via any number of additional connectors, to the receiver.
+1. The receiver checks that the destination address matches the loopback address he announced to the sender over IL-DCP earlier, and forwards the ILPv4 `prepare` message to the sender, over the loopback link.
```
emschwartz: What if the receiver doesn't want it? How would they communicate that to the sender?

0029-loopback-transport/0029-loopback-transport.md
```diff
@@ -0,0 +1,33 @@
...
+1. The sender's LT module accepts or rejects the ILPv4 payment that looped back over the loopback link.
+1. If for whatever reason the payment does not arrive in time, the LT module responds to the application layer protocol with an error message instead
+
+## How to use
+
+When designing an application layer protocol, it's advisable to support both LT and PSK. Even though LT is far simpler than PSK, the extra loopback hop means that LT payments inherently have a higher latency than PSK payments, where the receiver can immediately fulfill incoming prepared payments. In situations where the sender can easily formulate instructions for the receiver about when to accept a payment and when not, and the sender trusts the receiver to execute these instructions in good faith, the speed up achieved with PSK means money needs to be tied up for less time, and this may translate to lower transaction fees, depending on connector pricing policies. So the application layer protocol needs to provide a choice between LT and PSK.
```
emschwartz: > When designing an application layer protocol, it's advisable to support both LT and PSK

I disagree with this statement. Even if there are two options I think it would make much more sense to pick one for your specific use case. On the internet, HTTP doesn't let you use either TCP or UDP.

0029-loopback-transport/0029-loopback-transport.md
```diff
@@ -0,0 +1,33 @@
...
+
+If LT is chosen, the loopback link needs to be established; this will usually require some data exchange (e.g., one of the two parties sends connection details to the other party).
+
+The flow described above assumes one party only acts as the sender, and the other party only acts as the receiver, but once the loopback link is established, it can easily be used in both directions (in that case, each participant would send an IL-DCP request to the other participant, to ask for a loopback address).
```
emschwartz: Only if the spec says that's what the implementations need to support. Can the receiver assume that the client has implemented ILDCP?

0029-loopback-transport/0029-loopback-transport.md
```diff
@@ -0,0 +1,33 @@
+---
+title: The Loopback Transport Protocol (LT)
+draft: 1
+---
+# Loopback Transport Protocol (LT)
+
+The Loopback Transport protocol is a technique for transporting value over the Interledger. It relies on the establishment of a temporary, direct, ILPv4 link between sender and receiver (the loopback link). The flow is as follows.
```
emschwartz: Can the BTP link be used for sending data as well? If so, please describe how that works

0029-loopback-transport/0029-loopback-transport.md
michielbdejong: loopback address = sender's address at the receiver

sure, i can add a description of how the receiver would behave (and as discussed in our 1:1, i'll also add a receiver in the demo, forgot to do that earlier)

0029-loopback-transport/0029-loopback-transport.md
michielbdejong: that's part of a higher-level protocol; for this spec we assume Alice wants to pay Bob and Bob wants to be paid.

0029-loopback-transport/0029-loopback-transport.md
michielbdejong: i would see it more like browsers support both http (the simpler, loopback-like protocol) and https (the more complex, psk-like one).

0029-loopback-transport/0029-loopback-transport.md
michielbdejong: i guess they can just try sending an IL-DCP request and see if they get a reply. any extra protocol for signalling whether that request will probably fail will not really add much information?

0029-loopback-transport/0029-loopback-transport.md
michielbdejong: it's described in the BTP spec, just set `protocolName` to something other than `'ilp'`

emschwartz: Is choosing that protocolName and message format in or out of scope for this protocol?

0029-loopback-transport/0029-loopback-transport.md
emschwartz: Ok maybe I'm misunderstanding what is in scope for this protocol then. Are you saying that this requires an authenticated, encrypted, connection-oriented, framed communication protocol but that senders and receivers can choose whatever protocol they want that provides all of those features? Where does the protocol negotiation happen?

Can implementors make any assumptions about what communication protocol others will support?

> It's the same as for how sender and connector talk to each other, it's a local decision.

Actually it's very different, precisely because it's not a local decision. I have one connector that I talk to and we agree on the protocol when we peer. There are many receivers I want to pay in the world and in order to make this protocol work I need to be able to communicate to each of them. We all need to coordinate about what communication protocols we're going to support and expect other people to support.

0029-loopback-transport/0029-loopback-transport.md
emschwartz: Can that higher level protocol communicate over the channel established for loopback or does it need a separate channel?

0029-loopback-transport/0029-loopback-transport.md
emschwartz: If implementations are expected to implement both and one of them is lower-latency, as you describe, why would you use the other?

0029-loopback-transport/0029-loopback-transport.md
michielbdejong: out of scope, but it could be part of the setup protocol

0029-loopback-transport/0029-loopback-transport.md
michielbdejong: sure, as long as there's a way to distinguish the different types of messages

emschwartz: Does this protocol give you a way to do that or if that also out of scope?

0029-loopback-transport/0029-loopback-transport.md
emschwartz: Could you please make a list of the inputs to this protocol (the things that need to be negotiated or communicated up front) like there is in the [PSK2 spec](https://github.com/interledger/rfcs/blob/master/0025-pre-shared-key-2/0025-pre-shared-key-2.md#setup)?

michielbdejong: see https://github.com/interledger/rfcs/pull/413/files#diff-33e3b9e69eaad2fce9a64285ee83f39dR35 (follow-up PR to this one, which deals with the setup flow)

0029-loopback-transport/0029-loopback-transport.md
michielbdejong: I think what you're trying to say with this thread is that there are advanced features (like updating the requested amount of a payment that is already in progress) that are implemented in https://github.com/interledgerjs/ilp-protocol-stream (which by the way, is awesome! I really like what you're working on there!) that are not implemented in LT. And you're right, LT is a very simple transport protocol, that does not have such advanced features. Let's chat more about it at our face-to-face meeting next week!

0029-loopback-transport/0029-loopback-transport.md
emschwartz: If those are the default options, are implementations expected to include both? If you try to connect to one that uses an incompatible protocol or version, how do you know and what error if any should you expect?

In general, if you design a protocol that has multiple options you need to specify how the negotiation happens and what the errors are if incompatible versions try talking to one another.

0029-loopback-transport/0029-loopback-transport.md
emschwartz: I'm not suggesting another protocol for signaling whether it supports IL-DCP, my point is that this spec should specify what is expected of implementations and what others can rely on. If I implement it such that it awaits an IL-DCP call and throws an error if it doesn't hear anything back, have I implemented this spec (im)properly?

0029-loopback-transport/0029-loopback-transport.md
michielbdejong: Discovering the ledger-layer details of how the sender can connect to the receiver is described in the follow-up PR, #413. Let's continue this discussion there!

0029-loopback-transport/0029-loopback-transport.md
michielbdejong: You're right, I should probably leave the specific way in which the sender discovers the loopback address out of scope. For instance, it could be that the sender announces their own loopback address to the receiver using a route broadcast, or that their ILPv4 link uses hard-configured addresses.

michielbdejong: Hm, after trying that in https://github.com/interledger/rfcs/pull/416/commits/08e86ac2d68d46165dd7827a23e79b5530627aa8 I reverted it again because I realized it only makes the setup harder. So let's say the receiver always supports IL-DCP, and the sender always uses IL-DCP to discover the loopback address.

So then yes, the sender can throw an error if the IL-DCP call fails.

emschwartz:
In general I think the idea of using a real-time communication link between the sender and receiver may have some advantages over sending data (especially larger amounts of data) through ILP. However, I think there are some major issues with this spec as it currently stands.

## "Simplicity" and protocol design

The goal is not only to make each protocol layer as simple as possible. It is to find a right set of shared features and the right degree of flexibility to expose useful functionality to protocols or applications built on top. It is trivial to build the simplest protocol at any layer simply by removing all of the features, but that would not be a useful addition to the stack. It is also important to think about the interfaces that each layer expects from the one below it and provides to the one above it. Finally, every option a protocol provides is both an additional degree of implementation complexity and an opportunity for incompatibilities to arise.

I think this iteration of this spec and the responses to some of my questions above suggest an overemphasis on simplicity without enough consideration of what features would be useful for a transport protocol to provide. Anything that is out of scope for this protocol is in scope for something else, so claiming the protocol is simple because many features and issues are unaddressed is like saying one's house is clean because all of the dirt has been moved under the rugs. Some features can be debated, but others are either going into the transport layer or will need to be reimplemented by every protocol built on top of it.

## Transport layer features

These are features that transport protocols probably should handle:
- Condition generation and fulfillment
- Determining the path exchange rate
- Fulfilling packets only if enough money arrives
- Determining the path Maximum Packet Amount

These are more debatable, but I think probably should be considered part of this layer:
- Automatically chunking payments into packets depending on the path Maximum Packet Amount
- Providing an encrypted, authenticated communication channel between the sender and receiver (whether over ILP or IP)
- Streaming payments such that the amount to send/receive can be adjusted over the course of an application-level interaction (such as watching a movie and paying as you go)

## PSK2 and Loopback are both incomplete

Both PSK2 and Loopback implement the first set of features listed above, but neither handle the second set of features.

PSK2 uses a shared secret, some minimal cryptography (HMAC and AES-GCM, which are provided by most programming languages' standard crypto libraries), and literally [13 bytes](https://github.com/interledger/rfcs/blob/master/0025-pre-shared-key-2/0025-pre-shared-key-2.md#data-packet) of data to implement these features.

Loopback uses an unspecified connection-oriented, bidirectional, encrypted, authenticated channel to implement these features (but [does not expose a data channel](https://github.com/interledgerjs/ilp/pull/134#discussion_r178597607) to the protocols built on top of it). Since the exact protocol is used is not part of the spec, protocols built on top must add a negotiation step.

In any case, **implementing the first set of features is simple, because the set of features is incomplete.**

Any protocol or application that builds on top of either one would need to handle chunking payments to account for different paths' Maximum Packet Amounts. Also, most applications will want some sender-receiver communication, so transport protocols can either provide that or we can have every application (protocol) handle that in its own way.

## What data, if any, should go over ILP?

I think there is a reasonable question to be asked about what data should be transmitted through ILP versus over the Internet.

ILP is obviously not currently optimized for sending large amounts of data and would be considerably slower for sending something like a large video stream than an Internet connection.

However, it seems crazy to require a connection over the Internet _just_ to transmit the amount that arrived at the receiver and the fulfillment and adding additional latency to every packet merely to avoid standardizing the 6 fields used in PSK2 (3 excluding the encryption envelope).

The main benefit I see of establishing an Internet connection between the sender and receiver is to enable larger amounts of data to be sent more efficiently than if it went through ILP. Even so, we might question whether every application protocol will want such a connection (versus using a more context-specific communication method).

Nevertheless, Loopback currently requires such a communication channel but does not even provide one to higher-level protocols. Users of the protocol would need to determine for themselves how to differentiate between loopback messages and other data sent over the same channel. Loopback thus seems more complicated to use and even implement than PSK2, while still missing the actually complex but useful features that PSK2 also lacks.

michielbdejong:
> Loopback thus seems more complicated to use and even implement than PSK2

Why would you say that?  [LT](https://github.com/interledgerjs/ilp/blob/mj-lt/src/lib/lt.js) is 64 lines of code, yet [PSK2](https://github.com/interledgerjs/ilp-protocol-psk2) is about 2000, so that alone shows you did a great job implementing PSK2, but I think if you say it's easy to implement, that's just because you're an above-average programmer, not because it would be easy for average programmers to implement! :)

> it seems crazy to require a connection over the Internet just to transmit the amount that arrived at the receiver and the fulfillment and adding additional latency to every packet merely to avoid standardizing the 6 fields used in PSK2 (3 excluding the encryption envelope)

I have to correct you there, sorry, you're still seeing Interledger from a PSK-centric perspective. You might just as well say that it seems crazy to require 6 standardized fields used in PSK2 (3 excluding the encryption envelope)  just to transmit the amount that arrived at the receiver and the fulfillment and adding additional complexity to every packet merely to avoid a connection over the Internet.

If you still think LT seems crazy now, then wait for the enlightenment ceremony next week! ;)

adrianhopebailie:
@emschwartz I think a lot of this criticism is premature as it focuses on questions that are best answered with further experimentation.

They are good questions but a lot of the statements you are making about what is and isn't required in a transport protocol seem subjective and mostly based on what is in the one transport protocol we have today as opposed to what has been demonstrated to be important requirements through use in applications. For example, the establishment of the condition and fulfillment could be done entirely by the receiver and the condition passed to the sender via SPSP and then used in a variety of transport protocols. The fact that PSK defines a way to establish these doesn't mean all transport protocols must.

@michielbdejong this does beg the question, is Loopback really a transport protocol if there is no definition of how the `data` in the ILP Packets are formatted? Is our whole model of transport protocols appropriate when a lot of what we are describing is actually "setup"?

To my mind, a transport protocol is a collection of features that address common requirements of a number of applications (i.e. it saves them all re-implementing those features in their own application layer protocols). We can't say what those requirements and features are for sure until we have a **variety** of applications using the transport protocol.

The design of PSK (and it seems ILP Streams too, although I am still looking at that) are being driven by a very specific set of applications and use cases like streaming content and paid APIs. The majority of the world's payments don't work that way so I don't see any value in evaluating Loopback by comparing it to PSK. All that does is perpetuate this tunnel vision where the transport design is being driven by a limited set of applications. We should be evaluating a transport on how easy it is to use within a wider set of applications and preferably applications that are actually in use already.

> What data, if any, should go over ILP?

Why is answering this question useful?

emschwartz:
> Why would you say that? LT is 64 lines of code

As mentioned, I think Loopback is incomplete and appears "simple" because it's lacking features that anyone using it will just have to implement themselves.

Users of the current Loopback spec/implementation need to:
- Negotiate about the communication protocol to be used
- Handle errors when one side tries to use an unsupported protocol and/or deal with the case where two implementations do not support the same protocol
- Negotiate the `protocolName` they will use for sending messages over the loopback channel, unless they will use yet another channel for sending application protocol data between them
- Modify existing plugins to support sending custom data because none of them currently do
- Set up two loopback channels if they want to be able to send money in both directions (currently the receiver does not have a way to get the address of the sender)
- Determine the path's Maximum Packet Amount for themselves and make sure they send packets that are smaller
- Figure out which class of ILP errors they will retry packets on and which are unrecoverable
- The receiver needs a valid TLS certificate
- The receiver needs a hardened connector instance and sufficient DoS mitigations for it to stay up on the public internet

> additional complexity to every packet merely to avoid a connection over the Internet

Are you critiquing the implementation complexity or the inclusion of 20 bytes in the packet? I would happily take the 20 bytes if it meant one fewer trip across the internet.

> I think a lot of this criticism is premature as it focuses on questions that are best answered with further experimentation.

@adrianhopebailie We have done such experimentation. We have worked with IPR, PSK, PSK2, paystream, and now STREAM and we have tried building each of those into SPSP (including invoices), HTTP-ILP, and Web Monetization. My comments are based on the results of those experiments.

> For example, the establishment of the condition and fulfillment could be done entirely by the receiver

When I say "Condition generation and fulfillment" I mean that the transport protocol needs to stipulate how these are generated, I am not assuming it's done in the PSK way. IPR handles condition and fulfillment generation.

> Why is answering this question useful?

The main difference between encrypting data and putting it in the ILP packet and establishing a channel over the Internet, is whether you are sending data over ILP or over IP. You could send whatever protocol data you want over either channel.

michielbdejong:
> Is our whole model of transport protocols appropriate when a lot of what we are describing is actually "setup"?

Good question! Indeed, LT is not an end-to-end protocol, and I think that is part of the "tunnel vision" which is making this such a long discussion about such a short spec.

Let's park this PR until Monday, and have a face-to-face discussion about it then?

0029-loopback-transport/0029-loopback-transport.md
michielbdejong: Right, sorry, reading this back, what I meant to say is that a *setup protocol* (like http-ilp, payment-pointers, web-monetization, web-payments, ...) should provide access to both. An application layer protocol in the sense of "the paid audio/video streaming protocol" which is used as a building block for an application, should probably pick one of the two, and then it makes sense to have separately implemented building blocks like that for developers to use.

PR #419, opened 2018-04-06
Update 0001-interledger-architecture.md

BobWay:
Again, commenting inline. I didn't make direct edits, because I suspect others may disagree with my characterization.

### DIFF
--- a/0001-interledger-architecture/0001-interledger-architecture.md
+++ b/0001-interledger-architecture/0001-interledger-architecture.md
@@ -96,6 +96,14 @@ The Interledger stack is separated into four layers:
 
 ### Application Layer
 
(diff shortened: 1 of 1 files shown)


0001-interledger-architecture/0001-interledger-architecture.md
```diff
@@ -96,6 +96,14 @@ The Interledger stack is separated into four layers:
 
 ### Application Layer
 
+> An *application layer* protocol (like HTTPS) presents an API for the actual application to use. It does not specify an algorithm that an application must implement. As such, I think our layering metaphor is miss-layered. (see comments in SPSP)
+>
+> The SPSP protocol provides session information and destination account discovery. It does not do the rest of what is described below.
+>
+> As described below, I don't see any ILP RFC's which actually specifiy an application layer protocol.
```
emschwartz: I agree that we've specified setup protocols rather than application layer protocols. I think an application layer protocol would do things like communicate details about an invoice or receiver via an Interledger transport layer protocol. The setup is how you kick off the ILP interaction

PR #424, opened 2018-04-19
[postponed] docs: ILPv4 whitepaper

michielbdejong:
We talked a while ago about publishing an updated whitepaper, here's a strawman of what that might look like.


michielbdejong:
To read as pdf: https://github.com/interledger/rfcs/blob/mj-whitepaper-ilpv4/whitepaper/interledger.pdf

michielbdejong:
List of authors is an estimate :) roughly based on  https://github.com/interledger/rfcs/graphs/contributors

michielbdejong:
Some open questions:
* is anything missing that should be in scope?
* do we want to use the current three-step structure of first explaining HTLCs, then explaining transfers with destinations on top of that, and then explaining packets with data on top of that, or do we want to consider and compare alternative argument structures?
* is the description of ILP addresses correct and complete?
* for timing disputes, do we want to say the fulfiller gets the benefit of the doubt, or discuss the options of a) using post-transaction messaging and b) using ilp-connector's "disputes are ok" approach? or do we even want to leave accounting semantics out of scope?
* do we want to say bilateral transfer protocols should in general be idempotent, even though BTP/2.0 is actually defined not to be? should we consider that a relaxed implementation of the proper definition, or describe both options?
* should we describe both bilateral transfer protocols with and without acknowledgement, or pick one? (BTP/1.0 did have them, BTP/2.0 doesn't)
* should add references (to the v1 whitepaper, to OER, and probably some others) and fix overfull hboxes
* should probably add a diagram of messaging in e.g. a three-transfer payment, and one visualizing 
 Alice, Bob and Charlie in a network graph
* any other topics that the whitepaper needs to discuss? relationship to other payment systems, with discussion of pros and cons, for instance?

sharafian:
> for timing disputes, do we want to say the fulfiller gets the benefit of the doubt, or discuss the options of a) using post-transaction messaging and b) using ilp-connector's "disputes are ok" approach? or do we even want to leave accounting semantics out of scope?

I strongly feel that 'disputes are ok' is the best approach, because disputes are the mechanism by which the risk against your peer is limited. If the fulfiller gets the benefit of the doubt, they can steal an infinite amount of money from you.

BobWay:
@sharafian as is tradition, I feel strongly the other way. I think the difference depends on your expectation for the size of the payment and relationship between direct parties.

When you "timeout" a transaction you are making a direct statement the the party to the right of you. You are saying rather assertively, "I'm not going to pay you. Period. If you make the mistake of paying the person to the right of you, then that is your loss." You are also saying, "I'm changing the agreed upon state of our accounting relationship unilaterally (probably without even telling you). If you think the state is different, then you are wrong." 

To make this process even plausibly work, each party needs to take the concept of "time" very seriously. Clock synchronization and telescoping timeouts become mandatory. The shorter the end-to-end timeout the more certainty you need in understanding the "time slop" at each bi-lateral step.

I haven't heard much discussion about monitoring time, mostly because when that seems important, the research team just says "payments are going to get smaller, and that won't be as important anymore." However, those presumptions do not hold for the xCurrent development team.

In my SNAP deck, I propose a "split the difference" process. It involves a "last call" message that propagates left to right, just like the prepare propagation. While that message is propagating, one of two things will happen. Either it will cross a fulfill message currently propagating right to left, or it will make it all the way to the right end. If the "last call" crosses the fulfill message, then its propagation can be stopped. The fulfill message will naturally inform everyone waiting. If the "last call" makes it to the right most end, then it is obviously too late for the next to last party to accept a new fulfillment. At that point, they can just propagate an authoritative "reverse/reject" message from right to left informing everyone waiting.

If you are thinking about this problem academically and abstractly, either solution would seem to work. However, if you are explaining the options to banks handling non-micropayments, then only the later is likely to be received warmly.

michielbdejong:
>  If the fulfiller gets the benefit of the doubt, they can steal an infinite amount of money from you.

That's why I added the remark "Note that this allows them to be owed the amount of the transfer, but they still rely on the other party's cooperation to actually net this debt against a transfer in the opposite direction, so it will not be a lucrative way for one agent to steal from a neighboring one".

And the question is, do we want to mention
The way disputed balance was used in PluginVirtual, is effectively a cleverly simple implementation of that limit on the "other party's cooperation"! :)

I think if we already say that at some point you want to stop cooperating, then it doesn't add much to describe our (effective, but pretty confusing) implementation of it. So I agree with Bob that it's unintuitive to mention that implementation to people who think of disputes as a bug (which is probably most readers). We could mention it, but the amount of confusion added would probably be bigger than the amount of useful information conveyed?

whitepaper/interledger.aux
```diff
@@ -0,0 +1,46 @@
+\relax
```
dappelt: When working with git + tex, I prefer to commit only source files like .tex and .bib. I put intermediary and compiled files like .dvi, .aux, .pdf etc. into .gitignore.

sharafian:
> So I agree with Bob that it's unintuitive to mention that implementation to people who think of disputes as a bug (which is probably most readers). We could mention it, but the amount of confusion added would probably be bigger than the amount of useful information conveyed?

If you stop cooperating, then it's because you disagree with your peer. I don't understand how having a dispute isn't the most intuitive way to represent a disagreement about the balance.

On the flip-side, an issue that shatters the basic security guarantees of the system by allowing any peer to steal an unbounded amount of money is a much larger red flag. We can't just hand-wave that away by saying "you'll accept every fulfill, but also have some system in place to make sure your peer doesn't steal too much"

What would an alternative implementation look like? Would you just accept the fulfill but then add to a pool of disputed balance? However you shuffle the numbers around it's fundamentally based on your view of the balance differing from your peer's.

> I haven't heard much discussion about monitoring time, mostly because when that seems important, the research team just says "payments are going to get smaller, and that won't be as important anymore." However, those presumptions do not hold for the xCurrent development team.

If you can establish a bound for clock skew and latency, then you can set the min message window to some value larger than that. A clock skew of less than a second is reasonable to expect if people are using NTP. There's an optimal point where risk of failing to pass on the fulfillment balances against currency risk from the free option that you get.

> In my SNAP deck, I propose a "split the difference" process. It involves a "last call" message that propagates left to right, just like the prepare propagation. While that message is propagating, one of two things will happen. Either it will cross a fulfill message currently propagating right to left, or it will make it all the way to the right end. If the "last call" crosses the fulfill message, then its propagation can be stopped. The fulfill message will naturally inform everyone waiting. If the "last call" makes it to the right most end, then it is obviously too late for the next to last party to accept a new fulfillment. At that point, they can just propagate an authoritative "reverse/reject" message from right to left informing everyone waiting.

It sounds like that 'last call' is effectively cancelling your outgoing escrow. What's to stop someone from sending out a last call message right after they send a prepare in order to make payments for free?

michielbdejong:
> it's fundamentally based on your view of the balance differing from your peer's.

The big difference is that if you record late fulfills explicitly, instead of effectively lumping them in with the rejects, then you *know* a dispute exists, and you know for how much, so failed payments will become easier to debug. If your peer just records your late fulfills as failed payments, then later they will start rejecting your payments and that will come as a surpise to you. You're throwing away debugging information, and blurring the lines between intended behavior of the system as a whole, and intended behavior of each node individually. Let's have a video call to discuss the paragraphs about Timing Disputes further! Maybe we can find a phrasing that does right to both views.

What did you think of the other paragraphs?

BobWay:
I really like how you’ve described this as agreement between two parties on a sequence of HTLA’s. That is much better in my mind than focusing on the post HTLA view of the balance from each perspective.

michielbdejong:
@justmoon @emschwartz @adrianhopebailie any feedback from you?

@dappelt @sharafian @BobWay anything you would like to see changed apart from what was already discussed?

whitepaper/interledger.tex
```diff
@@ -0,0 +1,351 @@
...
+\date{April 2018}
+\maketitle
+\begin{abstract}
+In 2015, Evan Schwartz and Stefan Thomas published version 1 of the Interledger protocol (ILP). After three years of research,
+we are proud to present its evolution, ILP version 4. ILPv4 uses a chain of Hash Time Lock Agreements (HTLAs) between network neighbors, to form
+multi-transfer Interledger payments that remotely interconnect non-neighbors.
+\end{abstract}
+
+\section{Conditional Transfers between Neighbors}
+Consider a network of agents, where each agent has accounting relationships with at least one other agent in the network. An accounting
```
adrianhopebailie: I prefer standard network graph terminology:

s/agents/nodes/
s/neighbors/peers/

whitepaper/interledger.tex
```diff
@@ -0,0 +1,351 @@
...
+relationship between two agents is defined by a unit of value and a log of HTLAs whose amounts are expressed in this unit of value, so that they
+can be netted against each other (total amount in one direction, minus total amount in the other direction), to define what one agent owes the other.
+
+\subsection{Hash Time Lock Agreements}
+Suppose Alice and Bob are two agents who are neighbors in an Interledger network. A Hash Time Lock Agreement (HTLA) between them is then
```
adrianhopebailie: s/two agents who are neighbors/two nodes who are peers/

whitepaper/interledger.tex
```diff
@@ -0,0 +1,351 @@
...
+defined by a {\tt hash}, a {\tt time}, and an agreed {\tt amount}. If Alice sends such a HTLA to Bob, and Bob rejects the agreement, the state of their accounting
```
adrianhopebailie: s/a {\tt time}/an expiry {\tt timestamp}/

whitepaper/interledger.tex
```diff
@@ -0,0 +1,351 @@
...
+relationship does not change. However, if Bob presents the SHA256 preimage of the {\tt hash} before {\tt time}, then the HTLA is added to the relationship's log, to say Alice
```
adrianhopebailie: s/before {\tt time}/before expiry {\tt timestamp}/

whitepaper/interledger.tex
```diff
@@ -0,0 +1,351 @@
...
+An ILP address is a string, consisting of one or more segments, separated by dots ({\tt .}). Each segment may contain alphanumeric characters (upper or lower case, case-sensitive),
+underscores ({\tt \_}),
+tildes ({\tt \textasciitilde}),
+and hyphens ({\tt -}).
+
+\subsection{Static Routing}
+In Interledger networks where the set of agents participating in the network and the connections between them don't change,
+an easy way to make sure that all agents in the network know how to route multi-transfer payments towards their destination, would be to assign an ILP address to each agent, and statically configure the forwarding rules for each destination in each agent. In any case, the first two agents to form a new network can choose their own ILP addresses that way.
+But when a new agent joins, one of two things can happen:
+they may become a child of the agent through which they get joined to the network, or become a first-class citizen.
```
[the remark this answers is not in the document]
adrianhopebailie: -1 to "first-class citizen"
+1 to "parent", "child", "peer" - this is pretty standard terminology for describing a network topology

whitepaper/interledger.tex
```diff
@@ -0,0 +1,351 @@
...
+
+\subsection{Sub-Addresses}
+If the new agent becomes a child, then it takes the address of the parent through which it joins the network, and adds one or more segments at the end. By using its parent's address as a
+prefix, other agents in the network will always be able to route payments to the new agent, provided they know how to route a payment to the new agent's parent. This is the hierarchical
+way to generate new ILP addresses.
+
+\subsection{Route Broadcasts}
```
adrianhopebailie: I think we need to model the case of conflicting broadcasts more thoroughly and be more prescriptive about the behaviour of a node that receives conflicting broadcasts.

michielbdejong: how would you distinguish conflict (so address collision, two destinations call themselves by the same name) from the (desirable!) situation where there are two or more alternative routes to the same destination?

michielbdejong: cc @justmoon

michielbdejong:
@justmoon could you quickly give your first impression of whether we're on the right track with this?

whitepaper/interledger.tex
```diff
@@ -0,0 +1,351 @@
...
+\subsection{End-to-end Data}
+As multiple transfers are chained into an Interledger payment, the initiator of the first transfer in the chain (the 'sender') may want to send some data to the
+beneficiary of the last transfer in the chain (the 'receiver'). For this, Prepare packets have a {\tt data} field. Each agent who participates in a chain of Interledger
+transfers is expected to copy this data unchanged from their incoming transfer's Prepare packet to their outgoing transfer's Prepare packet. Likewise,
+just like one {\tt Prepare} packet triggers the next, and its data is carried over, each {\tt Fulfill} packet will generally trigger another {\tt Fulfill} packet, and each {\tt Reject}
+packet will trigger another {\tt Reject} packet, on the return path. For these, the {\tt data} field should also be copied over, so that the receiver can relay data back
+to the sender. This transportation of end-to-end data is not guaranteed by the chain of hashlock conditions, but can still be useful, for instance if sender and receiver sign
+the data payload cryptographically.
+
+\subsection{Prepare}
```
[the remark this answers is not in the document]
michielbdejong: also, as @justmoon requested, I'll move these three sections to the appendix, so the main text will only say something like "there is an OER message to prepare/fulfill/reject an ILP transfer, see appendix A,B, and C, respectively" and then only explain that a successful transfer is `>prepare` + `<fulfill` and a failed one is `>prepare` + `<reject`

michielbdejong:
@justmoon I have a pretty good idea of how I want to move the three packet descriptions to the appendix, but leave the description of ILP address format inline, do you agree with that intention?

@justmoon about adding more text about motivation: i think we can do a problem statement that explains that we want to do multi-hop, and then present hashlock chaining as the solution. And then second step is choice for sha256, and third step is to explain we want to standardize the OER packets with fixed-length fields, so that they can be forwarded efficiently. What do you think of that general setup? How would you word it yourself?

michielbdejong:
Closing this PR temporarily, until we can gather more feedback.

PR #426, opened 2018-05-07
FIX: Changes to reflect ILPv4 model

### LINKED ISSUE #410: RFC - 15 - peer.config is an invalid address
According to the ILP Address spec the address `peer.config` is invalid.

> Destination addresses MUST NOT end in a period (.) character, and MUST contain at least two segments after the allocation scheme prefix.

Since this address is used in protocols such as ILDCP it's clear that not all implementations are strictly following the ILP Address specification.

The specification should be updated or ILDCP should use a valid address.

This was uncovered in the Quilt implementation when the ILP Address parsing libs were upgraded to strictly follow the rules in the spec. See https://github.com/hyperledger/quilt/issues/128

### LINKED ISSUE #422: Update ILP Address spec based on discussions at community summit
Some insights from the community summit that need to be incorporated into the ILP Addresses RFC:

 - No more address prefixes (all nodes have an address)
 - We don't address ledgers (ledgers are not directly part of the network)
 - Routing info is exchanged using ILP packets using the special `peer.route` 
 - Packets addressed with `peer.*` are never forwarded (these are bilateral)

NOTE: Some of this may be better suited to be added to a routing protocol spec

adrianhopebailie:
- Remove concept of "Address Prefixes"
- Remove reference to ledgers in addressing
- Remove concept of local vs forwarding behaviour at connectors

Fixes #422 
Fixes #410

### DIFF
--- a/0015-ilp-addresses/0015-ilp-addresses.md
+++ b/0015-ilp-addresses/0015-ilp-addresses.md
@@ -106,7 +106,7 @@ As an example, the sender may encode the identifier of a shared key in the addre
 
 A node can be addressed relative to another _"parent"_ connector. Smaller and lesser-known connectors may find it easier to advertise routes to themselves in this manner. That way, rather than needing to have their address known to every connector in the same large neighborhood, only the parent connector needs to know how to route to the smaller node.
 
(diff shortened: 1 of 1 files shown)


adrianhopebailie:
Addresses: #410

PR #428, opened 2018-05-10
Add Stale app to automatically close old issues

emschwartz:
https://probot.github.io/apps/stale/

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,28 @@
+# Config for the Stale app https://probot.github.io/apps/stale/
+
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 30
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,24 @@
...
+exemptLabels:
+  - pinned
+  - security
+
+# Label to use when marking an issue as stale
+staleLabel: wontfix
+
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
```
adrianhopebailie: Let's change this default text to suggest that if an issue is important to someone they raise it on a community call or via Gitter

PR #438, opened 2018-06-13
add missing local address scheme to ABNF spec and regex

uroshercog:
The address spec is missing the `local` allocation scheme as defined in the following [table](https://github.com/interledger/rfcs/blob/master/0015-ilp-addresses/0015-ilp-addresses.md#allocation-schemes).


adrianhopebailie:
Thanks for the fix @uroshercog

PR #441, opened 2018-07-18
Feat(WIP): update for new thinking on web monetization

sharafian:
- Changes from the SPSP-based API to a STREAM-based one.
- Adds description of the polyfill that can be used for WM on mobile and without extensions.
- Uses Web Monetization Handlers, iframes that can process ILP packets for the user.

Places where it could use more work:

- Currently, the STREAM API is packaged up via webpack and an identical API is exposed. Should we make it more web-y?
- Are all of the method names reasonable for a Web standard?

I'd love to feel confident about all of our choices here so that it'll be ready when it gets integrated into a lot of different sites.


0028-web-monetization/0028-web-monetization.md
```diff
@@ -1,79 +1,296 @@
...
 
 With advertisements, the browser decides whether to display the ads and the user decides whether to engage with the ads. With Web Monetization, the browser decides whether to pay the site and, if so, how much to pay.
 
-Web Monetization makes use of [Payment Pointers](../0026-payment-pointers/0026-payment-pointers.md) in order to associate a site with an ILP destination, and [SPSP](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md) in order to set up ILP payments.
-
-The browser or browser extension which provides Web Monetization will likely use the [Ledger Plugin Interface](../0004-ledger-plugin-interface/0004-ledger-plugin-interface) in order to trigger payment when a site requests it.
+Web Monetization makes use of [ILP/STREAM](../0029-stream/0029-stream.md) provide a high-level way to send money and data, while still providing tremendous flexibility.
```
emschwartz: Missing a "to" after the link

0028-web-monetization/0028-web-monetization.md
```diff
@@ -1,79 +1,296 @@
 ---
 title: Web Monetization
```
emschwartz: Why is this called Web Monetization instead of something like the Web Micropayment API?

0028-web-monetization/0028-web-monetization.md
```diff
@@ -1,79 +1,296 @@
...
+- The user visits their provider's webpage.
+- The provider's webpage calls `window.WebMonetization.register`.
+- The browser will display a native pop-up box asking the user whether they want to set the handler proposed by the page.
+  - If the user confirms, `handlerUrl` is stored in the browser's state and `window.WebMonetization.register` resolves to true.
+  - If the user cancels the registration, `window.WebMonetization.register` will resolve to false.
+
+### Monetization
+
+- The user visits a webpage.
+- When the page wants to open an ILP/STREAM connection, it calls `window.WebMonetization.monetize()` with a `destinationAccount` and `sharedSecret`.
```
emschwartz: Why `.monetize` instead of `.createConnection` or `.createStream`?

0028-web-monetization/0028-web-monetization.md
```diff
@@ -1,79 +1,296 @@
...
+increasing values.
+
+##### Parameters
+
+- `amount: String` - The amount to send (in our units).
+
+#### Wait for Money to be Sent
+
+```ts
+stream.sendTotal(amount: String, timeout?: Number): Promise<void>
```
emschwartz: In the stream module the timeout is in an options object

0028-web-monetization/0028-web-monetization.md
```diff
@@ -1,79 +1,296 @@
...
+```
 
-Rejects with `Error` if the user's browser decides not to send money to this page.
+Sends data over the stream, which will be emitted as `data` events on the other
+side. The data may be broken into chunks or buffered; one `send` call doesn't
+necessarily correspond to a single `data` event.
 
+##### Parameters
+
+- `msg: Uint8Array | String` - The data to send over the stream.
```
emschwartz: If the user sends a string, will the other side get a string or just bytes? (I assume it's bytes but it would be worth mentioning)

0028-web-monetization/0028-web-monetization.md
emschwartz: It would be worth describing what type of flexibility this provides (e.g. opening connections to multiple destinations)

On that note, can a site open an unlimited number of connections?

0028-web-monetization/0028-web-monetization.md
```diff
@@ -1,79 +1,296 @@
...
 ## Specification
 
-#### monetize
+### Browser API / Polyfill API
+
+#### Register
 
-`window.monetize({ receiver: string }) -> Promise<void>`
+```ts
+window.WebMonetization.register({
```
adrianhopebailie: Based on the work done on the payment handler API I would recommend that this is done using [`Workers`](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API). Payment Handler is the only standard I am aware of that has a pattern of:
 1. Registration of service from Origin A
 2. Invocation from website at Origin B where browser mediates the communication from website to service

The pattern for extending the [`ServiceWorkerGlobalScope`](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerGlobalScope) is to add the methods, events, and properties you need and then extend [`ServiceWorkerGlobalScope.registration`](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration) with the necessary methods/properties that will be called during registration.

In the case of monetization there is no need to provide service meta-data other than the name so maybe we can skip the extra ServiceWorkerRegistration hooks for now and just identify the provider by origin.

This might look like this in WebIDL:

```webidl

partial interface ServiceWorkerGlobalScope {
  attribute EventHandler onmonetize;
};

[Constructor(DOMString type, MonetizeEventInit eventInitDict), Exposed=ServiceWorker]
interface MonetizeEvent : ExtendableEvent {
  readonly attribute USVString origin;
  readonly attribute DOMString destinationAccount;
  readonly attribute DOMString sharedSecret;
  void respondWith(Promise<IlpConnectionInit> handlerResponsePromise);
};

dictionary MonetizeEventInit : ExtendableEventInit {
  readonly attribute USVString origin;
  readonly attribute DOMString destinationAccount;
  readonly attribute DOMString sharedSecret;
};

dictionary IlpConnectionInit {
  readonly attribute DOMString connectionTag;
  readonly attribute double lastPacketExchangeRate;
  readonly attribute double minimumAcceptableExchangeRate;
  readonly attribute DOMString totalDelivered;
  readonly attribute DOMString totalSent;
  readonly attribute DOMString totalReceived;
}
```

It would be used by a provider like this:

```js
self.addEventListener('monetize', monetizeEvent => {
  console.log('Provider got monetize request', monetizeEvent)

  // 1. Peform work to create uplink (open WS to provider)
  // 2. Get handle on channel to send ilp packets back to website
  // 3. Return connection meta-data to be used to construct connection object
  const channel = monetizeEvent.ilpChannel
  const uplink = createUplink()

  uplink.on('data', reqFromUplink => {
    reqFromUplink.replyWith(channel.sendRequest(reqFromUplink))
  })

  channel.on('data', reqFromWebsite => {
    reqFromWebsite.replyWith(uplink.sendRequest(reqFromWebsite))
  })

  monetizeEvent.replyWith({
    connectionTag,
    lastPacketExchangeRate,
    minimumAcceptableExchangeRate,
    totalDelivered,
    totalSent,
    totalReceived	
  })
})
```

sharafian: Is the idea that the service worker running on the provider's origin would be pulling in the STREAM library? Then we would need a more complex RPC protocol in order to allow any ILP/STREAM method call to be triggered on the service worker from a page that uses web monetization

adrianhopebailie: No, it will only need to establish the BTP uplink to it's own server and then pass packets between that link and the website (via postMessage)

PR #442, opened 2018-07-24
Added statements of accounts

dora-gt:
Added statements of accounts. If you are uncomfortable of some expressions, please modify as you think because my English is not so good.


adrianhopebailie:
@dora-gt thanks for prodding on this. This document is intended to be a requirements specification for a node so I have changed the text you provided to reflect that.

PR #448, opened 2018-08-02
Update 0020-explain-like-im-five.md

fjx2000:
Remove a redundant word "about",  there were two, just removed one.


sharafian:
LGTM; can you bump the draft version so it passes CI?

adrianhopebailie:
Thanks for the fix @fjx2000

Issue #432, opened 2018-05-18
Document route broadcasting protocol


adrianhopebailie:
Document the route broadcasting protocol as implemented in [interledgerjs/ilp-connector](https://github.com/interledgerjs/ilp-connector/)

dora-gt:
Which is the best file to refer to?

adrianhopebailie:
.ts - The .js is auto-generated

dora-gt:
I started to write this.
If anyone else is doing the same thing, please let me know.

adrianhopebailie:
Most of the code is in either:
 - https://github.com/interledgerjs/ilp-connector/tree/master/src/routing , or
 - https://github.com/interledgerjs/ilp-protocol-ccp

@justmoon May have some other suggestions.

dora-gt:
Thank you for the information. I'm reading the code.
This is a bit complicated than ILDCP, so give me a few days to understand the whole view.

dora-gt:
My recognition:

- Routing Table ID
  - UUID that is reset everytime the connector starts up
  - If the UUID is changed, the epoch for the table is reset to `0`
- Epoch
  - How many times the routing table is modified since routing table ID is set.

dora-gt:
I'm writing [here](https://github.com/dora-gt/rfcs/blob/feature/dora-gt/route-broadcasting-protocol/0000-route-broadcasting-protocol/0000-route-broadcasting-protocol.md), though it is WIP.

sappenin:
pinging the to keep the issue open - we should document this protocol.

sappenin:
This is currently being worked on here: https://github.com/interledger/rfcs/pull/455

sappenin:
Re-opening so we keep working on #455

PR #455, opened 2018-08-13
Added the first draft of Route Broadcasting Protocol

dora-gt:
Added the first draft of Route Broadcasting Protocol based on this issue: #432, including ASN.1 definition.
Please check the description.

It seems that there are some unused fields, so either of removing or implementing may be required.


0000-route-broadcasting-protocol/0000-route-broadcasting-protocol.md
```diff
@@ -0,0 +1,160 @@
...
+  - The `code` of the ILP packet is arbitrary, depending on the situation
+  - The `message` of the ILP packet is arbitrary, depending on the situation
+  - The `triggeredBy` of the ILP packet is the ILP address of the node where the error occuered
+  - The `data` of the ILP packet is empty (size: 0)
+
+### Route Update
+#### Procedure
+Route Update is done in the following procedure:
+
+- A node requests an update to the corresponded node with information of route update logs.
```
sappenin: One thing that's unclear here is which node does what, and how this relates to the different relationships defined in [Node Requirements (draft)](https://github.com/interledger/rfcs/blob/fd5eafec4688c16ae2331cf5fe5f633adb29d398/0000-node-requirements/0000-node-requirements.md#link-relations).

If my understanding is correct, NodeA will send a `RouteControlRequest` to NodeB, and then NodeB will enter the `SYNC` mode. Next, NodeA will send a `RouteUpdateRequest` with added or withdrawn routes.

I suppose that if NodeA is a child of NodeB, however, then the original `RouteControlRequest` would fail?

0000-route-broadcasting-protocol/0000-route-broadcasting-protocol.md
```diff
@@ -0,0 +1,160 @@
...
+- The corresponded node updates its routing table, and respond that the request is done.
+- If the request cannot be deserialized or interpreted as appropriate, the corresponded node responds error.
```
sappenin: It would be good to also add summary notes about what should happen between Child, Peer, and Parent nodes with regard to the overall protocol.

0000-route-broadcasting-protocol/RouteBroadcastingProtocol.asn
```diff
@@ -0,0 +1,60 @@
...
+  knownEpochIndex UInt32, -- the epoch that the requesting node knows at that point
+  features SEQUENCE OF Feature -- reserved for the future, currently not used
+}
+
+-- Route Update Request
+
+Auth ::= OCTET STRING (SIZE (32))
+
+Property ::= SEQUENCE {
+  meta UInt8,
```
sappenin: It might be worth defining the meaning of `meta`. See [ilp-protocol-ccp](https://github.com/interledgerjs/ilp-protocol-ccp/blob/master/src/index.ts#L272) for some examples of the data encoded into this byte field.

From what I can tell, the following fields exist: `optional`, `transitive`, `partial`, `isUtf8`.

0000-route-broadcasting-protocol/RouteBroadcastingProtocol.asn
```diff
@@ -0,0 +1,60 @@
...
+  props SEQUENCE OF Property -- reserved for the future, currently not used
+}
+
+-- NOTICE: The log of "toEpochIndex" will NOT be included.
+-- e.g. with this log: [ a, b, c, d ], update request from 0 to 2 means [ a, b ]
+RouteUpdateRequestData ::= SEQUENCE {
+  routingTableId RoutingTableId, -- the routing table ID that the requesting node knows at that point
+  currentEpochIndex UInt32, -- current epoch index of the node that requests the update for the other nodes
+  fromEpochIndex UInt32, -- the epoch index that the log of this update request starts from
+  toEpochIndex UInt32, -- the epoch index that the log of this update request ends at
```
sappenin: I assume this is `inclusive` (as opposed to `exclusive`), but might be worth stating directly. For example, if `fromEpoch=1`, and `toEpoch=3`, then I would expect 3 routes (1, 2, and 3).

0000-route-broadcasting-protocol/RouteBroadcastingProtocol.asn
```diff
@@ -0,0 +1,60 @@
...
+  holdDownTime UInt32, -- reserved for the future, currently not used
```
sappenin: Despite not being used, can we document what this means?

0000-route-broadcasting-protocol/RouteBroadcastingProtocol.asn
```diff
@@ -0,0 +1,60 @@
...
+  speaker Address, -- the ILP address of the requesting node
+  addedRoutes SEQUENCE OF NewRoute, -- the added routes
```
sappenin: Consider changing to `New Routes that the speaker is advertising to the counterpart node`

0000-route-broadcasting-protocol/RouteBroadcastingProtocol.asn
```diff
@@ -0,0 +1,60 @@
...
+  withdrawnRoutes SEQUENCE OF Address -- the withdrawn routes
```
sappenin: Consider changing to: `Withdrawn Routes that the speaker is advertising as being no longer valid or useful.`

dora-gt:
@sappenin Thank you so much for the input.
I'll fix soon.

0000-route-broadcasting-protocol/0000-route-broadcasting-protocol.md
dora-gt: Added `Operation Model` section to make the roles clear.
Does this help?

0000-route-broadcasting-protocol/0000-route-broadcasting-protocol.md
dora-gt: Added `Link Relations and Routing` section. What do you think?

0000-route-broadcasting-protocol/RouteBroadcastingProtocol.asn
dora-gt: I actually knew the options but I hesitated to write it because it seemed that the options were not used. Should I write it nevertheless?

0000-route-broadcasting-protocol/RouteBroadcastingProtocol.asn
dora-gt: I referred to this code:
https://github.com/interledgerjs/ilp-connector/blob/master/src/routing/ccp-sender.ts#L182-L186
and this specification:
https://www.ecma-international.org/ecma-262/5.1/#sec-15.4.4.10

As a conclusion, in my opinion, `3` is not included in your case. See the picture below.
image

Am I wrong?

0000-route-broadcasting-protocol/RouteBroadcastingProtocol.asn
dora-gt: I wanted to specify, but I could not find any reference to determine the function.
@justmoon Could you give me some reference?

dora-gt:
@sappenin I fixed some of your inputs as far as I could at this point.
Some of your inputs need more arguments or references from someone who knows well.

0000-route-broadcasting-protocol/0000-route-broadcasting-protocol.md
```diff
@@ -86,11 +86,36 @@ The left image shows how epochs are stored. The right image shows an example of
...
+- `Child` node
+  - `g.parent-node.child-node`
+  - this is a sub-address of `g.parent-node`
+
+In this case, the parent node already knows the address of the child node, so the routing is a truism. The parent node knows when to transfer ILP packets to the child node, and the child node transfers ILP packets basically to the parent node because it is the gateway to the other addresses for the child node.
+
+On the other hand, in a `Peer` and `Peer` relation, the two nodes have different addresses respectively, and the further connected addresses are unclear. So in this case, it is generally helpful to configure to exchange routing information each other.
+
+Conclusively, the following is the general configuration of nodes depending on the relations.
+- `Peer` and `Peer`
```
sappenin: Consider using `to` instead of `and` in these cases, so it would read like this: `Peer` to `Peer`.

0000-route-broadcasting-protocol/0000-route-broadcasting-protocol.md
```diff
@@ -162,6 +188,7 @@ Route Update is done in the following procedure:
 
 - A node sends a route update request to the corresponded node with information of route update logs.
 - The receiver node updates its routing table, and responds that the request is done.
+- If the receiving node is not configured to receive route updates, the receiving node responds error.
```
sappenin: Minor nit, should read: ...`the receiving node responds with an error`

0000-route-broadcasting-protocol/RouteBroadcastingProtocol.asn
sappenin: I think it would be fine if these definitions were written later (e.g., after merging this PR), although the terms are used in the source-code (though undocumented), and so it would be good to outline them in the event some implementation wanted to start using them.

0000-route-broadcasting-protocol/RouteBroadcastingProtocol.asn
sappenin: Yes, I think you're right. In that case, we should use the term `exclusive`, like this:

`the epoch index that the log of this update request ends at, exclusive`

0000-route-broadcasting-protocol/0000-route-broadcasting-protocol.md
dora-gt: BTP connections could be opened from each side, and `to` makes us imagine it is **directional**, so I thought `and` was more appropriate in this case. Though I don't have the native sense of English, I'll follow your opinion. Amended the lines.

0000-route-broadcasting-protocol/0000-route-broadcasting-protocol.md
dora-gt: Fixed to use prepositions.

sappenin:
bumping to keep this from getting closed by Stalebot.

0000-route-broadcasting-protocol/0000-route-broadcasting-protocol.md
```diff
@@ -0,0 +1,226 @@
...
+  - `g.parent-node`
+- `Child` node
+  - `g.parent-node.child-node`
+  - this is a sub-address of `g.parent-node`
+
+In this case, the parent node already knows the address of the child node, so the routing is a truism. The parent node knows when to transfer ILP packets to the child node, and the child node transfers ILP packets basically to the parent node because it is the gateway to the other addresses for the child node.
+
+On the other hand, in a `Peer` to `Peer` relation, the two nodes have different addresses respectively, and the further connected addresses are unclear. So in this case, it is generally helpful to configure to exchange routing information each other.
+
+Conclusively, the following is the general configuration of nodes depending on the relations.
```
justmoon: s/general/default/

0000-route-broadcasting-protocol/0000-route-broadcasting-protocol.md
```diff
@@ -0,0 +1,226 @@
...
+- `Peer` to `Peer`
+  - Configured to exchange routing information.
+  - Does send route control requests and route update requests that are explained later in the [Requests](#requests) section.
+- `Parent` to `Child`
+  - Configured **NOT** to exchange routing information.
+  - Does **NOT** send route control requests and route update requests.
+  - That said, the nodes MAY be configured to exchange routing information for some specific situations.
+
+### Node Finite State Machine (FSM)
+A node is kind of an FSM that has a status of either:
```
justmoon: Should be clearer that the IDLE/SYNC status is **per counterpart**.

justmoon: Would avoid the wording "kind of" which is imprecise. Would say something like: "Each node consists of a state machine which tracks for each routing counterpart node:"

0000-route-broadcasting-protocol/0000-route-broadcasting-protocol.md
```diff
@@ -0,0 +1,226 @@
...
+  - The `type` of the ILP packet is `ILP Prepare` (type id: 12)
+  - The `amount` of the ILP packet is `0`
+  - The `expiresAt` of the ILP packet is arbitrary
+  - The `executionCondition` of the ILP packet is `Zmh6rfhivXdsj8GLjp+OIAiXFIVu4jOzkCpZHQ1fKSU=` in Base64 format
+  - The `destination` address of the ILP packet is `peer.route.control`
+  - The `data` of the ILP packet is described at `RouteControlRequestData` in [ASN.1 definition](#asn1-definition)
+- Response
+  - The `type` of the ILP packet is `ILP Fulfill` (type id: 13)
+  - The `fulfillment` of the ILP packet is 32byte octet string all filled with zeros
+  - The `data` of the ILP packet is empty (size: 0)
```
justmoon: For forward compatibility, nodes should ignore if the packet data is non-empty.

0000-route-broadcasting-protocol/0000-route-broadcasting-protocol.md
```diff
@@ -0,0 +1,226 @@
...
+- If the request cannot be deserialized or interpreted as appropriate, the corresponded node responds with an error.
+
+#### Packet
+- Request
+  - The `type` of the ILP packet is `ILP Prepare` (type id: 12)
+  - The `amount` of the ILP packet is `0`
+  - The `expiresAt` of the ILP packet is arbitrary
+  - The `executionCondition` of the ILP packet is `Zmh6rfhivXdsj8GLjp+OIAiXFIVu4jOzkCpZHQ1fKSU=` in Base64 format
+  - The `destination` address of the ILP packet is `peer.route.control`
+  - The `data` of the ILP packet is described at `RouteControlRequestData` in [ASN.1 definition](#asn1-definition)
```
justmoon: For forward compatibility, nodes should ignore any extra data that they do not understand.

0000-route-broadcasting-protocol/0000-route-broadcasting-protocol.md
```diff
@@ -0,0 +1,226 @@
...
+  - When the node is shutdown
+  - When the node is restarted
+
+### Routing Table ID
+A routing table has its identifier, that is called **routing table ID**. A routing table ID is used for identifying its instance, and the ID is given when the table is created as empty. A routing table ID is a [UUID v4](https://tools.ietf.org/html/rfc4122#section-4.1.3).
+
+### Epoch
+Every time the routing table is modified, the update is logged and the revision number of the table is increased. The revision number of the table is called an **epoch**.
+
+A node maintains the epoch that is already known to the counter node, and considers it when requesting route updates so that the node can send only the difference (a stack of adding or withdrawing routes) from the last update with the last known epoch as the starting point. So the node doesn't need to send the whole routing table.
```
justmoon: If we use the term "counterpart node", let's use it here as well, instead of "counter node".

0000-route-broadcasting-protocol/RouteBroadcastingProtocol.asn
```diff
@@ -0,0 +1,68 @@
...
+Property ::= SEQUENCE {
+  meta UInt8,
+  id UInt16,
+  value OCTET STRING
+}
+
+NewRoute ::= SEQUENCE {
+  addedRoutePrefix Address, -- the prefix of the added route
+  path SEQUENCE OF Address, -- the hops destinated to the prefix node
+  auth Auth, -- reserved for the future, currently not used
```
justmoon: It's not just reserved for the future! Routes are not yet filtered based on this but correct implementations will generate and correctly update the `auth` field.

Basically, the node originating a route will use a random value. Each node forwarding a route will SHA-256 hash this value.

It's not as good as BGPsec or other heavier schemes for routing security but it is better than nothing. An attacker can choose not to hash the auth value and pass it on unchanged, to make themselves look one hop closer than they are. But they cannot reverse the hash to make themselves look two or more hops closer. They can use an unrelated hash chain but connectors could be set such that this triggers a manual review.

In the future, we may add a routing PKI system to establish the correct base hash so that every knows which is the correct hash chain. Beyond that we may add full BGPsec style signing.

adrianhopebailie: @kruisdraad was discussing authenticated route updates in https://github.com/interledger/network/issues/4#issuecomment-427302643

My proposal to him was to look at using Distributed IDentifiers (DIDs) as a way to identify nodes and to root these in the distributed ledgers that we already use for settlement.

I have an XRP-biased proposal that could be adapted to be more generic: https://github.com/interledger/network/blob/master/docs/xrpl-pki.md

0000-route-broadcasting-protocol/0000-route-broadcasting-protocol.md
dora-gt: Indeed, my bad. Thank you for the input.

0000-route-broadcasting-protocol/0000-route-broadcasting-protocol.md
dora-gt: Yes, the terminology should be unified, thank you.

0000-route-broadcasting-protocol/RouteBroadcastingProtocol.asn
dora-gt: I could not understand the whole security model of `auth`.
How does the receiver verify the auth value?

dora-gt: I think what makes ILP beautiful is its independency. So I think it needs careful discussions to use some ledgers for the base of identities, that makes ILP dependent.

dora-gt:
I'll add some explanations that @justmoon mentioned.
Give me a few days. I'm sorry.

khoslaventures:
@dora-gt this looks like a worthy effort. Still interested in working on this?

dora-gt:
@khoslaventures I'm sorry I could not notice.
Yes, I want to finish this but recently I don't have enough time to work on this... 😞 
I still have incentive to work on this.

adrianhopebailie:
Re-opening this and planning to contribute over the next few weeks with @matdehaast and @DonChangfoot

dora-gt:
I pushed the latest version anyway. Thank you all.

dora-gt:
We have to define how `auth` field is used.
I met Stefan at SF and heard about it but I could not understand the exact security model of it. I'm sorry.

adrianhopebailie:
This issue was moved to https://forum.interledger.org/t/added-the-first-draft-of-route-broadcasting-protocol/283

emschwartz:
I'm finally trying to dig into CCP. Thanks @dora-gt for putting this together -- it's super helpful! Much easier to understand than the code in `ilp-connector` :)

PR #458, opened 2018-08-14
Fix URL re-writer to handle fragments

### LINKED ISSUE #452: Dead links on the website
Some links work when clicked on GitHub, but not on the website.

For example, in the file https://github.com/interledger/rfcs/blob/master/0027-interledger-protocol-4/0027-interledger-protocol-4.md there is a link to "Interledger Architecture overview" with the URL "../0001-interledger-architecture/0001-interledger-architecture.md#interledger-protocol-suite".

On [that page on the website](https://interledger.org/rfcs/0027-interledger-protocol-4/), this gives a 404 because it points directly to an md file.

It seems that when a link is `../some-page/some-page.md` then the `some-page.md` is removed and the link works on the website, BUT, if it's `../some-page/some-page.md#some-anchor` - then the `.md` section isn't removed and results in a 404.

adrianhopebailie:
Fixes #452


davidgilbertson:
LGTM

Issue #445, opened 2018-08-01
Mismatch between RFC and implementation, BTP


dora-gt:
In the [BTP RFC](https://github.com/interledger/rfcs/blob/master/0023-bilateral-transfer-protocol/0023-bilateral-transfer-protocol.md#overview), the structure of BTP packet is explained as the below.

1. Type ID
2. Request ID
3. Packet Specific Data
4. Sub-Protocol Count
5. Sub-Protocol Data

While the implementation is like the below, not having `packet specific data`. `writeProtocolData` just writes out 1) sub-protocol count, 2) sub-protocol data.
[btp-packet/index.js#L125](https://github.com/interledgerjs/btp-packet/blob/master/index.js#L125)

We need to make it clear which is correct, or, if it is correct, I think RFC should be updated to avoid leading to misunderstand.

adrianhopebailie:
I would recommend that the RFC be corrected to reflect reality if the implementation is different.

@sharafian can you provide some insight into which BTP packet types and sub-protocols are now widely used so we can document these. We should avoid documenting stale code that is no longer used.

sharafian:
Currently, the plugin BTP is only making use of `MESSAGE` and `RESPONSE` packets. `MESSAGE` and `RESPONSE` have no packet specific data, which is why it goes straight to sub-protocol count.

dora-gt:
I see. So, maybe RFC is wrong (...or at least misleading). I checked [ASN.1 definition](https://github.com/interledger/rfcs/blob/master/asn1/BilateralTransferProtocol.asn) it reflects what you told. So I suggest amending RFC according to the ASN.1. I'll try.

adrianhopebailie:
@earizon the protocols operate at different layers of the stack. An ILP error should not be carried in a BTP error.

earizon:
@adrianhopebailie It's weird because the BTP errors can be mostly mapped to ILPv4 errors.

T00 UnreachebleError can be mapped to T01 Peer Unreachable
F01 InvalidFiledsError can be mapped to F00 Bad Request
F03 TransferNotFoundError can be mapped to F06 Unexpected Payment
F08 InsufficientBalanceError can be mapped to F04 Insufficient Liquidity

F00 NotAcceptedError, F04InvalidFulfillmentError, F06 AlreadyRolledBackError, F07AlreadyFulfilledError have no equivalent in 0027 ILPv4 but can be added to reuse in BTP and non-BTP protocols (ILP-over-HTTP).

[...]

PR #470, opened 2018-08-27
Amended BTP RFC according to the implementation

dora-gt:
Based on this discussion #445, amended BTP RFC according to the implementation.


dora-gt:
Thank you @sharafian for the review.

PR #472, opened 2018-08-28
Fix headings on RFC 32

adrianhopebailie:


0032-peering-clearing-settlement/0032-peering-clearing-settlement.md
```diff
@@ -1,13 +1,15 @@
 ---
 title: Peering, Clearing and Settling on the Interledger network
-draft: 1
+draft: 2
 ---
 
+# Peering, Clearing and Settling
+
```
dora-gt: nit: it differs from the above title

PR #489, opened 2018-10-30
Assign RFC number

adrianhopebailie:


0034-connector-requirements/0034-connector-requirements.md
```diff
@@ -0,0 +1,150 @@
...
+
+The specific schedule and mechansim for doing this will be specific to the settlement system used by the peers (e.g. payment channels on a distributed ledger, traditional wire transfers via the banking system, etc.), therefor the functionality required to do this is not included in the core node but rather in settlement-system-specific plugins or adaptors.
+
+It is necessary for the node to have a view of the current outstanding obligations with the peer (the peer's account balance) in order to apply appropriate risk management measures when processing packets from the peer.
+
+The logic that determines when to perform a settlement is currently implemented within the node (as opposed to in the plugin) and as such the node must instruct the plugin when to perform a settlement. This reduces the outstanding obligations with the peer to a limit that the node considers safe such that it can forward further packets for the peer.
```
dora-gt: It seems that the entity that exactly determines when to settle (when to execute on-ledger transaction) is the plug-in (`_claimFunds` is not called externally). Is that ok?

adrianhopebailie: In this context `_claimFunds` is not settlement.

The sendMoney from the connector is what triggers settlement which is just the exchange of claims.

Issue #298, opened 2017-09-14
Support bulk-fulfill?


michielbdejong:
Something I thought of when we were talking about streaming payments yesterday: if Alice wants to pay Bob, and they have real-time two-way out-of-band communication open, Bob can "loop back" the conditions of all incoming "payment chunks" to Alice. She can check whether all of them arrived, and then atomically share all the fulfillments with Bob. This would be a transport protocol, like PSK.

Situation in which Chunk-Loop Fulfillment (CLF) payments are useful:
* Alice wants to send money to Bob, because, for instance, Bob has sent Alice an invoice.
* She wants one of two things to happen within a hard timeout limit: total-success, or total-failure
* In the case of total-failure, she wants *all* her money back; it's ok to put money on hold until the timeout limit, but after that, all money that was on hold should be back in her source account.
* In the case of total-success, she wants to know that Bob was able to claim the full destination amount
* Whether the destination ledger actually increased Bob's balance or not with the full amount doesn't matter, it's enough if Alice knows that Bob was in the possession of the necessary fulfillment(s) at the right time.
* Alice has a maximum source amount she is willing to pay to achieve total-success, but no minimum. It's ok if she doesn't get charged the full amount.
* Alice and Bob have two-way out-of-band communication
* Alice and Bob are unable or unwilling to use PSK, for instance because there is no single path through the network that has enough liquidity to complete the whole payment at the full amount.

Rough description of the CLF procedure:
* Alice decides on a number of chunks to use, `n`
* Alice generates two random numbers: a CLF `paymentId` and a CLF `secret`.
* Alice sends `n` ILP payments to Bob, putting `'CLF:' + i + '/' + n + ':' + paymentId` in the packet data of each chunk `i`, as well as some information about what she is paying for, for instance the ID of the invoice she is paying for.
* Alice creates the condition of each chunk payment by taking the HMAC of the full ILP packet with the CLF `secret`, and computing the SHA-256 hash of it.
* Alice prepares the `n` chunks.
* If (and only if) Bob sees all `n` out of `n` prepared payments coming in, he looks at the invoice ID or other information mentioned in the packets, calculates the total destination amount of all chunks together, and decides if he wants to accept this CLP payment from Alice.
* If Bob decides to accept the payment, he sends an out-of-band 'accept' message to Alice, mentioning the CLF `paymentId`, and asking her for the CLF `secret`.
[...]

PR #493, opened 2018-11-08
Add connector risk information

khoslaventures:
At worst, a connector can withhold a payment to its counterparty in
the fulfill phase but get paid by another connector. This change adds some details related to this.

### DIFF
--- a/0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
+++ b/0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
@@ -1,12 +1,10 @@
 ---
 title: Connector Risk Mitigations
-draft: 1
+draft: 3
 ---
 # Connector Risk Mitigations
 
-Interledger connectors take some risk in exchange for the revenue they generate from facilitating payments. This document outlines the major categories of risks connectors face and suggests some possible mitigations.
-
-This is a work in progress and is not an exhaustive list.
+Interledger connectors take some risk in exchange for the revenue they generate from facilitating payments. There are also scenarios in which connectors will facilitate payments without taking margin and need to account for greater risk. This document outlines the major categories of risks connectors face and suggests some possible mitigations. This is a work in progress and is not an exhaustive list.
 
 **Monitoring is a must! Even connectors that implement all of these strategies should monitor their transaction patterns and use warnings or kill switches to avoid losing money in the case of an unexpected attack.**
 
@@ -16,29 +14,68 @@ The main risk connectors face in Interledger is being unable to fulfill the inco
 
 Possible mitigations include:
 
-* **Connecting Reliable Ledgers** - Connectors should choose the ledgers they connect in part based on the reliability and speed of fulfillment notifications and the time the ledgers require to process fulfillments once they are submitted
-* **Adjusting Transfer Expiry Window** - Connectors determine the window of time they require between when the incoming and outgoing transfers expire. Connectors should set this window such that they are confident that they will be able to deliver the fulfillment in time even if the outgoing transfer is executed at the last possible moment.
-* **Denial of Service Protections** - Connectors should take standard [Denial of Service](https://en.bitcoin.it/wiki/Weaknesses#Denial_of_Service_.28DoS.29_attacks) attack prevention steps to prevent attackers from overloading the connectors' servers with network traffic
-* **Prioritizing Fulfillments** - Connectors should prioritize fulfilling transfers over preparing new ones or responding to quote requests and may even have separate processes or machines responsible for those different behaviors
-* **Redundant Instances** - Connectors should run redundant processes and machines to increase the difficulty of interfering with their processing
-* **Avoiding Public Endpoints** - Connectors may reduce attack vectors by using communication channels and protocols that do not entail having a publicly-accessible server (for example using [WebRTC](https://webrtc.org/) or [Virtual Private Networks](https://en.wikipedia.org/wiki/Virtual_private_network) to communicate with peers
+* **Adjusting Transfer Expiry Window** - Connectors determine the window of expiration for a transfer (within a full prepare-fulfill payment cycle) during the prepare phase. The expiration is dependent on the amount of time it takes for a connector to receive a fulfill from from their bilateral party. Forwarding connectors should set the window such that they will be able to receive the fulfillment and deliver it to the previous connector in time even if the outgoing transfer is executed at the last possible moment.
+* **Denial of Service Protections** - Connectors should take standard [Denial of Service](https://en.bitcoin.it/wiki/Weaknesses#Denial_of_Service_.28DoS.29_attacks) attack prevention steps to prevent attackers from overloading the connectors' servers with network traffic. This includes IP-level protections and ILP-level protections.
+* **Prioritizing Fulfillments** - Connectors should prioritize fulfilling transfers over preparing new ones and may even have separate processes or machines responsible for those different behaviors.
+* **Redundant Instances** - Connectors should run redundant processes and machines to increase the difficulty of interfering with their processing.
+* **Avoiding Public Endpoints** - Connectors may reduce attack vectors by using communication channels and protocols that do not entail having a publicly-accessible server (for example using [WebRTC](https://webrtc.org/) or [Virtual Private Networks](https://en.wikipedia.org/wiki/Virtual_private_network) to communicate with peers.
 
-## Liquidity Exhaustion
+Note that unconditional payment channels do not require longer timeouts while enabling peers to rebalance frequently and minimize their bilateral trust.
 
-Attackers could tie up connectors' liquidity by preparing payments through them that the attacker knows will fail. For example, an attacker could prepare numerous payments to itself and then not fulfill any of them.
+## Payments Griefing
+
+Liquidity Exhaustion: Attackers could tie up connectors' liquidity by preparing payments through them that the attacker knows will fail. For example, an attacker could prepare numerous payments to itself and then not fulfill any of them.
 
 Possible mitigations include:
 
 * **Managing "Payment Bandwidth"** - Connectors should monitor the percentage of their total liquidity that is on hold at any given time for a given peer, customer, or destination prefix and may reject incoming payment requests if that party exceeds their allocated "payment bandwidth". Connectors may allocate less bandwidth to unknown or untrustworthy senders or receivers. Note that ILP Packets contain the destination account but not the source account, so connectors should apply this logic to the immediate peer or customer that they receive the payments from and possibly the destination account or prefix.
-* **Preferring Smaller Payments** - Smaller payments place a smaller percentage of the connector's liquidity on hold and thus each one carries less risk that the payment being prepared and then failing would create significant opportunity cost
-* **Blacklisting Senders and Receivers** - Connectors may refuse to facilitate payments from certain sources or to certain destinations if their rate of failed payments is unusually high
+* **Preferring Smaller Payments** - Smaller payments place a smaller percentage of the connector's liquidity on hold and thus each one carries less risk that the payment being prepared and then failing would create significant opportunity cost - in ILPv4, this is the default.
+* **Blacklisting Senders and Receivers** - Connectors may refuse to facilitate payments from certain sources or to certain destinations if their rate of failed payments is unusually high.
 
-## Exchange Rate Volatility
+Counterparty Theft: Since connectors forward packets on credit, they take a risk that the counterparty doesn't or can't settle later.
 
-Once connectors prepare their outgoing transfers, they are committed to the payment even if the exchange rate between the assets fluctuates.
+Example:
+Suppose you have connectors A, B, C and D, and the payment is going from A->B->C->D. Take the case of C. During the fulfill phase, D will pass a fulfill packet to C, meaning that C will have to pay. It's important to note that the balances have been updated, but settlement can occur at a later time if dealing with a trusted counterparty. In the best case scenario, we settle in payment channels on every balance update. But because ILP is using unconditional payment channels for settlements, C can write a malicious client that could forward a fulfill packet to B, and avoid paying D by withholding settlement. In this case, A would pay B, B would pay C, but C didn't pay D.
 
-Possible approaches include:
+Possible mitigations include:
+* **Limiting total counterparty risk** - When forwarding packets for a peer,
+    limiting the total value of unsettled payments is by leveraging pre-emptive
+    payment channel updates after the fulfill phase limits the amount a
+    counterparty could steal. If the settlement receiving in a payment channel
+    is less than that of an on chain-fee, a connector that cheats its counterparty would lose much
+    more capital in the process. For example, a connector that has been deprived of
+    payment from its counterparty could refuse to facilitate future payments,
+    and the malicious counterparty would have to close their payment channel to
+    make the stolen micropayment amount liquid. Given that on-chain transaction fees
+    are much higher than that of a micropayment in Interledger (i.e. 1 satoshi), an
+    attacker would lose more money opening and closing channels to cheat their
+    counterparty. Therefore, while this griefing attack is possible, if
+    connectors are leveraging smaller payments and are requiring settlement down
+    to the unit of a packet (ideally less than a on-chain fee),
+    the attacking connector is more incentivized to facilitate payments to make
+    profit on the exchange rate/fees.
+* **Use an external arrangement** - Previously, we've mentioned that
+    connectors are more likely to cooperate so they can earn margin on packet
+    transfers. However, there are some scenarios in which they are not seeking a margin
+    (i.e. a connector could have a model for charging its customers a flat rate for
+    a certain amount of bandwidth at 0% margins). In these scenarios, it makes more
+    sense for the connectors involved to leverage legal agreements (similar to what
+    ISPs do for forwarding BGP packets) among connectors that are on-path and by KYCing
+    any future counterparties that may want to forward packets to them.
 
-* **Accounting for Slippage** - Connectors may add a buffer to their expected exchange rate to account for some movement in the price, and they may add a premium for especially volatile currencies
-* **Preferring Shorter Timeouts** - Connectors may provide better exchange rates for payments with shorter timeouts because they carry less risk that the price will move drastically while the payment is in flight
+## Solved Issues
 
+These were risks that were present in older versions of the protocol and are
+no longer the case.
+
+### Exchange Rate Volatility
+
+Once connectors prepare their outgoing transfers, they are committed to the payment even if the exchange rate between the assets fluctuates. The support of low-value packets allows for connectors to do exchange rates on a very fine-grained basis. Additionally, by having shorter timeouts for the prepare packets connectors are limiting their risk of being taken advantage from by the free option problem.
+
+In the free option problem, a connector during the fulfill phase can receive a fulfill packet but decide whether or not proceed with payment when receiving the packet based on the current exchange rate of whole market. In the optimal case (for the payer), connectors receiving the fulfill packet can hold their payment until just before the expiry of the packet and then complete only if the exchange rate moved in the sender's favor.
+
+### Unreliable Ledgers
+
+In the Lightning Network and older versions of ILP, HTLCs (Hash-Timelock Contracts) are used for multi-hop payment channels. The pre-image condition that is agreed upon between sender and receiver is enforceable on chain, and allows for an interoperability method known as an atomic swap.
+
+Unfortunately, HTLCs require timing assumptions as atomically locking up funds and settling on-chain involves locking up funds for time from when a transaction is sent until block finality. This means that if we have a payment from A->B->C, when C claims the money from B, the protocol must prevent race conditions in which A and B can pull out funds if C is not responsive. An example of a race condition would be where C claims the funds from B, but A tries to close its channel with B before B is able to claim funds from A. In order to get around this, atomic swaps between payment channels of connectors are secured under a synchrony assumption, in which the time it takes to submit a transaction to the main chain in case of unresponsive counterparty is heavily considered. We've previously suggested connectors choose the ledgers they plug-in based on the reliability and speed of fulfillment notifications and the time the ledgers require to process fulfillments once they are submitted, but this is no longer needed for risk mitigation because of simplified payment structure.


0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
```diff
@@ -17,22 +17,25 @@ The main risk connectors face in Interledger is being unable to fulfill the inco
 Possible mitigations include:
 
 * **Connecting Reliable Ledgers** - Connectors should choose the ledgers they connect in part based on the reliability and speed of fulfillment notifications and the time the ledgers require to process fulfillments once they are submitted
-* **Adjusting Transfer Expiry Window** - Connectors determine the window of time they require between when the incoming and outgoing transfers expire. Connectors should set this window such that they are confident that they will be able to deliver the fulfillment in time even if the outgoing transfer is executed at the last possible moment.
+* **Adjusting Transfer Expiry Window** - Connectors determine the window of time they require between when the incoming and outgoing transfers expire. Connectors should set this window such that they are confident that they will be able to deliver the fulfillment in time even if the outgoing transfer is executed at the last possible moment. TODO: Not sure if still relevant because we are not using HTLCs
```
khoslaventures: Is this still relevant? We're not using HTLA/Cs right, so I think it's not

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
```diff
@@ -17,22 +17,25 @@ The main risk connectors face in Interledger is being unable to fulfill the inco
...
+
+In the worst case, during the fulfill phase a connector along the path fulfilling the payment could withhold the fulfill packet such that their corresponding connector would not receive a payment. The corresponding connector would at worst lose the unit amount being streamed in a single packet (i.e. 1 satoshi). The fulfilling connector would gain the unit amount and would put their reputation as a connector on the line and would miss out on future fee gains. Additionally, the fulfilling connector would have to close the payment channel between them and the corresponding connector, as the corresponding connector would choose not to facilitate payments through the channel after being cheated. This would result in an on-chain transaction which would make the economics for the attacker not worth it, as they could be making more by continuing to facilitate fees.
```
khoslaventures: I think this should be in a different section, but not sure where. Let me know what works best!

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
khoslaventures: @emschwartz

khoslaventures:
@adrianhopebailie @emschwartz mind reviewing for me? I think this is ready!

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
adrianhopebailie: `Connecting Reliable Ledgers` is not relevant anymore.

However, I'd say `Adjusting Transfer Expiry Window` is. While this is no longer enforced by the ledger, it is still part of the protocol rules. A connector does still need to adjust the expiry in outgoing prepare packets to give themselves time to process and pass on the fulfill.

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
```diff
@@ -17,25 +17,28 @@ The main risk connectors face in Interledger is being unable to fulfill the inco
 Possible mitigations include:
 
 * **Connecting Reliable Ledgers** - Connectors should choose the ledgers they connect in part based on the reliability and speed of fulfillment notifications and the time the ledgers require to process fulfillments once they are submitted
```
adrianhopebailie: This is no longer relevant and can be removed.

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
```diff
@@ -17,25 +17,28 @@ The main risk connectors face in Interledger is being unable to fulfill the inco
...
-* **Adjusting Transfer Expiry Window** - Connectors determine the window of time they require between when the incoming and outgoing transfers expire. Connectors should set this window such that they are confident that they will be able to deliver the fulfillment in time even if the outgoing transfer is executed at the last possible moment.
```
adrianhopebailie: This is still relevant. 

How two connectors reconcile and/or enforce the settlement of their Interledger accounts is outside the scope of the protocol but the protocol does still expect them to respect the expiry times and deliver fulfillments before the corresponding prepare expires.

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
```diff
@@ -17,25 +17,28 @@ The main risk connectors face in Interledger is being unable to fulfill the inco
...
 * **Denial of Service Protections** - Connectors should take standard [Denial of Service](https://en.bitcoin.it/wiki/Weaknesses#Denial_of_Service_.28DoS.29_attacks) attack prevention steps to prevent attackers from overloading the connectors' servers with network traffic
 * **Prioritizing Fulfillments** - Connectors should prioritize fulfilling transfers over preparing new ones or responding to quote requests and may even have separate processes or machines responsible for those different behaviors
```
adrianhopebailie: `responding to quote requests` - this is no longer part of the protocol and can be removed

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
```diff
@@ -17,25 +17,28 @@ The main risk connectors face in Interledger is being unable to fulfill the inco
...
 * **Redundant Instances** - Connectors should run redundant processes and machines to increase the difficulty of interfering with their processing
 * **Avoiding Public Endpoints** - Connectors may reduce attack vectors by using communication channels and protocols that do not entail having a publicly-accessible server (for example using [WebRTC](https://webrtc.org/) or [Virtual Private Networks](https://en.wikipedia.org/wiki/Virtual_private_network) to communicate with peers
 
+Note that newer unconditional payment channels in ILPv4 do not require longer timeouts while enabling peers to rebalance frequently and minimize their bilateral trust.
```
adrianhopebailie: I think we could expand on this note to explain how ILPv4 changed the whole model.

The ledger is no longer in the flow so the nature of the protocol is different. In the past, ILP worked best over ledgers that could prepare and fulfill transfers quickly as this time was additive along the path of a payment.

With ILPv4 we prefer ledgers that can simple complete a transfer quickly and inexpensively so that we can do frequent bilateral settlements and keep the settlement risk between peers low.

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
```diff
@@ -17,25 +17,28 @@ The main risk connectors face in Interledger is being unable to fulfill the inco
...
 * **Blacklisting Senders and Receivers** - Connectors may refuse to facilitate payments from certain sources or to certain destinations if their rate of failed payments is unusually high
 
+In the worst case, during the fulfill phase a connector along the path fulfilling the payment could withhold the fulfill packet such that their corresponding connector would not receive a payment. The corresponding connector would at worst lose the unit amount being streamed in a single packet (i.e. 1 satoshi). The fulfilling connector would gain the unit amount and would put their reputation as a connector on the line and would miss out on future fee gains. Additionally, the fulfilling connector would have to close the payment channel between them and the corresponding connector, as the corresponding connector would choose not to facilitate payments through the channel after being cheated. This would result in an on-chain transaction which would make the economics for the attacker not worth it, as they could be making more by continuing to facilitate fees.
```
adrianhopebailie: > The fulfilling connector would gain the unit amount

This is not true. By withholding the fulfillment the upstream connector is denying themselves a payment (and all connectors downstream).

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
karzak: Maybe clarify that Transfer Expiry Window is no longer dependent on execution of on-ledger transaction, just on the amount of time it takes to receive a fulfill from their bilateral party.

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
```diff
@@ -1,10 +1,10 @@
 ---
 title: Connector Risk Mitigations
-draft: 1
+draft: 2
 ---
 # Connector Risk Mitigations
 
-Interledger connectors take some risk in exchange for the revenue they generate from facilitating payments. This document outlines the major categories of risks connectors face and suggests some possible mitigations.
+Interledger connectors take some risk in exchange for the revenue they generate from facilitating payments. This document outlines the major categories of risks connectors face and suggests some possible mitigations. Since the introduction of ILPv4, some mitigations are no longer needed.
```
emschwartz: The ones that aren't needed should probably just be taken out. Having them in there is just confusing

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
```diff
@@ -17,25 +17,28 @@ The main risk connectors face in Interledger is being unable to fulfill the inco
 Possible mitigations include:
 
 * **Connecting Reliable Ledgers** - Connectors should choose the ledgers they connect in part based on the reliability and speed of fulfillment notifications and the time the ledgers require to process fulfillments once they are submitted
-* **Adjusting Transfer Expiry Window** - Connectors determine the window of time they require between when the incoming and outgoing transfers expire. Connectors should set this window such that they are confident that they will be able to deliver the fulfillment in time even if the outgoing transfer is executed at the last possible moment.
 * **Denial of Service Protections** - Connectors should take standard [Denial of Service](https://en.bitcoin.it/wiki/Weaknesses#Denial_of_Service_.28DoS.29_attacks) attack prevention steps to prevent attackers from overloading the connectors' servers with network traffic
```
emschwartz: It may be worth mentioning that this is both IP-level DoS protections and within-ILP protections

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
emschwartz: Very few people knew what ILP was when we were talking about ILPv1 so it feels like it might be better to take out everything that's v1-related and just focus on v4.

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
```diff
@@ -17,25 +17,28 @@ The main risk connectors face in Interledger is being unable to fulfill the inco
...
 * **Avoiding Public Endpoints** - Connectors may reduce attack vectors by using communication channels and protocols that do not entail having a publicly-accessible server (for example using [WebRTC](https://webrtc.org/) or [Virtual Private Networks](https://en.wikipedia.org/wiki/Virtual_private_network) to communicate with peers
 
+Note that newer unconditional payment channels in ILPv4 do not require longer timeouts while enabling peers to rebalance frequently and minimize their bilateral trust.
+
 ## Liquidity Exhaustion
 
-Attackers could tie up connectors' liquidity by preparing payments through them that the attacker knows will fail. For example, an attacker could prepare numerous payments to itself and then not fulfill any of them.
+Attackers could tie up connectors' liquidity by preparing payments through them that the attacker knows will fail. For example, an attacker could prepare numerous payments to itself and then not fulfill any of them. In the case of ILPv4, this would require a significant amount of payments to cause damage because of the nature of streaming micropayments.
```
emschwartz: > this would require a significant amount of payments to cause damage because of the nature of streaming micropayments

That sounds like a mitigation but it's not really, because it's not too difficult for an attacker to send many packets.

The real mitigation here is that with ILPv4, we adjusted the senders' expectations so that they can function even if the payment bandwidth / liquidity they are allocated is very low. This changed expectation allows the connectors to split their liquidity among their customers such that one client can't interfere with another's connection. (If you're doing large atomic or atomic multi-path payments, you can't limit the amount of liquidity you're allocating to each customer as easily)

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
```diff
@@ -17,25 +17,28 @@ The main risk connectors face in Interledger is being unable to fulfill the inco
...
+
 ## Exchange Rate Volatility
 
-Once connectors prepare their outgoing transfers, they are committed to the payment even if the exchange rate between the assets fluctuates.
+Once connectors prepare their outgoing transfers, they are committed to the payment even if the exchange rate between the assets fluctuates. In ILPv4, the support of low-value packets allows for connectors to do exchange rates on a per-packet basis. By having shorter timeouts for the prepare packets, senders have stronger guarantees for quick resolution of their packets and connectors are limiting their risk of being taken advantage from by the "free option problem" standpoint (ensure that packets are held until just before their expiry and then fulfilled only if the exchange rate moved in the sender's favor).
```
emschwartz: I'm not sure this section is super relevant anymore. If you have packets with very short timeouts (and relatively low amounts of money), this kind of becomes a non-issue for the connector

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
khoslaventures: What about in the case of a connector getting a fulfillment from a previous connector but not sending a fulfill to the next connector?

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
khoslaventures: @emschwartz I wonder if it's better to take things out or leave them as non-issues somewhere in the RFCs. Some people probably have questions as to whether ILP is susceptible to such things based on looking through other protocols and I think having answers to those type of questions is good for legitimacy

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
khoslaventures: Also, it seems like an attacker can't really tie up liquidity like they can in HTLCs - unless they are sending multiple prepares from multiple open channels (pretty expensive to do) and the packet amounts are small (the default), technically you can aggressively ban senders that don't complete their prepare-fulfill cycles? Unless I'm missing something.

khoslaventures: > If you're doing large atomic or atomic multi-path payments

Also, what are "atomic" payments in Interledger? Are these conditional payment channels?

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
emschwartz: Let's say you have connectors A, B, and C and the payment is going from A -> B -> C. Take the case of B. C passes the fulfillment to B, meaning that B now needs to pay. If B doesn't pass the fulfillment on to A, it's _B_ that looses the money. A doesn't care, because A hasn't paid anything yet.

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
emschwartz: > Is this the default or does a connector have to specify it?

Connectors specify the bandwidth they want to allocate to each user (in the `ilp-connector` implementation this is governed by the throughput middleware). Right now on the network the bandwidth is low so it's reasonable for the sender to assume they have pretty low bandwidth.

> technically you can aggressively ban senders that don't complete their prepare-fulfill cycles?

We considered that earlier but decided that isn't the way to go. A) it isn't necessarily the sender's fault if packets are rejected b) rejecting packets may sometimes be useful for application layer protocols c) an attacker could run the same attack where they send money to themselves and do actually fulfill it, but after a delay.

> Also, it seems like an attacker can't really tie up liquidity like they can in HTLCs

If you're not using HTLCs it may be a little more flexible but not completely. Let's say I'm your upstream provider, so you send some big chunk of your packets through me. There's some limit to how much I'll let you send at once (that's bound by our credit limit). If one of your users prepares a ton of packets that go through our connection, they might be able to use up all of your upstream bandwidth, which would block other users from sending.

> Also, what are "atomic" payments in Interledger? Are these conditional payment channels?

There aren't any. I was referring to Lightning's Atomic Multipath Payment (AMP) protocol. We went down the [bulk fulfill](https://github.com/interledger/rfcs/issues/298) path earlier and decided not to implement it.

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
emschwartz: Sure. This description sounds off to me.

> In ILPv4, the support of low-value packets allows for connectors to do exchange rates on a per-packet basis.

You can always set rates on a per-packet basis, it's just that the packet amounts are lower now and more importantly, the time you're holding it for is shorter.

> senders have stronger guarantees for quick resolution of their packets

This feels out of place in a doc on connector risk mitigation

> "free option problem" standpoint (ensure that packets are held until just before their expiry and then fulfilled only if the exchange rate moved in the sender's favor).

Trying to introduce this problem and say that it's solved in one sentence is a bit rushed. If you want to include it, it would be better to have an explanation that's a little longer and clearer (or link to another description of that issue).

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
khoslaventures: >  C passes the fulfillment to B, meaning that B now needs to pay. 
I need some clarity on this. So fulfillments don't imply payments which is why my description is incorrect?

Ok, let's say B passes the fulfillment on to A. But B hasn't paid C. Does A pay B? Does A know that C hasn't been paid and act accordingly? (and would that apply if we had another node, D as the receiver, and C as a connector)?

I just need to change this description to talk about not paying rather than withholding a fulfillment right?

emschwartz: Each settlement agreement is handled on a bilateral basis. Once B has gotten the fulfillment from C, B owes C the money. If B screws C, that's C's problem. If B gives the fulfillment to A, B has met their side of the agreement with A and A owes B the money.

The not paying problem is simply the counterparty risk that is limited by the total value of the Interledger packets you're willing to forward for your counterparty before they send you the corresponding on-ledger money (or payment channel claims).

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
austin-king: Given the path sender -> connector A -> connector B -> receiver if connector B fails to pass along a fulfill to connector A in time, then only connector B will lose money because connector A will not update their balance with connector B (A paying B).  Therefore, a connector has nothing to gain by not passing along a fulfill because when they pass along a fulfill they are essentially getting paid back.

austin-king: Ahh, ignore this.  Didn't see thread below.

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
khoslaventures: > senders have stronger guarantees for quick resolution of their packets
>> This feels out of place in a doc on connector risk mitigation

Agreed, I think this was because of my misunderstanding about payments above.

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
```diff
@@ -16,29 +16,53 @@ The main risk connectors face in Interledger is being unable to fulfill the inco
...
+    more capital in the process. For example, a connector that has been deprived of
+    payment from its counterparty could refuse to facilitate future payments,
+    and the malicious counterparty would have to close their payment channel to
+    make the stolen micropayment amount liquid. Given that on-chain transaction fees
+    are much higher than that of a micropayment in Interledger (i.e. 1 satoshi), an
+    attacker would lose more money open and closing channels to cheat their
+    counterparty. Therefore, while this griefiing attack is possible, if
+    connectors are leveraging smaller payments, the attacker is more incentivized
+    to facilitate payments to make profit on the exchange rate/fees.
+* **Leverage payment channels with fraud proofs** - (TODO: Wondering if this is possible?)
```
khoslaventures: Wondering if there was any insights into leveraging payment channels with fraud proofs to punish griefers/prevent them from continuing. i.e. if someone were to be slashed on the basis of this, that would be pretty cool.

This would basically mean that ILP's construction would start to bleed more into payment channels and ledgers would have to support this kind of proof (likely only Ethereum because of turing complete smart contract language).

khoslaventures:
also @emschwartz @adrianhopebailie ready for another review!

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
austin-king: One perspective I would change throughout this document is that connectors are incentivized to continue participating in the network because they will take a margin on packets that pass through.  That is only one of the possible reasons for connectors to continue participating.  

You can imagine a relationship where a connector has a business model similar to an ISP where they charge their customers a flat rate for a certain amount of bandwidth and 0% margins.  In this scenario connectors are not incentivized to stay connected to the greater network to earn some percentage profit, but instead to provide a good customer experience where all of their customers can communicate value to any other entity in the greater network.

khoslaventures: Ah I'm seeing what you're getting at now. I misread what you initially said.
I just pushed something that should add some details regarding this.

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
```diff
@@ -16,29 +14,70 @@ The main risk connectors face in Interledger is being unable to fulfill the inco
...
+    have integrations for this, but it would be possible to construct ILP-aware
+    payment channel implementations such that if a counterparty refused to pay
+    its next hop despite receiving payment, then the payment channel
+    implementation could have a proof system in which the connector that lost
+    money could submit a dispute on ledger such that the payment channel would
+    close with the stolen payment recovered in the final balance. Such a proof
+    would require proofs of payment from the other parties along the path and
+    proof of non-payment between the stealing counterparty and the reporting
+    connector.
+* **Use a restricted connector setup** - Previously, we've mentioned that
```
khoslaventures: @austin-king look alright?

austin-king: LGTM

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
```diff
@@ -16,29 +14,70 @@ The main risk connectors face in Interledger is being unable to fulfill the inco
...
+    more capital in the process. For example, a connector that has been deprived of
+    payment from its counterparty could refuse to facilitate future payments,
+    and the malicious counterparty would have to close their payment channel to
+    make the stolen micropayment amount liquid. Given that on-chain transaction fees
+    are much higher than that of a micropayment in Interledger (i.e. 1 satoshi), an
+    attacker would lose more money open and closing channels to cheat their
+    counterparty. Therefore, while this griefing attack is possible, if
+    connectors are leveraging smaller payments, the attacking connector is more incentivized
+    to facilitate payments to make profit on the exchange rate/fees.
+* **Provide fraud proof support on the ledger** - At the moment, ILP does not
```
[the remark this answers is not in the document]
khoslaventures: > I think fraud proofs, as they are mentioned here, are out of scope for this RFC, but it could lead to an interesting new proposal if you have a good idea for how it could be constructed.

Yea I'll move it into another proposal. I think this may only work in certain cases and I don't expect most chains to provide endpoints to do it. I'll write up a proposal and publish if it makes sense. Thanks!

khoslaventures:
@emschwartz @adrianhopebailie Can you give this another review?

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
```diff
@@ -16,29 +14,60 @@ The main risk connectors face in Interledger is being unable to fulfill the inco
...
+* **Prioritizing Fulfillments** - Connectors should prioritize fulfilling transfers over preparing new ones and may even have separate processes or machines responsible for those different behaviors.
+* **Redundant Instances** - Connectors should run redundant processes and machines to increase the difficulty of interfering with their processing.
+* **Avoiding Public Endpoints** - Connectors may reduce attack vectors by using communication channels and protocols that do not entail having a publicly-accessible server (for example using [WebRTC](https://webrtc.org/) or [Virtual Private Networks](https://en.wikipedia.org/wiki/Virtual_private_network) to communicate with peers.
+
+Note that unconditional payment channels do not require longer timeouts while enabling peers to rebalance frequently and minimize their bilateral trust.
 
-## Liquidity Exhaustion
+## Payments Griefing
```
adrianhopebailie: Spelling mistake?

khoslaventures: haha, Griefing is I suppose urban dictionary for games. We can use Grieving.

khoslaventures:
> > Counterparty Theft: Suppose you have connectors A, B, C and D, and the payment is going from A->B->C->D. Take the case of C. During the fulfill phase, D will pass a fulfill packet to C, meaning that C will have to pay. But because ILP is using unconditional payment channels, C can write a malicious client that could forward a fulfill packet to B, and avoid paying D. In this case, A would pay B, B would pay C, but C didn't pay D.
> 
> This conflates settlement with the updating of balances. D will only pass a fulfill back to C if they have updated the balance between the two of them, which could cause a settlement event but it does not necessitate it.
> 
I think my explanation doesn't talk about the separation of the two but assumes the best case scenario of settlement on every balance update. Will see if I can add details to explain the separation of the two, but want to keep the idea as simple as possible. A good goal for the RFCs would be to explain this as part of the core concepts.

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
```diff
@@ -16,29 +14,60 @@ The main risk connectors face in Interledger is being unable to fulfill the inco
...
-* **Prioritizing Fulfillments** - Connectors should prioritize fulfilling transfers over preparing new ones or responding to quote requests and may even have separate processes or machines responsible for those different behaviors
-* **Redundant Instances** - Connectors should run redundant processes and machines to increase the difficulty of interfering with their processing
-* **Avoiding Public Endpoints** - Connectors may reduce attack vectors by using communication channels and protocols that do not entail having a publicly-accessible server (for example using [WebRTC](https://webrtc.org/) or [Virtual Private Networks](https://en.wikipedia.org/wiki/Virtual_private_network) to communicate with peers
+* **Adjusting Transfer Expiry Window** - Connectors determine the window of expiration for a transfer (within a full prepare-fulfill payment cycle) during the prepare phase. The expiration is dependent on the amount of time it takes for a connector to receive a fulfill from from their bilateral party. Forwarding connectors should set the window such that they will be able to receive the fulfillment in time even if the outgoing transfer is executed at the last possible moment.
```
emschwartz:
```suggestion
* **Adjusting Transfer Expiry Window** - Connectors determine the window of expiration for a transfer (within a full prepare-fulfill payment cycle) during the prepare phase. The expiration is dependent on the amount of time it takes for a connector to receive a fulfill from from their bilateral party. Forwarding connectors should set the window such that they will be able to receive the fulfillment and deliver it to the previous connector in time even if the outgoing transfer is executed at the last possible moment.
```

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
```diff
@@ -16,29 +14,60 @@ The main risk connectors face in Interledger is being unable to fulfill the inco
...
+* **Prioritizing Fulfillments** - Connectors should prioritize fulfilling transfers over preparing new ones and may even have separate processes or machines responsible for those different behaviors.
+* **Redundant Instances** - Connectors should run redundant processes and machines to increase the difficulty of interfering with their processing.
+* **Avoiding Public Endpoints** - Connectors may reduce attack vectors by using communication channels and protocols that do not entail having a publicly-accessible server (for example using [WebRTC](https://webrtc.org/) or [Virtual Private Networks](https://en.wikipedia.org/wiki/Virtual_private_network) to communicate with peers.
+
+Note that unconditional payment channels do not require longer timeouts while enabling peers to rebalance frequently and minimize their bilateral trust.
 
-## Liquidity Exhaustion
+## Payments Grieving
```
emschwartz:
```suggestion
## Liquidity Exhaustion / Payment Griefing
```

The aim of the attack is to exhaust your liquidity, so I think we should call it that. For comparison, a "Denial of Service" is what happens to you, not how it's done

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
```diff
@@ -16,29 +14,60 @@ The main risk connectors face in Interledger is being unable to fulfill the inco
...
 
-Attackers could tie up connectors' liquidity by preparing payments through them that the attacker knows will fail. For example, an attacker could prepare numerous payments to itself and then not fulfill any of them.
+Liquidity Exhaustion: Attackers could tie up connectors' liquidity by preparing payments through them that the attacker knows will fail. For example, an attacker could prepare numerous payments to itself and then not fulfill any of them.
```
emschwartz:
```suggestion
Attackers could tie up connectors' liquidity by preparing payments through them that the attacker knows will fail. For example, an attacker could prepare numerous payments to itself and then not fulfill any of them.
```

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
```diff
@@ -16,29 +14,60 @@ The main risk connectors face in Interledger is being unable to fulfill the inco
...
-
-Possible approaches include:
-
-* **Accounting for Slippage** - Connectors may add a buffer to their expected exchange rate to account for some movement in the price, and they may add a premium for especially volatile currencies
-* **Preferring Shorter Timeouts** - Connectors may provide better exchange rates for payments with shorter timeouts because they carry less risk that the price will move drastically while the payment is in flight
+    Counterparty Theft: Suppose you have connectors A, B, C and D, and the payment is going from A->B->C->D. Take the case of C. During the fulfill phase, D will pass a fulfill packet to C, meaning that C will have to pay. It's important to note that the balances have been updated, but settlement can occur at a later time if dealing with a trusted counterparty. In the best case scenario, we settle in payment channels on every balance update. But because ILP is using unconditional payment channels for settlements, C can write a malicious client that could forward a fulfill packet to B, and avoid paying D by withholding settlement. In this case, A would pay B, B would pay C, but C didn't pay D.
```
emschwartz: This seems like a complicated way to describe a straightforward counterparty risk. If you forward packets on credit, you are taking a risk that the counterparty doesn't or can't settle later. Possible solutions are requiring your peer to pre-fund each transaction (possibly with a pre-emptive payment channel update), keeping the total amount unfunded low, or having an external arrangement that ensures they will pay up later.

emschwartz:
```suggestion
*  **Counterparty Risk** - If connectors forward ILP packets on credit, there is a risk that a peer will not settle their debts later, either out of malice or because they have run out of money.
```

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
```diff
@@ -16,29 +14,60 @@ The main risk connectors face in Interledger is being unable to fulfill the inco
...
+Possible mitigations include:
+* **Accept only smaller payments** - If the payment per packet is less than that
+    of an on chain-fee, a connector that cheats its counterparty would lose much
+    more capital in the process. For example, a connector that has been deprived of
+    payment from its counterparty could refuse to facilitate future payments,
+    and the malicious counterparty would have to close their payment channel to
+    make the stolen micropayment amount liquid. Given that on-chain transaction fees
+    are much higher than that of a micropayment in Interledger (i.e. 1 satoshi), an
+    attacker would lose more money opening and closing channels to cheat their
+    counterparty. Therefore, while this grieving attack is possible, if
```
emschwartz:
```suggestion
    counterparty. Therefore, while this griefing attack is possible, if
```

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
```diff
@@ -16,29 +14,60 @@ The main risk connectors face in Interledger is being unable to fulfill the inco
...
-
-Possible approaches include:
-
-* **Accounting for Slippage** - Connectors may add a buffer to their expected exchange rate to account for some movement in the price, and they may add a premium for especially volatile currencies
-* **Preferring Shorter Timeouts** - Connectors may provide better exchange rates for payments with shorter timeouts because they carry less risk that the price will move drastically while the payment is in flight
+    Counterparty Theft: Suppose you have connectors A, B, C and D, and the payment is going from A->B->C->D. Take the case of C. During the fulfill phase, D will pass a fulfill packet to C, meaning that C will have to pay. It's important to note that the balances have been updated, but settlement can occur at a later time if dealing with a trusted counterparty. In the best case scenario, we settle in payment channels on every balance update. But because ILP is using unconditional payment channels for settlements, C can write a malicious client that could forward a fulfill packet to B, and avoid paying D by withholding settlement. In this case, A would pay B, B would pay C, but C didn't pay D.
 
+Possible mitigations include:
```
emschwartz: The most important one is limiting the total value of unfunded or unsecured payments you're willing to forward for a peer. Accepting smaller payments isn't useful if they send a ton of small payments and you aren't limiting the total

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
khoslaventures: I like your suggestion, but I think having an example somewhere below would be helpful for understanding - maybe this is a greater problem of how to convey concepts in ILP to understand what you wrote concisely

khoslaventures: Agreed, it is complex, but it's an example. We can try moving the example below.

0018-connector-risk-mitigations/0018-connector-risk-mitigations.md
khoslaventures: I placed a section here because I think you could add a scenario where someone just connects to you, opens channels, and then will take payment but not forward it to their counterparties. Thoughts?

khoslaventures: That wouldn't be liquidity exhaustion but rather "trolling" in the sense of delaying payments. It could be expanded to a denial of service if you had everyone connect to your malicious nodes and would not forward their payments. In essence, you would waste an on-chain close fee in the case of using payment channels for settlement.

khoslaventures: Actually, the section is for Payments Griefing, which contains two types of attacks: Liquidity Exhaustion and Counterparty Theft. I think they are separate because liquidity can be exhausted through holding up credit, whereas theft is basically taking a settlement but not passing it on.

emschwartz:
@adrianhopebailie can you re-review or approve this PR so it can be merged?

PR #498, opened 2018-11-19
Add CC-BY-SA badge to footer

adrianhopebailie:


adrianhopebailie:
Added an override for the version checking for cases like this where we change the template.

sappenin:
@adrianhopebailie Is this something we should re-open?

PR #502, opened 2019-01-04
Add CC-BY-SA badge to footer

adrianhopebailie:


tmpl/rfc.ejs.html
```diff
@@ -104,13 +104,14 @@
         <div class="row">
           <div class="col-md-12">
             <div class="col-sm-4 col-xs-12">
-              <p>Interledger Team</p>
+              <p>Copyright © 2018<br/> W3C Interledger Payments CG</p>
```
sappenin: Is there any way to dynamically get this date? It's already outdated...unless you want it to be pegged to 2018?

sappenin: Is there any way to dynamically get this date? It's already outdated...unless you want it to be pegged to 2018?

adrianhopebailie: > Is there any way to dynamically get this date?

There may be but I'll have to log that as a new issue. Don't have the time to investigate now.

PR #504, opened 2019-01-24
Add ILP Over HTTP

emschwartz:
Here is a minimal spec for the protocol proposed in [this blog post](https://medium.com/interledger-blog/thoughts-on-scaling-interledger-connectors-7e3cad0dab7f).

### DIFF
--- a/0035-ilp-over-http/0035-ilp-over-http.md
+++ b/0035-ilp-over-http/0035-ilp-over-http.md
@@ -1,0 +1,56 @@
+---
+title: ILP Over HTTP
+draft: 1
+---
+
+# ILP Over HTTP
+> A bilateral communication protocol for server-to-server connections
+
+## Motivation
+
+Scaling Interledger infrastructure to handle large volumes of ILP packets requires horizontally scaling connectors. Using HTTP for bilateral communication enables service providers to leverage standard tools and services for hosting, load balancing, Distributed Denial of Service (DDoS) protection, and monitoring.
+
+## Overview
+
+In an ILP Over HTTP connection, both peers run HTTP servers with accessible HTTPS endpoints. When peering, the peers exchange their respective URLs, authentication tokens or TLS certificates, ILP addresses, and settlement-related details.
+
+Each ILP Prepare packet is sent as the body of an HTTP request to the peer's server endpoint. ILP Fulfill or Reject packets are returned as the body of the HTTP response.
+
+## Specification
+
+This is a minimal protocol built on HTTP. HTTP/2 is HIGHLY RECOMMENDED for performance reasons, although HTTP/1.1 MAY also be used. Implementations SHOULD support HTTP version negotiation via Application Protocol Negotiation (ALPN).
+
+### Authentication
+
+Peers MAY use any standard HTTP authentication mechanism to authenticate incoming requests. TLS Client Certificates are RECOMMENDED between peers for security and performance, though bearer tokens such as JSON Web Tokens (JWTs) or Macaroons MAY be used instead. Basic authentication (username and password) is NOT RECOMMENDED, because of the additional delay introduced by securely hashing the password.
+
+### Request
+
+```http
+POST /ilp HTTP/x.x
+Accept: application/octet-stream
+Content-Type: application/octet-stream
+Authorization: Bearer zxcljvoizuu09wqqpowipoalksdflksjdgxclvkjl0s909asdf
+Host: connector.example
+
+< Body: Binary OER-Encoded ILP Prepare Packet >
+```
+
+**Field Details:**
+
+- **Path** - A connector MAY specify any HTTP path for their peer to send ILP packets to.
+- **Body** - ILP Packet encoded using OER, as specified in [RFC 27: Interledger Protocol V4](./0027-interledger-protocol-4/0027-interledger-protocol-4.md).
+
+### Response
+
+```http
+HTTP/x.x 200 OK
+Content-Type: application/octet-stream
+
+< Body: Binary OER-Encoded ILP Fulfill or Reject Packet >
+```
+
+All ILP Packets MUST be returned with the HTTP status code `200: OK`.
+
+An endpoint MAY return standard HTTP errors, including but not limited to: a malformed or unauthenticated request, rate limiting, or an unresponsive upstream service. Connectors SHOULD either retry the request, if applicable, or relay an ILP Reject packet back to the original sender with an appropriate [Final or Temporary error code](./0027-interledger-protocol-4/0027-interledger-protocol-4#error-codes). Server errors (status codes 500-599) SHOULD be translated into ILP Reject packets with `T00: Temporary Error` codes.
+


0000-ilp-over-http.md
```diff
@@ -0,0 +1,57 @@
...
+This is a minimal protocol built on HTTP. HTTP/2 is RECOMMENDED for performance reasons, although HTTP/1.1 MAY also be used. Implementations SHOULD support HTTP version negotiation via Application Protocol Negotiation (ALPN).
+
+### Authentication
+
+Peers MAY use any standard HTTP authentication mechanism to authenticate incoming requests. Bearer Tokens or TLS Client Certificates are RECOMMENDED.
+
+### Request
+
+```http
+POST /ilp HTTP/2.0
```
emschwartz: Is there a canonical way of writing out HTTP2 requests? It's a binary format so maybe not...

0000-ilp-over-http.md
```diff
@@ -0,0 +1,57 @@
...
+Accept: application/octet-string
+Content-Type: application/octet-string
```
emschwartz: Should the content type be `application/octet-string` or `application/ilp+octet-string`?

0000-ilp-over-http.md
```diff
@@ -0,0 +1,57 @@
...
+Authorization: Bearer zxcljvoizuu09wqqpowipoalksdflksjdgxclvkjl0s909asdf
+Host: connector.example
+X-ILP-Destination: example.usd.connector.account
```
sharafian: The `X-` is no longer standard practice: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers "Custom proprietary headers can be added using the 'X-' prefix, but this convention was deprecated in June 2012, because of the inconveniences it caused when non-standard fields became standard in RFC 6648;"

0000-ilp-over-http.md
sharafian: :+1: `application/ilp+octet-stream` will still be recognized as an octet string by most applications, while still allowing us to determine if an ILP packet is being sent by content type

0000-ilp-over-http.md
sharafian: I think that we should specify this protocol without saying a specific HTTP version. Even if HTTP2 is used on the outgoing side, it's likely that the HTTP2 connection will be made to a load balancer and then a regular HTTP connection will be made to the application that actually implements this protocol. A client might also use a proxy like envoy (https://www.envoyproxy.io/) to upgrade outgoing requests to HTTP2 while using an ordinary HTTP client themselves.

0000-ilp-over-http.md
sharafian: Octet-stream is the mime-type, not octet-string https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types#applicationoctet-stream

0000-ilp-over-http.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Motivation
+
+Scaling Interledger infrastructure to handle large volumes of ILP packets requires horizontally scaling connectors. Using HTTP for bilateral communication enables service providers to leverage standard tools and services for hosting, load balancing, Distributed Denial of Service (DDoS) protection, and monitoring.
+
+## Overview
+
+In an ILP Over HTTP connection, both peers run HTTP servers with accessible HTTPS endpoints. When peering, the peers exchange their respective URLs, authentication tokens or TLS certificates, ILP addresses, and settlement-related details.
+
+Each ILP Prepare packet is sent as the body of an HTTP request to the peer's server endpoint. The HTTP response body will contain either the ILP Fulfill or Reject packet.
```
sharafian: The HTTP server might return an error code rather than returning an ILP packet, if it couldn't be processed/authenticated

0000-ilp-over-http.md
```diff
@@ -0,0 +1,57 @@
...
+Host: connector.example
+X-ILP-Destination: example.usd.connector.account
+
+< Body: Binary OER-Encoded ILP Prepare Packet >
+```
+
+**Field Details:**
+
+- **Path** - A connector MAY specify any HTTP path for their peer to send ILP packets to.
+- **X-ILP-Destination Header** - (Optional) It is RECOMMENDED to set this header to the destination ILP address from the ILP packet. This enables standard HTTP load balancers to route packets to particular connector instances without parsing the ILP packet.
```
sharafian: Can we also add `ILP-Peer-Name` as in https://github.com/interledgerjs/ilp-plugin-http ? It can be a useful feature in a situation where a stateful cluster of connectors is talking to a stateless cluster of connectors.

0000-ilp-over-http.md
emschwartz: I'll change the example, but I think it's important to mention what versions implementations SHOULD support or default to. If you're using this internally, all bets are off, but the purpose of the spec is to make sure that messages that cross administrative boundaries work.

adrianhopebailie: As discussed on the community group call, using HTTP pre-2 introduces HOL blocking that makes this likely to be significantly worse than BTP. 

I don't think using this protocol is worth it unless you use HTTP2.

0000-ilp-over-http.md
```diff
@@ -0,0 +1,59 @@
...
+### Response
+
+```http
+HTTP/x.x 200 OK
+Content-Type: application/ilp+octet-stream
+
+< Body: Binary OER-Encoded ILP Fulfill or Reject Packet >
+```
+
+All ILP Packets MUST be returned with the HTTP status code `200: OK`. An endpoint MAY return HTTP errors if a request is malformed or unauthenticated.
```
sentientwaffle: * Specify how an HTTP error code response should be converted to an `IlpReject` to be relayed back in the chain.
* `"if a request is malformed or unauthenticated"` makes it sound like those are the only 2 cases where an endpoint could return non-`200`, but there are plenty of others (e.g. the load balancer rate limiting, a reverse proxy sending "service unavailable", etc.).

0000-ilp-over-http.md
emschwartz: Quite possibly, and I think in most cases the libraries people would use will support HTTP2. Do you think the language around its use should be stronger?

0000-ilp-over-http.md
```diff
@@ -0,0 +1,59 @@
...
+### Response
+
+```http
+HTTP/x.x 200 OK
+Content-Type: application/ilp+octet-stream
+
+< Body: Binary OER-Encoded ILP Fulfill or Reject Packet >
+```
+
+All ILP Packets MUST be returned with the HTTP status code `200: OK`. An endpoint MAY return standard HTTP errors, including but not limited to: a malformed or unauthenticated request, rate limiting, or an unresponsive upstream service. Connectors SHOULD either retry the request, if applicable, or relay an ILP Reject packet back to the original sender with an appropriate [Final or Temporary error code](./0027-interledger-protocol-4/0027-interledger-protocol-4#error-codes).
```
sappenin: > Connectors SHOULD either retry the request, if applicable, or relay an ILP Reject packet back to the original sender

I think it would be worthwhile to clarify (non-normatively) what some of these applicable cases _might_ be. Probably this means just suggesting how to handle ambiguous HTTP error-code responses, which I think limited to the `5xx` family. 

I.e., cases where it's ambiguous what the receiving peer did with the ILP-prepare packet. For example, an HTTP 500 doesn't tell me if my peer meant to respond with a `FULFILL` but blew-up before being able to do so; or...if it instead meant to `REJECT`, but fell over before it could say that.

sharafian: It might actually be OK to say that a `5xx` error code should trigger a `T00` error, and retries should never be done at this layer. STREAM already handles retrying so we shouldn't put unnecessary work onto the connector.

kincaidoneil: Clarification: are HTTP error codes ever sent *in lieu of* an ILP reject (e.g. in the case the request was authenticated, got through the load balancer, and the PREPARE was correctly formed)?

0000-ilp-over-http.md
```diff
@@ -0,0 +1,59 @@
...
+This is a minimal protocol built on HTTP. HTTP/2 is HIGHLY RECOMMENDED for performance reasons, although HTTP/1.1 MAY also be used. Implementations SHOULD support HTTP version negotiation via Application Protocol Negotiation (ALPN).
+
+### Authentication
+
+Peers MAY use any standard HTTP authentication mechanism to authenticate incoming requests. Bearer Tokens or TLS Client Certificates are RECOMMENDED.
+
+### Request
+
+```http
+POST /ilp HTTP/x.x
```
adrianhopebailie: I'm still not convinced about this format vs simply putting the packet headers as HTTP headers and the `data` payload as the data.

The currently method has 2 major disadvantages:
  1. Lots of redundancy if you want to put the ILP address in a header and the packet (this is often the bulk of the packet size already)
  2. Double serialization. You are serializing the packet as OER and then again into an HTTP message.

@justmoon cited the risk of packet corruption but I'm not sure that risk is realistic. This is intended to only be used over HTTPS so the data can't be changed at all between the endpoints.

I realize there is a risk that SSL terminating proxies or load balancers may add or remove headers but they never change existing headers, especially ones they don't understand.

I think it would be worth testing both formats and validating some of these assumptions but in theory this is going to be a lot less efficient.

0000-ilp-over-http.md
```diff
@@ -0,0 +1,59 @@
...
+Accept: application/ilp+octet-stream
+Content-Type: application/ilp+octet-stream
```
adrianhopebailie: I don't think custom content types are necessary unless you plan on doing some kind of protocol negotiation or having other protocols served from the same endpoint.

This is just adding additional implementation burden on top of standard libraries IMO.

On the other hand, if you were sending the packet headers as HTTP headers you could use the types `application/ilp-prepare+octet-stream`, `application/ilp-fulfill+octet-stream`, and `application/ilp-reject+octet-stream` to indicate the type (possibly useful in the load balancer processing too to have this surfaced in the headers).

0000-ilp-over-http.md
adrianhopebailie: Possible case may be that the HTTP module on the connector is overloaded and is simply responding to all requests with a 503. I agree with Ben that the safest thing to do is always handle a 5XX response by triggering a T00 error (or simply letting the request time out?).

0000-ilp-over-http.md
adrianhopebailie: This is (I think) a symptom of the problem with the current plugin/connector architecture.

The connector assumes that a plugin represents a single account BUT a plugin uses protocols like HTTP and WebSockets where it hosts a server and could have many clients.

The result is you have to have mini-accounts (i.e. a lite connector) implemented inside the plugin.

The correct way to do this is to authenticate a connection and then you can assume that all messages on that connection are from the same client (who has now been authenticated as a specific peer).

In lieu of this I think Ben is suggesting that a peer can identify themselves using this extra header.

emschwartz: Unless this type of header is strictly necessary, I'd prefer to leave this to the authentication mechanism chosen. If a peer wants to identify themselves, a username or a specific TLS client cert seems like a better way to do that.

0000-ilp-over-http.md
emschwartz: What about T01: Peer Unreachable, T02: Peer Busy, or T05: Rate Limited? Those sound like they _could_ be translations of HTTP 5xx errors.

That said, it would be easiest to implement if you just translate all 5xx errors into T00 ILP errors.

0000-ilp-over-http.md
```diff
@@ -0,0 +1,59 @@
...
+
+## Overview
+
+In an ILP Over HTTP connection, both peers run HTTP servers with accessible HTTPS endpoints. When peering, the peers exchange their respective URLs, authentication tokens or TLS certificates, ILP addresses, and settlement-related details.
+
+Each ILP Prepare packet is sent as the body of an HTTP request to the peer's server endpoint. ILP Fulfill or Reject packets are returned as the body of the HTTP response.
+
+## Specification
+
+This is a minimal protocol built on HTTP. HTTP/2 is RECOMMENDED for performance reasons, although HTTP/1.1 MAY also be used. Implementations SHOULD support HTTP version negotiation via Application Protocol Negotiation (ALPN).
```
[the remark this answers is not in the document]
emschwartz: I think it's worth allowing HTTP/1.1 for convenience reasons, even though I'd expect most, if not all, real deployments of this to use HTTP2 (and switch to HTTP3 when it's available).

That said, a big part of the motivation is being able to load balance and horizontally scale connector clusters. HTTP/1.1 doesn't prevent that, it just means that a dropped packet would slow down other requests sent over the same connection. That may not be a huge issue if you're either not sending requests over the same TCP connection (which would incur a different performance hit) or if you're on a closed network that's less likely to drop packets than the open internet.

0000-ilp-over-http.md
```diff
@@ -0,0 +1,59 @@
...
+
+### Authentication
+
+Peers MAY use any standard HTTP authentication mechanism to authenticate incoming requests. Bearer Tokens or TLS Client Certificates are RECOMMENDED.
```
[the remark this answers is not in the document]
emschwartz: With HTTP2 header compression, it wouldn't actually send the full string over the wire each time. 

Would using TLS client certs do what you're suggesting? If so, that sounds good to me. However, as far as I understand, client certs can be somewhat inconvenient to setup so I'd prefer to give options and let people determine what makes sense for their setup.

0000-ilp-over-http.md
emschwartz: I don't feel strongly about this. @sharafian want to weigh in?

sharafian: I don't think we should be putting the internals of the packet into the headers. I guess custom content types may not be used but I can imagine some cases where you might want to use them (i.e. a middleware that catches all ILP packets but doesn't require a special pathname, just like we do for SPSP).

0000-ilp-over-http.md
sharafian: > As discussed on the community group call, using HTTP pre-2 introduces HOL blocking that makes this likely to be significantly worse than BTP.

HTTP with keepalive has head of line blocking, but you can just put each request on a different connection. This adds round trips, especially if TLS is used, which is why I would only imagine it inside of a datacenter/cluster or even between processes on a single machine. But I don't want to deal with the headache of certs and HTTP2 support everywhere inside of our cluster when we could just have the load balancer deal with HTTP2 and do HTTP/1.1 inside.

0000-ilp-over-http.md
sharafian: > Double serialization. You are serializing the packet as OER and then again into an HTTP message.

This change would cause us to deserialize the packet more times, only for it to be reserialized when it's passed to the connector. Unless we changed the ledger plugin interface to pass around the parsed packet (rather than buffers), I don't think this is a good idea.

0000-ilp-over-http.md
```diff
@@ -0,0 +1,59 @@
...
+Host: connector.example
+ILP-Destination: example.usd.connector.account
+
+< Body: Binary OER-Encoded ILP Prepare Packet >
+```
+
+**Field Details:**
+
+- **Path** - A connector MAY specify any HTTP path for their peer to send ILP packets to.
+- **ILP-Destination Header** - (Optional) It is RECOMMENDED to set this header to the destination ILP address from the ILP packet. This enables standard HTTP load balancers to route packets to particular connector instances without parsing the ILP packet.
```
sharafian: I'm not really sold on this. The load balancers in google cloud and AWS don't support features this advanced. It's possible you could do it with HAproxy or something but if it requires a lot of customization then you might as well write a load balancer that understands ILP packets (basically a connector at that point)

0000-ilp-over-http.md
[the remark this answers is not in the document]
adrianhopebailie: Client certs are definitely harder to setup but they'd be more efficient.

My point here is to avoid suggesting bearer tokens or using them in the example because the result will be that it becomes the de-facto standard for this protocol.

0000-ilp-over-http.md
adrianhopebailie: > Unless we changed the ledger plugin interface to pass around the parsed packet (rather than buffers), I don't think this is a good idea.

We've already done this in the connector refactor for this exact reason. The current interface is too opinionated on how the packet will be sent over the wire.

emschwartz: I'm also a fan of changing the implementations to pass around a deserialized representation (or making this change in any new implementations if we don't want to change the JS one). We made the interface "send data" to begin with with the idea that you might pass different types of data through that. However, there are so many places that deserialize the packet and only really work if it is an ILP packet that the separation of the layers doesn't really work.

0000-ilp-over-http.md
emschwartz: You're right, I actually misread something in the Google Cloud Load Balancer and thought they did support this. Client IP-based session affinity should be fine.

0000-ilp-over-http.md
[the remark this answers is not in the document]
emschwartz: > the result will be that it becomes the de-facto standard for this protocol

I don't think that's true. As far as I know, most HTTP libraries support both methods and peers will use whatever setup makes sense for them.

0000-ilp-over-http.md
sappenin: > ...unless you plan on doing some kind of protocol negotiation or having other protocols served from the same endpoint.

I prefer using custom content-types to make room for ideas like the quote cited above...for example, I plan to build both "packet serialization" variants of this proposal (HTTP headers vs OER Packet in the body) to better understand performance, and having a narrower content-type would be helpful here to allow room for other types (e.g., `application/ilp-header+octet-stream`).

0000-ilp-over-http.md
```diff
@@ -0,0 +1,57 @@
...
+
+Each ILP Prepare packet is sent as the body of an HTTP request to the peer's server endpoint. ILP Fulfill or Reject packets are returned as the body of the HTTP response.
+
+## Specification
+
+This is a minimal protocol built on HTTP. HTTP/2 is HIGHLY RECOMMENDED for performance reasons, although HTTP/1.1 MAY also be used. Implementations SHOULD support HTTP version negotiation via Application Protocol Negotiation (ALPN).
+
+### Authentication
+
+Peers MAY use any standard HTTP authentication mechanism to authenticate incoming requests. TLS Client Certificates are RECOMMENDED for security and performance.
```
sharafian: TLS client certs are a little tricky when the HTTP server in the plugin exists behind a load balancer (it's a plain HTTP server rather than HTTPS)

0000-ilp-over-http.md
sharafian: Regardless of which way you lean, this is really a larger discussion that probably shouldn't be decided in this PR. If we decide to make that change across the board then it would be big enough to justify a change in this protocol

0000-ilp-over-http.md
emschwartz: I'm eager to resolve this discussion because ultimately it doesn't make a big difference whether we use a custom content type or not and we could bikeshed this topic indefinitely.

If we want to use a custom type in the future for a different sort of protocol (maybe along the lines of the ILP packet headers in the HTTP headers), that would be enough to differentiate the OER-serialized version from the later version.

Does anyone feel _very strongly_ that it should be `application/ilp+octet-stream`?

sappenin: FWIW, I don't feel _very_ strongly. My implementation supports all three content-types.

;)

0000-ilp-over-http.md
adrianhopebailie: > put each request on a different connection

Connection establishment is the biggest culprit in most network perf issues. The TCP connection establishment handshake is one of the main drivers behind QUIC.

> I don't want to deal with the headache of certs and HTTP2 support everywhere inside of our cluster when we could just have the load balancer deal with HTTP2 and do HTTP/1.1 inside.

What you do internally is up to you. It definitely shouldn't drive what's in the standard

emschwartz: The current draft says:

```
HTTP/2 is HIGHLY RECOMMENDED for performance reasons, although HTTP/1.1 MAY also be used. Implementations SHOULD support HTTP version negotiation via Application Protocol Negotiation (ALPN).
```

@adrianhopebailie Is that okay with you?

0000-ilp-over-http.md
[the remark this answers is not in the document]
adrianhopebailie: Peers could have agreed to auth BTP using proper WebSocket handshakes but they all settled on credentials in the URL translated into an `auth` subprotocol because that's what was in the examples...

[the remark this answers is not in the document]
emschwartz: Actually the BTP spec is pretty specific about how it expects peers to authenticate one another, and it doesn't make it sound like there are options:

> Before anything else, when a client connects to a server, it sends a special Message request. Its primary protocolData entry MUST have name 'auth', content type MIME_APPLICATION_OCTET_STREAM, and empty data, and among the secondary entries, there MUST be a UTF-8 'auth_token' entry. The further secondary protocol data entries of this Message request MAY also be used to send additional information to the server. In situations where no authentication is needed, the 'auth_token' data can be set to the empty string, but it cannot be omitted.

> **If the client sends any BTP call that is not a Message, or sends a Message call whose primary sub-protocol is not auth, the server should respond with an Error and close the connection.** [emphasis added]

0000-ilp-over-http.md
adrianhopebailie: @sappenin that's a good reason to use custom content types and actually means we could allow both serializations (OER in body AND packet headers in HTTP headers) because the `Accept` and `Content-Type` headers will tell the peer what is supported.

Can I suggest the following types:
 - `application/ilp-prepare+octet-stream`
 - `application/ilp-fulfill+octet-stream`
 - `application/ilp-reject+octet-stream`

 - `application/ilp+octet-stream` (where packet is OER encoded in body, therefor type is determined from type indicator byte in packet)

Given that most HTTP modules already have easy ways to bind application logic content types tit seems like supporting both serialisations would be pretty trivial. @sappenin, is that your experience?

0000-ilp-over-http.md
adrianhopebailie: HAProxy definitely does support this (and I think nginx) so I wouldn't exclude it. 
I agree that we shouldn't make it mandatory. 

If you did the header + payload serialisation you'd have this header anyway so that could be a way for peers to expose it if they decide they need it.

0000-ilp-over-http.md
adrianhopebailie: I assume you can't even do client cert validation on a load balancer anyway. If you could (more likely on some kind of TLS appliance) you'd almost certainly expect the identity of the client to be passed along somehow, maybe in a header.

That said, writing a spec like this with strong assumptions about the deployment scenarios is a bad idea. The spec shouldn't assume this protocol will only be used by people behind a load balancer.

The reality is that client certs ARE better for performance so saying so is not wrong.

0000-ilp-over-http.md
sappenin: @adrianhopebailie I agree with your type suggestions in principle, and also like the point by @emschwartz about keeping things simple. 

To that end, let's discuss the details of the header-variant of this protocol in a different PR (I don't see anything in the current PR that would block any of that).

adrianhopebailie:
This issue was moved to https://forum.interledger.org/t/add-ilp-over-http/276

PR #508, opened 2019-02-11
Fixes #507

### LINKED ISSUE #507: Rename RFC-0014 to "0014-paid-http-apis"
[RFC-0014](https://github.com/interledger/rfcs/blob/master/0014-http-ilp/0014-http-ilp.md) is currently called "Http ILP", but this doesn't really describe what this RFC is about, and will likely cause confusion when [ILP over HTTP](https://github.com/interledger/rfcs/pull/504) is merged. 

Instead, we should rename this RFC to "**Paid HTTP APIs**" and rename the folder structure to be `0014-paid-http-apis`.

adrianhopebailie:
Fixes #507 

@sappenin I don't think changing the file name is worth the work.


sappenin:
@adrianhopebailie I renamed the folder and the file to be `paid-http-apis`. Is that OK?

PR #509, opened 2019-02-11
Fixes #506

### LINKED ISSUE #506: Cleanup naming of RFC 0000
Currently, [RFC-0000](https://github.com/interledger/rfcs/blob/master/0000-node-requirements/0000-node-requirements.md) is called "Node Requirements" but the RFC text says "Connector Requirements". This should be cleaned up. Also, shouldn't this RFC have a number other than 0000?

adrianhopebailie:
@sappenin - 0034 is the connector requirements RFC, this should have just been deleted


sappenin:
Oh nice, my bad.

PR #510, opened 2019-02-13
fix code block formatting

Unihedro:
Currently two code blocks are getting merged into one, with the subtitle consumed in the block


emschwartz:
@adrianhopebailie Should this PR also update the draft number? I thought circle wouldn't let anything through if the draft number wasn't incremented

Unihedro:
Are you talking about this check?

https://github.com/interledger/rfcs/blob/master/scripts/generate_web.js#L139

I don't know why but it seems that circle didn't mind the new line, as it logged `"No changes in 0028-web-monetization/0028-web-monetization.md"`

adrianhopebailie:
@emschwartz it does a compare of the rendered version with the current rendered version so they are clearly not rendering differently

PR #511, opened 2019-02-20
docs(0027): add Invalid Peer Response error code

emschwartz:
for cases when the next peer responds to an ILP Prepare with something invalid, such as a fulfillment that does not match the Prepare

### DIFF
--- a/0027-interledger-protocol-4/0027-interledger-protocol-4.md
+++ b/0027-interledger-protocol-4/0027-interledger-protocol-4.md
@@ -1,6 +1,6 @@
 ---
 title: Interledger Protocol V4 (ILPv4)
-draft: 5
+draft: 6
 ---
 
 # Interledger Protocol V4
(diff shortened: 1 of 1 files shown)


adrianhopebailie:
Interesting scenario. Is this a final error though?

I'm trying to determine what logic one would implement in a connector that received this error?

emschwartz:
Good question. I put it as final because it indicates that something is
pretty wrong with that peer. You could try it again but it seems unlikely
that it would just start working if something more fundamental is wrong
with their system.

adrianhopebailie:
I assume that in the following (badly drawn) scenario:

```
A ---> B ---> C ---> E
       |             |
		----> D ---->|
```

If `B` gets an invalid response from `C` they will generate the reject message to `A`?
In that case, I'd expect `B` to consider the link to `C` as down and to then start routing packets to `E` via `D` instead. So `A` could send the same packet again and expect to get a new result.

So I'd agree that `B` considers `C` as permanently down until further notice but that's an internal routing consideration. They could try routing the same packet via `D` without even responding to `A` if the expiry allowed this.

From `A`'s perspective this is not a final error (or maybe not an error they even need to understand as being different from a generic temp error?)

Maybe we should pick this up on the forum?

emschwartz:
I could go either way on this question.

It does seem a bit unlikely that connectors would have logic to disconnect
a peer automatically if they get a bad fulfillment back.

I think this kind of very specific technical detail is fine to discuss on
github, but I can open a forum thread if you want to broaden the discussion.

adrianhopebailie:
This issue was moved to https://forum.interledger.org/t/docs-0027-add-invalid-peer-response-error-code/278

emschwartz:
@adrianhopebailie Any objections to merging this?

adrianhopebailie:
> Any objections to merging this?

I still don't know why this is a Final error? 

That indicates to the sender that they should not retry the request but I don't think that is correct. Nothing about the request is wrong.

PR #514, opened 2019-02-27
Pull payment spec

sabinebertram:
As promised when closing #501, this is the stand-alone pull-payment spec. It extends the basic SPSP spec by explaining how a pull payment is conducted once a STREAM connection has been established using SPSP. It also specs out optional response body fields.

### DIFF
--- a/0036-spsp-pull-payments/0036-spsp-pull-payments.md
+++ b/0036-spsp-pull-payments/0036-spsp-pull-payments.md
@@ -1,0 +1,135 @@
+---
+title: SPSP Pull Payments
+draft: 1
+---
+# SPSP Pull Payments
+
+## Preface
+
+This document describes how to conduct pull payments via [STREAM](../0029-stream/0029-stream.md) once the payment details have been exchanged via the [Simple Payment Setup Protocol](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md) (SPSP).
+
+## Introduction
+
+### Motivation
+
+The Simple Payment Setup Protocol (SPSP) only describes how the details required for a STREAM connection are exchanged and how a simple push payment is made. However, it lacks a detailed explaination of how a pull payment is conducted using this connection. 
+
+### Scope
+
+This document specifies how to conduct a pull payment once a STREAM connection has been set up. It is intended for use by end-user applications.
+
+### Operation Overview
+
+Any SPSP server will expose an HTTPS endpoint called the SPSP Endpoint. The client can query this endpoint to get information about about the server's connection details, namely ILP address and a shared secret. The endpoint MAY also contain additional information for displaying purposes. The client can set up a STREAM connection and pull multiple ILP payments using the details provided by the server. 
+
+## Model of Operation
+
+### Creating a token
+
+A pull payment token SHOULD be be opaque. For the purpose of offline generation, it MAY include negotiated pull payment agreement parameters that SHOULD become obsolete once the SPSP endpoint corresponding to that token has been created. The SPSP server SHOULD store the negotiated pull payment agreement parameters.
+
+The pull payment agreement parameters that SHOULD be negotiated are those defined in the `agreement` object within the [Response Body](#Response-Body). 
+
+### Conducting the pull payment
+
+We assume that the client knows the server's SPSP endpoint (see [Payment Pointers](../0026-payment-pointers/0026-payment-pointers.md)).
+
+1. The user's SPSP client queries the server's SPSP Endpoint.
+
+2. The SPSP endpoint responds with the server info, including the server's ILP address (`destination_account`) and the shared secret (`shared_secret`) to be used in STREAM. 
+    * The `destination_account` SHOULD be used as the STREAM destinationAccount.
+    * The `shared_secret` SHOULD be decoded from base64 and used as the STREAM sharedSecret.
+
+3. The SPSP client establishes a STREAM connection to the server using the server's ILP address and shared secret and constructs a stream to the client on the ILP/STREAM connection.
+
+4. The SPSP server begins sending ILP packets to fulfill the payment.
+    1. The server will adjust their `sendMax` to reflect the amount they're willing to send.
+        * `sendMax` SHOULD be derived from the pull payment agreement parameters.
+    2. The client will adjust their `receiveMax` to reflect the amount they're willing to receive.
+        * `receiveMax` MAY be `Infinity`.
+    3. The client's and server's STREAM modules will move as much value as possible while staying inside these bounds.
+    4. If the server reaches their `sendMax`, they end the stream and the connection. If the client reaches their `receiveMax`, they will end the stream and the connection.
+
+## Specification
+
+### Query (`GET <SPSP Endpoint>`)
+
+The client queries the SPSP endpoint to get information about the server:
+
+#### Request
+
+``` http
+GET /0f09dc92-84ad-401b-a7c9-441bc6173f4e HTTP/1.1
+Host: alice.com
+Accept: application/spsp4+json, application/spsp+json
+```
+
+#### Response
+
+``` http
+HTTP/1.1 200 OK
+Content-Type: application/spsp4+json
+
+{
+  "destination_account": "alice.ilpdemo.red~0f09dc92-84ad-401b-a7c9-441bc6173f4e",
+  "shared_secret": "k5nubgM6zpb88NPGVnI/tVjRdgpUh+JvMueRFEMvPcY=",
+  "balance": {
+    "total": "5000",
+    "interval": "1000",
+    "available": "1000"
+  },
+  "cycle": 3,
+  "agreement": {
+    "amount": "2000",
+    "start": "2019-01-01T08:00Z",
+    "expiry": "2021-01-02T00:00Z", 
+    "interval": "P0Y1M0DT0H0M0S",
+    "cycles": 12,
+    "cap": false,
+    "asset": {
+      "code": "USD",
+      "scale": 2
+    }
+  }
+}
+``` 
+
+##### Response Body
+
+The response body is a JSON object that includes basic account details necessary for setting up a STREAM connection as well as optional parameters for displaying purposes.
+
+| Field | Type | Description |
+|---|---|---|
+| `destination_account` | [ILP Address](../0015-ilp-addresses/0015-ilp-addresses.md) | ILP Address of the server. In case of push payments, this is the receiver, in case of pull payments, this is the sender. |
+| `shared_secret` | 32 bytes, [base64 encoded](https://en.wikipedia.org/wiki/Base64) (including padding) | The shared secret to be used by this specific HTTP client in the [STREAM](../0029-stream/0029-stream.md). Should be shared only by the server and this specific HTTP client, and should therefore be different in each query response. |
+| `balance`  | Object | _(OPTIONAL)_ Details of this server account's balance. |
+| `balance.total` | Integer String | _(OPTIONAL)_ Total amount, denoted in `agreement.asset.code`, which has been pulled by the client since `start` to date. It is the sum of all outgoing chunks. |
+| `balance.interval` | Integer String | _(OPTIONAL)_ Amount, denoted in `agreement.asset.code`, which has been pulled by the client since the start of the current `cycle`. It is the sum of all outgoing chunks. |
+| `balance.available` | Integer String | _(OPTIONAL)_ Amount, denoted in `agreement.asset.code`, which is still available to be pulled by the client until the end of the current `cycle`. |
+| `cycle` | Integer | _(OPTIONAL)_ Current interval cycle out of a total of `agreement.cycles`. |
+| `agreement` | Object | _(OPTIONAL)_ Details about the pull payment agreement. |
+| `agreement.amount` | Integer String | _(OPTIONAL)_ Amount, denoted in `agreement.asset.code`, which can pulled by the client each `agreement.interval`. |
+| `agreement.start` |	String | _(OPTIONAL)_ [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) UTC timestamp, e.g. `"2019-01-01T08:00Z"`, representing the start of the pull payment agreement. |
+| `agreement.expiry` |	String | _(OPTIONAL)_ [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) UTC timestamp, e.g. `"2021-01-02T00:00Z"`, representing the expiry of the pull payment agreement. It is the time when the SPSP endpoint is destroyed, i.e. remaining funds cannot be pulled after that point in time. |
+| `agreement.interval` | String | _(OPTIONAL)_ [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601) duration, e.g. `"P0Y1M0DT0H0M0S"` = 1 month, which describes how often `agreement.amount` can be pulled. |
+| `agreement.cycles` | Integer | _(OPTIONAL)_ Number of times that `agreement.interval` is applied, starting at `agreement.start`. If `agreement.interval` is 1 month and `agreement.cycles` is 12, then `agreement.amount` can be pulled 12 times within the year starting at `agreement.start`. |
+| `agreement.cap` | Boolean | _(OPTONAL)_ Defines whether any balance not pulled before the start of the next interval cycle is accumulated or expires. If `agreement.cap = true`, the maximum pullable amount per `agreement.interval` is `agreement.amount`. If `agreement.cap = false`, the maximum pullable amount per `agreement.interval` is `agreement.amount` plus any remaining funds accumulated but not pulled over the last interval cycles.|
+| `agreement.asset` | Object | _(OPTIONAL)_ Details about the agreement's asset. |
+| `agreement.asset.code` | String |  _(OPTIONAL)_ Asset code to identify the agreement's currency. Currencies that have [ISO 4217](https://en.wikipedia.org/wiki/ISO_4217) codes should use those. |
+| `agreement.asset.scale` | Integer |  _(OPTIONAL)_ The scale of the amounts `balance.maximum` and `balance.current` are stated in (e.g. an amount of `"1000"` with a scale of `2` translates to `10.00` units of the server's asset/currency) |
+
+**Note:** Currency amounts are denominated as integer strings instead of native JSON numbers to avoid losing precision during JSON parsing. Applications MUST represent these numbers in a data type that has precision equal or greater than an unsigned 64-bit integer.
+
+##### Errors
+
+###### Token Does Not Exist
+
+``` http
+HTTP/1.1 404 Not Found
+Content-Type: application/spsp4+json
+
+{
+  "id": "InvalidPointerError",
+  "message": "Token does not exist."
+}
+```
--- a/0036-spsp-pull-payments/0036-spsp-pull-payments.md
+++ b/0036-spsp-pull-payments/0036-spsp-pull-payments.md
@@ -131,7 +131,7 @@ The response body is a JSON object that includes basic account details necessary
 
 ##### Errors
 
-###### pointer Does Not Exist
+###### Pointer Does Not Exist
 
 ``` http
 HTTP/1.1 404 Not Found
--- a/0036-spsp-pull-payments/0036-spsp-pull-payments.md
+++ b/0036-spsp-pull-payments/0036-spsp-pull-payments.md
@@ -44,7 +44,7 @@ Implementations MAY try to parse the Pull Payment Pointer, however, it MUST supp
 
 ### Conducting the Pull Payment
 
-The Payee's SPSP Client opens a [STREAM](../0029-stream/0029-stream.md) connection to the Payer's SPSP Server as described in the [Simple Payment Setup Protocol](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md). Note that if there is already an existing open STREAM connection to the `destination_account` the Pull Payment Pointer is resolving to, the connection SHOULD be closed. 
+The Payee's SPSP Client opens a [STREAM](../0029-stream/0029-stream.md) connection to the Payer's SPSP Server as described in the [Simple Payment Setup Protocol](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md). Note that if there is already an existing open STREAM connection to the `destination_account` the Pull Payment Pointer is resolving to, the connection SHOULD be closed. The server SHOULD only allow for one stream at a time.
 
 Once this connection is established, the process continues as follows: 
 
--- a/0036-spsp-pull-payments/0036-spsp-pull-payments.md
+++ b/0036-spsp-pull-payments/0036-spsp-pull-payments.md
@@ -58,7 +58,7 @@ The SPSP Server begins sending ILP packets to complete the Pull Payment.
 
 The STREAM parameters - `sendMax` and `receiveMax` - are defined in [STREAM's frame encoding](../0029-stream/0029-stream.md#53-frames).
 
-Note that if there are multiple open STREAM connections to the `destination_account` the Pull Payment Pointer is resolving to, the SPSP Server MUST only set one `sendMax` at a time.
+Note that if there are multiple open streams (via one or more STREAM connections) to the `destination_account` the Pull Payment Pointer is resolving to, the SPSP Server MUST only set one `sendMax` at a time.
 
 ## Specification
 
--- a/0036-spsp-pull-payments/0036-spsp-pull-payments.md
+++ b/0036-spsp-pull-payments/0036-spsp-pull-payments.md
@@ -44,7 +44,9 @@ Implementations MAY try to parse the Pull Payment Pointer, however, it MUST supp
 
 ### Conducting the Pull Payment
 
-The Payee's SPSP Client opens a [STREAM](../0029-stream/0029-stream.md) connection to the Payer's SPSP Server as described in the [Simple Payment Setup Protocol](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md). Once this connection is established, the process continues as follows: 
+The Payee's SPSP Client opens a [STREAM](../0029-stream/0029-stream.md) connection to the Payer's SPSP Server as described in the [Simple Payment Setup Protocol](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md). Note that if there is already an existing open STREAM connection to the `destination_account` the Pull Payment Pointer is resolving to, the connection SHOULD be closed. 
+
+Once this connection is established, the process continues as follows: 
 
 The SPSP Server begins sending ILP packets to complete the Pull Payment.
   1. The SPSP Server will adjust their `sendMax` to reflect the amount they're willing to send.
--- a/0036-spsp-pull-payments/0036-spsp-pull-payments.md
+++ b/0036-spsp-pull-payments/0036-spsp-pull-payments.md
@@ -44,7 +44,7 @@ Implementations MAY try to parse the Pull Payment Pointer, however, it MUST supp
 
 ### Conducting the Pull Payment
 
-The Payee's SPSP Client opens a [STREAM](../0029-stream/0029-stream.md) connection to the Payer's SPSP Server as described in the [Simple Payment Setup Protocol](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md). Note that if there is already an existing open STREAM connection to the `destination_account` the Pull Payment Pointer is resolving to, the connection SHOULD be closed. The server SHOULD only allow for one stream at a time.
+The Payee's SPSP Client opens a [STREAM](../0029-stream/0029-stream.md) connection to the Payer's SPSP Server as described in the [Simple Payment Setup Protocol](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md). 
 
 Once this connection is established, the process continues as follows: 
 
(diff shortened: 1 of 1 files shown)


adrianhopebailie:
This issue was moved to https://forum.interledger.org/t/pull-payment-spec/281

0036-spsp-pull-payments/0036-spsp-pull-payments.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Operation Overview
+
+Any SPSP server will expose an HTTPS endpoint called the SPSP Endpoint. The client can query this endpoint to get information about about the server's connection details, namely ILP address and a shared secret. The endpoint MAY also contain additional information for displaying purposes. The client can set up a STREAM connection and pull multiple ILP payments using the details provided by the server. 
+
+## Model of Operation
+
+### Creating a token
+
+A pull payment token SHOULD be be opaque. For the purpose of offline generation, it MAY include negotiated pull payment agreement parameters that SHOULD become obsolete once the SPSP endpoint corresponding to that token has been created. The SPSP server SHOULD store the negotiated pull payment agreement parameters.
```
adrianhopebailie: This is a difficult paragraph to parse. There is a lot there that I think needs some explanation.

What is a token?
What are "negotiated pull payment agreement parameters"?

0036-spsp-pull-payments/0036-spsp-pull-payments.md
```diff
@@ -0,0 +1,135 @@
...
+
+The pull payment agreement parameters that SHOULD be negotiated are those defined in the `agreement` object within the [Response Body](#Response-Body). 
+
+### Conducting the pull payment
+
+We assume that the client knows the server's SPSP endpoint (see [Payment Pointers](../0026-payment-pointers/0026-payment-pointers.md)).
+
+1. The user's SPSP client queries the server's SPSP Endpoint.
+
+2. The SPSP endpoint responds with the server info, including the server's ILP address (`destination_account`) and the shared secret (`shared_secret`) to be used in STREAM.
```
adrianhopebailie: I would reference the SPSP spec here and not redefine that part of the flow or you risk the specs getting out of sync.

0036-spsp-pull-payments/0036-spsp-pull-payments.md
sabinebertram: I agree, I should define what a _token_ is. There is a brief description of what the _negotiated pull payment agreement parameters_ are in the paragraph following this one but yes, there could be more information on that, too.

I'm wondering, however, how detailed this section should be. The way I understood our previous discussions is that there should not be too much detail about the negotiation phase. If anything, that should become a different spec.

0036-spsp-pull-payments/0036-spsp-pull-payments.md
sabinebertram: Agreed.

0036-spsp-pull-payments/0036-spsp-pull-payments.md
adrianhopebailie: > there should not be too much detail about the negotiation phase. If anything, that should become a different spec.

That's true but there is a lot of new language in here without context. Some light background is needed, especially while the other spec doesn't exist and can't be referenced.

0036-spsp-pull-payments/0036-spsp-pull-payments.md
```diff
@@ -0,0 +1,135 @@
+---
+title: SPSP Pull Payments
+draft: 1
+---
+# SPSP Pull Payments
+
+## Preface
+
+This document describes how to conduct pull payments via [STREAM](../0029-stream/0029-stream.md) once the payment details have been exchanged via the [Simple Payment Setup Protocol](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md) (SPSP).
```
sharafian: what's a pull payment? some readers might not be familiar with the term or what it means in an ILP context

0036-spsp-pull-payments/0036-spsp-pull-payments.md
sharafian: I don't think this phrasing of "SHOULD be opaque" makes sense. The token _is_ opaque unless the spec guarantees that it'll be parsed in a certain way. You could just say that a pull payment pointer can be any valid payment pointer. Implementations MAY try to parse details out of that payment pointer, but they MUST support payment pointers that are totally opaque to them.

0036-spsp-pull-payments/0036-spsp-pull-payments.md
```diff
@@ -0,0 +1,135 @@
...
+
+### Creating a token
+
+A pull payment token SHOULD be be opaque. For the purpose of offline generation, it MAY include negotiated pull payment agreement parameters that SHOULD become obsolete once the SPSP endpoint corresponding to that token has been created. The SPSP server SHOULD store the negotiated pull payment agreement parameters.
+
+The pull payment agreement parameters that SHOULD be negotiated are those defined in the `agreement` object within the [Response Body](#Response-Body). 
+
+### Conducting the pull payment
+
+We assume that the client knows the server's SPSP endpoint (see [Payment Pointers](../0026-payment-pointers/0026-payment-pointers.md)).
```
sharafian: This SPSP endpoint is the token resolved to a URL as per the payment pointer spec, right? Could use clarification

0036-spsp-pull-payments/0036-spsp-pull-payments.md
```diff
@@ -0,0 +1,135 @@
...
+
+1. The user's SPSP client queries the server's SPSP Endpoint.
```
sharafian: We should link to the SPSP spec here so that you can read exactly what an SPSP query looks like

0036-spsp-pull-payments/0036-spsp-pull-payments.md
```diff
@@ -0,0 +1,135 @@
...
+
+2. The SPSP endpoint responds with the server info, including the server's ILP address (`destination_account`) and the shared secret (`shared_secret`) to be used in STREAM. 
+    * The `destination_account` SHOULD be used as the STREAM destinationAccount.
+    * The `shared_secret` SHOULD be decoded from base64 and used as the STREAM sharedSecret.
```
sharafian: MUST

0036-spsp-pull-payments/0036-spsp-pull-payments.md
```diff
@@ -0,0 +1,135 @@
...
+
+3. The SPSP client establishes a STREAM connection to the server using the server's ILP address and shared secret and constructs a stream to the client on the ILP/STREAM connection.
```
sharafian: And the server constructs a stream to the client or the client constructs a stream to the server?

0036-spsp-pull-payments/0036-spsp-pull-payments.md
```diff
@@ -0,0 +1,135 @@
...
+
+4. The SPSP server begins sending ILP packets to fulfill the payment.
```
sharafian: fulfill already means something in interledger. "to complete the pull payment" might be more accurate.

0036-spsp-pull-payments/0036-spsp-pull-payments.md
```diff
@@ -0,0 +1,135 @@
...
+    1. The server will adjust their `sendMax` to reflect the amount they're willing to send.
+        * `sendMax` SHOULD be derived from the pull payment agreement parameters.
+    2. The client will adjust their `receiveMax` to reflect the amount they're willing to receive.
+        * `receiveMax` MAY be `Infinity`.
+    3. The client's and server's STREAM modules will move as much value as possible while staying inside these bounds.
```
sharafian: could link to STREAM here for a description of how these parameters work

0036-spsp-pull-payments/0036-spsp-pull-payments.md
```diff
@@ -0,0 +1,135 @@
...
+    4. If the server reaches their `sendMax`, they end the stream and the connection. If the client reaches their `receiveMax`, they will end the stream and the connection.
```
sharafian: Is the client allowed to re-open a connection using the same token? Or is this one-time use? If both versions are allowed, how can you tell the difference?

0036-spsp-pull-payments/0036-spsp-pull-payments.md
```diff
@@ -0,0 +1,135 @@
...
+### Query (`GET <SPSP Endpoint>`)
+
+The client queries the SPSP endpoint to get information about the server:
+
+#### Request
+
+``` http
+GET /0f09dc92-84ad-401b-a7c9-441bc6173f4e HTTP/1.1
+Host: alice.com
+Accept: application/spsp4+json, application/spsp+json
```
sharafian: should only be `application/spsp4+json`; `application/spsp+json` refers to the deprecated PSK2 version which this document doesn't specify a flow for.

0036-spsp-pull-payments/0036-spsp-pull-payments.md
```diff
@@ -0,0 +1,135 @@
...
+
+##### Response Body
+
+The response body is a JSON object that includes basic account details necessary for setting up a STREAM connection as well as optional parameters for displaying purposes.
+
+| Field | Type | Description |
+|---|---|---|
+| `destination_account` | [ILP Address](../0015-ilp-addresses/0015-ilp-addresses.md) | ILP Address of the server. In case of push payments, this is the receiver, in case of pull payments, this is the sender. |
+| `shared_secret` | 32 bytes, [base64 encoded](https://en.wikipedia.org/wiki/Base64) (including padding) | The shared secret to be used by this specific HTTP client in the [STREAM](../0029-stream/0029-stream.md). Should be shared only by the server and this specific HTTP client, and should therefore be different in each query response. |
+| `balance`  | Object | _(OPTIONAL)_ Details of this server account's balance. |
```
sharafian: Is the balance denominated in the agreement asset or in receiver units?

0036-spsp-pull-payments/0036-spsp-pull-payments.md
sabinebertram: As @adrianhopebailie suggested, I'll remove this whole part and only link the SPSP spec to not have 2 places to add updates in the future.

0036-spsp-pull-payments/0036-spsp-pull-payments.md
sabinebertram: I'll change that in the basic SPSP spec. Are these a MUST for both bullet points?

0036-spsp-pull-payments/0036-spsp-pull-payments.md
sabinebertram: The server opens connection and stream. Thanks for catching that.

0036-spsp-pull-payments/0036-spsp-pull-payments.md
sabinebertram: I was thinking of the token kind of like an id for the pull agreement with an expiry time. As long as you have it and it has not expired, you can use it to open a new connection and pull. If the endpoint is in the money, you will actually pull value, otherwise you pull 0. 
But maybe that is too insecure?

0036-spsp-pull-payments/0036-spsp-pull-payments.md
sabinebertram: Well, I specified it to be denominated in agreement asset (see line 106-108) but I guess you can argue about that. For it to be denoted in the server's asset, there would have to be another response field which includes this information.

0036-spsp-pull-payments/0036-spsp-pull-payments.md
sabinebertram: I'll define it.

0036-spsp-pull-payments/0036-spsp-pull-payments.md
```diff
@@ -0,0 +1,140 @@
...
+* **Pull Payment Agreement** - An agreement between the Payer and the Payee that specifies how much value can be pulled how often and for how long. 
+* **Token** - An opaque string that represents an identifier for the Pull Payment Agreement. It is part of the [Payment Pointer](../0026-payment-pointers/0026-payment-pointers.md) the SPSP Client needs to connect to the SPSP Server. 
+
+### Operation Overview
+
+The Payee's SPSP Client will set up a [STREAM](../0029-stream/0029-stream.md) connection to the Payer's SPSP Server using the Token, as described by the [Simple Payment Setup Protocol](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md). On connection, the SPSP Client tries to pull as much value from the SPSP server as possible given the Pull Payment Agreement.
+
+## Model of Operation
+
+### Negotiating the Pull Payment Agreement
```
wilsonianb: _forgot to click add review comment_
I'm on the fence deciding if this should be described as a negotiation. With Codius, the user asks host for the price. The host provides the amount and interval at which they would pull/charge. The user either sets the details and token (because they accept the price) or does nothing (because they don't).
What if this section just describes that the Payer sets the agreement/schedule/rules (with or without input from the Payee) and creates the corresponding token, etc?

0036-spsp-pull-payments/0036-spsp-pull-payments.md
```diff
@@ -0,0 +1,140 @@
...
+
+### Scope
+
+This document specifies how to conduct a Pull Payment once a STREAM connection has been set up. It is intended for use by end-user applications.
+
+### Definitions
+* **Payer** - The entity that units of value are pulled from by the Payee. It is running the [SPSP Server](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md#Definitions).
+* **Payee** - The entity that is pulling units of value from the Payer. It is running the [SPSP Client](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md#Definitions).
+* **Pull Payment** - The process of pulling units of value from one account into another account.
+* **Pull Payment Agreement** - An agreement between the Payer and the Payee that specifies how much value can be pulled how often and for how long.
```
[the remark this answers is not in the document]
sabinebertram: Well, it also includes price, asset code, and scale. Thus, there are not just timing variables included. 
However, I don't have strong feelings about the term `Pull Payment Agreement`.

0036-spsp-pull-payments/0036-spsp-pull-payments.md
```diff
@@ -0,0 +1,140 @@
...
+* **Payee** - The entity that is pulling units of value from the Payer. It is running the [SPSP Client](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md#Definitions).
+* **Pull Payment** - The process of pulling units of value from one account into another account.
+* **Pull Payment Agreement** - An agreement between the Payer and the Payee that specifies how much value can be pulled how often and for how long. 
+* **Token** - An opaque string that represents an identifier for the Pull Payment Agreement. It is part of the [Payment Pointer](../0026-payment-pointers/0026-payment-pointers.md) the SPSP Client needs to connect to the SPSP Server. 
+
+### Operation Overview
+
+The Payee's SPSP Client will set up a [STREAM](../0029-stream/0029-stream.md) connection to the Payer's SPSP Server using the Token, as described by the [Simple Payment Setup Protocol](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md). On connection, the SPSP Client tries to pull as much value from the SPSP server as possible given the Pull Payment Agreement.
```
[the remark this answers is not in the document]
sabinebertram: You are right. I'll rephrase that saying something like 

> On connection, the SPSP Client tries to pull value from the SPSP server, staying within the terms of the Pull Payment Agreement.

Feel free to suggest something better than that.

0036-spsp-pull-payments/0036-spsp-pull-payments.md
sabinebertram: I don't see a reason why there could not be more of a negotiation. It is just very simple in the Codius use case. In other use cases, it could be something like 
* the merchant suggests an asset 
* the user does not like that and proposes another asset
* the merchant agrees
* the user creates the token
In essence, however, we may just need another spec for the "negotiation" or however we want to call it.

0036-spsp-pull-payments/0036-spsp-pull-payments.md
```diff
@@ -0,0 +1,140 @@
...
+#### Response
+
+``` http
+HTTP/1.1 200 OK
+Content-Type: application/spsp4+json
+
+{
+  "destination_account": "alice.ilpdemo.red~0f09dc92-84ad-401b-a7c9-441bc6173f4e",
+  "shared_secret": "k5nubgM6zpb88NPGVnI/tVjRdgpUh+JvMueRFEMvPcY=",
+  "balance": {
```
sharafian: how would you feel about putting all the fields (`balance`, `cycle`, and `agreement`) under a `pull` field? There's no field that says the type of payment this is, so someone might get confused deducing between an invoice and a pull payment if they both have balance fields or something

0036-spsp-pull-payments/0036-spsp-pull-payments.md
```diff
@@ -0,0 +1,140 @@
...
+
+Implementations MAY try to parse the Token, however, it MUST support totally opaque tokens. 
+
+### Conducting the Pull Payment
+
+The Payee's SPSP Client opens a [STREAM](../0029-stream/0029-stream.md) connection to the Payer's SPSP Server as described in the [Simple Payment Setup Protocol](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md). Once this connection is established, the process continues as follows: 
+
+The SPSP Server begins sending ILP packets to complete the Pull Payment.
+  1. The SPSP Server will adjust their `sendMax` to reflect the amount they're willing to send.
+      * `sendMax` SHOULD be derived from the Pull Payment Agreement parameters.
```
sharafian: It's worth specifying how we derive the sendMax

0036-spsp-pull-payments/0036-spsp-pull-payments.md
```diff
@@ -0,0 +1,140 @@
...
+  2. The SPSP Client will adjust their `receiveMax` to reflect the amount they're willing to receive.
+      * `receiveMax` MAY be `Infinity`.
+  3. The SPSP Client's and Server's [STREAM Modules](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md#Definitions) will move as much value as possible while staying inside these bounds.
```
sharafian: at what point is the `balance` updated? If you query mid-way through pulling will the balance be accurate or will it only update after the stream is closed?

0036-spsp-pull-payments/0036-spsp-pull-payments.md
```diff
@@ -0,0 +1,140 @@
...
+  4. If the SPSP Server reaches their `sendMax`, they end the stream and the connection. If the SPSP Client reaches their `receiveMax`, they will end the stream and the connection.
```
sharafian: How does this behave when there's multiple open connections pulling?

0036-spsp-pull-payments/0036-spsp-pull-payments.md
```diff
@@ -0,0 +1,140 @@
...
+### Scope
+
+This document specifies how to conduct a Pull Payment once a STREAM connection has been set up. It is intended for use by end-user applications.
+
+### Definitions
+* **Payer** - The entity that units of value are pulled from by the Payee. It is running the [SPSP Server](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md#Definitions).
+* **Payee** - The entity that is pulling units of value from the Payer. It is running the [SPSP Client](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md#Definitions).
+* **Pull Payment** - The process of pulling units of value from one account into another account.
+* **Pull Payment Agreement** - An agreement between the Payer and the Payee that specifies how much value can be pulled how often and for how long. 
+* **Token** - An opaque string that represents an identifier for the Pull Payment Agreement. It is part of the [Payment Pointer](../0026-payment-pointers/0026-payment-pointers.md) the SPSP Client needs to connect to the SPSP Server.
```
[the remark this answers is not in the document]
sabinebertram: So you are suggesting(?):
**Token** - A [Payment Pointer](../0026-payment-pointers/0026-payment-pointers.md) that includes an opaque string representing the identifier and for the Pull Payment Agreement and the authorization to pull.

0036-spsp-pull-payments/0036-spsp-pull-payments.md
sabinebertram: I'm ok with doing that as long as we do the same for invoices, i.e. the balance moves within the invoice object.

0036-spsp-pull-payments/0036-spsp-pull-payments.md
sabinebertram: Is that a spec or implementation detail? The balance is an optional field so it may not be included in the query response.

0036-spsp-pull-payments/0036-spsp-pull-payments.md
sabinebertram: Pulling from the same endpoint? Should that be allowed?

sharafian: if it's not allowed we should say so. I think it would be nice if it were but I can't think of how we can make it work

0036-spsp-pull-payments/0036-spsp-pull-payments.md
```diff
@@ -0,0 +1,144 @@
...
+| `pull.agreement.asset` | Object | _(OPTIONAL)_ Details of the agreement's asset. |
+| `pull.agreement.asset.code` | String |  _(OPTIONAL)_ Asset code to identify the agreement's currency. Currencies that have [ISO 4217](https://en.wikipedia.org/wiki/ISO_4217) codes should use those. |
+| `pull.agreement.asset.scale` | Integer |  _(OPTIONAL)_ The scale of the amounts `pull.balance.total`, `pull.balance.interval`, and `pull.balance.available` are stated in (e.g. an amount of `"1000"` with a scale of `2` translates to `10.00` units of the server's asset/currency) |
+
+**Note:** Currency amounts are denominated as integer strings instead of native JSON numbers to avoid losing precision during JSON parsing. Applications MUST represent these numbers in a data type that has precision equal or greater than an unsigned 64-bit integer.
+
+##### Errors
+
+###### pointer Does Not Exist
```
wilsonianb:
```suggestion
###### Pointer Does Not Exist
```

0036-spsp-pull-payments/0036-spsp-pull-payments.md
wilsonianb: Should the server also only allow one stream at a time?

sabinebertram: I don't know if that works when the server is hosted by a wallet provider. Multiple entities may want to pull from different endpoints at the same time.

wilsonianb: It seems like the SPSP server (wallet provider or otherwise) should only allow a single connection and stream at a time to each pull payment pointer.

sharafian:
This looks good to me once we clarify that you can only have one open connection/stream per pull pointer

sabinebertram:
Where is the best place to say that there should only be one connection per pull pointer at a time?

sharafian:
Probably "The Payee's SPSP Client opens a [STREAM](../0029-stream/0029-stream.md) connection to the Payer's SPSP Server" we could say that if there's already an open connection to that pay pointer (have to be careful that it's one per pointer not one per server) then the connection should be closed

0036-spsp-pull-payments/0036-spsp-pull-payments.md
```diff
@@ -0,0 +1,146 @@
...
+
+Prior to a Pull Payment, the Payer and the Payee have to negotiate a Pull Payment Agreement. This agreement SHOULD be comprised of the parameters defined in the `agreement`-object within the [Response Body](#Response-Body). The Payer's SPSP Server SHOULD store the negotiated Pull Payment Agreement parameters.
+
+The result of the negotiation is a Pull Payment Pointer representing the Pull Payment Agreement. The Pull Payment Pointer is required by the Payee's SPSP Client to connect to the Payer's SPSP server. 
+
+Implementations MAY try to parse the Pull Payment Pointer, however, it MUST support totally opaque Pull Payment Pointers. 
+
+### Conducting the Pull Payment
+
+The Payee's SPSP Client opens a [STREAM](../0029-stream/0029-stream.md) connection to the Payer's SPSP Server as described in the [Simple Payment Setup Protocol](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md). Note that if there is already an existing open STREAM connection to the `destination_account` the Pull Payment Pointer is resolving to, the connection SHOULD be closed.
```
wilsonianb: Should this also say that the server should only allow one stream at a time?

sabinebertram: You mean specifically saying that? I.e.
> Note that if there is already an existing open STREAM connection to the `destination_account` the Pull Payment Pointer is resolving to, the connection SHOULD be closed. The server SHOULD only allow for one STREAM at a time.

wilsonianb: :+1: 
`STREAM` -> `stream`

0036-spsp-pull-payments/0036-spsp-pull-payments.md
```diff
@@ -0,0 +1,146 @@
...
+
+The response body is a JSON object that includes basic account details necessary for setting up a STREAM connection as well as optional parameters for displaying purposes. The fields are defined in the following: 
+
+| Field | Type | Description |
+|---|---|---|
+| `destination_account` | [ILP Address](../0015-ilp-addresses/0015-ilp-addresses.md) | (see [Simple Payment Setup Protocol](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md#Response-Body)) |
+| `shared_secret` | 32 bytes, [base64 encoded](https://en.wikipedia.org/wiki/Base64) (including padding) | (see [Simple Payment Setup Protocol](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md#Response-Body)) |
+| `pull`  | Object | _(OPTIONAL)_ Details of this Pull Payment Pointer. |
+| `pull.balance`  | Object | _(OPTIONAL)_ Details of this Pull Payment Pointer's balance. |
+| `pull.balance.total` | Integer String | _(OPTIONAL)_ Total amount, denoted in `pull.agreement.asset.code`, which has been pulled by the client since `pull.agreement.start` to date. It is the sum of all outgoing chunks. |
```
wilsonianb:
```suggestion
| `pull.balance.total` | Integer String | _(OPTIONAL)_ Total amount, denoted in `pull.agreement.asset`, which has been pulled by the client since `pull.agreement.start` to date. It is the sum of all outgoing chunks. |
```
or some other way of denoting both the code AND scale.
Ditto for

- `pull.balance.interval`
- `pull.balance.available`
- `pull.agreement.amount`

0036-spsp-pull-payments/0036-spsp-pull-payments.md
wilsonianb: Actually... after just trying to implement that limit, I think it is safe to have multiple connections and streams.
The SPSP server should only adjust their `sendMax` (step 1 above) for a single open stream at a time per pull payment pointer.
(over) simple example:
https://github.com/sabinebertram/ilp-spsp-pull-payment-server/commit/a58afe3470c926cb7c0ea1417e887e9cd76735ad

sabinebertram: Do you mean it is ok to have multiple connections and streams per server or per pull pointer?

0036-spsp-pull-payments/0036-spsp-pull-payments.md
sabinebertram: Basically, you want less objects and more individual fields with precise names? I'm open for suggestions.

0036-spsp-pull-payments/0036-spsp-pull-payments.md
wilsonianb: It is safe to have multiple connections and streams per pull pointer, but it SHOULD (MUST?) only set `sendMax` for one stream at a time per pull pointer.

sabinebertram: I would say MUST. I just don't see what we would gain by allowing multiple connections per pull pointer at a time.

wilsonianb: The pull server probably doesn't need to allow multiple connections or streams (unless maybe there's also data streaming happening), but I don't think this needs to comment on it since the basic concern is to not set `sendMax` on multiple streams at a time (for a single pull pointer).

0036-spsp-pull-payments/0036-spsp-pull-payments.md
```diff
@@ -0,0 +1,148 @@
...
+  1. The SPSP Server will adjust their `sendMax` to reflect the amount they're willing to send.
+      * `sendMax` SHOULD be derived from the Pull Payment Agreement parameters, i.e., `sendMax` SHOULD be equal to `pull.balance.available`, converted to the SPSP Server's operating asset, taking exchange rate fluctuations into account.
+  2. The SPSP Client will adjust their `receiveMax` to reflect the amount they're willing to receive.
+      * `receiveMax` MAY be `Infinity`.
+  3. The SPSP Client's and Server's [STREAM Modules](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md#Definitions) will move as much value as possible while staying inside these bounds.
+  4. If the SPSP Server reaches their `sendMax`, they end the stream and the connection. If the SPSP Client reaches their `receiveMax`, they will end the stream and the connection.
+
+The STREAM parameters - `sendMax` and `receiveMax` - are defined in [STREAM's frame encoding](../0029-stream/0029-stream.md#53-frames).
+
+Note that if there are multiple open STREAM connections to the `destination_account` the Pull Payment Pointer is resolving to, the SPSP Server MUST only set one `sendMax` at a time.
```
wilsonianb:
```suggestion
Note that if there are multiple open streams (via one or more STREAM connections) to the `destination_account` the Pull Payment Pointer is resolving to, the SPSP Server MUST only set one `sendMax` at a time.
```
Thinking about this some more, you could technically set lower `sendMax`'s across multiple simultaneous streams and make sure the sum of them doesn't exceed the amount you're willing to send.
But I think this is fine as is since we're already saying this above:
>The SPSP Server will adjust their `sendMax` to reflect the amount they're willing to send.

PR #516, opened 2019-03-01
Ping Protocol

sappenin:
Following [discussion in the ILP Forum here](https://forum.interledger.org/t/document-the-ping-protocol/279), this PR introduces the **Ping Protocol**, which specifies a mechanism for Interledger nodes to test uptime and fulfill/reject response-time for a particular payment path, from the perspective of a payment sender.

### DIFF
--- a/0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
+++ b/0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
@@ -1,7 +1,6 @@
 # Interledger Ping Protocol
 
 ## Introduction
-
 In order to monitor the health of an internetworked system, it is often useful to be able to send probe messages. In the Internet system, ICMP [echo](https://en.wikipedia.org/wiki/Ping_(networking_utility)#Echo_request) request and response packets are used together to verify connectivity (colloquially known as a "pinging" a host).  
 
 This RFC specifies a similar mechanism for Interledger nodes to test `uptime` and fulfill/reject `response-time` for a particular payment path, from the perspective of a payment sender.
@@ -23,25 +22,21 @@ All Interledger implementations _SHOULD_ respond to ping protocol requests, unle
 * The **Recipient** is the Interledger node responding to the ping request.
 * **Source amount** is the amount debited from the sender of an ILP payment.
 * **Destination amount** is the amount credited to the recipient of an ILP payment.
-* The **Known Preimage** is the [ASCII](https://tools.ietf.org/html/rfc20) string `pingpingpingpingpingpingpingping`, which in Base64 is `cGluZ3BpbmdwaW5ncGluZ3BpbmdwaW5ncGluZ3Bpbmc`. 
-* The **Known Condition** is the SHA256 hash of the Known Preimage, which when encoded using Base64 is `jAC8DGFPZPfh4AtZpXuvXFe2oRmpDVSvSJg2oT+bx34=`*[]: 
+* The **Known Preimage** is the [ASCII](https://tools.ietf.org/html/rfc20) string `pingpingpingpingpingpingpingping`, which in Base64 is `cGluZ3BpbmdwaW5ncGluZ3BpbmdwaW5ncGluZ3Bpbmc=`. 
+* The **Known Condition** is the SHA256 hash of the Known Preimage, which when encoded using Base64 is `jAC8DGFPZPfh4AtZpXuvXFe2oRmpDVSvSJg2oT+bx34=` 
 
 ## Protocol
 1. Sender sends a Prepare packet using the following details:
    - Amount: Any amount chosen by the sender.
    - Condition: The Known Condition, as defined above.
    - Destination ILP address: Recipient's Interledger address.
-   - Packet Data: Unspecified.
   
-2. Upon receiving the Prepare packet, the Recipient _SHOULD fulfill the payment using the following information:
+2. Upon receiving the Prepare packet, the Recipient MUST fulfill the payment using the following information:
    - Fulfillment: The Known Fulfillment, as defined above.
-   - Packet Data: Unspecified.
 
 Note that a recipient _MAY_ reject the payment if appropriate, for example due to an insufficient amount, invalid expiry, or for other factors.
-
  
 ## Recommended Uses
-
 * Interledger participants may use this mechanism to test their own connectivity by sending pings to one or more destinations that are known to support the protocol described in this document.
 
 * A monitoring service that is testing the availability of different ILP addresses may also send pings to these destinations from various sources. Interledger nodes that wish to be listed in such a monitoring service must support the protocol described in this document.
--- a/0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
+++ b/0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
@@ -8,7 +8,7 @@ This RFC specifies a similar mechanism for Interledger nodes to test `uptime` an
 ## Scope
 It is already possible to test unidirectional, end-to-end connectivity in Interledger by sending a small payment using existing mechanisms. However, doing so requires higher-level protocols, such as [SPSP](https://github.com/interledger/rfcs/blob/master/0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md) and [STREAM](https://github.com/interledger/rfcs/blob/master/0029-stream/0029-stream.md).
 
-Instead, this protocol uses only [Interledger-layer](https://github.com/interledger/rfcs/blob/master/0001-interledger-architecture/0001-interledger-architecture.md#interledger-protocol) technologies, plug a a known condition/fulfillment pair, thus allowing a sender to control the expected response type while at the same time preventing a sender from utilizing bandwidth without paying for it
+Instead, this protocol uses only [Interledger-layer](https://github.com/interledger/rfcs/blob/master/0001-interledger-architecture/0001-interledger-architecture.md#interledger-protocol) technologies, plus a known condition/fulfillment pair, thus allowing a sender to control the expected response type while at the same time preventing a sender from utilizing bandwidth without paying for it
 
 The ping protocol is not designed to debug routing issues and does not provide additional diagnostic information about the state of routing. That use case would be better served by a separate `traceroute` mechanism.
 
--- a/0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
+++ b/0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
@@ -3,7 +3,7 @@
 ## Introduction
 In order to monitor the health of an internetworked system, it is often useful to be able to send probe messages. In the Internet system, ICMP [echo](https://en.wikipedia.org/wiki/Ping_(networking_utility)#Echo_request) request and response packets are used together to verify connectivity (colloquially known as a "pinging" a host).  
 
-This RFC specifies a similar mechanism for Interledger nodes to test `uptime`, `response-time` and approximate `path cost` for a particular sender/Receiver pair.
+This RFC specifies a similar mechanism for Interledger nodes to test `uptime`, `response-time` and approximate `path cost` for a particular Initiator/Recipient pair.
 
 ## Scope
 It is already possible to test end-to-end connectivity in Interledger by sending a small payment using existing mechanisms. However, doing so requires higher-level protocols, such as [SPSP](https://github.com/interledger/rfcs/blob/master/0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md) and [STREAM](https://github.com/interledger/rfcs/blob/master/0029-stream/0029-stream.md).
--- a/0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
+++ b/0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
@@ -50,7 +50,7 @@ Note that a Recipient _MAY_ reject the payment if appropriate, for example due t
 
 ![bidirectional-flow](images/bidirectional-flow.svg)
 
-1. Initiator generates a new random 32-byte value to use as a Fulfilment (`F`) and generates the corresponding condition (`C`).
+1. Initiator generates a new random 32-byte value to use as a fulfilment (`F`) and generates the corresponding condition (`C`).
 2. Initiator sends an ILP Prepare packet to the Recipient with the following details (1):
    - **Destination**: The Recipient's Interledger address.
    - **Amount**: Any amount chosen by the Initiator.
--- a/0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
+++ b/0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
@@ -29,7 +29,7 @@ The primary difference between the two modes is that unidirectional mode allows 
 * **Source amount** is the amount debited from the sender of an ILP payment.
 * **Destination amount** is the amount credited to the Recipient of an ILP payment.
 * The **Ping Protocol Fulfillment** is the [ASCII](https://tools.ietf.org/html/rfc20) string `pingpingpingpingpingpingpingping`, which is `0x70696E6770696E6770696E6770696E6770696E6770696E6770696E6770696E67` in hexadecimal encoding. This value is used in Unidirectional mode only. 
-* The **Known Condition** is the SHA256 hash of the Known Preimage, which when encoded using Base64 is `jAC8DGFPZPfh4AtZpXuvXFe2oRmpDVSvSJg2oT+bx34=`. This value is used in Unidirectional mode only.
+* The **Ping Protocol Condition** is the SHA256 hash of the Ping Protocol Fulfillment, which is `0x8c00bc0c614f64f7e1e00b59a57baf5c57b6a119a90d54af489836a13f9bc77e` in hexadecimal encoding. This value is used in Unidirectional mode only.
 
 ## Unidirectional Mode
 
--- a/0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
+++ b/0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
@@ -3,40 +3,76 @@
 ## Introduction
 In order to monitor the health of an internetworked system, it is often useful to be able to send probe messages. In the Internet system, ICMP [echo](https://en.wikipedia.org/wiki/Ping_(networking_utility)#Echo_request) request and response packets are used together to verify connectivity (colloquially known as a "pinging" a host).  
 
-This RFC specifies a similar mechanism for Interledger nodes to test `uptime` and fulfill/reject `response-time` for a particular payment path, from the perspective of a payment sender.
+This RFC specifies a similar mechanism for Interledger nodes to test `uptime`, `response-time` and approximate `path cost` for a particular sender/Receiver pair.
 
 ## Scope
 It is already possible to test unidirectional, end-to-end connectivity in Interledger by sending a small payment using existing mechanisms. However, doing so requires higher-level protocols, such as [SPSP](https://github.com/interledger/rfcs/blob/master/0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md) and [STREAM](https://github.com/interledger/rfcs/blob/master/0029-stream/0029-stream.md).
 
-Instead, this protocol uses only [Interledger-layer](https://github.com/interledger/rfcs/blob/master/0001-interledger-architecture/0001-interledger-architecture.md#interledger-protocol) technologies, plus a known condition/fulfillment pair, thus allowing a sender to control the expected response type while at the same time preventing a sender from utilizing bandwidth without paying for it.
-
+Instead, this protocol uses only [Interledger-layer](https://github.com/interledger/rfcs/blob/master/0001-interledger-architecture/0001-interledger-architecture.md#interledger-protocol) technologies, plus a known condition/fulfillment pair, to test connectivity between two nodes in the Interledger.
+ 
 The ping protocol is not designed to debug routing issues and does not provide additional diagnostic information about the state of routing. That use case would be better served by a separate `traceroute` mechanism.
-
-Additionally, the ping protocol is not designed to test bi-directional connectivity. For example, when a node encounters a ping protocol request, it either fulfills or rejects the payment, but does not perform any kind of "mirror payment" in response. That use case would be better served by a separate `echo` mechanism. 
 
 All Interledger implementations _SHOULD_ respond to ping protocol requests, unless a node has a specific reason not to, such as security or privacy concerns.
 
+## Overview
+This RFC defines two ping protocol modes: unidirectional and bidirectional.
+
+* **Unidirectional**: A ping Initiator sends a specially formed ILP Prepare packet, and the Receiver either fulfills or rejects the packet according to the rules defined in [Unidirectional Mode](#Unidirectional Mode).
+* **Bidirectional**: An Initiator sends a specially formed "ping" packet that prompts the Receiver to send an additional "pong" packet. Upon receiving this pong packet, the Initiator either fulfills or rejects according to the rules defined in [Bidirectional Mode](#Bidirectional Mode), thus prompting the Receiver to fulfill or reject the original ping packet.
+
+The primary difference between the two modes is that unidirectional mode allows an Initiator to test connectivity from the Initiator to the Recipient only, whereas bidirectional mode allows the Initiator to test connectivity in both directions.
+
 ## Terminology
 
-* The **Sender** is the Interledger node initiating the ping request.
-* The **Recipient** is the Interledger node responding to the ping request.
+* The **Initiator** is the Interledger node initiating the ping request.
+* The **Recipient** is the Interledger node responding to the Initiator's request.
 * **Source amount** is the amount debited from the sender of an ILP payment.
-* **Destination amount** is the amount credited to the recipient of an ILP payment.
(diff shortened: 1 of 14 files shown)


0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
```diff
@@ -0,0 +1,42 @@
...
+
+This RFC specifies a similar mechanism for Interledger nodes to test `uptime` and fulfill/reject `response-time` for a particular payment path, from the perspective of a payment sender.
+
+## Scope
+It is already possible to test unidirectional, end-to-end connectivity in Interledger by sending a small payment using existing mechanisms. However, doing so requires higher-level protocols, such as [SPSP](https://github.com/interledger/rfcs/blob/master/0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md) and [STREAM](https://github.com/interledger/rfcs/blob/master/0029-stream/0029-stream.md).
+
+Instead, this protocol uses only [Interledger-layer](https://github.com/interledger/rfcs/blob/master/0001-interledger-architecture/0001-interledger-architecture.md#interledger-protocol) technologies, plug a a known condition/fulfillment pair, thus allowing a sender to control the expected response type while at the same time preventing a sender from utilizing bandwidth without paying for it
```
dora-gt:
```suggestion
Instead, this protocol uses only [Interledger-layer](https://github.com/interledger/rfcs/blob/master/0001-interledger-architecture/0001-interledger-architecture.md#interledger-protocol) technologies, plug a a known condition/fulfillment pair, thus allowing a sender to control the expected response type while at the same time preventing a sender from utilizing bandwidth without paying for it.
```

Could you explain concretely the meaning of `preventing a sender from utilizing bandwidth without paying for it`?

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
```diff
@@ -0,0 +1,42 @@
...
+
+## Protocol
+1. Sender sends a Prepare packet using the following details:
+   - Amount: Any amount chosen by the sender.
+   - Condition: The Known Condition, as defined above.
+   - Destination ILP address: Recipient's Interledger address.
+  
+2. Upon receiving the Prepare packet, the Recipient MUST fulfill the payment using the following information:
+   - Fulfillment: The Known Fulfillment, as defined above.
+
```
dora-gt: Could you describe the behavior of intermediary nodes?

sappenin:
Bump

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
```diff
@@ -0,0 +1,42 @@
...
+This RFC specifies a similar mechanism for Interledger nodes to test `uptime` and fulfill/reject `response-time` for a particular payment path, from the perspective of a payment sender.
+
+## Scope
+It is already possible to test unidirectional, end-to-end connectivity in Interledger by sending a small payment using existing mechanisms. However, doing so requires higher-level protocols, such as [SPSP](https://github.com/interledger/rfcs/blob/master/0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md) and [STREAM](https://github.com/interledger/rfcs/blob/master/0029-stream/0029-stream.md).
+
+Instead, this protocol uses only [Interledger-layer](https://github.com/interledger/rfcs/blob/master/0001-interledger-architecture/0001-interledger-architecture.md#interledger-protocol) technologies, plus a known condition/fulfillment pair, thus allowing a sender to control the expected response type while at the same time preventing a sender from utilizing bandwidth without paying for it
```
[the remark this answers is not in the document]
sappenin: > What do you mean by this: "thus allowing a sender to control the expected response type"

Just that the sender of a ping request can control whether they get a `FULFILL` or a `REJECT` response back using the known condition/fulfillment pair. This was just on extra way of rationalizing why this protocol has a known fulfillment (as opposed to instead just using an unfulfillable request).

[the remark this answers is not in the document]
sappenin: > Also, nit: missing a period at the end of the paragraph.

Fixed.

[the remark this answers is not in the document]
sappenin: > There's no way to stop senders from sending unfulfillable packets (because you can't distinguish a real condition from a random one)

Yeah, that's not what I meant, but this is probably poorly worded ATM - In the `echo` protocol, an initiator (i.e., sender) can trigger outgoing mirror payments from the receiver, which could take-up bandwidth (but only in extreme cases). In the `ping` protocol, that can't happen.

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
```diff
@@ -0,0 +1,78 @@
...
+The primary difference between the two modes is that unidirectional mode allows an Initiator to test connectivity from the Initiator to the Recipient only, whereas bidirectional mode allows the Initiator to test connectivity in both directions.
+
+## Terminology
+
+* The **Initiator** is the Interledger node initiating the ping request.
+* The **Recipient** is the Interledger node responding to the Initiator's request.
+* **Source amount** is the amount debited from the sender of an ILP payment.
+* **Destination amount** is the amount credited to the Recipient of an ILP payment.
+* The **Known Preimage** is the [ASCII](https://tools.ietf.org/html/rfc20) string `pingpingpingpingpingpingpingping`, which in binary is `70696E6770696E6770696E6770696E6770696E6770696E6770696E6770696E67`. This value is used in Unidirectional mode only. 
+* The **Known Condition** is the SHA256 hash of the Known Preimage, which when encoded using Base64 is `jAC8DGFPZPfh4AtZpXuvXFe2oRmpDVSvSJg2oT+bx34=`. This value is used in Unidirectional mode only.
```
emschwartz:
```suggestion
* The **Ping Protocol Condition** is the SHA256 hash of the Ping Protocol Fulfillment, which is `0x8c00bc0c614f64f7e1e00b59a57baf5c57b6a119a90d54af489836a13f9bc77e` in hexadecimal encoding. This value is used in Unidirectional mode only.
```

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
```diff
@@ -0,0 +1,78 @@
...
+
+The primary difference between the two modes is that unidirectional mode allows an Initiator to test connectivity from the Initiator to the Recipient only, whereas bidirectional mode allows the Initiator to test connectivity in both directions.
+
+## Terminology
+
+* The **Initiator** is the Interledger node initiating the ping request.
+* The **Recipient** is the Interledger node responding to the Initiator's request.
+* **Source amount** is the amount debited from the sender of an ILP payment.
+* **Destination amount** is the amount credited to the Recipient of an ILP payment.
+* The **Known Preimage** is the [ASCII](https://tools.ietf.org/html/rfc20) string `pingpingpingpingpingpingpingping`, which in binary is `70696E6770696E6770696E6770696E6770696E6770696E6770696E6770696E67`. This value is used in Unidirectional mode only.
```
emschwartz:
```suggestion
* The **Ping Protocol Fulfillment** is the [ASCII](https://tools.ietf.org/html/rfc20) string `pingpingpingpingpingpingpingping`, which is `0x70696E6770696E6770696E6770696E6770696E6770696E6770696E6770696E67` in hexadecimal encoding. This value is used in Unidirectional mode only. 
```

emschwartz: If this is only used in Unidirectional mode, it should go in that section rather than in the terminology section

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
```diff
@@ -0,0 +1,78 @@
...
+* **Destination amount** is the amount credited to the Recipient of an ILP payment.
+* The **Known Preimage** is the [ASCII](https://tools.ietf.org/html/rfc20) string `pingpingpingpingpingpingpingping`, which in binary is `70696E6770696E6770696E6770696E6770696E6770696E6770696E6770696E67`. This value is used in Unidirectional mode only. 
+* The **Known Condition** is the SHA256 hash of the Known Preimage, which when encoded using Base64 is `jAC8DGFPZPfh4AtZpXuvXFe2oRmpDVSvSJg2oT+bx34=`. This value is used in Unidirectional mode only.
+
+## Unidirectional Mode
+
+1. Initiator sends a ILP Prepare packet using the following details:
+   - **Destination**: Recipient's Interledger address.
+   - **Amount**: Any amount chosen by the Initiator.
+   - **Expiry**: Any appropriate expiry.
```
emschwartz: Want to give a recommendation here? If I were implementing this, I might not know what an appropriate expiry might be

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
```diff
@@ -0,0 +1,78 @@
...
+   - **Condition**: The Known Condition, as defined in [Terminology](#terminology).
+   
+2. Upon receiving the Prepare packet, the Recipient MUST fulfill the payment using the following information:
+   - **Fulfillment**: The Known Fulfillment, as defined above.
+
+Note that a Recipient _MAY_ reject the payment if appropriate, for example due to an insufficient amount, invalid expiry, or for other factors.
```
emschwartz: Should they use any specific errors? How should the initiator interpret the different possible errors?

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Bidirectional Mode
+
+1. Initiator generates a new random 32-byte value to use as a Fulfilment (`F`) and generates the corresponding condition (`C`).
```
emschwartz:
```suggestion
1. Initiator generates a new random 32-byte value to use as a fulfilment (`F`) and generates the corresponding condition (`C`).
```

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
```diff
@@ -0,0 +1,78 @@
...
+2. Initiator sends an ILP Prepare packet to the Recipient with the following details:
+   - **Destination**: The Recipient's Interledger address.
+   - **Amount**: Any amount chosen by the Initiator.
```
emschwartz: Can you describe how the initiator would choose the amount? Should it be zero? Should it be non-zero? Are they going to get that money back?

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
```diff
@@ -0,0 +1,78 @@
...
+   - **Expiry**: An appropriate expiry that is long enough to allow for two round trips.
+   - **Condition**: The condition `C`.
+   - **Data**: A concatenated series of bytes with the following information:
+      - the bytes of the ASCII string `ECHOECHOECHOECHO`. In binary this value is `0x4543484F4543484F4543484F4543484F`.
```
emschwartz:
```suggestion
      - the bytes of the ASCII string `ECHOECHOECHOECHO`. In hexadecimal this value is `0x4543484F4543484F4543484F4543484F`.
```

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
```diff
@@ -0,0 +1,78 @@
...
+      - the byte `0x00`
+      - The ILP address of the Initiator (i.e., the return address) as an OER-encoded, variable length IA5 string.
```
emschwartz: Maybe link to the OER spec here

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
```diff
@@ -0,0 +1,78 @@
...
+4. Upon receiving the Prepare packet, the Recipient identifies that it is a bidirectional Ping request by confirming the following:
+  1. The packet is addressed directly to itself (i.e. it has no tx or child suffix).
```
emschwartz: What's a tx suffix? It might be worth giving an example of what this means, along the lines of: "for example, a node with the address `example.node1` would respond to ping packets addressed to `example.node1` but **not** those addressed to `example.node1.child`"

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
```diff
@@ -0,0 +1,78 @@
...
+  1. The packet has a payload conforming to the specification above (e.g., the byte following the `ECHOECHOECHOECHO` prefix in the payload is `0x00`).
+5. The Recipient DOES NOT immediately send an ILP Fulfill response. Instead the Recipient sends a new ILP `Prepare` packet addressed to the address parsed from the original original packet's payload. This new packet has following details:original
```
emschwartz:
```suggestion
5. The Recipient DOES NOT immediately send an ILP Fulfill response. Instead the Recipient sends a new ILP `Prepare` packet addressed to the address parsed from the original original packet's payload. This new packet has following details: 
```

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
```diff
@@ -0,0 +1,78 @@
...
+   - **Data**: A concatenated series of bytes with the following information:
+      - the bytes of the ASCII string `ECHOECHOECHOECHO`. In binary this value is `0x4543484F4543484F4543484F4543484F`.
+      - the byte `0x00`
+      - The ILP address of the Initiator (i.e., the return address) as an OER-encoded, variable length IA5 string.
+4. Upon receiving the Prepare packet, the Recipient identifies that it is a bidirectional Ping request by confirming the following:
+  1. The packet is addressed directly to itself (i.e. it has no tx or child suffix).
+  1. The packet has a payload conforming to the specification above (e.g., the byte following the `ECHOECHOECHOECHO` prefix in the payload is `0x00`).
+5. The Recipient DOES NOT immediately send an ILP Fulfill response. Instead the Recipient sends a new ILP `Prepare` packet addressed to the address parsed from the original original packet's payload. This new packet has following details:original 
+    - *Destination*: The Interledger address found in the original packet's data payload.
+    - *Amount*: Any amount chosen by the Initiator.
```
emschwartz: Is this amount chosen by the Initiator or does the recipient apply some kind of rate to this?

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
```diff
@@ -0,0 +1,78 @@
...
+    - *Expiry*: An expiry that is marginally smaller than the expiry on the original ILP Prepare.
+    - *Condition*: The same condition `C`.
+    - *Data*: A concatenated series of bytes with the following information:
+      - the bytes of the ASCII string `ECHOECHOECHOECHO` (i.e., `0x4543484F4543484F4543484F4543484F`).
```
emschwartz:
```suggestion
      - the bytes of the ASCII string `ECHOECHOECHOECHO` (`0x4543484F4543484F4543484F4543484F` in hexadecimal).
```

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
```diff
@@ -0,0 +1,78 @@
...
+      - the byte `0x01`
+7. Upon receiving this packet, the Initiator identifies that it is a Pong by the fact that it is addressed to the address used in the original Ping request.
```
emschwartz: Don't they identify that it's a pong based on the `0x01` byte in the data rather than the address? With the address, they just check that it's sent to their own address

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
```diff
@@ -0,0 +1,78 @@
...
+8. The Initiator fulfills this second packet using the fulfilment `F`.
```
emschwartz: Do they need to fulfill it? What does that tell them that they didn't already learn from getting the Prepare packet?

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
```diff
@@ -0,0 +1,78 @@
...
+9. Upon receiving the ILP `Fulfill` packet from the Initiator, the Recipient then fulfills the original Ping request using the same fulfillment `F`.
+
+## Implementation Recommendations
+
+1. The Initiator SHOULD choose a unique address that is only used for a single particular pinging session.
```
emschwartz: I think you mean that they should append a unique suffix to their normal ILP address. If they choose a random one, the packet won't actually reach them.

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
```diff
@@ -0,0 +1,78 @@
...
+1. Ping packet amounts SHOULD use amounts that are as small as possible such that any agregate losses due to ping traffic would not affect the operation of a Connector.
```
emschwartz: What does that mean? Is that a recommendation to the Initiator or the intermediary connectors?

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
```diff
@@ -0,0 +1,78 @@
...
+1. Because a Bidirectional Initiator controls the destination of all packets in the flows defined by this RFC, the Recipient must SHOULD be careful to apply proper exchange-rate logic to all packets.
```
emschwartz: What does that mean? Sounds like you're hinting at some potential problem without stating it directly

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
martinlowinski: connector.land, moneyd-gui and ilp-ping implement the bidirectional mode and have a default expiration set to 30000ms. So maybe a recommendation could be 15000ms for the unidirectional mode.

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
sappenin: This was fixed by the changes from @dora-gt

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
sappenin: After considering this change for a while, I don't think the RFC should really take a position on a recommended expiry. 

I say this because I was about to write, "This value will vary by path, but should generally not exceed 30 seconds." But then, after seeing it in there, it felt artificially proscriptive. For example, there might be times where this value should not exceed some other value, like maybe 5 seconds.

Personally, I like the text as-is here - it's basically saying that the initiator can pick whatever value they find appropriate. For very long paths, I supposed there might even be a time where 30s isn't enough.

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
sappenin: I'm leaning towards not adding anything specific about error-codes here, for two reasons:

First, [IL-RFC-27](https://github.com/interledger/rfcs/blob/master/0027-interledger-protocol-4/0027-interledger-protocol-4.md#error-codes) does a pretty good job defining ILP error codes and potential meanings. In general, I've found using those pretty intuitive on their own.

Second, the happy-path here is that a ping should be fulfilled, and if it's not being fulfilled, then the two broad categories for "why" are either: standard-ILP error messages apply (e.g., F04 - the initiator didn't send enough in the packet's amount for the packet to make it)....or....the receiver has disabled "ping" requests.

That last case (Connector has disabled ping protocol requests) is one where it _might_ make sense to be proscriptive, but here again, it feels like a typical connector would just return an `F00` or an `F02`, which sort of stands on its own without extra verbiage here.

Thoughts?

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
sappenin: I'm not sure I have a single answer here. In general I prefer the idea of picking something > zero here, because I think a Connector should generally reject all 0-value packets (this is Interledger after all, and it should be trivial to send _some_ amount of value).

I'm not sure what the value should be though for pinging. Intuitively, it seems like the rule should be "high enough to make it through the path, but as low as possible."

Anything more specific come to mind?

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
sappenin: Hmmm...I was about to answer "yes", but I remembered [this comment](https://forum.interledger.org/t/document-the-ping-protocol/279/9) by @justmoon.

Reading that comment, it seems like an implementation should do the following:

1. Look to see if the destination address of an incoming packet _starts with_ its own ILP address. If so, then route the packet to some sort of "ping protocol" handler.
1. At some point, check for the "pong" byte as a verification.
1. Correlate the actual incoming address to an outgoing "ping" that was already sent.

E.g., Alice pings Bob using a packet containing the address `g.alice.abc123`. Bob will send a "pong" to `g.alice.abc123` and Alice can fulfill that.

All that said though, is this type of thing even needed? In the forum post, the comment that triggered the response was:

> It’s not clear how Alice could differentiate a Pong from other packets right now as she uses her address with no additional suffix.

Does Alice need to differentiate here?

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
sappenin: Good question - I didn't invent this flow, so I'm speculating a bit here (there could be other rationale here that I'm overlooking), but if the Initiator doesn't fulfill the "pong", then the Receiver will likely _also_ reject the "ping", and so the packets won't actually have ever been paid for.

This isn't the end of the world because the amount of money earned from ping/pong requests should be negligible. However, as a matter of good Connector operation and ecosystem health, I feel pretty strongly that _every_ packet should allow a Connector to earn a fee (as much as possible - there are of course always Rejects). But as a matter of good protocol design, I prefer the mutual "fulfill" over a mutual "reject".

I suppose also if both sides reject, it's sort of unclear what _actually_ happened. From the perspective of the Initiator, that entity could _probably_ infer connectivity, but they wouldn't be able to see other things, like possibly a ping/pong rate....(?)

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
sappenin: Added.

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
sappenin: I was referring to something like a transaction-id, but I think that's probably confusing. I've updated the text with your example.

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
sappenin: Great question! For future readers, see [here](https://forum.interledger.org/t/deprecate-the-echo-protocol-and-replace-with-one-way-ping/520/3).

I've updated the spec to refer to your observation that the packet must be processed through the Connector's FX logic.

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
sappenin: Agreed!

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
sappenin: Agree. I've updated this verbiage -- though see my comment above. We need to decide if the protocol should be triggered on ILP addresses that _start with_ the Connector address, or instead  _equal_ the Connector address.

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
sappenin: Good point - that was meant for the Initiator.

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
sappenin: I've updated this verbiage to be clearer (hopefully).

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
sappenin: This is fixed.

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
sappenin: This part of the discussion is a bit out of date since the RFC has changed a bit. I'm going to resolve this discussion, but feel free to make a new comment if you still have questions.

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
emschwartz: > because I think a Connector should generally reject all 0-amount packets (this is Interledger after all, and it should be trivial to send some amount of value)

I very much disagree with this. Zero amount packets are used in STREAM to send data, and I think there is no point in disallowing them. If you disallow it, I would just send an unfulfillable packet of a non-zero amount to achieve the same effect while unnecessarily tying up liquidity. You can't stop people from sending packets just to transmit data so you might as well embrace it and just rate limit the number of packets sent for any reason if that is a concern.

0000-ilp-ping-protocol/0000-ilp-ping-protocol.md
emschwartz: I think it's easiest for the initiator to identify ping responses if they use a special suffix appended to their ILP address

PR #519, opened 2019-03-07
Invoice spec

sabinebertram:
This specifies how to conduct invoice payments. It used to be part of the general SPSP spec but has been broken out into its own spec when creating the basic SPSP spec (#513) and the spec on pull payments (#514). It is a suggestion derived from the discussions on the [ILP Forum](https://forum.interledger.org/t/push-payments-and-invoices/103).


0035-spsp-invoices/0035-spsp-invoices.md
```diff
@@ -0,0 +1,124 @@
...
+#### Response
+
+``` http
+HTTP/1.1 200 OK
+Content-Type: application/spsp4+json
+
+{
+  "destination_account": "example.ilpdemo.red~invoice123",
+  "shared_secret": "6jR5iNIVRvqeasJeCty6C+YB5X9FhSOUPCL/5nha5Vs=",
+  "balance": "5360",
```
sharafian: Why isn't `balance` inside of the `invoice` object?

sabinebertram: It makes it consistent with the pull payment spec. Also, the variables within the `invoice` objects don't change on incoming money whereas balance does.

0035-spsp-invoices/0035-spsp-invoices.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Introduction
+
+### Invoice Payments
+
+An Invoice Payment is a push payment to an Invoice SPSP Endpoint that allows the Payer (sender) to settle outstanding balances with the Payee (receiver). The Invoice SPSP Endpoint MAY expose the details about the Invoice and, when receiving value, updates its outstanding balance accordingly.
+
+### Motivation
+
+The [Simple Payment Setup Protocol](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md) (SPSP) only descirbes how the details required for a STREAM connection are exchanged and how simple push payments are made. Invoices require the SPSP server to keep track of incoming payments, i.e. act as an accounting server.
```
sublimator: > descirbes

0035-spsp-invoices/0035-spsp-invoices.md
```diff
@@ -0,0 +1,130 @@
...
+        "priceCurrency": "USD"
+      },
+      "paymentDueDate": "2019-01-30T00:00Z",
+      "paymentStatus": "http://schema.org/PaymentDue",
+      "description": "Chair model 'Rustic'",
+      "provider": "The Red Furniture Store",
+      "asset": {
+        "code": "USD",
+        "scale": 2
+    }
```
sharafian: tabbing here needs to be fixed

0035-spsp-invoices/0035-spsp-invoices.md
```diff
@@ -0,0 +1,130 @@
...
+
+The response body is a JSON object that includes basic account details necessary for setting up a STREAM connection as well as optional parameters for displaying purposes. The fields are defined in the following: 
+
+| Field | Type | Description |
+|---|---|---|
+| `destination_account` | [ILP Address](../0015-ilp-addresses/0015-ilp-addresses.md) | (see [Simple Payment Setup Protocol](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md#Response-Body)) |
+| `shared_secret` | 32 bytes, [base64 encoded](https://en.wikipedia.org/wiki/Base64) (including padding) | (see [Simple Payment Setup Protocol](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md#Response-Body)) |
+| `push`  | Object | _(OPTIONAL)_ Details of this specific push payment pointer. |
+| `push.balance`  | String Integer | _(OPTIONAL)_ Amount, denoted in `push.invoice.asset.code`, which completes the invoice payment. |
+| `push.invoice` | Object | _(OPTIONAL)_ Invoice details. It SHOULD follow the invoice schema proposed by [schema.org](https://schema.org/Invoice). Additionally, it SHOULD include the followin parameters:|
```
sharafian: This makes the spec quite confusing to implement. We also don't have any dependency on schema.org in any other specs. Better to just describe the fields you used in your example and maybe just note that they came from this schema on schema.org

sabinebertram: What I don't like about the schma.org spec is that the amounts are not integers with scale as we have it everywhere else. 
```
"totalPaymentDue": {
        "price": 199.99,
        "priceCurrency": "USD"
      },
```
@sharafian Do you think we should "redefine" that?

sharafian: I think I'd just scrap the `totalPaymentDue`, `paymentDueDate`, `paymentStatus`, and `provider` fields. We could replace it with `amount` (which is an integer amount in units defined by `asset`, and we could add an `additional_details` field that you can put arbitrary extra information into.

sabinebertram: Basically, you are suggesting to not use schema.org at all but to try to come up with the fields that are absolutely necessary (maybe inspired by schema.org) and have an `additional_fields`.

0035-spsp-invoices/0035-spsp-invoices.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Operation Overview
+
+The Payer's SPSP Client will set up a [STREAM](../0029-stream/0029-stream.md) connection to the Payee's SPSP Server using the Invoice Payment Pointer, as described by the [Simple Payment Setup Protocol](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md). On connection, the SPSP Client pushes as much value to the SPSP Server as necessary to complete the Invoice.
+
+## Model of Operation
+
+### Creating an Invoice Payment Pointer
+
+Prior to the Invoice Payment, the Payee has to create the Invoice Payment Pointer representing the Invoice. This Invoice Payment Pointer MUST be unique to the Invoice. The Payee's SPSP Server SHOULD store the Invoice details, which are defined in the `invoice`-object within the [Response Body](#Response-Body), within the Invoice SPSP Endpoint in order to account for balance changes.
```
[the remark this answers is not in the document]
sabinebertram: I think I just forgot to delete something when I rephrased that paragraph.

0035-spsp-invoices/0035-spsp-invoices.md
```diff
@@ -0,0 +1,130 @@
...
+    "balance": "5360",
+    "invoice": {
+      "totalPaymentDue": {
+        "price": 199.99,
+        "priceCurrency": "USD"
+      },
+      "paymentDueDate": "2019-01-30T00:00Z",
+      "paymentStatus": "http://schema.org/PaymentDue",
+      "description": "Chair model 'Rustic'",
+      "provider": "The Red Furniture Store",
```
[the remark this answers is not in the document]
sabinebertram: Yes, this is how schema.org calls the merchant.

0035-spsp-invoices/0035-spsp-invoices.md
```diff
@@ -0,0 +1,130 @@
...
+
+The Payer's SPSP Client opens a [STREAM](../0029-stream/0029-stream.md) connection to the Payee's SPSP Server as described in the [Simple Payment Setup Protocol](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md). Once this connection is established, the process continues as follows: 
+
+The SPSP Client begins sending ILP packets to complete the Invoice.
+  1. The SPSP Client will adjust their `sendMax` to reflect the amount they're willing to send.
+      * `sendMax` SHOULD be derived from the Invoice.
+  2. The SPSP Server will adjust their `receiveMax` to reflect the amount they're willing to receive.
+      * `receiveMax` SHOULD be derived from the Invoice.
+  3. The SPSP Client's and Server's [STREAM Modules](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md#Definitions) will move as much value as possible while staying inside these bounds.
+  4. If the SPSP Client reaches their `sendMax`, they end the stream and the connection. If the SPSP Server reaches their `receiveMax`, they will end the stream and the connection.
```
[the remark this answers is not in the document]
sabinebertram: I guess if the `sendMax` and `receiveMax` are set in a way that payer won't send more than what he needs to pay the invoice in full, then this does not need to be a separate requirement. But I agree that I need to define _how_ you derive `sendMax` and `receiveMax` in order for this to hold.

0035-spsp-invoices/0035-spsp-invoices.md
```diff
@@ -0,0 +1,130 @@
...
+
+##### Response Body
+
+The response body is a JSON object that includes basic account details necessary for setting up a STREAM connection as well as optional parameters for displaying purposes. The fields are defined in the following: 
+
+| Field | Type | Description |
+|---|---|---|
+| `destination_account` | [ILP Address](../0015-ilp-addresses/0015-ilp-addresses.md) | (see [Simple Payment Setup Protocol](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md#Response-Body)) |
+| `shared_secret` | 32 bytes, [base64 encoded](https://en.wikipedia.org/wiki/Base64) (including padding) | (see [Simple Payment Setup Protocol](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md#Response-Body)) |
+| `push`  | Object | _(OPTIONAL)_ Details of this specific push payment pointer. |
```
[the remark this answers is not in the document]
sabinebertram: That is a good question. What is optional and what is not. This also guides if I can even say that `sendMax` is derived from the Invoice.

0035-spsp-invoices/0035-spsp-invoices.md
sharafian: yeah, I think the fields that are absolutely necessary are just the `amount` and `asset`. Everything else can go into a `additional_fields` object because they're just metadata that describes the payment but doesn't really affect the logic of how you pay.

0035-spsp-invoices/0035-spsp-invoices.md
```diff
@@ -0,0 +1,129 @@
...
+
+### Conducting an Invoice Payment
+
+The Payer's SPSP Client opens a [STREAM](../0029-stream/0029-stream.md) connection to the Payee's SPSP Server as described in the [Simple Payment Setup Protocol](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md). Once this connection is established, the process continues as follows: 
+
+The SPSP Client begins sending ILP packets using the STREAM protocol to complete the Invoice.
+  1. The SPSP Client will adjust their `sendMax` to reflect the amount they're willing to send.
+      * `sendMax` SHOULD be derived from the Invoice, i.e., `sendMax` SHOULD be equal to `push.invoice.amount - push.balance`, converted to the SPSP Client's operating asset, taking exchange rate fluctuations into account.
+  2. The SPSP Server will adjust their `receiveMax` to reflect the amount they're willing to receive.
+      * `receiveMax` SHOULD be derived from the Invoice, i.e., `sendMax` SHOULD be equal to `push.invoice.amount - push.balance`, converted to the SPSP Server's operating asset, taking exchange rate fluctuations into account.
```
emschwartz: This should be set slightly over the amount they want. Because of different exchange rates and asset scales, the rounding error may make it impossible to deliver the exact number the receiver is requesting.

0035-spsp-invoices/0035-spsp-invoices.md
```diff
@@ -0,0 +1,129 @@
...
+  3. The SPSP Client's and Server's [STREAM Modules](../0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md#Definitions) will move as much value as possible while staying inside these bounds.
+  4. If the SPSP Client reaches their `sendMax`, they end the stream and the connection. If the SPSP Server reaches their `receiveMax`, they will end the stream and the connection. Furthermore, when the SPSP Server has received enough value to fully pay the invoice, it ends the stream and the connection.
```
emschwartz: It might be worth mentioning that they should end the connection with a `NoError` code in the `ConnectionClose` frame

0035-spsp-invoices/0035-spsp-invoices.md
sabinebertram: I would argue that `receiveMax` does not need to many adjustments because the receiver only needs to convert the price to her underlying currency and maybe add some value x to account for exchange rate risk. The `sendMax`, on the other side, should account for the exchange rate risk (invoice asset and sender's underlying asset) plus connector fees.

PR #523, opened 2019-03-19
docs(0027): packets must be canonical OER

sublimator:


0027-interledger-protocol-4/0027-interledger-protocol-4.md
```diff
@@ -94,7 +94,7 @@ In the case of post-funded accounts, senders transfer value to their connector a
 
 ### Packet Format
 
-ILPv4 packets SHOULD be encoded using the ASN.1 Octet Encoding Rules.
+ILPv4 packets MUST be encoded using the ASN.1 Canonical Octet Encoding Rules.
```
sappenin: Are we sure we want this to read `MUST`? 

For example, if two peers (`A` & `B`) wanted to exchange ILP packet information using some non-OER encoding (e.g., HTTP headers), and then `B` and `C` wanted to exchange packets using OER encoded "packets", would this not still be considered "ILP"?

sublimator: Speaking for myself, no, not sure :) In that case, seems neither SHOULD or MUST are really appropriate? 

Just reading this:
https://medium.com/interledger-blog/thoughts-on-scaling-interledger-connectors-7e3cad0dab7f

It's interesting how much information in ILP is implied by the connection

emschwartz: I think SHOULD is the right way to say it.

> 3. SHOULD   This word, or the adjective "RECOMMENDED", mean that there
   may exist valid reasons in particular circumstances to ignore a
   particular item, but the full implications must be understood and
   carefully weighed before choosing a different course.

From [RFC 2119](https://tools.ietf.org/html/rfc2119)

sublimator: Seems so

justmoon: The [robustness principle](https://en.wikipedia.org/wiki/Robustness_principle) says a compliant implementation MUST produce correct encodings but MAY accept other, invalid encodings.

I think it should say that you MUST encode with Canonical OER but you MAY accept packets that are not encoded canonically.

sublimator: That's a good, and in my view, different point. 

It seemed like there were a lot of good reasons enumerated for using C-OER. If you have a well defined standard for canonical encodings it makes sense to mandate them to me. 

Otherwise, in practice you'll have to define (read: arise ad-hoc) what are semantically equivalent/meaningful encodings. ([See criticisms](https://en.wikipedia.org/wiki/Robustness_principle#Criticism)) I'm not advocating to "fail fast" and crash if you are being sent non-conformant packets by malevolent actors though. 

I couldn't quite articulate what I meant before when I said "seems neither are really appropriate". You aren't really saying much with SHOULD other than "this is the default encoding format, but we might support others", though in a manner contorting to conform to FORMAL WORDING. 

wrt: encoding packets in HTTP headers, I'm not really up-to-speed with why you would want to do that, but sticking to one way of doing things seems wise. 

I'll appeal to Tim Peters
```
In the face of ambiguity, refuse the temptation to guess.
There should be one-- and preferably only one --obvious way to do it.
Although that way may not be obvious at first unless you're Dutch.
```

Do we know any Dutch ?

sappenin: From my perspective, I'm good with [either](https://github.com/interledger/rfcs/pull/523#discussion_r268683409) / [option](https://github.com/interledger/rfcs/pull/523#discussion_r274223268) proposed by @emschwartz  or @justmoon - I think both of them leave room for alternate implementations, which was my main concern.

sublimator:
As far as allowing wiggle room for different encodings, it's easier to
loosen than tighten. Shrug.

I don't have the big picture to have any real opinion.

sublimator:
Any further thoughts from anyone on this? I was thinking just closing the PR

adrianhopebailie:
I'm supportive of @justmoon 's proposal to differentiate between what you MUST be able to send and what you MAY be prepared to receive.

Even the ILP-over-HTTP spec we're using today encodes the packets using canonical OER so I am unaware of any variances in implementation.

sublimator:
> I'm supportive of @justmoon 's proposal to differentiate between what you MUST be able to send and what you MAY be prepared to receive.

I don't really see the need if you've chosen a canonical encoding method but will update the PR.

sublimator:
I don't think I got the wording right to express the intent, so would welcome anyone pushing something over it, or proposing a fix :) It's not logically symmetrical to say **MUST** send in a given format but MAY receive in another, so it just doesn't sit right with me, and thus not something I can express well. 

I think @adrianhopebailie seemed closer with `MUST *be able* to send` vs `MUST send` 

We need to make the distinction between:
1. accepting slight changes in packets (as an example the timestamp format) for what are essentially C-OER in intent 
2. completely different encodings (e.g. http headers with text, or JSON) 

On the larger debate of whether tolerance vs intolerance (adversity) helps with robustness (strength), I'm in the latter camp, especially when we are talking about how to encode packets when a well specified canonical encoding method has already been chosen.

PR #525, opened 2019-04-08
feat(0035): update to include account identification

matdehaast:
Add an optional header to ILP-http to allow for the association of the request to account on the underlaying host receiving the request.

### DIFF
--- a/0035-ilp-over-http/0035-ilp-over-http.md
+++ b/0035-ilp-over-http/0035-ilp-over-http.md
@@ -1,6 +1,6 @@
 ---
 title: ILP Over HTTP
-draft: 2
+draft: 3
 ---
 
 # ILP Over HTTP
@@ -32,6 +32,7 @@ Host: example.com
 Accept: application/octet-stream
 Content-Type: application/octet-stream
 Authorization: Bearer zxcljvoizuu09wqqpowipoalksdflksjdgxclvkjl0s909asdf
+ILP-Account-Name: alice
 
 < Body: Binary OER-Encoded ILP Prepare Packet >
 ```
@@ -40,7 +41,8 @@ Authorization: Bearer zxcljvoizuu09wqqpowipoalksdflksjdgxclvkjl0s909asdf
 
 - **Path** - A connector MAY specify any HTTP path for their peer to send ILP packets to.
 - **Host Header** - The standard [HTTP Host Header](https://tools.ietf.org/html/rfc2616#section-14.23) indicating the domain of the HTTP server the Request is sent to.
-- **Content-Type / Accept Headers** - MUST be set to `application/octet-stream`.
+- **Content-Type / Accept Headers** - MUST be set to `application/octet-stream`.\
+- **ILP-Account-Name** - An optional header used to identify the account for which the request is being made.
 - **Body** - ILP Packet encoded using OER, as specified in [RFC 27: Interledger Protocol V4](./0027-interledger-protocol-4/0027-interledger-protocol-4.md).
 
 ### Response


0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -32,6 +32,7 @@ Host: example.com
 Accept: application/octet-stream
 Content-Type: application/octet-stream
 Authorization: Bearer zxcljvoizuu09wqqpowipoalksdflksjdgxclvkjl0s909asdf
+ILP-Account-Name: alice
```
sappenin: Can we call this `ILP-Account-Id` instead?

matdehaast: I prefer the id suffix over name. I know @adrianhopebailie was a fan of using `ilp-peer-id`. Do we have a consensus on `peer` vs `account`

sappenin: I think we should use `account` here, but perhaps it's a naming ambiguity thing. If `peer` == `account`, then it doesn't matter. 

However, in my mind, `peer` != `account` because a peer could have multiple accounts potentially (e.g., imagine you and me are peered with one credential that can operate across currencies/accounts).

However, in ILP, I don't think we should introduce that type of peering because it's much simpler to just have accounts. So, I vote for `account` and we can (maybe never) use the `peer` term in the future.

sappenin: > But, in this context I think we are talking about a peer not an account
Can you clarify why? I'd like to _not_ introduce peering into ILP if we can help it. In my mind it's actually simpler and more secure to only have accounts, with each account using a different auth credential.

0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -40,7 +41,8 @@ Authorization: Bearer zxcljvoizuu09wqqpowipoalksdflksjdgxclvkjl0s909asdf
 
 - **Path** - A connector MAY specify any HTTP path for their peer to send ILP packets to.
 - **Host Header** - The standard [HTTP Host Header](https://tools.ietf.org/html/rfc2616#section-14.23) indicating the domain of the HTTP server the Request is sent to.
-- **Content-Type / Accept Headers** - MUST be set to `application/octet-stream`.
+- **Content-Type / Accept Headers** - MUST be set to `application/octet-stream`.\
+- **ILP-Account-Name** - An optional header used to identify the account for which the request is being made.
```
emschwartz: Could you expand on what this header is for? I'm wondering about a couple of points:
- When you say it's optional, does that mean that it's optional from the perspective of this protocol in general or optional from the perspective of every ILP-over-HTTP server? 
- If a server is set up to use this header, how should it behave if that header is not present?
- How do you find out this value? Is that another field that needs to be communicated when peering?
- How do you know whether a peer requires this or not (if it can be required) and do all client implementations need to support sending this field?

sappenin: I think we should either close this PR or else move this header into a section that details (non-normatively) how peers might use extra headers to help with authentication. 

For example, the current Java implementation of ILP-over-HTTP supports a JWT authentication mode where the entire token can trivially be used in a performant manner without _any_ additional headers.

Also, read more about my current opinion here: https://forum.interledger.org/t/connector-auth-model/517/4?u=sappenin

adrianhopebailie: Agree. I don't think having this as a normative requirement makes sense.

I do think we should have at least one auth scheme that all implementations MUST support and this should be a well defined standard.

That manifests as some changes to the "Authentication" section to include some normative requirements. I'll write those up.

adrianhopebailie:
After extensive discussion we're decided this is not a normative requirement

PR #527, opened 2019-05-04
docs(0028): update spec with monetization tag observation changes

sublimator:
Overview
* Specify observing head for monetization tags
  * Introduce concept of a meta "Tags Set"
    * if tags are injected they must be ordered carefully to work with meta observer
* Rename Request-ID to Monetization-ID
  * leave `requestId` field the same for the moment
    * Add comment near field for correlation
* Consistently use `browser` rather than mix of `browser` and `agent`
* Bar iframe support until have a sensible working implementation
* meta[content] can specify a payment pointer or an SPSP url
  * Generically referred to as `Payment Setup Endpoint`
* Make it clear that the amount is the destination received amount

### DIFF
--- a/0028-web-monetization/0028-web-monetization.md
+++ b/0028-web-monetization/0028-web-monetization.md
@@ -41,16 +41,16 @@ Web Monetization makes use of [SPSP](../0009-simple-payment-setup-protocol/0009-
 
 This flow refers to the user's **browser** and the user's **provider**, [defined above](#terminology).
 
-- The user's agent navigates a webpage.
+- The user's agent navigates to a webpage.
 - The user's agent sets `document.monetization` to an Object which implements [EventTarget](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget).
 - The user's agent sets `document.monetization.state` to `pending`.
-- The user's agent observes the `<head>` and looks for a Web Monetization `<meta name="monetization">` tag ([specified below](#meta-tags)).
+- The user's agent observes the `<head>` and looks for a Web Monetization `<meta name="monetization">` tag ([specified below](#meta-tags-set)).
   - There MUST be only one `<meta name="monetization">` tag at any given time
-  - The `<meta>` tags MUST be in the `<head>` of the document.
-  - The `<meta>` tags MUST be in the top level window (i.e. not inside an iframe)  
+  - The `<meta>` Tags Set MUST be in the `<head>` of the document.
+  - The `<meta>` Tags Set MUST be in the top level window (i.e. not inside an iframe)
 
 - Below is repeated for every semantically (consider `meta.content = meta.content`) new  `<meta name="monetization">` tag seen for the life of the page:
-  - If any of the Web Monetization `<meta>` tags are malformed, the browser will stop here. The user's agent SHOULD report a warning via the console.
+  - If any of the Web Monetization `<meta>` Tags Set are malformed, the browser will stop here. The user's agent SHOULD report a warning via the console.
   - If the Web Monetization `<meta name="monetization">` tag is well-formed, the browser should extract the Payment Setup Endpoint ([Payment Pointer](../0026-payment-pointers) or [SPSP](../0009-simple-payment-setup-protocol) Url).
   - The user's agent generates a fresh **random** UUID (version 4) and uses this as a Monetization ID. This MUST be regenerated upon dynamic changes to the meta tags and MUST be unique per page load in order to avoid user tracking.
   - The user's agent passes these payment details to the user's provider (see [Payment Handler Flow](#payment-handler-flow)).
@@ -58,13 +58,14 @@ This flow refers to the user's **browser** and the user's **provider**, [defined
     - On the SPSP query to resolve the Payment Setup Endpoint, a `Web-Monetization-Id` header is sent, containing the Monetization ID. The server running the web-monetized site may use this to associate future requests by the user with their payments.
     - With the `destination_account` and `shared_secret` fetched from the SPSP query, a STREAM connection is established. A single STREAM is opened on this connection, and a positive SendMax is set.
     - The provider SHOULD set their SendMax high enough that it is never reached, and make payment adjustments by limiting throughput.
-  - Once the STREAM connection has fulfilled an ILP packet with a non-zero amount, the provider notified the browser, and the browser dispatches an event on `document.monetization`. Payment SHOULD continue.
-    - The user's agent sets `document.monetization.state` to `started`. This MUST occur before the `monetizationstart` event is fired.
-    - The event's type is `monetizationstart`. The event has a `detail` field with an object containing the Payment Setup Endpoint and the Monetization ID ([specified below](#monetizationstart)).
+  - Once the STREAM connection has fulfilled an ILP packet with a non-zero amount, the provider notifies the browser:
+    - The user's agent sets `document.monetization.state` to `started`.
+    - This MUST occur before the `monetizationstart` event is fired.
+    - This event has a `detail` field with an object containing the Payment Setup Endpoint and the Monetization ID ([specified below](#monetizationstart)).
     - The user's agent also emits a `monetizationprogress` ([specified below](#monetizationprogress)) event from `document.monetization`, corresponding to this first packet. If there are no listeners the event MAY NOT be emitted.
   - Payment continues for the lifetime of a given meta tag (or semantically equivalent)
     - The provider MAY decide to stop/start payment at any time, e.g. if the user is idle, backgrounds the page, or instructs the browser to do so.
-    - If the STREAM connection is severed, the provider will redo the SPSP query to the same Payment Setup Endpoint as before with the same Monetization ID. The user's agent MUST NOT re-process the `<meta>` tags.
+    - If the STREAM connection is severed, the provider will redo the SPSP query to the same Payment Setup Endpoint as before with the same Monetization ID. The user's agent MUST NOT re-process the `<meta>` Tags Set.
     - Each time a packet with a nonzero amount is fulfilled, the provider notifies the browser, and the browser emits an event on `document.monetization`. The event's type is `monetizationprogress`. The event has a `detail` field containing the details of the packet ([specified below](#monetizationprogress)). If there are no listeners the event MAY NOT be emitted.
     - When a stream is closed the `document.monetization.state` MUST be set back to 'pending'
 
@@ -72,7 +73,7 @@ This flow refers to the user's **browser** and the user's **provider**, [defined
 
 A provider can be implemented as a Payment Handler supporting the 'webmonetization' payment method (The payment method specification for this payment method is still under development.). Communication between the browser and the provider would use this flow.
 
-- After parsing the `<meta>` tags, the browser creates a new [PaymentRequest](https://www.w3.org/TR/payment-request/#paymentrequest-interface) object with the following [PaymentMethodData](https://www.w3.org/TR/payment-request/#dom-paymentmethoddata) argument.
+- After parsing the `<meta>` Tags Set, the browser creates a new [PaymentRequest](https://www.w3.org/TR/payment-request/#paymentrequest-interface) object with the following [PaymentMethodData](https://www.w3.org/TR/payment-request/#dom-paymentmethoddata) argument.
 
 ```json
 {
@@ -89,11 +90,12 @@ A provider can be implemented as a Payment Handler supporting the 'webmonetizati
 
 ## Specification
 
-### Meta Tags
+### Meta Tags Set
 
-This `<meta>` tags MUST be in the document's `<head>`. The `<meta>` tags allows the user's agent to pay a site via Web Monetization by specifying a [Payment Setup Endpoint](../0026-payment-pointers/0026-payment-pointers.md) or SPSP url.
+The `<meta>` Tags Set MUST be in the document's `<head>`. The `<meta>` Tags Set allows the user's agent to pay a site via Web Monetization by specifying a [Payment Pointer](../0026-payment-pointers/0026-payment-pointers.md) or [SPSP](../0009-simple-payment-setup-protocol) url.
 
-The `name` of the `<meta>` tags all start with `monetization`. The table below lists the different `name`s and the formats of their `content`. Currently there is only one tag, but this may be expanded in the future.
+The `name` of the `<meta>` tags all start with `monetization`. The table below lists the different `name`s and the formats of their `content`. Currently there is only one tag, but this may be expanded in the future so care MUST be given to altering a Tags Set such that `<meta name="monetization">` is the last one modified.   
+
 
 | Name | Required? | Format | Description |
 |:--|:--|:--|:--|
@@ -113,7 +115,7 @@ The `name` of the `<meta>` tags all start with `monetization`. The table below l
 
 ```ts
 document.monetization: EventTarget
-document.monetization.state: String
+document.monetization.state: string
 ```
 
 `document.monetization.state` can be one of two values.
@@ -133,14 +135,14 @@ Dispatched once the first ILP packet with a non-zero amount has been fulfilled b
 {
   detail: {
     // Payment Setup Endpoint
-    paymentPointer: String,
-    // Monetization ID
-    requestId: String
+    paymentPointer: string,
+    // Monetization ID / Web-Monetization-ID header
+    requestId: string
   }
 }
 ```
 
-The `paymentPointer` matches the one in the `<meta>` tags. The `requestId` matches the UUID generated by the browser (see [Flow](#flow)).
+The `paymentPointer` matches the one in the `<meta name="monetization">` tag. The `requestId` matches the Monetization ID (UUID V4) generated by the browser (see [Flow](#flow)).
 
 #### `monetizationprogress`
 
@@ -150,14 +152,14 @@ Dispatched every time an ILP packet with a non-zero amount has been fulfilled by
 {
   detail: {
     // Received amount
-    amount: String,
-    assetCode: String,
-    assetScale: Number
+    amount: string,
+    assetCode: string,
+    assetScale: number
   }
 }
 ```
 
-- `amount` is a String containing the amount contained in the ILP packet.
+- `amount` is a String containing the destination amount **received** as specified in the ILP packet.
 - `assetCode` contains the three letter asset code describing the amount's units.
 - `assetScale` contains a number representing the scale of the amount. For example, cents would have an assetScale of `2`.
 
@@ -165,7 +167,7 @@ Dispatched every time an ILP packet with a non-zero amount has been fulfilled by
 
 #### `Web-Monetization-Id`
 
-Contains the `Monetization ID` that the browser generated. This header MUST always be sent on SPSP queries for Web Monetization. This value MUST be a UUID version 4.
+Contains the `Monetization ID` (currently referred to as `requestId` in the events) that the browser generated. This header MUST always be sent on SPSP queries for Web Monetization. This value MUST be a UUID version 4.
 
 ```http
 Web-Monetization-Id: dcd479ad-7d8d-4210-956a-13c14b8c67eb
--- a/0028-web-monetization/0028-web-monetization.md
+++ b/0028-web-monetization/0028-web-monetization.md
@@ -5,7 +5,7 @@ draft: 8
 
 # Web Monetization
 
-Web Monetization is a proposed browser API that uses ILP micropayments to monetize a site. It can be provided through a polyfill or an extension, but the goal is to eventually implement it directly into the user's agent.
+Web Monetization is a proposed browser API that uses ILP micropayments to monetize a site. It can be provided through a polyfill or an extension, but the goal is to eventually implement it directly into the user's browser.
 
 ## Overview
 
@@ -20,7 +20,7 @@ Web Monetization is a proposed browser API that uses ILP micropayments to moneti
 - Should be extremely simple for webmasters to use in their site.
 - Backend infrastructure should be optional; should be usable on a static site.
 - Should not require any interaction with the user.
-- Should give user's agent a choice about how much to spend, and which sites to support.
+- Should give user's browser a choice about how much to spend, and which sites to support.
 - Should give advanced webmasters a way to associate payments with their users, in order to unlock premium experiences.
 - Should pay continuously as the user consumes content.
 - Should be compatible with existing application and transport protocols on Interledger.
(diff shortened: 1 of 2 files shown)


sublimator:
@sharafian

0028-web-monetization/0028-web-monetization.md
```diff
@@ -44,24 +44,30 @@ This flow refers to the user's **browser** and the user's **provider**, [defined
...
 - Once the STREAM connection has fulfilled an ILP packet with a non-zero amount, the provider notified the browser, and the browser dispatches an event on `document.monetization`. Payment SHOULD continue.
   - The user's agent sets `document.monetization.state` to `started`. This MUST occur before the `monetizationstart` event is fired.
-  - The event's type is `monetizationstart`. The event has a `detail` field with an object containing the Payment Pointer and the Request ID ([specified below](#monetizationstart)).
+  - The event's type is `monetizationstart`. The event has a `detail` field with an object containing the Payment Pointer and the Monetization ID ([specified below](#monetizationstart)).
   - The user's agent also emits a `monetizationprogress` ([specified below](#monetizationprogress)) event from `document.monetization`, corresponding to this first packet. If there are no listeners the event MAY NOT be emitted.
-- Payment continues until the user closes/leaves the page.
+- Payment continues for the lifetime of a given meta tag (or semantically equivalent)
```
sublimator: should really be a set of meta tags I guess, but yeah ...

sharafian:
We should also have some tips for implementors about CORS and the Web-Monetization-Id header. This is something I didn't even think about at first because extensions ignore CORS but in some cases you need to serve an OPTIONS endpoint that has Access-Control-Allow-Headers set to allow Web-Monetization-Id

0028-web-monetization/0028-web-monetization.md
```diff
@@ -140,12 +136,12 @@ Dispatched once the first ILP packet with a non-zero amount has been fulfilled b
 {
   detail: {
     paymentPointer: String,
-    requestId: String
+    id: String
```
sharafian: What's the rationale for changing this field? This turns this into a breaking change

sublimator:
Putting aside the fact it is a breaking change:
I think "id" and  "Monetization ID" are more appropriate. The former is
nice and short (leaving wiggle room?? Actually, I could be easily moved to
s/requestId/monetizationId) and the latter matches the existing
Web-Monetization-ID header.

The meta tags and are no longer bound to a single page "request" and
streams can stop and start.

Back to "breaking":
Could we live with the pain of the change? Is it worth it for conceptual
clarity? How many people would it effect? When would be the time to make
changes like this?

Does "request" still make sense to you? Would it make sgnse to the
uninitiated?

I guess one could define non enumerable properties on the events with
deprecation warnings. Shrug.

Your call

sharafian:
> The meta tags and are no longer bound to a single page "request" and
streams can stop and start.

The main reason to say requestId was to make it clear that the ID was ephemeral (i.e. unique to each SPSP request). A monetizationId might seem like it refers to the user rather than to the SPSP request.

I could maybe get `webMonetizationId` (because it matches the header) or just `id` because it's pretty general. But I still feel like `requestId` isn't innapropriate.

> - monetization stop event ?

:+1: I think this makes sense. We might also want to add a new monetization state: `stopped`. So you can tell the difference between initializing (`"pending"`), stopped (`"stopped"`), and not supported (`undefined`). This also makes it less confusing when you check if web monetization is supported on a page that doesn't use it. "pending" is a confusing state if there's no meta tag present.

> - specify both sent and received currency code/scale explicitly

:+1: STREAM exposes both of these so it makes sense to expose them to the page. You'd want to use the source currency preferences for display (like counters) and you'd want to use the destination currency preferences to check how much the user has paid (e.g. to adjust paid video quality or to trigger loading of a piece of exclusive content)

traviscrist:
> > The meta tags and are no longer bound to a single page "request" and
> > streams can stop and start.
> 
> The main reason to say requestId was to make it clear that the ID was ephemeral (i.e. unique to each SPSP request). A monetizationId might seem like it refers to the user rather than to the SPSP request.
> 
> I could maybe get `webMonetizationId` (because it matches the header) or just `id` because it's pretty general. But I still feel like `requestId` isn't innapropriate.

If we make it `id` or `webMonetizationId` it no longer feels short lived, the update above which removes that its only on page loads also sorta implies this id is a long lived id. I feel like what we put here should not just be id since its a short lived id, its not long lived. `webMonetizationSessionId` could work but that's getting really long.

sublimator:
@sharafian 
> just id because it's pretty general.

Clarity through ambiguity TM :)

@traviscrist 
> it no longer feels short lived,

I think I understand. 

We moved to allow dynamic `<meta>` tags as a way of supporting SPA, such as coil.com (which I suppose will end up using server side rendering for SEO, but regardless, I'm sure you can imagine the use cases, such as a long page of videos that stop/start as you scroll ).  

----

For better or worse, I suppose I've come to think of a `monetization` as a concept/thing.

I'll try and illustrate why `monetizationId` made sense to me:

The event names:
**monetization**start 
**monetization**stop

The meta tag:
`<meta name='`**`monetization`**`' />`

The (content request) header:
Web-**Monetization-Id**

Look at the relationships between the various entities:

|Entities|Relationship|
|--------|------------|
|`<meta name="monetization">`:requestId| 1:1|
|SPSP resolve request: requestId | M:1 (see: pause/resume)| 
|Content request: requestId | M:1 (see: reconnects while streaming videos) |

When do we generate a new `requestId` now that we have moved to supporting dynamic meta tags? Whenever we see a new monetization meta tag.

Note, that in the table above I'm using request in the sense of a http request, rather than in a more abstract sense. Kind of cheating, but there *is* a strict mapping of meta tag to an identifier. To me the current `request` in `requestId` maps more closely to the request for content than the SPSP request.

I recall that requestId was at first dubbed `correlationId`. Used for correlating a monetization stream with later requests for content/services/etc. 

If you guys aren't really board, given it's a break change, should probably just leave it as `requestId`

0028-web-monetization/0028-web-monetization.md
```diff
@@ -44,24 +44,30 @@ This flow refers to the user's **browser** and the user's **provider**, [defined
...
-  - If the Web Monetization `<meta>` tags are well-formed, the browser should extract the Payment Pointer.
-  - The browser will generate a fresh UUID (version 4) and use this as the Request ID from this point forward. **This Request ID MUST be unique per page load**, not per browser, session nor site.
+  - The `<meta>` tags MUST be in the top level window (i.e. not inside an iframe)  
+  - If any of the Web Monetization `<meta>` tags are malformed, the browser will stop here. The browser SHOULD report a warning via the console.
+  - If the Web Monetization `<meta name="monetization">` tag is well-formed, the browser should extract the Payment Pointer or SPSP url
+  (TODO: This would imply that when dynamically injecting multiple meta tags (in the sense of others than `<meta name="monetization">`, it would require careful ordering so as to be operating on the information defined by the set of tags as a whole, removal with document.head.querySelectorAll('meta[name^="monetization"]'), then adding the name="monetization" tag last)
+  
+  )
+  - The browser will generate a fresh UUID (version 4) and use this as the Monetization ID from this point forward.
```
sublimator: @traviscrist 
Yes, this needs fixing ^

sublimator:
@sharafian @traviscrist 
Thoughts re: monetizationpause/ monetizationresume events?
Would be interesting to get input from cinnamon guys.

Note that often people open a youtube tab to put some music on, change tabs while still listening and then do something else.

sublimator:
@sharafian 
> We should also have some tips for implementors about CORS
> "pending" is a confusing state if there's no meta tag present.

Noted

sublimator:
@traviscrist 
Do you have any specific concerns with easing the requirement of only one meta per "page load" (which I assume means a single http request)? In particular, any security concerns? That was one aspect that nags at me.

We were discussing an https requirement at one point, but then you'd really have to be the browser to be able to get to the raw request bytes to ensure it wasn't inserted dynamically. Failing that, you'd have to request the page again, which I doubt would be workable in all cases.

traviscrist:
> When do we generate a new requestId now that we have moved to supporting dynamic meta tags? 

Maybe it should have been called pageLoadId the whole idea behind the id changing on page load is to prevent tracking a user with this id by forcing it the be fresh on each page load. This is my main concern in relaxing the requirement that this be unique per page load. The name of it should just go with that requirement, I'm not really set on any name just that this id not be another vector for tracking user behavior since that could slow down the adoption of the web monetization spec. 

> Note that often people open a youtube tab to put some music on, change tabs while still listening and then do something else.

If the page does not reload then the id would not change, if it does then it would need to be correlated again with that user. 

> @traviscrist
> Do you have any specific concerns with easing the requirement of only one meta per "page load" (which I assume means a single http request)? In particular, any security concerns? That was one aspect that nags at me.

Where is this change in the proposed changes here? I don't see that requirement being eased any where in the proposed changes. Maybe I missed it though.

sublimator:
YouTube question wasn't really concerning the correlation id. Just
something that occurred to me. The extension currently pauses monetization
when a tab is not active.

We moved the random id generation from per page load to per meta content
change.

Late here, will answer tomorrow in more detail

traviscrist:
> We moved the random id generation from per page load to per meta content
change.

So this means that if the page is reloaded then the id stays the same? 

I understand changing the id on the meta content change but I think we should also keep the requirement that it change/(be unique) on a page load as well. From what I can tell these changes relax that requirement which makes it so the id no longer needs to be unique per page reload.

sublimator:
> So this means that if the page is reloaded then the id stays the same?
No. The requirements were not intended to be relaxed. It's just that the
language needs tightening up.

The idea (and latest impl in the extension) is to generate a new uuid.v4
every time it sees a new meta tag, and it's not persisted between page
loads. Reload the page and you'll get a new id.

0028-web-monetization/0028-web-monetization.md
```diff
@@ -44,24 +44,30 @@ This flow refers to the user's **browser** and the user's **provider**, [defined
...
-  - If the Web Monetization `<meta>` tags are malformed, the browser will stop here. The browser SHOULD report a warning via the console.
-  - If the Web Monetization `<meta>` tags are well-formed, the browser should extract the Payment Pointer.
-  - The browser will generate a fresh UUID (version 4) and use this as the Request ID from this point forward. **This Request ID MUST be unique per page load**, not per browser, session nor site.
+  - The `<meta>` tags MUST be in the top level window (i.e. not inside an iframe)  
+  - If any of the Web Monetization `<meta>` tags are malformed, the browser will stop here. The browser SHOULD report a warning via the console.
+  - If the Web Monetization `<meta name="monetization">` tag is well-formed, the browser should extract the Payment Pointer or SPSP url
+  (TODO: This would imply that when dynamically injecting multiple meta tags (in the sense of others than `<meta name="monetization">`, it would require careful ordering so as to be operating on the information defined by the set of tags as a whole, removal with document.head.querySelectorAll('meta[name^="monetization"]'), then adding the name="monetization" tag last)
```
traviscrist: Should the document portions here be code blocks?

traviscrist:
@sublimator 
Chatted with ben on this, I now see the idea that when the page is reloaded the meta tag would get picked up and then a new uuid would be generated each time. 

Right now I don't think that's very clear in the updated documentation. Its also not clear how this would work for a static site since they meta tag would be part of the page. If we can add some clarity around when the uuid is generated and the overall flow of the meta tag being picked up to the uuid being created that would resolve my concerns about it. 

I'm also good with calling it monetizationId, I think it makes sense, its a breaking change but if we are going to change it we should change it now.

sublimator:
> I don't think that's very clear in the updated documentation
Agreed. Still very rough. Thanks for highlighting it

> I'm also good with calling it monetizationId
@sharafian 
What about you?

0028-web-monetization/0028-web-monetization.md
sublimator: That's just some scribble thoughts I put in for discussion/thought ... 

People will have to be careful to make sure they have all their ducks (monetization meta tags used to convey info other than a payment pointer) in a row before updating.

sublimator:
@sharafian 
> Payment Pointer or SPSP url

Do we have a single term that covers both ?
Note that the emitted monetization events currently use `paymentPointer` 
If not, do we need to constantly put `Payment Pointer or SPSP url` or just a single well placed mention that it's possible to also use an SPSP url. 

Does the fact that you might need to use an SPSP url mean the design of payment pointers is lacking somehow?

0028-web-monetization/0028-web-monetization.md
```diff
@@ -41,33 +41,37 @@ Web Monetization makes use of [SPSP](../0009-simple-payment-setup-protocol/0009-
 
 This flow refers to the user's **browser** and the user's **provider**, [defined above](#terminology).
 
-- The user navigates their browser to a webpage.
-- The browser sets `document.monetization` to an Object which implements [EventTarget](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget).
-- The browser sets `document.monetization.state` to `pending`.
-- The browser observes the `<head>` and looks for a Web Monetization `<meta name="monetization">` tag ([specified below](#meta-tags)).
+- The user's agent navigates a webpage.
```
sublimator: I changed these to consistently use `user's agent` though perhaps it should have been the other way around?

0028-web-monetization/0028-web-monetization.md
traviscrist: oh its all in a todo

0028-web-monetization/0028-web-monetization.md
```diff
@@ -41,33 +41,37 @@ Web Monetization makes use of [SPSP](../0009-simple-payment-setup-protocol/0009-
...
-- Payment continues for the lifetime of a given meta tag (or semantically equivalent)
-  - The provider MAY decide to stop/start payment at any time, e.g. if the user is idle, backgrounds the page, or instructs the browser to do so.
-  - If the STREAM connection is severed, the provider will redo the SPSP query to the same Payment Pointer as before with the same Monetization ID. The browser MUST NOT re-process the `<meta>` tags.
-  - Each time a packet with a nonzero amount is fulfilled, the provider notifies the browser, and the browser emits an event on `document.monetization`. The event's type is `monetizationprogress`. The event has a `detail` field containing the details of the packet ([specified below](#monetizationprogress)). If there are no listeners the event MAY NOT be emitted.
-  - When a stream is closed the `document.monetization.state` MUST be set back to 'pending'
+
+  - The user's agent generate a fresh **random** UUID (version 4) and uses this as a Monetization ID. This MUST be regenerated upon dynamic changes to the meta tags and MUST be unique per page load in order to avoid user tracking.
```
traviscrist: I like this, I'm not sure if we should put in the to prevent user tracking, @sharafian curious about your thoughts on this.

sharafian: it's fine to call it out explicitly. There's a grammatical error though: `generate a fresh random UUID` should be `generates a fresh random UUID`

sublimator:
@sharafian
I believe that was already fixed in a subsequent fix

sharafian:
Is this PR blocked on anything? @sublimator @traviscrist

sublimator:
I think the plan was to get experience with things working in the extension
and then make this PR match that.

0028-web-monetization/0028-web-monetization.md
```diff
@@ -139,13 +139,15 @@ Dispatched once the first ILP packet with a non-zero amount has been fulfilled b
 ```ts
 {
   detail: {
+    // Payment Setup Endpoint
     paymentPointer: String,
+    // Monetization ID
```
traviscrist: I think the requestId here would need to change to reflect the new name monetization id right?

traviscrist:
> I think the plan was to get experience with things working in the extension and then make this PR match that.

I thought this was looking good, I left one comment about the request id, @sharafian if you and I review this one more time then I think it could be merged. I'd rather make this change from request id to monetization id sooner than wait. Its kinda a big change so the sooner we make it the better.

0028-web-monetization/0028-web-monetization.md
sharafian: I'm against renaming fields unless we absolutely have to

0028-web-monetization/0028-web-monetization.md
```diff
@@ -154,13 +156,18 @@ Dispatched every time an ILP packet with a non-zero amount has been fulfilled by
...
   detail: {
+    // Received amount
     amount: String,
     assetCode: String,
     assetScale: Number
   }
 }
 ```
 
+(TODO:
```
sharafian: Let's make it clear that these are the receiving amounts then. We could introduce sent amounts addititively in a later update but for now I think we just need to get the spec up to parity with the implementation

0028-web-monetization/0028-web-monetization.md
traviscrist: If we don't rename it and keep it as monetization id doesn't that feel a little odd, maybe it should be called "Monetization Request ID". The issue is that if I see requestId and I search for request id I will not find Monetization ID in this doc. Ill only be able to find this comment to link the 2 together.

traviscrist: Talked to Ben in person. Lets just keep this as it is in the PR with the Monetization ID comment.

sublimator:
This is still very much a draft, hence the TODOs. I guess we could just
merge this in now after a bit of cleanup though as it more closely matches
reality and it is taking time. I didn’t understand Ben’s question about
“what’s the hold up?” properly before.

As for a follow up:
I think there are ways we could change the names, while keeping the old
fields (non enumerable) which console.warn deprecation messages upon access
if in our heart of hearts we want to change some names.I think the
amount/asset etc, could do with clarifying specification of src/dest. I’d
do it now/soon before much adoption.

sublimator:
I watched the Sublime Text 2 to Sublime Text 3 transition, which went from
Python 2 to Python 3 and a naming convention change for the plugin api.
There was a transition period with compatibility layers but it was possible
to make a clean break.

Changes from v1 to v2 included using json for conf over xml. There were far
less users then.

In the end it was worth it.

I don’t think we should be afraid to make changes in the early days,
especially if it can be done in a non breaking way.

As for ‘requestId’, it’s kind of grown on me/could live with it, but that’s
typical once you no longer have fresh eyes.

adrianhopebailie:
@sublimator @sharafian @traviscrist 

This spec is likely to move from here to something in the W3C format as soon as @brucelawson and I have submitted this to the Web Platform Incubator CG (WICG) later this week.

I'd recommend we look at already migrating this over to that doc (incorporating these changes), linking to it from here, deprecating this and making further changes there.

@sublimator - do you want to have a stab at that or would you prefer I do it?

The WICG has a template for new proposals which I've already used to put the following together: https://github.com/adrianhopebailie/web-monetization

The main entry point for the community is the explainer which is expected to be updated regularly based on feedback and then the spec (index.html) should be updated whenever there is consensus on changes.

The spec will need to be a balance of "what is implemented already" and "what the community says should be in the spec". i.e. Initially the implementation drives the spec but as we get more implementations the spec will become the reference and implementations will need to follow.

Keep in mind that the spec is targeted at browser vendors (they are the implementors) not web developers. BUT, the API design should make sense to developers. So, for example, the choice of property names should consider only what the developer cares about (i.e. not SPSP, STREAM or ILP).

sublimator:
@adrianhopebailie

I defer to your greater expertise here.
I'm sure Ben and Travis will want to have some input before it's 'thrown to
the wolves' though

sharafian:
@adrianhopebailie Right now there's no accurate spec of WM other than this PR so I think it's important to get this PR merged. Your spec that you linked looks good, I'm wondering where I can leave feedback on it

0028-web-monetization/0028-web-monetization.md
```diff
@@ -41,33 +41,39 @@ Web Monetization makes use of [SPSP](../0009-simple-payment-setup-protocol/0009-
 
 This flow refers to the user's **browser** and the user's **provider**, [defined above](#terminology).
```
sublimator: *browser* vs *agent*

0028-web-monetization/0028-web-monetization.md
```diff
@@ -41,33 +41,39 @@ Web Monetization makes use of [SPSP](../0009-simple-payment-setup-protocol/0009-
...
+  - The user's agent passes these payment details to the user's provider (see [Payment Handler Flow](#payment-handler-flow)).
+  - The provider resolves the Payment Setup Endpoint and begins to pay. The provider MAY be acting from a different machine from the user. Cookies and session information MUST NOT be carried with any requests made to resolve the Payment Setup Endpoint, with the exception of the Monetization ID.
+    - On the SPSP query to resolve the Payment Setup Endpoint, a `Web-Monetization-Id` header is sent, containing the Monetization ID. The server running the web-monetized site may use this to associate future requests by the user with their payments.
+    - With the `destination_account` and `shared_secret` fetched from the SPSP query, a STREAM connection is established. A single STREAM is opened on this connection, and a positive SendMax is set.
+    - The provider SHOULD set their SendMax high enough that it is never reached, and make payment adjustments by limiting throughput.
+  - Once the STREAM connection has fulfilled an ILP packet with a non-zero amount, the provider notifies the browser:
```
sublimator: @sharafian @traviscrist 

The phrase:  "the provider notified (grammar error?) the browser, and the browser dispatches an event on `document.monetization`."  kind of threw me for bit, as I misread it as "provider notifies browser VIA dispatching an event on document.monetization"

In terms of "the provider notified the browser":

The oauth polyfill just "does it" and the "provider" in the case of the extension sends a wext runtime 'monetizationProgress' message to the content script.

0028-web-monetization/0028-web-monetization.md
```diff
@@ -139,13 +134,15 @@ Dispatched once the first ILP packet with a non-zero amount has been fulfilled b
 ```ts
 {
   detail: {
-    paymentPointer: String,
-    requestId: String
+    // Payment Setup Endpoint
+    paymentPointer: string,
+    // Monetization ID / Web-Monetization-ID header
```
sublimator: Added these comments here for correlation

0028-web-monetization/0028-web-monetization.md
sublimator: Also added Web-Monetization-ID header to the comment

adrianhopebailie:
@sharafian 

> Right now there's no accurate spec of WM other than this PR so I think it's important to get this PR merged.

Agree, let's get it merged and then we can migrate it to the W3C template. Can we aim to have that done this week?

> Your spec that you linked looks good, I'm wondering where I can leave feedback on it

The WICG process is to have a dedicated repo for proposals so use the issue list on that repo to log any comments about the spec or the explainer.

0028-web-monetization/0028-web-monetization.md
adrianhopebailie: This is the as-yet-undefined interface from provider to browser that we need to  design and implement more explicitly.

I don't think we need to do that in this PR but we should start thinking about soon and include in the W3C proposal.

sublimator: Take a crack :) 
I might have a go tomorrow or the weekend. 
Sleepy McJoe today :)

sharafian:
Merging on behalf of Niq

PR #531, opened 2019-05-16
Improve ILP-over-HTTP Auth

sappenin:
* Introduce formal Authentication Profiles for ILP-over-HTTP
* Clarify normative auth requirements for ILP-over-HTTP
* Provide recommendations around various best practices relating to performance, security, and operational deployment.

Fundamentally, we need at least one Authentication scheme that all Connectors running ILP-over-HTTP implement, and that scheme should attempt to satisfy the following:

- The scheme should be as simple as possible (emphasis: “as possible”. This doesn’t mean “simplicity over security”. Instead, it means there’s a tradeoff, and we want to find the right balance).
- The scheme should be required for all Connectors to implement in order to encourage broad deployment.
- The scheme should consider high-performance (to support high packet throughput).
- The scheme should resist replays (emphasis: resist, not necessarily be replay-proof, in order to aid in high-performance if required).
- The scheme should not negate or prevent the ability of operators to layer-on more security on top of it, depending on the requirements and characteristics of any given deployment.

Signed-off-by: sappenin <sappenin@gmail.com>


0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -56,3 +52,115 @@ All ILP Packets MUST be returned with the HTTP status code `200: OK`.
...
+
+In order to find this balance, this document defines two Authentication profiles, each with various trade-offs that should be considered before use:
+
+* `SIMPLE`: Allows two ILP nodes to utilize a previously agreed-upon shared-secret as a [Bearer token](https://tools.ietf.org/html/rfc6750) in all ILP-over-HTTP requests. Peers SHOULD consider this token to be opaque and SHOULD NOT derive any special meaning from the token. 
+
+* `JWT_HS_256`: Allows two ILP nodes to utilize a previously agreed-upon shared-secret in order to _derive_ a Bearer token that conforms to the JSON Web Token (JWT) specification. JWTs generated by this scheme can then be used as a Bearer in all ILP-over-HTTP requests.
+
+Peers MAY use any standard HTTP authentication mechanism to authenticate incoming requests, but SHOULD support both of the authentication profiles defined in this RFC, and MUST support `JWT_HS_256`.
+ 
+In addition, TLS Client Certificates are RECOMMENDED between peers for additional security. Basic authentication (username and password) is NOT RECOMMENDED for the same reasons that the `SIMPLE` profile is only reccommended for development and testing scenarios only.
```
adrianhopebailie: I'd consider TLS Client Certs to be a third authentication profile that I think negates the need for the others. It's the hardest to do but probably best because it doesn't require each request to be authenticated once the transport is authenticated.

0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -56,3 +52,115 @@ All ILP Packets MUST be returned with the HTTP status code `200: OK`.
...
+
+#### `SIMPLE` Authentication Profile
+This profile allows two ILP nodes to utilize a previously agreed-upon shared-secret that contains at least 32 bytes (256 bits) of randomly generated data, and is encoded using Base64.
+
+Because tokens in this profile do not inherently contain information about the identity of the caller, requests MUST contain an additional HTTP request-header named `Auth-Principal`.
```
adrianhopebailie: Good header name choice 😄

0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -56,3 +52,115 @@ All ILP Packets MUST be returned with the HTTP status code `200: OK`.
...
+ 
+This extra header allows for the identity of the authentication request to be separated from authentication token itself, which reduces computational overhead as well as data-management complexity (e.g., implementations do not need to create data-store indexes using derivations of tokens for lookup purposes).
+
+Depending on the implementation, the value of this header SHOULD represent an `ILP Account Id`, but future implementations MAY make this represent an `ILP Peer Id` or some other value.
```
adrianhopebailie: Don't think this paragraph is necessary.

0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -56,3 +52,115 @@ All ILP Packets MUST be returned with the HTTP status code `200: OK`.
...
+
+##### Example Usage
+An example shared-secret in this profiles is `HEiMCp0FoAC903QHueY89gAWJHo/izaBnJU8/58rlSI=`. This shared secret is passed as an `Authorization` header in each HTTP request, using the Bearer token scheme, along with an `Auth-Principal` header like this:
```
adrianhopebailie: s/profiles/profile/

0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -56,3 +52,115 @@ All ILP Packets MUST be returned with the HTTP status code `200: OK`.
...
+  * The shared-secret itself never expires, so if an implementation neglects to rotate the secret with its peer, this token will likely be very long-lived. This increases the chance of compromise by an attacker, and means compromised usage of this type of token could go undetected for very longer periods of time.
+  * Requires out-of-band communication for both peers to agree upon the shared secret.
+
+#### `JWT_HS_256` Authentication Profile
+This profile allows two ILP nodes to utilize a previously agreed-upon shared-secret, but then derive an [RFC-7519](https://tools.ietf.org/html/rfc7519) compliant JWT token in order to perform actual authentication.
+
+##### JWT Claims
+In order to be considered a valid JWT for this profile, the signed JWT MUST contain a `sub` (subject) claim containing the identifier of the "principal" that the token authenticates. Depending on the implementation, the value of this claim SHOULD represent an `ILP Account Id`, but future implementations MAY make this represent an `ILP Peer Id` or some other value.
```
adrianhopebailie: As above I'd drop this because there is a lot of implicit meaning in those names which we don't (need to) define:

> Depending on the implementation, the value of this claim SHOULD represent an `ILP Account Id`, but future implementations MAY make this represent an `ILP Peer Id` or some other value.

IMO what you have is enough, its standard authentication/identity terminology to simply say the "principal".

0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -56,3 +52,115 @@ All ILP Packets MUST be returned with the HTTP status code `200: OK`.
...
+
+Optionally, a JWT token SHOULD also include an `exp` (expiry) claim that indicates a date/time after which the token should be considered invalid. Note that tokens without this claim never expire.
```
adrianhopebailie: Maybe add:

> and only become invalid if the shared secret changes used to sign the JWT changes.

> Implementations SHOULD require that all tokens contain an expiry claim and MAY define a maximum value for this claim that they consider safe and reject any tokens with a missing or invalid expiry claim.

0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -56,3 +52,115 @@ All ILP Packets MUST be returned with the HTTP status code `200: OK`.
...
+
+* **Cons**
+  * Total transmitted bytes for authentication are more than the `SIMPLE` scheme (about 41 bytes, or ~50% more). However, HTTP/2 header compression should mitigate this differential.
+  * Requires out-of-band communication for both peers to agree upon the shared secret.
+
+## Appendix1: Security Best Practices
+This section outlines and clarifies some best practices for authentication-token security when using this protocol. Recommendations in this section are _Non-Normative_, but highly RECOMMENDED.
+
+### Follow Standardized Security Recommendations
+It is advisable to follow any all all applicable best practices when using a Bearer-token scheme for authentication. [Section 5.1 of RFC-6570](https://tools.ietf.org/html/rfc6750#section-5) contains a number of very good practices that should be considered on a per-deployment basis.
```
adrianhopebailie: s/any all all/any and all/

0035-ilp-over-http/0035-ilp-over-http.md
sappenin: Hmmm...great point! I think I'll just move this to the end under "Recommendations" as something extra to consider for a secure infrastructure. 

I think that using TLS Client Certs _could_ negate the other profiles, but there are deployment scenarios where both are necessary (specifically, edge-terminated TLS connections, which are unfortunately _very_ common).

Checkout my new verbiage to see what you think.

sappenin: Hmmm...great point! I think I'll just move this to the end under "Recommendations" as something extra to consider for a secure infrastructure. 

I think that using TLS Client Certs _could_ negate the other profiles, but there are deployment scenarios where both are necessary (specifically, edge-terminated TLS connections, which are unfortunately _very_ common).

Checkout my new verbiage to see what you think.

0035-ilp-over-http/0035-ilp-over-http.md
sappenin: 😄

0035-ilp-over-http/0035-ilp-over-http.md
sappenin: Dang, Yeah, I went back and forth on this one. I'd like to remove this paragraph actually, but much of the impetus of this PR is around me trying to peer the Java implementation with other implementations, and needing clarity on certain specifics of ILP-over-HTTP (the Auth systems here are subtly different between Rust, Java, and JS).

If we remove this paragraph, then the meaning of the Principal might be different between implementations (e.g., see [my comments here](https://forum.interledger.org/t/connector-auth-model/517/4?u=sappenin) for one way this could be).

I'm unclear if this is OK or not. For example, in [java-ilp-connector](https://github.com/sappenin/java-ilpv4-connector) I'm going to need to build a Java client that can talk to other connectors. Let's say I'm going to (somehow) work it out with this other Connector that I want to use one of the profiles in this PR. Great. Now let's imagine that my Java Connector wants to use ILP with the other connector over two different accounts (e.g., a USD account and an XRP account). In the Java impl, the current implementation will demand that there be two different HTTP connections, so I'm going to need two different Auth-Principals.

But what if the _other_ Connector in this example actually implements Peer-based Authentication where an ILP-over-HTTP connection can be initiated on a per-peer basis (i.e., One HTTP connection for two peers), and multiple ILP accounts between those peers can be used over the same ILP-over-HTTP connection? In this world, the Java implementation is immediately not going to work because in order to operate on the two accounts in my hypothetical above, the Java Client is going to need a third piece of data -- it will first need the `AuthPrincipal` (Identifies me as `Peer123`); It will then need an Account identifier (USD vs XRP account); and then it will need a credential (e.g., a shared-secret).

Ideally, I'd _love_ not to have to standardize this meaning, but is that reasonable/advisable? Doesn't this meaning need a normative definition?

So, my questions are:

1. How do Connector implementations know what the other party requires means by the `Auth-Principal`? 
1. Does it matter?
1. What happens when two different Connector Implementations decide to define the `Auth-Principal` differently?

0035-ilp-over-http/0035-ilp-over-http.md
sappenin: Fixed.

0035-ilp-over-http/0035-ilp-over-http.md
sappenin: Fixed.

0035-ilp-over-http/0035-ilp-over-http.md
sappenin: Agreed! I worked your wording into the existing text.

0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -56,3 +52,123 @@ All ILP Packets MUST be returned with the HTTP status code `200: OK`.
...
+  * Requires out-of-band communication for both peers to agree upon the shared secret.
+
+#### `JWT_HS_256` Authentication Profile
+This profile allows two ILP nodes to utilize a previously agreed-upon shared-secret, but then derive an [RFC-7519](https://tools.ietf.org/html/rfc7519) compliant JWT token in order to perform actual authentication.
+
+##### JWT Claims
+In order to be considered a valid JWT for this profile, the signed JWT MUST contain a `sub` (subject) claim containing the identifier of the "principal" that the token authenticates. Depending on the implementation, the value of this claim SHOULD represent an `ILP Account Id`, but future implementations MAY make this represent an `ILP Peer Id` or some other value.
+
+A JWT token SHOULD also include an `exp` (expiry) claim that indicates a date/time after which the token should be considered invalid. Implementations SHOULD reject any tokens with a missing or invalid expiry claim.
```
sappenin: In JWT-land, there are two _other_ **very common** claims that are typically included in a JWT, and they are probably worth talking about in the context of ILP-over-HTTP:

* The [`iss` claim](https://tools.ietf.org/html/rfc7519#section-4.1.1) identifies the principal that issued a JWT (in ILP-over-HTTP, this would be the URL of the sender of an ILP Prepare packet).
* The [`aud` claim](https://tools.ietf.org/html/rfc7519#section-4.1.3) narrows the scope of the token to only being valid at a particular peer (in ILP-over-HTTP, this would be the URL of the receiver of an ILP Prepare packet).

RFC-7519 clearly marks these two claims as optional, though it does define concrete behaviors _if_ these claims are present in a JWT.

For the purposes of this Authentication Profile in this RFC (i.e., `JWT_HS_256`), I can essentially think of two attack vectors where these claims _may_ have an impact:

* **Shared Secret is Compromised**: If an attacker were to compromise the underlying shared-secret used to generate the `HS_256` JWTs, then the attacker could simply forge the `iss` and `aud` claims, so including them in tokens conforming to this RFC doesn't seem to add any protection against that scenario.

* **Token Compromised**: If an attacker were to somehow intercept a JWT in-transit (or in a log file or wherever), the attacker would not be able to mint _new_ tokens, but the attacker could make calls against either Peer in an ILP-over-HTTP relationship. Adding the `aud` claim would help to narrow the places where this token could work to _only one of the peers_. This is because during JWT verification, each peer would verify the `aud` claim, and without the underlying shared-secret, this value could not be forged. For example, if my URL is `https://david.example.com`, and I receive a token with an `aud` of `https://alice.example.com`, I would reject this token because it was likely generated by me and meant for my peer (`alice` in this case). In other words, I would only accept tokens containing a signed `aud` claim with a value of `https://david.example.com`.

I'm inclined to update this proposal to include the `aud` claim as a normative requirement for two reasons:

1. It reduces the attack space between any two ILP-over-HTTP peers by 50% in the event of a compromised token.
1. Both Java libraries that I have investigated require the `iss` and `aud` claims, so I would like to confirm if there might be something deeper at play from a JWT security perspective relating to these two claims. I have an [outstanding question in one of the projects](https://github.com/auth0/java-jwt/issues/334) to see if that's true.

0035-ilp-over-http/0035-ilp-over-http.md
adrianhopebailie: > How do Connector implementations know what the other party means by the Auth-Principal?

They figure it out before peering

> Does it matter?

Not in a normative sense, I don't think so

> What happens when two different Connector Implementations decide to define the Auth-Principal differently?

They need to figure this out before peering because they need to figure out a bunch of other stuff like extra headers or claims in the JWT to indicate, for example, "the account" when the principal is a peer.

i.e. Let's stick with the simple case of two peers auth with each other and then internally map the principal of that auth to some internal accounting system. For now we will all have an implicit 1:1 mapping of auth principal to account but this may evolve in time at which point we'll need to define how that would work:

1. One auth and additional headers/claims to identify an account, or
2. A different auth principal per account

0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -150,6 +151,11 @@ One option is to encrypt any secrets at-rest using an Authenticated Encryption a
   
 ### Secrets In Memory
 Implementations SHOULD minimize the amount of time that an actual secret-value exists in-memory in unencrypted form. This includes narrowing the availability of secrets to only code that actually requires them; minimizing the time any secret might exist in memory; and zeroing out memory after a secret is no longer used, if possible. 
+
+### Mutual TLS
+All ILP-over-HTTP connections MUST be performed over a TLS session. However, it is also RECOMMENDED to use TLS Client Certificates between peers for additional _network-level_ security. 
+
+Note that while possible, it is not advised to establish the identity of an ILP Principle using Client Certificates because many network operators will terminate TLS connections at the network edge, and then forward traffic on to actual backends in a new connection (e.g., a new TLS session or simply unencrypted). In TLS-termination scenarios, the actual backend will not be able to validate the original Client Certificate identity, so a higher-level Authentication scheme (like tokens) is required.
```
adrianhopebailie: I disagree with this as it is making big assumptions about how someone will use and deploy TLS.

There are notes here that are good for people to consider when using Client Certs but I wouldn't go so far as to say this is not advised.

On the contrary it is likely to be most efficient way of doing auth as long as the parties do it correctly. 

From what I have observed for example, many cloud providers support Client Cert **validation** and then pass the cert along as a header for the application to then perform **authentication**. 

Once this is done once by the application the result can be cached for a long time because the cert will seldom change (Certificates have much longer lifetimes than JWT's for example).

sappenin: Fair points. I've removed this section entirely. This spec doesn't really need to cover any of this other than bringing Mutual TLS to the attention of the reader (IMHO).

0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -56,3 +52,123 @@ All ILP Packets MUST be returned with the HTTP status code `200: OK`.
...
+  * Allows both identity and authentication claims to be contained in single Bearer token, which eliminates the need for a second `Auth-Principal` header. This should simplify token validation and ease account-lookup (no need to create an HMAC or encrypted data-store index because the principal can be trivially decoded from the token).
+  * Requires only enough processing to perform an HMAC-SHA256 signing operation, which is very fast.
+  * Supports token expiry, which allows tokens to be generally short-lived so that peers can narrow the potential window of unauthorized usage in the event of token compromise.
+  * The actual shared-secret is _never_ transmitted "on the wire" during any request. Instead, authentication tokens are always _derived_ from the shared-secret, which eliminates the risk of an _actual_ shared-secret being intercepted  in transit.
+
+* **Cons**
+  * Total transmitted bytes for authentication are more than the `SIMPLE` scheme (about 41 bytes, or ~50% more). However, HTTP/2 header compression should mitigate this differential.
+  * Requires out-of-band communication for both peers to agree upon the shared secret.
+
+## Appendix1: Security Best Practices
```
sappenin: @adrianhopebailie Do you feel fine leaving this appendix in this RFC? I tried to keep the recommendations specific to tokens, so this is _probably_ the right place...but am open to other ideas.

0035-ilp-over-http/0035-ilp-over-http.md
sappenin: > Not in a normative sense, I don't think so

OK, cool. I like this flexibility and rationale, and [have removed](https://github.com/interledger/rfcs/pull/531/commits/24bca61866b9e7f4004fdd546cfc57f7515a5f7e) any normative definitions of "Authentication Principal" in the spec.

> Let's stick with the simple case of two peers auth with each other and then internally map the principal of that auth to some internal accounting system

Agree. But just to clarify your points, you're saying that even though this is what _most_ of the implementations will do today, we don't need to mandate this...and we don't even want to mention it in the RFC, correct?

I'm open to that path, though I would support a small sentence like this, as a guide to implementors:

`the value of this header SHOULD represent an "ILP Account Id"` `

But I don't feel strongly about including this here, especially given how much we've discussed this in the past. Just wanted to raise this one last time to be sure I've teased out your preference here.

0035-ilp-over-http/0035-ilp-over-http.md
sappenin: This is now [removed](https://github.com/interledger/rfcs/pull/531/commits/24bca61866b9e7f4004fdd546cfc57f7515a5f7e).

0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -56,3 +52,119 @@ All ILP Packets MUST be returned with the HTTP status code `200: OK`.
...
 
+### Authentication
+When authenticating requests between Interledger nodes, it is important to choose an algorithm that maintains an appropriate balance between usability and security, while at the same time maintaining high-performance _and_ interoperability.
+
+In order to find this balance, this document defines two Authentication profiles, each with various trade-offs that should be considered before use:
+
+* `SIMPLE`: Allows two ILP nodes to utilize a previously agreed-upon shared-secret as a [Bearer token](https://tools.ietf.org/html/rfc6750) in all ILP-over-HTTP requests. Peers SHOULD consider this token to be opaque and SHOULD NOT derive any special meaning from the token.
```
sappenin: This morning it occurred to me that since this RFC has added the `Auth-Principal` header, this profile is now _exactly_ equivalent from a security perspective to HTTP BASIC Auth. Except that it uses different headers. 

This had me wondering if we shouldn't just replace the `SIMPLE` bearer scheme with HTTP Basic Auth so that implementations don't have to implement something custom -- almost every HTTP client library in every language will support BASIC auth.

0035-ilp-over-http/0035-ilp-over-http.md
adrianhopebailie: > I'm open to that path, though I would support a small sentence like this, as a guide to implementors:

> `the value of this header SHOULD represent an "ILP Account Id" `

I think it is worse with it in for this reason; What IS an `ILP Account Id`? The sentence adds no value without a definition and I don't think this is the spec to define it.

0035-ilp-over-http/0035-ilp-over-http.md
adrianhopebailie: The goal is interop so we should define the minimum requirements of implementations to be interoperable knowing that that will become the most commonly used profile so it should be well designed.

Client certs are def too complex to make a normative requirement.

0035-ilp-over-http/0035-ilp-over-http.md
kincaidoneil: > I think it is worse with it in for this reason; What IS an ILP Account Id? The sentence adds no value without a definition and I don't think this is the spec to define it.

Agreed (and I know there's been a lot of back and forth on accounts vs peers), but I do think it's critical that we define what an account is.

The settlement engine spec requires an account identifier shared between itself, the accounting system, and the transport (or at least a way to map that account id to a peer to send messages to, and vice-versa). Also, to support a REST with the account id in the URL, e.g. `POST /account/:id`, we need the account ID to be URL encode-able, which is an additional constraint.

0035-ilp-over-http/0035-ilp-over-http.md
sappenin: > we should define the minimum requirements of implementations to be interoperable

Totally agree - I made the intro to this section (i.e., `Appendix1`) very overt to say that these are only recommendations, and are all non-normative. 

I guess my question was: is this RFC the right place to put a bunch of non-normative security recommendations, or does that kind of thing belong somewhere else?

0035-ilp-over-http/0035-ilp-over-http.md
sappenin: After building an implementation of this PR in Java today, it counterintuitively turns out that I can re-use the `JWT_HS_256` Authentication provider for _both_ profiles (`SIMPLE` and `JWT_HS_256`). 

This is because I can implement it once for `JWT_HS_256`, and then to support `SIMPLE`, I simply give my peer an `HS_256` JWT with a very long expiry. From my peer's perspective, this just an opaque token, which it happily re-uses for a very long time. 

From that perspective, keeping `SIMPLE` as a Bearer scheme actually reduces implementation work on the server. For the outgoing client though, it's a bit more of a pain, and using BASIC auth would be less work.

All that to say, the more I think about it, I feel like even defining `SIMPLE` is maybe a mistake. The PR is quite overt that nobody should use it in production, so I wonder if we just remove it entirely from the PR so that there's only one authentication profile that all implementations must normatively support.

Any strong thoughts or opinions about keeping or getting rid of the `SIMPLE` profile?

0035-ilp-over-http/0035-ilp-over-http.md
sappenin: > What IS an `ILP Account Id`?

Hmmm....I agree that it's technically not normatively defined anywhere, but the concept of an Account is [pretty-well-covered](https://github.com/interledger/rfcs/blob/master/0001-interledger-architecture/0001-interledger-architecture.md) by RFC, so it shouldn't be a stretch to know what this is. 

In a sentence, it's: "the unique identifier of an Account". 

What's an Account, one might ask? That's easy: go read RFC-0001.

sappenin: > I do think it's critical that we define what an account is (and maybe that's here, maybe that's somewhere else).

I agree with this -- though correct me if I'm wrong @kincaidoneil but I think you're saying, "I do think it's critical that we define what an **accountId** is", right?

sappenin: All that said, I think I'm of the persuasion right now that AccountId is just a String or whaterver two peers want to make it, and so maybe we don't need to normatively define it's _format_ or _structure_. 

But as a concept, I think it should be unambiguous in Interledger to use the term `AccountId` and have people know what that means (as a concept at least).

If that requires a spec somewhere, then I say we add it somewhere.

0035-ilp-over-http/0035-ilp-over-http.md
adrianhopebailie: No, I think this is a fine place. 

Although in some respects I'd say we could use this same auth stuff for WebSocket based transports too so maybe a stand-alone spec will be the way we go in the long run...

0035-ilp-over-http/0035-ilp-over-http.md
adrianhopebailie: Happy to get rid of it. It did occur to me on first reading it that a JWT with long expiry would probably be fine for testing and COULD be generated by the server anyway.

0035-ilp-over-http/0035-ilp-over-http.md
kincaidoneil: @sappenin Yes (thank you for clarifying!), I meant we should define what an _account id_ is.

> All that said, I think I'm of the persuasion right now that AccountId is just a String or whaterver two peers want to make it, and so maybe we don't need to normatively define it's format or structure.

But I'd say what the account ID represents is very important.

Suppose you have an incoming `peer.settle` packet, and the auth mechanism specifies a peer ID, but not an account ID (one previous topic of discussion). Suppose, if that peer has multiple accounts, when the connector forwards that packet to the settlement engine, which account ID should it specify?

Which is to say, I think it's important we have a way to link incoming packets to the account identifier shared with the settlement engine (which is very related to this spec).

adrianhopebailie: > I think it's important we have a way to link incoming packets to the account identifier shared with the settlement engine 

But not as part of the minimum requirements for auth between peers which is what we're trying to define here.

Right now we have an implicit 1:1 association between the entity that has auth'd with me and the account I track for them.

When the possibility exists one day that the entity that auth's with me may have multiple accounts then we'll need to define a way to identify the account but that may never be required so I don't think we need to define it.

Let's start with what we need now and add normative requirements as we discover we need them later.

0035-ilp-over-http/0035-ilp-over-http.md
sappenin: > No, I think this is a fine place.

Cool, I'll keep it for now.

0035-ilp-over-http/0035-ilp-over-http.md
sappenin: Ok, so summarizing this topic, I see three options. 

I'll list them here with what I see as Pros/Cons for each, then see how everyone feels via Thumbs-up/Thumbs down and/or comments (if you have a "pro" or a "con" that you want added, send it my way and I'll edit this comment).

# Choices for defining the `Auth Principal`
Summarizing this topic, I see three options:

1. **Stay Silent**:  
    1. The RFC is silent as to the meaning of the `sub` claim (aka, the `Auth-Principal`) in the JWTs defined in the `JWT_HS_256` Authentication profile.

1. **Define One JWT Auth Profile**: 
    1. The RFC normatively defines `AccountId` as being the unique identifier of an [Account](https://github.com/interledger/rfcs/blob/master/0032-peering-clearing-settlement/0032-peering-clearing-settlement.md#accounts-and-balances),
    1. For the `JWT_HS_256` auth profile, the RFC normatively defines the `sub` claim of the JWT as representing an `AccountId`. 
    1. If that definition ever needs to change, a new Authentication Profile will be defined.

1. **Define Two JWT Auth Profiles**: 
    1. In addition to defining the above, the RFC _also_  proactively defines a 2nd Authentication Profile (e.g., `JWT_HS_256_PEER`) that is identical to the above except that this 2nd Auth profile instead defines `AccountId` as being the identifier of a [Peer](https://github.com/interledger/rfcs/blob/master/0032-peering-clearing-settlement/0032-peering-clearing-settlement.md#accounts-and-balances).

# Pros/Cons of "Silent"
1. Pro: Allows the most flexibility for current implementations because they can choose any meaning for the Auth Principal that they would like.
1. Con: Creates ambiguity in implementations because implementations can't rely on any particular meaning of the Auth Principal when speaking to another supposedly ILP-over-HTTP compliant system (so any two peers would need to manage this meaning, and potentially code around this meaning, on a case-by-case basis).

# Pros/Cons of "1 Profile"
1. Pro: Clarifies the  meaning of the `sub` claim used in the `JWT_HS_256` profile; allows room for a future profile to be defined; doesn't prematurely define an Auth-Profile that is not currently used or being requested by anyone (i.e., choice 3 below).
1. Con: Precludes the implementation of Peer-wise Auth without an RFC revision to add a 2nd profile (for Peer-wise Auth).

# Pros/Cons of "2 Profiles"
1. Pro: Still clarifies meaning; Defines two different Auth profiles that implementations can choose from, allowing for future possibilities.
1. Con: Specifies Peer-based auth, which is not currently in-use today, and could easily be defined in a future RFC revision if broadly supported, implemented, or required.

0035-ilp-over-http/0035-ilp-over-http.md
sappenin: > Happy to get rid of it.

I'm leaning towards removing `SIMPLE`. I will wait a day or two in order to give others a chance to comment.

0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -56,3 +52,119 @@ All ILP Packets MUST be returned with the HTTP status code `200: OK`.
...
+
+One option is to encrypt any secrets at-rest using an Authenticated Encryption algorithm, such as AES-GCM-256. This [article](https://proandroiddev.com/security-best-practices-symmetric-encryption-with-aes-in-java-7616beaaade9) provides a nice overview using Java, along with various configuration options to consider.
+  
+### Secrets In Memory
+Implementations SHOULD minimize the amount of time that an actual secret-value exists in-memory in unencrypted form. This includes narrowing the availability of secrets to only code that actually requires them; minimizing the time any secret might exist in memory; and zeroing out memory after a secret is no longer used, if possible. 
+
+### Mutual TLS
+All ILP-over-HTTP connections MUST be performed over a TLS session. However, it is also RECOMMENDED to use TLS Client Certificates between peers for additional security.
```
tarcieri: Mutual TLS authentication is the industry standard for online payment APIs, particularly between things like banks and financial services companies.

If you're using mutual TLS, you've already established the peer identity, and there's no need to use a secondary mechanism (e.g. JWT claims) to do as such.

Unlike bearer credentials such as JWTs, X.509 peer identity is bound to the secure channel. This has pros and cons (mutual TLS makes authentication through a proxy difficult, where tokens generally carry higher exfiltration risk, and JWTs in particular risk immature/deficient implementations of a standard with a poorly thought out, overcomplicated cryptographic design), but I think the most important is that determining the validity of the remote peer can be handled directly by the TLS stack / application framework, as opposed to writing some ad hoc introspection logic for JWTs.

I think X.509 + token based solutions can be helpful, but much less so when the token is claims-bearing (which in my book means an AuthN credential). A much better use of tokens-over-mutual-TLS, IMO, is using tokens for fine-grained AuthZ and floating AuthZ policies. If you'd like to explore this, my recommendation would be to start by keeping the wire format of the tokens opaque, but stipulating a standard authorization workflow (though I dislike it for several reasons, OAuth2 fits the bill).

tl;dr: I would like to see this section expanded to include ways of obtaining identity information from the X.509 certificate presented by the peer during the TLS handshake. In particular I think it'd be interesting to explore ways of using Subject Alternative Names (SANs) in conjunction with things like ILP addresses.

sappenin: Hey @tarcieri, thanks for your input here!

> If you're using mutual TLS, you've already established the peer identity, and there's no need to use a secondary mechanism (e.g. JWT claims) to do as such.

In theory, this is true, and two years ago I would have agreed with you -- but in practice I've found that "there's no need to use a secondary mechanism" to be incorrect. The primary reason is that _many_ entities (you can use "banks" here, but it's broader than just banks) terminate TLS connections at the network edge, and then proxy those connections (sometimes over HTTP, believe it or not) to the actual backend.

In these deployments (where TLS is terminated), the statement that "X.509 peer identity is bound to the secure channel" is no longer true. The concept of authenticated identity has now been shifted to be: "Whoever the edge device says you are", which IMHO is a problem. 

> I think X.509 + token based solutions can be helpful

In my experience, because of the way entities outside of the control of software developers deploy their TLS connections, there _must_ be a higher-level authentication scheme that exists outside of TLS.

> If you'd like to explore this, my recommendation would be to start by keeping the wire format of the tokens opaque, but stipulating a standard authorization workflow

Setting aside authorization for a moment, I think your suggestion about "keeping the wire format of the tokens opaque" is the opinion I held about a month ago (i.e., keeping the tokens opaque), but after trying to build something, I ran into the trade-offs outlined in the PR (see [here](https://github.com/interledger/rfcs/pull/531/files#diff-ccfda9b7df50a9cfd8618bb9c08e360aR83)) when using a `SIMPLE` token. To your point, this is because that scheme lives "on top of" any Mutual TLS authentication that might exist.

I'm open to more feedback/ideas here though. The only reason I arrived at the `JWT_HS_256` profile defined in this PR is because of the constraints I outlined (i.e., can't rely upon TLS sessions not getting terminated + simple tokens have drawbacks).

> In particular I think it'd be interesting to explore ways of using Subject Alternative Names (SANs) in conjunction with things like ILP addresses.

I've explored similar ideas for other payment networks, and think this could be a very interesting option for a part of an ILP identity layer. The questions are many though - I've started a [forum thread here to discuss](https://forum.interledger.org/t/are-ilp-addresses-tied-to-identity-at-all-are-they-ephemeral/605). Would love to hear more thoughts from you and others on that one.

tarcieri: > In these deployments (where TLS is terminated), the statement that "X.509 peer identity is bound to the secure channel" is no longer true. The concept of authenticated identity has now been shifted to be: "Whoever the edge device says you are", which IMHO is a problem.

I noted this to a certain extent earlier when I enumerated pros and cons. However, I'll also note that with a token-based solution, compromise of an edge terminator has an analogous but slightly less severe problem in that attackers can replay observed tokens (unless they are sufficiently time limited).

This is the fundamental risk of edge termination. I'll quickly note an alternative to edge termination is SNI inspection.

All that said: the information being provided is peer identity, a.k.a. AuthN/"claims". To avoid compromise of the edge terminator becoming a catastrophe, as I mentioned earlier supplemental AuthZ credentials can also be required.

The advantage of relying on mutual TLS for identity and using tokens for AuthZ only is the token can be channel bound / "contextually confined" to a particular secure channel. This won't stop an attacker who pops the edge terminator, but it will stop one who exfiltrates a client-side token.

As written this suggests using mutual TLS, then ignoring the TLS peer information, in which case why are you using mutual TLS? Making the peer identity information encoded in the token a caveat/assertion (as in an AuthZ based scheme), rather than making the token the authority for it (as you're proposing), allows you to actually leverage the security provided by the MTLS channel.

The main disadvantage of relying on bearer credentials alone is exfiltration risk: there are many secure storage options for TLS private keys, including TPMs/TEEs/SEPs as well as HSMs and hardware tokens. It's significantly easier to keep them out-of-band of application logic and thereby prevent unintentional exposure.

I'll quickly note there's [RFC8471 Token Binding](https://tools.ietf.org/html/rfc8471) as a general purpose solution for binding tokens to a secure channel using an automatically provisioned self-signed certificate, however it's quite complex and not widely implemented (I believe Chrome has abandoned their efforts to implement it).

I'll footnote this by saying you can find me arguing against client certificates in an [unrelated conversation about API authentication for Rust's crates.io package repository](https://internals.rust-lang.org/t/ultra-pre-rfc-client-certificates-for-cargo-instead-of-shared-tokens/10173/8?u=bascule). The difference there is the tokens are being used to identify human principals rather than services.

But as I noted in that thread as well: "mutual TLS vs tokens" isn't a dichotomy. The two can be combined and leveraged as a force multiplier. I'd suggest you don't paint yourself into a corner and do attempt to leverage TLS peer identity info when available.

0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -56,3 +52,119 @@ All ILP Packets MUST be returned with the HTTP status code `200: OK`.
...
+One option is to encrypt any secrets at-rest using an Authenticated Encryption algorithm, such as AES-GCM-256. This [article](https://proandroiddev.com/security-best-practices-symmetric-encryption-with-aes-in-java-7616beaaade9) provides a nice overview using Java, along with various configuration options to consider.
+  
+### Secrets In Memory
+Implementations SHOULD minimize the amount of time that an actual secret-value exists in-memory in unencrypted form. This includes narrowing the availability of secrets to only code that actually requires them; minimizing the time any secret might exist in memory; and zeroing out memory after a secret is no longer used, if possible. 
+
+### Mutual TLS
+All ILP-over-HTTP connections MUST be performed over a TLS session. However, it is also RECOMMENDED to use TLS Client Certificates between peers for additional security. 
+  
+### High Security Deployments
+For deployments requiring very high security, it is recommended to utilize a secret-store deployed outside of the Connector runtime, such as [Vault](https://www.vaultproject.io/) or an [HSM](https://safenet.gemalto.com/data-encryption/hardware-security-modules-hsms/) or both.
```
sharafian: I don't think we should have this recommendation until someone actually does this

tarcieri: I'd suggest making the wording here a bit more generic, e.g. "key management service" and "hardware key storage"

sharafian: Could also clarify whether the suggestion to use a key management service rather than storing plaintext secrets in the config, or if the connector should be requesting signed JWTs from some external service rather than generating them itself and therefore having access to the secret in memory

sappenin:
Closing this in-favor of https://github.com/interledger/rfcs/pull/559.

Rather than mandating an auth mechanism as part of this RFC, #559 instead pitches the auth profiles as options instead of something normative so that any two peers can work out whatever auth they prefer, with the recs in #559 as a good starting point.

PR #532, opened 2019-05-30
Add images for the unidirectional-flow and bidirectional-flow.

dora-gt:
Add images and some minor fixes.

Signed-off-by: Taiga Nakayama <dora@dora-gt.jp>


dora-gt:
@sappenin Could you check?

PR #533, opened 2019-06-04
Fix the behavior of the Bidirectional Mode

dora-gt:
As I talked with @emschwartz  here: https://github.com/emschwartz/interledger-rs/pull/71#discussion_r289644371
it might be better to describe the criteria for determining echo packets.

It might be also better to refer to this from `0027-interledger-protocol-4.md` because the echo prefix is a kind of **reserved** prefix for the Interledger Protocol.


adrianhopebailie:
@sappenin and @dora-gt anything further here?

PR #549, opened 2019-10-07
Update template and deprecate IL-RFC 28

adrianhopebailie:
Update template to accept absolute URLs for deprecated specs

### DIFF
--- a/tmpl/rfc.ejs.html
+++ b/tmpl/rfc.ejs.html
@@ -129,7 +129,8 @@
           <p class="alert alert-warning" style="margin-top: 35px">
             <strong>This specification has been deprecated.</strong>
             <% if (typeof deprecated === 'string') { -%>
-              <br/>It has been replaced by <a href="../<%= deprecated %>">this</a> new specification.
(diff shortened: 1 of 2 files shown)


tmpl/rfc.ejs.html
```diff
@@ -129,7 +129,8 @@ <h4>On This Page</h4>
           <p class="alert alert-warning" style="margin-top: 35px">
             <strong>This specification has been deprecated.</strong>
             <% if (typeof deprecated === 'string') { -%>
-              <br/>It has been replaced by <a href="../<%= deprecated %>">this</a> new specification.
+              <% deprecated = deprecated.startsWith('http') ? deprecated : '../' + deprecated -%>
```
mDuo13: This may behave unexpectedly if we have an RFC with a name like "http-over-ilp" or something else that starts with "http". Right now that's unlikely because most of the RFCs start with the number, but there is already the `micropayments-w3c.md` one, so it's possible we may see more in the future.

A more foolproof test would be to look for a full scheme like `http:` or `https:` but that won't be quite as concise.

PR #551, opened 2019-10-08
Fixes #546

### LINKED ISSUE #546: STREAM RFC: codify the conditions for a ConnectionAssetDetails frame being returned from a receiver.
**Information that inspired this issue:**

- https://github.com/interledgerjs/ilp-protocol-stream/pull/69
- Discussion on https://github.com/interledger-rs/interledger-rs/issues/273
- Discussion on https://github.com/hyperledger/quilt/issues/261

Proposal:
STREAM RFC (il-rfc29) should codify the conditions for a `ConnectionAssetDetails` frame being returned from a receiver.

Question: What is the correct way to indicate a new Connection for a receiver?
Option1: Treat sequence of 1 as a new connection.
Option2: Whenever a ConnectionNewAddressFrame and a SendMoneyFrame are encountered.
Option3: Leave the RFC as is, which suggests only stateful receivers (see next question).

Question: should there be stateless receivers, or is this a protocol violation?

Question: What should a receiver reply with on a new connection?
Stateless:
- When receiver receives a `ConnectionNewAddress` and a `StreamMoneyFrame` frame from a sender, it should reply with a `ConnectionAssetDetails` frame.

Stateful:
- When receiver receives its first `StreamMoneyFrame` for a given connection, it should reply with a `ConnectionAssetDetails` frame.

sappenin:
Fixes #546 by clarifying the STREAM RFC.

* Clarify that `ConnectionAssetDetails` MUST not change during the lifetime of a Connection.

Signed-off-by: sappenin <sappenin@gmail.com>

### DIFF
--- a/0029-stream/0029-stream.md
+++ b/0029-stream/0029-stream.md
@@ -438,7 +438,7 @@ In other words, if a sender resends data (e.g. because a packet was lost), it MU
 | Source Asset Code | Utf8String | Asset code of endpoint that sent the frame. |
 | Source Asset Scale | UInt8 | Asset scale of endpoint that sent the frame. |
 
-Asset details exposed by this frame MUST not change during the lifetime of a Connection. Because of this, endpoints that honor `ConnectionAssetDetails` frames SHOULD ignore any repeat `ConnectionAssetDetails` frames for a given Connection.
+Asset details exposed by this frame MUST NOT change during the lifetime of a Connection. Because of this, endpoints that honor `ConnectionAssetDetails` frames SHOULD ignore any repeat `ConnectionAssetDetails` frames for a given Connection.
 
 ### 5.4. Error Codes
 
--- a/0029-stream/0029-stream.md
+++ b/0029-stream/0029-stream.md
@@ -178,8 +178,6 @@ Each endpoint MAY expose their asset details by sending a `ConnectionAssetDetail
 
 Asset details exposed by this frame MUST not change during the lifetime of a Connection.
 
-Endpoints that wish to discover the other endpoint's asset details MAY send a `ConnectionAssetDetails` frame containing its own asset details. The other endpoint MUST respond to this frame with a new `ConnectionAssetDetails` frame containing its own asset details.
-
 ### 4.4. Streams
 
 Once a connection is established, either endpoint can create streams to send money or data.
@@ -435,8 +433,6 @@ In other words, if a sender resends data (e.g. because a packet was lost), it MU
 |---|---|---|
 | Source Asset Code | Utf8String | Asset code of endpoint that sent the frame. |
 | Source Asset Scale | UInt8 | Asset scale of endpoint that sent the frame. |
-
-If an endpoint receives a `ConnectionAssetDetails` frame, the endpoint MUST respond with a new `ConnectionAssetDetails` frame containing its own asset details.
 
 Asset details exposed by this frame MUST NOT change during the lifetime of a Connection.
 
--- a/0029-stream/0029-stream.md
+++ b/0029-stream/0029-stream.md
@@ -178,7 +178,7 @@ Each endpoint MAY expose their asset details by sending a `ConnectionAssetDetail
 
 Asset details exposed by this frame MUST not change during the lifetime of a Connection. Because of this, endpoints that honor `ConnectionAssetDetails` frames SHOULD ignore any repeat `ConnectionAssetDetails` frames for a given Connection.
 
-Endpoints that wish to discover the other endpoint's asset details MAY send a `ConnectionNewAddress` frame to prompt the endpoint to respond with a `ConnectionAssetDetails` frame.
+Endpoints that wish to discover the other endpoint's asset details MAY send a `ConnectionAssetDetails` frame containing their own asset details. The other endpoint MUST respond to this frame with a new `ConnectionAssetDetails` frame containing its own asset details. In order to avoid infinite loops, this response frame MUST be encoded into the Stream packet response that correlates to request containing the initial `ConnectionAssetDetails` frame. 
 
 ### 4.4. Streams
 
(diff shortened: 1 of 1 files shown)


0029-stream/0029-stream.md
```diff
@@ -174,7 +174,11 @@ Implementations SHOULD wait for a valid response (encrypted with the same shared
 
 #### 4.3.1. Connection Asset Details
 
-Each endpoint MAY expose their asset details by sending a `ConnectionAssetDetails` frame.
+Each endpoint MAY expose their asset details by sending a `ConnectionAssetDetails` frame. This frame is optional because some use-cases, such as Web Monetization, do not require it.
+
+Asset details exposed by this frame MUST not change during the lifetime of a Connection. Because of this, endpoints that honor `ConnectionAssetDetails` frames SHOULD ignore any repeat `ConnectionAssetDetails` frames for a given Connection.
+
+Endpoints that wish to discover the other endpoint's asset details MAY send a `ConnectionNewAddress` frame to prompt the endpoint to respond with a `ConnectionAssetDetails` frame.
```
kincaidoneil: Alternatively, could we change this so sending a `ConnectionAssetDetails` frame triggers the other endpoint to respond with a `ConnectionAssetDetails` frame? That way it'd be more symmetric ("I'll tell you my asset details if you tell me yours").

I know I proposed this 😆...but bundling this behavior to the `ConnectionNewAddress` frame does seem somewhat odd (e.g. how are asset details related to changing your address mid-connection?)

adrianhopebailie: Or we could move this out of STREAM entirely and make it part of SPSP?

I feel like we may have explored that in the past but I can't recall

sappenin: > Or we could move this out of STREAM entirely and make it part of SPSP?

I've had various discussions recently about about potentially deprecating the `ConnectionAssetDetails` frame in STREAM.

Happy to be argued out of this position, but currently I don't prefer this because it forces whatever gets used instead of SPSP to have to provide `asset_code` and `asset_scale`, which _might_ be undesirable. 

E.g., imagine an alternative to SPSP that enables a sender and receiver who have advertised their public keys to arrive at a derived shared secret and receiver address. Forcing this future protocol to _also_ define some endpoint to privately get the asset details seems unnecessary since  these can be obtained privately inside of a Stream Connection.

sappenin: > could we change this so sending a ConnectionAssetDetails frame triggers the other endpoint to respond with a ConnectionAssetDetails frame

I actually like this a lot. Originally I wanted to avoid a circular loop, but as long as the response-frame is sent back in the fulfill/reject packet, I think that can work, and it's a nicer, more generic querying mechanism for getting the other endpoint's asset details.

0029-stream/0029-stream.md
```diff
@@ -174,7 +174,11 @@ Implementations SHOULD wait for a valid response (encrypted with the same shared
 
 #### 4.3.1. Connection Asset Details
 
-Each endpoint MAY expose their asset details by sending a `ConnectionAssetDetails` frame.
+Each endpoint MAY expose their asset details by sending a `ConnectionAssetDetails` frame. This frame is optional because some use-cases, such as Web Monetization, do not require it.
+
+Asset details exposed by this frame MUST not change during the lifetime of a Connection. Because of this, endpoints that honor `ConnectionAssetDetails` frames SHOULD ignore any repeat `ConnectionAssetDetails` frames for a given Connection.
+
+Endpoints that wish to discover the other endpoint's asset details MAY send a `ConnectionAssetDetails` frame containing their own asset details. The other endpoint MUST respond to this frame with a new `ConnectionAssetDetails` frame containing its own asset details. In order to avoid infinite loops, this response frame MUST be encoded into the Stream packet response that correlates to request containing the initial `ConnectionAssetDetails` frame.
```
sappenin: > In order to avoid infinite loops, this response frame MUST be encoded into the Stream packet response that correlates to request containing the initial `ConnectionAssetDetails` frame. 

This last sentence is kind of wordy, and the prior sentence is clear to say, "The other endpoint MUST respond to this frame..."

I'm thinking maybe we strike the last sentence, but I wanted to make sure somebody doesn't read this section and decide to respond with a `ConnectionAssetDetails` in a new request packet.

0029-stream/0029-stream.md
kincaidoneil: > Or we could move this out of STREAM entirely and make it part of SPSP?

How would the sender inform the recipient of their asset details? Query params?

emschwartz: > Or we could move this out of STREAM entirely and make it part of SPSP?

:thinking: https://github.com/interledgerjs/ilp-protocol-stream/pull/69 @adrianhopebailie

0029-stream/0029-stream.md
sappenin: I removed the wordy sentence.

sappenin:
> I don't think we should make any changes to the spec until we've implemented said changes and have some use case that requires this to be done the same way in all implementations.

Sounds fine to me. I have removed any breaking changes from this PR - the only thing left is non-breaking clarifications to the spec.

In the Java STREAM implementation, we're going to support what JS is doing (and apparently [what Rust is going to do](https://github.com/interledger-rs/interledger-rs/issues/395)). We'll _might_  experiment with what was proposed earlier in this PR (i.e., respond to a `ConnectionAssetDetails` frame with a corresponding `ConnectionNewAddress` frame) and see if it adds any value.

As a community, we can revisit the desired behavior here sometime in the future if we ever decide this is worth changing (see https://github.com/interledger/rfcs/issues/554, which is just a tracking issue, if you're interested).

0029-stream/0029-stream.md
```diff
@@ -174,7 +174,9 @@ Implementations SHOULD wait for a valid response (encrypted with the same shared
 
 #### 4.3.1. Connection Asset Details
 
-Each endpoint MAY expose their asset details by sending a `ConnectionAssetDetails` frame.
+Each endpoint MAY expose their asset details by sending a `ConnectionAssetDetails` frame. This frame is optional because some use-cases, such as Web Monetization, do not require it.
```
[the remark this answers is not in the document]
sappenin: Clarified this on the ILP call with @adrianhopebailie and @matdehaast. Remove the specific example.

sappenin:
@adrianhopebailie The build doesn't seem to be working for whatever reason. Once you determine what the problem is, let me know and I can push another change.

adrianhopebailie:
@mDuo13 @ryangyoung any idea what could be happening here. Could this be related to the change of template?

PR #556, opened 2019-10-14
Link to paymentpointers.org

adrianhopebailie:


0026-payment-pointers/0026-payment-pointers.md
```diff
@@ -1,6 +1,7 @@
 ---
 title: Payment Pointers
-draft: 2
+draft: FINAL
+deprecated: https://paymentpointers.org/
```
sappenin: I'm confused about what authority for this spec is going to be. If I look [here](https://paymentpointers.org/about/), the website points back to this RF, seems circular.

adrianhopebailie: Good point! The website is a copy of this spec so it seems silly to have two versions that have to stay in sync.

I'll drop the reference to this spec from the website which will ultimately need a proper spec like: https://webmonetization.org/specification.html

sappenin: Want to bump this thread when paymentpointers.org is updated?

adrianhopebailie: Done

adrianhopebailie: By the way, we want to also re-publish that site to use Docusaurus like webmonetization.org

sappenin: Sounds good - @adrianhopebailie will add a link on paymentpointers.org to the GH repo for paymentpointers.org so that people can file bugs and such in that repo.

PR #559, opened 2019-10-16
Propose HTTP Auth Recommendations

sappenin:
Replaces and enhances the work done as part of https://github.com/interledger/rfcs/pull/531


0038-http-auth-profiles/http-auth-profiles.md
```diff
@@ -0,0 +1,191 @@
...
+
+##### Trade-off Summary
+* **Pros**
+  * Allows both identity and authentication claims to be contained in single Bearer token, which eliminates the need for a second `Auth-Principal` header.
+  * Requires only enough processing to perform an HMAC-SHA256 signing operation, which is very fast.
+  * Supports token expiry, which allows tokens to be generally short-lived so that peers can narrow the potential window of unauthorized usage in the event of token compromise.
+  * The actual shared-secret is _never_ transmitted "on the wire" during any request. Instead, authentication tokens are always _derived_ from the shared-secret, which eliminates the risk of an _actual_ shared-secret being intercepted  in transit.
+
+* **Cons**
+  * More complex than the `SIMPLE` profile.
```
theotherian: Maybe I'm being pedantic, but I almost don't even think this is worth calling out given the rather more serious cons of the `SIMPLE` approach.

0038-http-auth-profiles/http-auth-profiles.md
```diff
@@ -0,0 +1,191 @@
...
+ ```
+
+##### Trade-off Summary
+* **Pros**
+  * Similar benefits to the `JWT_HS_256` profile, but more securely supports the issuer claim because each party signs tokens with its own private key. This allows token verifiers to know exactly who generated the token, including allowing for 3rd-party signing.
+  * Supports token expiry, which allows tokens to be generally short-lived so that peers can narrow the potential window of unauthorized usage in the event of token compromise.
+  * Allows for asymmetric key rotation without forcing a peer to change a shared-secret. 
+
+* **Cons**
+  * More complex than the `SIMPLE` and `JWT_HS_256` profile.
```
theotherian: Similar to the last comment, I personally feel like calling this out as a con doesn't offer a lot of value, but I'm happy to be ignored on this feeling.

0038-http-auth-profiles/http-auth-profiles.md
```diff
@@ -0,0 +1,191 @@
...
+### Avoid HTTP Basic and Form-based Auth
+HTTP Auth schemes using a username and password are NOT RECOMMENDED for the same reasons that the `SIMPLE` profile is only recommended for development and testing scenarios.
+ 
+### Use Reasonable Token Expiries
+Tokens generators should choose a reasonable token expiry. Considerations in this choice include the ability to cache and re-use tokens for a limited time in order to enable very fast authentication decisions. However, this should be balanced against a desire for shorter token lifetimes, which will limit the attack surface caused by a compromised token.
+
+As a best practice, implementations SHOULD use tokens that expire. For example, consider generating tokens with a lifetime that doesn't exceed 5 minutes.
+
+### Secrets At Rest
+Implementations SHOULD protect secret-values that can be used to generate authentication tokens by encrypting them prior to storage. This will help prevent actual shared-secrets or other sensitive data from being captured by unauthorized parties, increasing the chances that only Interledger software will be able to generate tokens.
```
theotherian: This seems like it's worth defining as a `MUST` to me.

0038-http-auth-profiles/http-auth-profiles.md
sappenin: I agree - though in the [other PR](https://github.com/interledger/rfcs/pull/531) that much of this work was debated inside of, this complexity was listed as a suggestion we should add as a negative trade-off for this profile.

0038-http-auth-profiles/http-auth-profiles.md
sappenin: See previous comment.

0038-http-auth-profiles/http-auth-profiles.md
theotherian: Taking an opportunity to expand on this since I was perhaps being a bit vague.

Since encrypting data at rest implies that it require some key to be able to decrypt and make the data usable, you can apply an encryption key either on the data itself before being stored on a disk, or you can have the partition be encrypted itself (which would also require a key to mount and decrypt).

Since the verbiage above says `prior to storage` it made me think that the latter condition should be considered as well since it could also be used as a means of encryption at rest, but wouldn't be predicated on `prior` since you're letting the storage encrypt for you.

sappenin: Ah, I see now. How about this change in verbiage:

```text
Implementations SHOULD protect secret-values that can be used to generate authentication tokens by encrypting them prior to storage, and/or storing them in an encrypted storage device.
```

sappenin: Just pushed this change. Feel free to suggest something else if you don't think the new text is sufficient.

theotherian: Yes; I like that verbiage. My sticking point is still `SHOULD` vs `MUST` I guess. Part of me thinks we should be more forceful since compromising this information is a big deal.

0038-http-auth-profiles/http-auth-profiles.md
```diff
@@ -0,0 +1,191 @@
...
+
+* `JWT_HS_256`: Allows two ILP nodes to utilize a previously agreed-upon shared-secret in order to _derive_ a `Bearer token` that conforms to the JSON Web Token (JWT) specification as defined in [RFC-7519](https://www.rfc-editor.org/rfc/rfc7519.html) using the `HS_256` signing algorithm defined in section 3.2 of [RFC-7518](https://www.rfc-editor.org/rfc/rfc7518#section-3.2).
+
+* `JWT_RS_256`: Allows two ILP nodes to utilize public-key pairs to _derive_ a `Bearer token` that conforms to the JSON Web Token (JWT) specification as defined in [RFC-7519](https://www.rfc-editor.org/rfc/rfc7519.html) using the `RS_256` signing algorithm defined in section 3.3 of [RFC-7518](https://www.rfc-editor.org/rfc/rfc7518#section-3.3).
+
+Peers MAY use any standard HTTP authentication mechanism to authenticate incoming requests, but SHOULD support `JWT_HS_256` at a minimum. It is RECOMMENDED to use `JWT_RS_256` for production deployments.
```
sharafian: The advantages of JWT_RS_256 over JWT_HS_256 are kinda niche, and the performance hit is significant when you're potentially doing an RSA signature verification on every packet. Overall I think HS_256 is the better option. Also worth noting that JWT_HS_256/JWT_RS_256 is only more secure than SIMPLE when you're worried about your HTTPS connection being man-in-the-middled

0038-http-auth-profiles/http-auth-profiles.md
```diff
@@ -0,0 +1,191 @@
...
+A minimal set of extensible, token-based profiles for securing Interledger relationships using HTTP.
+
+## Motivation
+Many Interledger protocols and software work well using HTTP. For example, [ILP-over-HTTP](https://github.com/interledger/rfcs/blob/master/0035-ilp-over-http/0035-ilp-over-http.md) utilizes HTTP at the [Link Layer](https://interledger.org/rfcs/0001-interledger-architecture/). Additionally, Interledger software often uses HTTP for administrative functionality and other purposes.
+
+This document defines common profiles that can be used to ensure implementations of Interledger software work securely in an interoperable fashion, while simultaneously allowing Interledger software operators to more clearly understand the trade-offs involved with various security choices.
+
+The profiles defined in this RFC should:
+
+* ...be as simple as possible while also being highly secure.
```
emschwartz: The term "highly secure" makes me cringe a bit. Everything that follows is about security but each make different tradeoffs and assume different threat models

0038-http-auth-profiles/http-auth-profiles.md
```diff
@@ -0,0 +1,191 @@
...
+
+Peers MAY use any standard HTTP authentication mechanism to authenticate incoming requests, but SHOULD support `JWT_HS_256` at a minimum. It is RECOMMENDED to use `JWT_RS_256` for production deployments.
+
+#### `SIMPLE` Authentication Profile
+This profile allows two ILP nodes to utilize a previously agreed-upon shared-secret that contains at least 32 bytes (256 bits) of randomly generated data, and is encoded using Base64.
+
+Because tokens in this profile do not inherently contain information about the identity of the caller, requests MUST contain an additional HTTP request-header named `Auth-Principal`.
+ 
+This extra header allows for the identity of the authentication request to be separated from authentication token itself, which reduces computational overhead as well as data-management complexity (e.g., implementations do not need to create data-store indexes using derivations of tokens for lookup purposes).
```
emschwartz: Couldn't the token just include the user ID? As long as it's the server that's being called that generates the token, they can include whatever details they need in that

0038-http-auth-profiles/http-auth-profiles.md
```diff
@@ -0,0 +1,191 @@
...
+
+* `JWT_HS_256`: Allows two ILP nodes to utilize a previously agreed-upon shared-secret in order to _derive_ a `Bearer token` that conforms to the JSON Web Token (JWT) specification as defined in [RFC-7519](https://www.rfc-editor.org/rfc/rfc7519.html) using the `HS_256` signing algorithm defined in section 3.2 of [RFC-7518](https://www.rfc-editor.org/rfc/rfc7518#section-3.2).
+
+* `JWT_RS_256`: Allows two ILP nodes to utilize public-key pairs to _derive_ a `Bearer token` that conforms to the JSON Web Token (JWT) specification as defined in [RFC-7519](https://www.rfc-editor.org/rfc/rfc7519.html) using the `RS_256` signing algorithm defined in section 3.3 of [RFC-7518](https://www.rfc-editor.org/rfc/rfc7518#section-3.3).
+
+Peers MAY use any standard HTTP authentication mechanism to authenticate incoming requests, but SHOULD support `JWT_HS_256` at a minimum. It is RECOMMENDED to use `JWT_RS_256` for production deployments.
+
+#### `SIMPLE` Authentication Profile
+This profile allows two ILP nodes to utilize a previously agreed-upon shared-secret that contains at least 32 bytes (256 bits) of randomly generated data, and is encoded using Base64.
```
emschwartz: Should a specific party create the token?

Is there a max length?

Should both sides use the same token to authenticate requests to the other?

0038-http-auth-profiles/http-auth-profiles.md
```diff
@@ -0,0 +1,191 @@
...
+
+Because tokens in this profile do not inherently contain information about the identity of the caller, requests MUST contain an additional HTTP request-header named `Auth-Principal`.
```
emschwartz: What's the advantage of using this header over including the account or user identifier in the token, or using HTTP Basic Auth?

0038-http-auth-profiles/http-auth-profiles.md
```diff
@@ -0,0 +1,191 @@
...
+
+##### Example Usage
+An example shared-secret in this profile is `HEiMCp0FoAC903QHueY89gAWJHo/izaBnJU8/58rlSI=`. This shared secret is passed as an `Authorization` header in each HTTP request, using the Bearer token scheme, along with an `Auth-Principal` header like this: 
+
+```
+Auth-Principal: alice-usd-123
+Authorization: Bearer HEiMCp0FoAC903QHueY89gAWJHo/izaBnJU8/58rlSI=
+``` 
+
+Implementations MAY support this profile, but SHOULD consider it for development purposes only.
```
emschwartz: Why? Plenty of services use this with real money today

Also, this doesn't specify how that token is created, so it could be the result of an OAuth flow. The defining characteristic of this profile is that the token is opaque to one party

0038-http-auth-profiles/http-auth-profiles.md
```diff
@@ -0,0 +1,191 @@
...
+ 
+ ```
+ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQC0H891JhR+Stgx81JyZeU48F4VUAS7E/OKvaVG5OjE9c+iIp2WcYFHqWjeBfrcPS1ADnTbxiKDd5D7EGWLDkBGha9H7m2hH/4PhywHomltp4Z1W7HJISzIS5JvPWFctKeKmhEekMi24yhtf44NkZg2zQijzLMQuxfaPGoW/88omtuDVaqQUmt3/Vx3v8D5ejQ2N8p7BrvpiUPQy+ZakAJf7MG0+EnaCjgnGAc9Q9wEBgMq6ifAENLne6BtQvA34jiWEIGDuD/veUwe0r0Ao/ZipZfcRJKybYNHbs5YQoxXOI2qo8qPwFrF2AJzak8+MwaiFYrDzGk8nV3e3i38RH0p test@example.com
+ ```
+
+##### Trade-off Summary
+* **Pros**
+  * Similar benefits to the `JWT_HS_256` profile, but more securely supports the issuer claim because each party signs tokens with its own private key. This allows token verifiers to know exactly who generated the token, including allowing for 3rd-party signing.
+  * Supports token expiry, which allows tokens to be generally short-lived so that peers can narrow the potential window of unauthorized usage in the event of token compromise.
+  * Allows for asymmetric key rotation without forcing a peer to change a shared-secret.
```
emschwartz: How?

0038-http-auth-profiles/http-auth-profiles.md
```diff
@@ -0,0 +1,191 @@
...
+
+##### Trade-off Summary
+* **Pros**
+  * Similar benefits to the `JWT_HS_256` profile, but more securely supports the issuer claim because each party signs tokens with its own private key. This allows token verifiers to know exactly who generated the token, including allowing for 3rd-party signing.
+  * Supports token expiry, which allows tokens to be generally short-lived so that peers can narrow the potential window of unauthorized usage in the event of token compromise.
+  * Allows for asymmetric key rotation without forcing a peer to change a shared-secret. 
+
+* **Cons**
+  * More complex than the `SIMPLE` and `JWT_HS_256` profile.
+  * Total transmitted bytes for authentication are more than the `SIMPLE` and `JWT_HS_256` schemes. However, HTTP/2 header compression should mitigate this differential as well.
```
emschwartz: How much more?

0038-http-auth-profiles/http-auth-profiles.md
sappenin: > The term "highly secure" makes me cringe a bit

I was thinking about wording tweaks, but am now thinking we can just strike this sentence.

0038-http-auth-profiles/http-auth-profiles.md
sappenin: > the performance hit is significant when you're potentially doing an RSA signature verification on every packet

I think the idea is that you would combine a short token expiration (e.g., 5 mins) with a server-local cached authentication decision to improve performance.

sappenin: > Also worth noting that JWT_HS_256/JWT_RS_256 is only more secure than SIMPLE when you're worried about your HTTPS connection being man-in-the-middled

MITM is one worry, but there are plenty of other worries. E.g., logging of secrets, TLS termination (which is very common). Pretty much _any_ place that a token can be leaked expands the attack surface when tokens don't expire.

I'm uncomfortable with any auth system involving real money that doesn't expire it's auth tokens.

0038-http-auth-profiles/http-auth-profiles.md
sappenin: > Should a specific party create the token? Is there a max length? Should both sides use the same token to authenticate requests to the other?

I don't think it matters. If it works for two peers, it's fine.

0038-http-auth-profiles/http-auth-profiles.md
sappenin: > What's the advantage of using this header over including the account or user identifier in the token,

The advantages are minor IMO, but 1) it's explicit, so two peers don't have to invent a new token-scheme. This helps interop. 2) It allows the actual auth token to be opaque. If two peers need to infer meaning into the token, it's not really opaque anymore.

>  or using HTTP Basic Auth?

The disadvantage of `BASIC Auth` is that it's not a Bearer token scheme. This may not matter depending on what we decide here, but at one point in time I was able to peer a Java implementation that only implements `JWT_HS_256` with the prior versions of the Rust Connector without issue because I could generate a `JWT_HS_256` token, and just configure the Rust connector to use that as an opaque token.

0038-http-auth-profiles/http-auth-profiles.md
sappenin: > Couldn't the token just include the user ID

Yeah, I think it could. 

It does feel a _bit_ strange to be using non-opaque tokens, but need to think a bit more about it.

0038-http-auth-profiles/http-auth-profiles.md
sappenin: > Why?

See trade-offs and comments elsewhere. It's a trade-off, but I consider non-expiring tokens to be a very difficult security surface to defend against.

sappenin: > Also, this doesn't specify how that token is created, so it could be the result of an OAuth flow. The defining characteristic of this profile is that the token is opaque to one party

I'm not clear on what you're suggesting. We should probably first clarify if the token is opaque or not per the other threads.

0038-http-auth-profiles/http-auth-profiles.md
sappenin: Using a combination of [RFC-7517](https://tools.ietf.org/html/rfc7517) and a pre-defined publishing location. I find Auth0's documentation and implementation to be pretty easy to understand: https://auth0.com/docs/jwks

Basically, when verifying a JWT, you can find the token-signer's public-key at some pre-defined endpoint (in Auth0's case secured by TLS). If the signing party (i.e., Auth0) wants to rotate their signing keys, they can advertise new ones.

In the case of two Connector peers using the `JWT_RS_256` profile, this is _one_ way this could be accomplished. This document doesn't go too deeply here because something like this is likely going to be different depending on whether or not two peers want to involve a 3rd party for signing purposes.

0038-http-auth-profiles/http-auth-profiles.md
```diff
@@ -0,0 +1,188 @@
...
+
+##### JWT Claims
+In order to be considered a valid JWT for this profile, signed JWTs MUST contain the following claims as defined in RFC-7519:
+ 
+ * `sub` (subject):   * `sub` (issuer): A claim that claim identifies the principal that is the subject of the JWT. See [Section 4.1.2 of RFC-7519](https://www.rfc-editor.org/rfc/rfc7519.html#section-4.1.2) for more details.
+* `exp` (expiry): A claim that indicates a date/time after which the token should be considered invalid. Implementations SHOULD reject any tokens with a missing or invalid expiry claim.  See [Section 4.1.4 of RFC-7519](https://www.rfc-editor.org/rfc/rfc7519.html#section-4.1.4) for more details.
+
+Implementations SHOULD reject any tokens with a missing or invalid expiry claim because tokens without this claim never expire, and only become invalid if the shared-secret used to sign the JWT changes or is otherwise invalidated.
```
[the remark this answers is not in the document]
sappenin: Removed.

sappenin:
I've update the PR to soften the language around the `SIMPLE` profile. I think it's sufficient to keep the RFC to "tradeoffs" and allow any two peers to decide which of the three profiles to use. Feel free to call out anything I missed in this regard.

sappenin:
> Is there scope to include a profile that is more like OAuth 2.0. i.e. the token is opaque but it is not "pre agreed" between the parties.

I tried to limit the spec to only normalizing "authentication" claims, but also tried to leave room for _any other_ claims if desired (e.g., authorization claims). Speak up if you think something is missing, but any of the tokens (well, the JWT's at least) should be usable in an OAuth 2.0 flow.

0038-http-auth-profiles/http-auth-profiles.md
```diff
@@ -0,0 +1,188 @@
...
+  * Requires out-of-band communication for both peers to agree upon the shared secret.
+
+#### `JWT_HS_256` Authentication Profile
+This profile allows two ILP nodes to utilize a previously agreed-upon shared-secret, but then derive an [RFC-7519](https://tools.ietf.org/html/rfc7519) compliant JWT token signed using the `HS_256` algorithm defined in section 3.2 of [RFC-7518](https://www.rfc-editor.org/rfc/rfc7518#section-3.2). This allows either party holding the shared secret to perform actual authentication by verifying the JWT using these algorithms.
+
+##### JWT Claims
+In order to be considered a valid JWT for this profile, signed JWTs MUST contain the following claims as defined in RFC-7519:
+ 
+ * `sub` (subject):   * `sub` (issuer): A claim that claim identifies the principal that is the subject of the JWT. See [Section 4.1.2 of RFC-7519](https://www.rfc-editor.org/rfc/rfc7519.html#section-4.1.2) for more details.
```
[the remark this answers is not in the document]
sappenin: Fixed.

adrianhopebailie:
@sappenin I wasn't looking to include other claims but rather change the way you describe the SIMPLE profile.

I think this may also make @emschwartz and @sharafian happier.

Basically, OAuth allows the access token to be either a "handle" (opaque) or something like a JWT which encodes the data.

When a resource server gets a request carrying the token it performs "introspection" to determine things like the subject, expiry, audience, scopes etc of the token.

Introspection can mean:
 1. querying the auth system for the details using the token as the lookup value, or
 2. it could mean decoding and validating the JWT

> Out of interest, the OAuth provider we have used for our POC does not use JWT's as access tokens, it uses an opaque token and has some security reasons why they recommend that but I can't find those docs right now!

So, perhaps we can describe SIMPLE as something else that is more along the lines of an "opaque token plus introspection" and then provide guidance on the issuance of the token such as setting an expiry?

adrianhopebailie:
Found the link: https://www.ory.sh/docs/hydra/advanced#json-web-tokens

0038-http-auth-profiles/http-auth-profiles.md
sappenin: Removed.

0038-http-auth-profiles/http-auth-profiles.md
sappenin: > Loads of services use API keys (including payment companies like Stripe and PayPal) and don't seem to be too worried about the issues you describe with the SIMPLE profile. 

I've adjusted the language in the spec in this regard -- in reality if two peers want to use the SIMPLE profile, they should be able to. In fairness to the `SIMPLE` scheme, I have been biasing towards the use-case of two nodes exchanging large values of payments (e.g., two FI's) and I think this has been a mistake in my assumption model. 

I'm going to resolve this issue since I've removed most of the problematic wording around `SIMPLE` - let's work on defining the details of the `SIMPLE` scheme a bit more clearly though.

0038-http-auth-profiles/http-auth-profiles.md
```diff
@@ -0,0 +1,183 @@
...
+
+### Overview
+This document defines three authentication profiles for securing HTTP network connections in Interledger.
+
+### Authentication
+When authenticating requests between Interledger nodes, it is important to choose an algorithm that maintains an appropriate balance between usability and security, while at the same time maintaining high-performance _and_ interoperability.
+
+In order to find this balance, this document defines three Authentication profiles, each with various trade-offs that should be considered before use:
+
+* `SIMPLE`: Allows two ILP nodes to utilize a previously agreed-upon shared-secret as a [Bearer token](https://tools.ietf.org/html/rfc6750) in all HTTP requests. Peers SHOULD consider this token to be opaque and SHOULD NOT derive any special meaning from the token.
```
sappenin: > So, perhaps we can describe SIMPLE as something else that is more along the lines of an "opaque token plus introspection" and then provide guidance on the issuance of the token such as setting an expiry?

Hmm....I like this angle, although I'm trying to figure out how to map it to RFC verbiage. Things like expiry are really a peer-wise decision, and I don't think I can make a reasonable recommendation here, especially since _most_ `SIMPLE` tokens will _never_ expire.

Also, recommending a specific deployment model seems a bit improper/premature for this RFC. If one set of peers want to use something like Hydra/Oathkeeper, and a different set of peers want to have the connector generate and validate auth tokens directly with no proxy, it seems like both should be valid choices assuming people know the tradeoffs for their use-case.

In my mind the only remaining question is: when using a SIMPLE auth token, how does the resource-server determine identity (whatever that means for the resource server). I _think_ the answer is that it doesn't matter for this RFC. Take the Rust connector as an example: it will concatenate a `UserId` + `Random bytes` to create a bearer token. Only it knows how to understand this token, but because it generated the token, this should be fine.

0038-http-auth-profiles/http-auth-profiles.md
```diff
@@ -0,0 +1,183 @@
...
+### Overview
+This document defines three authentication profiles for securing HTTP network connections in Interledger.
+
+### Authentication
+When authenticating requests between Interledger nodes, it is important to choose an algorithm that maintains an appropriate balance between usability and security, while at the same time maintaining high-performance _and_ interoperability.
+
+In order to find this balance, this document defines three Authentication profiles, each with various trade-offs that should be considered before use:
+
+* `SIMPLE`: Allows two ILP nodes to utilize a previously agreed-upon shared-secret as a [Bearer token](https://tools.ietf.org/html/rfc6750) in all HTTP requests. Peers SHOULD consider this token to be opaque and SHOULD NOT derive any special meaning from the token. 
+
```
sappenin: > Out of interest, the OAuth provider we have used for our POC does not use JWT's as access tokens, it uses an opaque token and has some security reasons why they recommend that but I can't find those docs right now!

@adrianhopebailie Moving this comment to a discussion here so we can resolve it once we're done. 

So, I think both models can work. There's just a tradeoff to be made during the "introspection" step. Either 1) a token can be validated in-memory without calling another system (JWT does this in a widely deployed and understood fashion), or 2) the Resource Server needs to call out to an Auth system or database and ask it if the token is valid. 

[Hydra Oathkeer](https://www.ory.sh/docs/oathkeeper/) is an interesting example here. It's an Auth proxy so it introspects the opaque non-JWT token and makes an Auth decision, but then the upstream service still needs to know how to trust the positive or negative assertion here. I can't tell how it does that, but it seems flexible.

0038-http-auth-profiles/http-auth-profiles.md
theotherian: To add to @sappenin's comment, another attack vector that can have its surface area reduced via expiry is replay attacks. I'm still not adept enough with ILP to eloquently articulate how that would happen, but it's a common concern with tokens that don't have expiration and/or aren't opaque.

0038-http-auth-profiles/http-auth-profiles.md
adrianhopebailie: > Things like expiry are really a peer-wise decision

I think the issue here is a "shared" token. Just because a token is opaque doesn't mean both sides should use the same token.

I think we should rather recommend that Alice gets a token from Bob and Bob decides how long it's valid and when to revoke it. Bob then also gets a token from Alice and Alice decides how long its valid and when to revoke it.

In many cases there is a client/server relationship (WebSockets) so it's okay  for just the server to issue a token to the client.

Are there implementations where ILP-over-HTTP is using the same token from both clients?

sappenin: > I think the issue here is a "shared" token. Just because a token is opaque doesn't mean both sides should use the same token.

Perhaps the proposed language can be improved here, but each of the profiles defined in this PR are "one-way", meaning that for any peer-relationship, either peer could choose a different profile. For example, were the Java connector to peer with the Rust connector, the Rust connector might be using `SIMPLE` for its tokens, whereas the Java Connector might only produce `JWT_HS_256` tokens. This type of flow should work fine because each "authenticating peer" is in charge of only accepting tokens it decides to accept.

From this perspective, the token is "shared" (i.e., if I'm making a request to you using the `SIMPLE` token you provided me, then we both know the token). But "shared" doesn't mean we have to use the same token in both directions.

> I think we should rather recommend that Alice gets a token from Bob and Bob decides how long it's valid and when to revoke it. Bob then also gets a token from Alice and Alice decides how long its valid and when to revoke it.

Yes, I agree - that was my intention with this (although if two peers want to use the same token in both directions, that's up to them). Let me see if I can edit `SIMPLE` to be a bit clearer here.

0038-http-auth-profiles/http-auth-profiles.md
sappenin: I update the `SIMPLE` profile to allow for the embedded userId case, and to be clearer that it is not necessarily a shared-secret scheme. I'm going to resolve this comment, but feel free to make another one on a specific part of the new text.

0038-http-auth-profiles/http-auth-profiles.md
sappenin: I've updated the text for `SIMPLE` to more clearly define who does what in that scheme. The new verbiage allows for a single token to be used by both peers, but it also clarifies the responsibilities of the `token-produce` vs the `token-user`.

Resolving this for now - plz make another comment in the new text if you want.

0038-http-auth-profiles/http-auth-profiles.md
sappenin: This header is now removed due to the new `SIMPLE` verbiage.

0038-http-auth-profiles/http-auth-profiles.md
sappenin: With the new `SIMPLE` profile definition I just committed, I've removed any mention of specific size-deltas between the 3 profiles because it's now possible for a `token-producer` in the `SIMPLE` profile to utilize an arbitrary encoding scheme when generating tokens. 

This, combined with the fact that JWTs allow arbitrary claims, means that we can't guarantee tokens in one profile will be smaller or larger than any other profile (Though generally `SIMPLE` tokens will be smaller if they just use _only_ 32-bytes of random data) -- the smallest `JWT_HS_256` token is about 41 bytes, but this depends on the identifiers in the claims (e.g., the accountId or whatever).

Resolving this for now, but feel free to make a new comment in the latest version.

0038-http-auth-profiles/http-auth-profiles.md
sappenin: I think all comments in this section have been addresses with the new `SIMPLE` profile wording. Resolving this here - make a comment inline if you have more thoughts.

sappenin:
Closing this in-favor of https://github.com/interledger/rfcs/pull/565

HTTP Auth is now a proposal in the /proposals folder per the new [CONTRIBUTING](https://github.com/interledger/rfcs/blob/master/CONTRIBUTING.md) guidelines.

PR #560, opened 2019-10-18
Fix typos

KevinWMatthews:


KevinWMatthews:
How do you like to receive small suggestions such as these? I'll start with a PR in case you want to merge it directly.

PR #567, opened 2019-12-16
Add idempotence/retry behavior to ILP-over-HTTP

kincaidoneil:
ILP-over-HTTP has a known issue that if the response containing a Fulfill packet is dropped (e.g. due to a network error), there's no mechanism to retry it, so the balances between the two peers will get out of sync and require manual reconciliation. Also, if the request with an ILP Prepare is dropped, it won't be rejected until the packet expires, which is undesirable since payments may take longer.

This proposes a backwards compatible change to ILP-over-HTTP to separate the Prepare and Fulfill/Reject into two separate HTTP request/responses. Each packet has a corresponding idempotency key enabling clients to safely retry packets without triggering the same side effect multiple times.

The flow looks roughly like:
1. Alice sends `POST` with the ILP Prepare to Bob.
1. Bob begins processing ILP Prepare and responds with a `200` status. The response is dropped due to a network error.
1. Alice doesn't receive any response after 5 seconds (for example), and retries sending the Prepare.
1. Bob ignores the Prepare since the request has an idempotency key he's already seen, and responds again with a `200`.
1. Alice gets the response, and knows no further retries are necessary.
1. Bob gets the corresponding ILP Fulfill, and sends a `POST` to Alice with the Fulfill packet and same idempotency key as the ILP Prepare.
1. Alice gets the request and responds with a `200` status.
1. Bob receives the `200` response and knows no retry is necessary.

### DIFF
--- a/0035-ilp-over-http/0035-ilp-over-http.md
+++ b/0035-ilp-over-http/0035-ilp-over-http.md
@@ -15,7 +15,7 @@ Scaling Interledger infrastructure to handle large volumes of ILP packets requir
 
 In an ILP Over HTTP connection, both peers run HTTP servers with accessible HTTPS endpoints. When peering, the peers exchange their respective URLs, authentication tokens or TLS certificates, ILP addresses, and settlement-related details.
 
-Each ILP Prepare packet is sent as the body of an HTTP request to the peer's server endpoint. ILP Fulfill or Reject packets are returned as the body of the HTTP response in the synchronous mode, or sent in a separate HTTP request in the asynchronous mode.
+Each ILP Prepare packet is sent as the body of an HTTP request to the peer's server endpoint. ILP Fulfill or Reject packets are returned as the body of the HTTP response in synchronous mode, or sent in a separate HTTP request in asynchronous mode.
 
 ## Specification
 
@@ -31,11 +31,12 @@ Peers MAY use any standard HTTP authentication mechanism to authenticate incomin
 
 ```http
 POST /ilp HTTP/x.x
-Host: example.com
+Host: bob.example
 Accept: application/octet-stream
 Content-Type: application/octet-stream
 Authorization: Bearer zxcljvoizuu09wqqpowipoalksdflksjdgxclvkjl0s909asdf
-Callback-Url: https://example.com/incoming/ilp
+Prefer: respond-async
+Callback-Url: https://alice.example/incoming/ilp
 Request-Id: 42ee09c8-a6de-4ae3-8a47-4732b0cbb07b
 
 < Body: Binary OER-Encoded ILP Prepare Packet >
@@ -46,14 +47,15 @@ Request-Id: 42ee09c8-a6de-4ae3-8a47-4732b0cbb07b
 - **Content-Type / Accept Headers** &mdash; MUST be set to `application/octet-stream`.
 - **Body** &mdash; ILP Prepare encoded using OER, as specified in [RFC 27: Interledger Protocol V4](./0027-interledger-protocol-4/0027-interledger-protocol-4.md).
 
-Requests expecting an asynchronous reply SHOULD also include the following headers:
+Asynchronous mode uses these additional headers:
 
+- **Prefer** &mdash; MUST be set to `respond-async`. If omitted, the reply behavior defaults to synchronous mode.
 - **Callback URL Header** &mdash; _Optional_. Callback URL of the origin connector to send an asynchronous HTTP request with the ILP Fulfill/Reject.
 - **Request Id Header** &mdash; _Optional_. UUIDv4 to uniquely identify this ILP Prepare, and correlate the corresponding ILP Fulfill/Reject.
 
 #### Response
 
-In the synchronous mode, the raw OER-encoded ILP Fulfill or Reject is returned within the body of the response:
+In synchronous mode, the raw OER-encoded ILP Fulfill or Reject is returned within the body of the response:
 
 ```http
 HTTP/x.x 200 OK
@@ -62,9 +64,9 @@ Content-Type: application/octet-stream
 < Body: Binary OER-Encoded ILP Fulfill or Reject Packet >
 ```
 
-If the request included a `Callback-Url` header and `Request-Id` header, the recipient handling the ILP Prepare SHOULD choose to handle the packet asynchronously and return the corresponding ILP Fulfill/Reject in a separate outgoing HTTP request.
+If the request includes a `Prefer: respond-async` header, the recipient handling the ILP Prepare SHOULD choose to handle the packet asynchronously and return the corresponding ILP Fulfill/Reject in a separate outgoing HTTP request.
 
-In the asynchronous mode, if the request is semantically valid, the recipient MUST respond immediately that the ILP Prepare is accepted for processing, even if the packet will ultimately be rejected:
+If the request is semantically valid and the recipient chooses to handle it asynchronously, they MUST respond immediately that the ILP Prepare is accepted for processing, even if the packet will ultimately be rejected:
 
 ```http
 HTTP/x.x 202 Accepted
@@ -75,8 +77,8 @@ HTTP/x.x 202 Accepted
 #### Request
 
 ```http
-POST /ilp HTTP/x.x
-Host: example.com
+POST /incoming/ilp HTTP/x.x
+Host: alice.example
 Content-Type: application/octet-stream
 Authorization: Bearer zxcljvoizuu09wqqpowipoalksdflksjdgxclvkjl0s909asdf
 Request-Id: 42ee09c8-a6de-4ae3-8a47-4732b0cbb07b
--- a/0035-ilp-over-http/0035-ilp-over-http.md
+++ b/0035-ilp-over-http/0035-ilp-over-http.md
@@ -35,18 +35,21 @@ Host: example.com
 Accept: application/octet-stream
 Content-Type: application/octet-stream
 Authorization: Bearer zxcljvoizuu09wqqpowipoalksdflksjdgxclvkjl0s909asdf
+Callback-Url: https://example.com/incoming/ilp
 Request-Id: 42ee09c8-a6de-4ae3-8a47-4732b0cbb07b
-Idempotency-Key: 8988dd17-55e4-40e0-9c57-419d81a0e3a5
 
 < Body: Binary OER-Encoded ILP Prepare Packet >
 ```
 
 - **Path** &mdash; A connector MAY specify any HTTP path for their peer to send ILP packets to.
-- **Host Header** &mdash; The standard [HTTP Host Header](https://tools.ietf.org/html/rfc2616#section-14.23) indicating the domain of the HTTP server the Request is sent to.
+- **Host Header** &mdash; The standard [HTTP Host Header](https://tools.ietf.org/html/rfc2616#section-14.23) indicating the domain of the HTTP server the request is sent to.
 - **Content-Type / Accept Headers** &mdash; MUST be set to `application/octet-stream`.
-- **Request Id Header** &mdash; _Optional_. UUIDv4 to correlate the corresponding ILP Fulfill/Reject.
-- **Idempotency Key Header** &mdash; _Optional_. UUIDv4 to uniquely identify this ILP Prepare packet.
 - **Body** &mdash; ILP Prepare encoded using OER, as specified in [RFC 27: Interledger Protocol V4](./0027-interledger-protocol-4/0027-interledger-protocol-4.md).
+
+Requests expecting an asynchronous reply SHOULD also include the following headers:
+
+- **Callback URL Header** &mdash; _Optional_. Callback URL of the origin connector to send an asynchronous HTTP request with the ILP Fulfill/Reject.
+- **Request Id Header** &mdash; _Optional_. UUIDv4 to uniquely identify this ILP Prepare, and correlate the corresponding ILP Fulfill/Reject.
 
 #### Response
 
@@ -59,23 +62,13 @@ Content-Type: application/octet-stream
 < Body: Binary OER-Encoded ILP Fulfill or Reject Packet >
 ```
 
-If the request included a `Request-Id` header, the recipient handling the ILP Prepare SHOULD choose to handle the packet asynchronously and return the corresponding ILP Fulfill/Reject in a separate outgoing HTTP request.
+If the request included a `Callback-Url` header and `Request-Id` header, the recipient handling the ILP Prepare SHOULD choose to handle the packet asynchronously and return the corresponding ILP Fulfill/Reject in a separate outgoing HTTP request.
 
 In the asynchronous mode, if the request is semantically valid, the recipient MUST respond immediately that the ILP Prepare is accepted for processing, even if the packet will ultimately be rejected:
 
 ```http
 HTTP/x.x 202 Accepted
 ```
-
-#### Retry Behavior
-
-Two peers MAY retry ILP Prepare packets, but MUST both agree to support retries out-of-band. If the sender retried sending a Prepare to a peer that doesn't support idempotence, multiple packets may mistakenly be fulfilled, and their accounting balances would desynchronize. Retrying ILP Prepare packets enables faster recovery when a packet is temporarily lost in transit instead of waiting for the packet to expire.
-
-If the recipient of the ILP Prepare responds with a `5xx` status or no HTTP response is received within a given timeout, the sender of the ILP Prepare SHOULD retry sending the packet with the same idempotency key.
-
-The sender MUST conclude retrying after receiving a response with a `2xx` status, `4xx` error, or receiving the corresponding ILP Fulfill/Reject packet.
-
-The sender SHOULD ensure there are multiple attempts to deliver the packet to the peer before the ILP Prepare expires.
 
 ### Async ILP Fulfill/Reject Reply
 
@@ -87,16 +80,14 @@ Host: example.com
 Content-Type: application/octet-stream
 Authorization: Bearer zxcljvoizuu09wqqpowipoalksdflksjdgxclvkjl0s909asdf
 Request-Id: 42ee09c8-a6de-4ae3-8a47-4732b0cbb07b
-Idempotency-Key: 6ff99499-008e-4499-8644-048450627496
 
 < Body: Binary OER-Encoded ILP Fulfill or Reject Packet >
 ```
 
-- **Path** &mdash; A connector MAY specify any HTTP path for their peer to send ILP packets to.
+- **Path** &mdash; HTTP path from the callback URL in the original request carrying the ILP Prepare.
 - **Host Header** &mdash; The standard [HTTP Host Header](https://tools.ietf.org/html/rfc2616#section-14.23) indicating the domain of the HTTP server the Request is sent to.
 - **Content-Type Header** &mdash; MUST be set to `application/octet-stream`.
-- **Request Id Header** &mdash; Request ID from the corresponding ILP Prepare, which is a UUIDv4.
-- **Idempotency Key Header** &mdash; _Optional_. UUIDv4 to uniquely identify this ILP Fulfill or Reject packet.
+- **Request Id Header** &mdash; Request ID from the corresponding ILP Prepare, which is a UUIDv4, matching this reply to the original request.
 - **Body** &mdash; ILP Packet encoded using OER, as specified in [RFC 27: Interledger Protocol V4](./0027-interledger-protocol-4/0027-interledger-protocol-4.md).
 
 #### Response
@@ -105,7 +96,7 @@ Idempotency-Key: 6ff99499-008e-4499-8644-048450627496
 HTTP/x.x 200 OK
 ```
 
-If the request ID doesn't correspond to an in-flight ILP Prepare, the connector should ignore it and return an error:
+If the request ID doesn't correspond to an in-flight ILP Prepare, or a reply was already processed, the connector should ignore it and return an error:
 
 ```http
 HTTP/x.x 400 Bad Request
@@ -113,23 +104,17 @@ HTTP/x.x 400 Bad Request
 
 #### Retry Behavior
 
-If the recipient of the ILP Fulfill/Reject responds with a `5xx` status or no HTTP response is received within a given timeout, the sender of the ILP Fulfill/Reject SHOULD retry sending the packet with the same idempotency key.
+If the recipient of the ILP Fulfill/Reject responds with a `5xx` status or no HTTP response is received within a given timeout, the sender of the ILP Fulfill/Reject SHOULD retry sending the request.
 
 The sender of the ILP Fulfill/Reject MUST conclude retrying after receiving a response with a `2xx` status or `4xx` error.
 
-The sender SHOULD ensure there are multiple attempts to deliver the reply packet to the peer before the corresponding ILP Prepare expires.
+The sender of the ILP Fulfill/Reject SHOULD ensure there are multiple attempts to deliver the reply packet to the peer before the corresponding ILP Prepare expires.
 
-### Idempotence
+#### Idempotence
 
-Every Interledger packet corresponds to a transaction that may affect financial accounting balances. If a request fails, such as due to a network connection error, retrying ILP requests and responses with [idempotence](https://en.wikipedia.org/wiki/Idempotence) can prevent balance inconsistencies between peers.
+An ILP Fulfill packet corresponds to a commitment which affects financial accounting balances. If an HTTP request carrying the ILP reply fails, such as due to a network connection error, retrying delivery of the ILP reply with [idempotence](https://en.wikipedia.org/wiki/Idempotence) can prevent balance inconsistencies between peers.
 
-If peers choose to retry ILP Prepare packets, they MUST both support idempotence. Peers SHOULD support idempotence for asynchronous ILP Fulfill/Reject replies.
-
-For connectors that support idempotence:
-
-- When a connector begins processing an incoming packet with an idempotency key it has not already tracked, it MUST persist that key.
-- If a subsequent request of the same type (ILP Prepare, or ILP Fulfill/Reject) is encountered with the same idempotency key, the packet MUST be ignored and responded to with the successful status.
-- For safety, the connector MUST persist each idempotency key until some amount of time after the corresponding ILP Prepare expires so it doesn't accidentally process a duplicate ILP Prepare packet.
+If the sender of an ILP Prepare expects an asynchronous reply, they should only process the first ILP reply they receive corresponding to the in-flight ILP Prepare.
 
 ### Error Handling
 
--- a/0035-ilp-over-http/0035-ilp-over-http.md
+++ b/0035-ilp-over-http/0035-ilp-over-http.md
@@ -15,7 +15,7 @@ Scaling Interledger infrastructure to handle large volumes of ILP packets requir
 
 In an ILP Over HTTP connection, both peers run HTTP servers with accessible HTTPS endpoints. When peering, the peers exchange their respective URLs, authentication tokens or TLS certificates, ILP addresses, and settlement-related details.
 
-Each ILP Prepare packet is sent as the body of an HTTP request to the peer's server endpoint. ILP Fulfill or Reject packets are returned as the body of the HTTP response.
+Each ILP Prepare packet is sent as the body of an HTTP request to the peer's server endpoint. ILP Fulfill or Reject packets are returned as the body of the HTTP response in the synchronous mode, or sent in a separate HTTP request in the asynchronous mode.
 
 ## Specification
 
@@ -25,7 +25,7 @@ This is a minimal protocol built on HTTP. HTTP/2 is HIGHLY RECOMMENDED for perfo
 
 Peers MAY use any standard HTTP authentication mechanism to authenticate incoming requests. TLS Client Certificates are RECOMMENDED between peers for security and performance, though bearer tokens such as JSON Web Tokens (JWTs) or Macaroons MAY be used instead. Basic authentication (username and password) is NOT RECOMMENDED, because of the additional delay introduced by securely hashing the password.
 
-### ILP Prepare
+### Send ILP Prepare
 
 #### Request
 
@@ -44,17 +44,13 @@ Idempotency-Key: 8988dd17-55e4-40e0-9c57-419d81a0e3a5
 - **Path** &mdash; A connector MAY specify any HTTP path for their peer to send ILP packets to.
 - **Host Header** &mdash; The standard [HTTP Host Header](https://tools.ietf.org/html/rfc2616#section-14.23) indicating the domain of the HTTP server the Request is sent to.
 - **Content-Type / Accept Headers** &mdash; MUST be set to `application/octet-stream`.
-- **Request ID Header** &mdash; UUIDv4 to correlate the corresponding ILP Fulfill/Reject with this ILP Prepare, generated from a cryptographically secure source of randomness.
-- **Idempotency Key Header** &mdash; UUIDv4 to uniquely identify this packet, generated from a cryptographically secure source of randomness.
-- **Body** &mdash; ILP Packet encoded using OER, as specified in [RFC 27: Interledger Protocol V4](./0027-interledger-protocol-4/0027-interledger-protocol-4.md).
+- **Request Id Header** &mdash; _Optional_. UUIDv4 to correlate the corresponding ILP Fulfill/Reject.
+- **Idempotency Key Header** &mdash; _Optional_. UUIDv4 to uniquely identify this packet.
+- **Body** &mdash; ILP Prepare encoded using OER, as specified in [RFC 27: Interledger Protocol V4](./0027-interledger-protocol-4/0027-interledger-protocol-4.md).
 
 #### Response
 
-```http
-HTTP/x.x 200 OK
-```
-
-If no `Idempotency-Key` header is included in the HTTP request, then the raw OER-encoded ILP Fulfill or Reject MUST be returned within the body of the response:
+In the synchronous mode, the raw OER-encoded ILP Fulfill or Reject is returned within the body of the response:
 
 ```http
 HTTP/x.x 200 OK
@@ -63,7 +59,25 @@ Content-Type: application/octet-stream
 < Body: Binary OER-Encoded ILP Fulfill or Reject Packet >
 ```
 
-### ILP Fulfill/Reject
+If the request included a `Request-Id` header, the recipient handling the ILP Prepare SHOULD choose to handle the packet asynchronously and return the corresponding ILP Fulfill/Reject in a separate outgoing HTTP request.
+
+In the asynchronous mode, if the request is semantically valid, the recipient MUST respond immediately that the ILP Prepare is accepted for processing, even if the packet will ultimately be rejected:
+
+```http
+HTTP/x.x 202 Accepted
+```
+
+#### Retry Behavior
+
+Two peers MAY retry ILP Prepare packets, but MUST both agree to support retries out-of-band to prevent duplicate side effects.
+
+If the recipient of the ILP Prepare responds with an `5xx` or `409 Conflict` status, or no HTTP response is received within a given timeout, the sender of the ILP Prepare SHOULD retry sending the packet with the same idempotency key.
+
+The sender MUST conclude retrying after receiving a response with a `2xx` status, `4xx` error, or receiving the corresponding ILP Fulfill/Reject packet.
+
+The sender SHOULD ensure there are multiple attempts to deliver the packet to the peer before the ILP Prepare expires.
+
+### Async ILP Fulfill/Reject Reply
 
 #### Request
 
@@ -81,8 +95,8 @@ Idempotency-Key: 6ff99499-008e-4499-8644-048450627496
 - **Path** &mdash; A connector MAY specify any HTTP path for their peer to send ILP packets to.
 - **Host Header** &mdash; The standard [HTTP Host Header](https://tools.ietf.org/html/rfc2616#section-14.23) indicating the domain of the HTTP server the Request is sent to.
 - **Content-Type Header** &mdash; MUST be set to `application/octet-stream`.
-- **Request ID Header** &mdash; Request ID from the corresponding ILP Prepare, which is a UUIDv4.
-- **Idempotency Key Header** &mdash; UUIDv4 to uniquely identify this packet, generated from a cryptographically secure source of randomness.
+- **Request Id Header** &mdash; Request ID from the corresponding ILP Prepare, which is a UUIDv4.
+- **Idempotency Key Header** &mdash; _Optional_. UUIDv4 to uniquely identify this packet.
 - **Body** &mdash; ILP Packet encoded using OER, as specified in [RFC 27: Interledger Protocol V4](./0027-interledger-protocol-4/0027-interledger-protocol-4.md).
 
 #### Response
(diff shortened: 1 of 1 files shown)


0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -24,26 +25,34 @@ This is a minimal protocol built on HTTP. HTTP/2 is HIGHLY RECOMMENDED for perfo
...
-- **Content-Type / Accept Headers** - MUST be set to `application/octet-stream`.
-- **Body** - ILP Packet encoded using OER, as specified in [RFC 27: Interledger Protocol V4](./0027-interledger-protocol-4/0027-interledger-protocol-4.md).
+#### Response
+
+```http
+HTTP/x.x 200 OK
+```
 
-### Response
+For backwards compatibility with [RFC 35: ILP over HTTP (draft 2)](https://interledger.org/rfcs/0035-ilp-over-http/draft-2.html), if no `Idempotency-Key` header is included, the raw OER-encoded ILP Fulfill or Reject MUST be returned within the body of the response:
```
emschwartz: Would the previous method be supported forever or would the intention be to deprecate it in the future?

0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -52,7 +61,48 @@ Content-Type: application/octet-stream
...
+Authorization: Bearer zxcljvoizuu09wqqpowipoalksdflksjdgxclvkjl0s909asdf
+Idempotency-Key: 8988dd17-55e4-40e0-9c57-419d81a0e3a5
+
+< Body: Binary OER-Encoded ILP Fulfill or Reject Packet >
+```
+
+- **Path** &mdash; A connector MAY specify any HTTP path for their peer to send ILP packets to.
+- **Host Header** &mdash; The standard [HTTP Host Header](https://tools.ietf.org/html/rfc2616#section-14.23) indicating the domain of the HTTP server the Request is sent to.
+- **Content-Type Header** &mdash; MUST be set to `application/octet-stream`.
+- **Idempotency Key Header** &mdash; Same idempotency key used in the corresponding ILP Prepare.
```
emschwartz: I'm not an expert on the use of idempotency keys, but this seems like it's overloading the same ID with both the idempotency behavior and a way to correlate a given request/response.

Is it better to:
- Have separate IDs for idempotency keys and request/response correlation
- Have one ID and call it the `Idempotency-Key`
- Have one ID and call it the `Request-Id`

0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -52,7 +61,48 @@ Content-Type: application/octet-stream
...
+
+If no HTTP response is received within a given timeout, clients should retry sending the packet with the same idempotency key. Clients SHOULD ensure there are multiple attempts to deliver the packet to the peer before the corresponding ILP Prepare expires.
 
-An endpoint MAY return standard HTTP errors, including but not limited to: a malformed or unauthenticated request, rate limiting, or an unresponsive upstream service. Connectors SHOULD either retry the request, if applicable, or relay an ILP Reject packet back to the original sender with an appropriate [Final or Temporary error code](./0027-interledger-protocol-4/0027-interledger-protocol-4#error-codes). Server errors (status codes 500-599) SHOULD be translated into ILP Reject packets with `T00: Temporary Error` codes.
+When a connector begins processing an incoming ILP Prepare with an idempotency key it has not already tracked, it MUST persist that key. If a subsequent request is encountered with the same idempotency key, the packet should be ignored, and the connector should respond with the same `200 OK` status. For safety, the connector MUST persist each idempotency key until the corresponding ILP Prepare expires.
```
emschwartz: > For safety, the connector MUST persist each idempotency key until the corresponding ILP Prepare expires.

@aanchal4 may have some thoughts about removing the records as soon as the packet expires

0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -52,7 +61,48 @@ Content-Type: application/octet-stream
...
+HTTP/x.x 200 OK
+```
+
+### Error Handling
+
+All semantically valid requests accepted for processing MUST be returned with the HTTP status code `200 OK`, even if the packet is ultimately rejected.
+
+An endpoint MAY return standard HTTP errors, including but not limited to: a malformed or unauthenticated request, rate limiting, or an unresponsive upstream service. Connectors SHOULD relay an ILP Reject packet back to the original sender with an appropriate [Final or Temporary error code](./0027-interledger-protocol-4/0027-interledger-protocol-4#error-codes). Server errors (status codes 500-599) SHOULD be translated into ILP Reject packets with `T00: Temporary Error` codes.
+
+### Idempotence Behavior
```
emschwartz: What happens if a `Fulfill` packet is submitted with an idempotency key that the node on the receiving end of that request doesn't have a record for?

0035-ilp-over-http/0035-ilp-over-http.md
kincaidoneil: While I don't think anyone would want to use the old version in production, there are a couple advantages to it:
1. Much easier for manual debugging
2. Doesn't require a publicly accessible endpoint to send outgoing packets

I'm indifferent, since (1) could be addressed by adding an admin endpoint to the node that makes it easier to send packets for debugging, and (2) might be best addressed using BTP or a WebSocket-based transport.

Should the old method eventually be deprecated?

0035-ilp-over-http/0035-ilp-over-http.md
kincaidoneil: Should nodes wait some additional constant before removing the key to account for network latency + clock skew between peers? (If the Prepare was retried at the last possible moment before it expired on the sending node, but the idempotency key was already deleted on the recipient so the Prepare was processed a second time)

matdehaast:
Async Request/Response!

That is how the Mojaloop API handles all REST API calls. I will say it does make life trickier for development and debugging. But alleviates all the statefulness of the actual request, which allows for better Load Balancing, queuing and generally a more scalable architecture IMO.

I am really in favour of this because when I think a packet goes 3-4 hops, then essentially you have this long chained sync http connection over multiple hops that could be the source of any number of errors.

0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -24,26 +25,34 @@ This is a minimal protocol built on HTTP. HTTP/2 is HIGHLY RECOMMENDED for perfo
...
+Idempotency-Key: 8988dd17-55e4-40e0-9c57-419d81a0e3a5
 
 < Body: Binary OER-Encoded ILP Prepare Packet >
 ```
 
-**Field Details:**
+- **Path** &mdash; A connector MAY specify any HTTP path for their peer to send ILP packets to.
+- **Host Header** &mdash; The standard [HTTP Host Header](https://tools.ietf.org/html/rfc2616#section-14.23) indicating the domain of the HTTP server the Request is sent to.
+- **Content-Type / Accept Headers** &mdash; MUST be set to `application/octet-stream`.
+- **Idempotency Key Header** &mdash; Globally unique string identifying this request, which MUST be derived from a cryptographically secure source of randomness.
```
matdehaast: Let be prescriptive and just say it needs to be UUID v4

0035-ilp-over-http/0035-ilp-over-http.md
matdehaast: I tend to agree with this.

Request-ID should be the request-response matching that is generated from the initial `POST` request. Idempotency key will be unique for each leg

kincaidoneil: > Request-ID should be the request-response matching that is generated from the initial POST request

If they're not randomly generated, what if two peers send incoming packets with colliding request IDs that a node routes to the same peer?

Should the `Request-Id` be a u32 to stay consistent across BTP hops as well?

matdehaast: Like I mentioned below. Lets just be normative here and say it must be a UUID v4. Is there any reason we don't just make that the requirement?

kincaidoneil: Making UUIDv4 normative sounds fine. Separately, are you saying the request ID should remain the same across hops? If so, how should collisions be handled?

matdehaast: Nope request-id is local between two nodes. It is used for the async request-response matching of a single leg and would be the SAME for a single request-response. The Idempotency key then would be unique for the request AND response.

kincaidoneil: Cool, gotcha 👍

kincaidoneil: Separated into two headers and prescribe UUIDv4: https://github.com/interledger/rfcs/compare/060732f4f52f286c020c9ebf085b453e91e51951...f4fc0020d82b46b6fdfde537a8f251906d95120c#diff-ccfda9b7df50a9cfd8618bb9c08e360aR47-R48

0035-ilp-over-http/0035-ilp-over-http.md
kincaidoneil: Added a note that a node should return a `400` in this case: https://github.com/interledger/rfcs/compare/060732f4f52f286c020c9ebf085b453e91e51951...f4fc0020d82b46b6fdfde537a8f251906d95120c#diff-ccfda9b7df50a9cfd8618bb9c08e360aR94-R99

kincaidoneil:
One caveat to the backwards compatibility: AFAIK, there's no safe way to retry sending ILP Prepares to a node supporting the old version of the spec without them processing the Prepare multiple times (right?). If the Prepare is retried as a result of a network error, this is probably fine, since it's no worse than the current version (money is lost)! _However_, if the client is retrying sending the Prepare e.g. every 5 seconds as long as it hasn't received any response to the original request, the older node would process the same Prepare multiple times.

I added a note that nodes must know whether their peer supports draft 3 or later before retrying sending Prepares, but how should this be implemented? A per-account flag?

https://github.com/interledger/rfcs/compare/060732f4f52f286c020c9ebf085b453e91e51951...f4fc0020d82b46b6fdfde537a8f251906d95120c#diff-ccfda9b7df50a9cfd8618bb9c08e360aR116

0035-ilp-over-http/0035-ilp-over-http.md
sharafian: 2) is something that we may consider in Coil's use case, where you have clients who only want to send packets. We use BTP right now which is OK but switching to HTTP could save on some connection logic. That said there's really no disadvantage to using nonstandard ILP HTTP when we have to so it's OK if this doesn't cover every use case for everyone

sharafian:
If we want to smoothly upgrade this we could add some version negotiation. you can carry an accept header on the prepare that informs whether you're OK with an async response, and the response will come back with a content type which communicates whether it's just an empty ACK or whether it's carrying the fulfillment.

I do think we should be careful about retrying packets at the link layer though. Imagine there's a network error preventing your fulfills from getting back to me (like my load balancer isn't accepting traffic from you but yours is OK). My connector is going to be accumulating all of these prepare packets that it continually retries until they time out. And the original sender is probably going to be sending more packets too. Your systems will be trying to send back an increasing number of fulfills. It could lead to a big slowdown where everyone gets mucked up with packets that keep getting rejected.

It might be good to investigate the prevalence of this error before we go about solving it. If you can measure how many packets you never got a fulfill/reject for, it would inform what we should be solving for. If it never happens in practice, then maybe it's not worth adding the memory footprint of tracking idempotence IDs for 100's to 1000's of packets per second. And if it happens a lot, maybe there's some network configuration upgrades we can do instead of a protocol change.

All that criticism is separate from splitting Prepare from Fulfill and Reject. I think an async flow will probably have benefits separate from idempotence (like Matt said, it leaves much fewer hanging connections). Although it would break Coil's setup, because we horizontally scale our connector and having the fulfill go to a different shard than the one the prepare came from will make it stop working. Does the rust implementation have that issue too when sharded?

sentientwaffle:
> If no HTTP response is received within a given timeout, clients SHOULD retry sending the packet with the same idempotency key.

When the sender has gotten a `200` from the receiver (but has not received a corresponding Fulfill or Reject), why is it the sender's responsibility to retry? The receiver already has the Prepare, and could keep retrying their fulfill/reject request until geting a `200` back from the original sender.


Aside: as both parties of a payment are simultaneously HTTP clients and servers it would be clearer to refer to them in the spec as "sender"/"receiver" rather than "client"/"server".

0035-ilp-over-http/0035-ilp-over-http.md
sappenin: I echo Ben's sentiment - I think we want to allow for ILP-over-HTTP to be used by a non-server in a "send-only" mode (e.g., a mobile app). I added a proposed change.

0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -24,26 +25,36 @@ This is a minimal protocol built on HTTP. HTTP/2 is HIGHLY RECOMMENDED for perfo
...
-- **Content-Type / Accept Headers** - MUST be set to `application/octet-stream`.
-- **Body** - ILP Packet encoded using OER, as specified in [RFC 27: Interledger Protocol V4](./0027-interledger-protocol-4/0027-interledger-protocol-4.md).
+#### Response
 
-### Response
+```http
+HTTP/x.x 200 OK
+```
+
+For backwards compatibility with [RFC 35: ILP over HTTP (draft 2)](https://interledger.org/rfcs/0035-ilp-over-http/draft-2.html), if no `Idempotency-Key` header is included, the raw OER-encoded ILP Fulfill or Reject MUST be returned within the body of the response:
```
sappenin:
```suggestion
If no `Idempotency-Key` header is included in the HTTP request, then the raw OER-encoded ILP Fulfill or Reject MUST be returned within the body of the response:
```

0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -52,7 +63,58 @@ Content-Type: application/octet-stream
...
+- **Idempotency Key Header** &mdash; UUIDv4 to uniquely identify this packet, generated from a cryptographically secure source of randomness.
+- **Body** &mdash; ILP Packet encoded using OER, as specified in [RFC 27: Interledger Protocol V4](./0027-interledger-protocol-4/0027-interledger-protocol-4.md).
+
+#### Response
+
+```http
+HTTP/x.x 200 OK
+```
+
+If the request ID doesn't correspond to an in-flight ILP Prepare, the connector should ignore it and return an error:
```
sappenin: Why not just return an ILP Reject packet here?

0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -52,7 +63,58 @@ Content-Type: application/octet-stream
...
+
+To enable this functionality, the two phases of an ILP packet lifecycle are separated into two different HTTP requests: one for the ILP Prepare, and one for the corresponding ILP Fulfill or ILP Reject.
+
+Responses with an HTTP `200 OK` status attest that the packet has been successfully accepted for processing, even if the packet may not necessarily be forwarded. If the connector responds with an HTTP `5xx` or `409 Conflict` status, the client SHOULD retry the request with the same idempotency key.
+
+If no HTTP response is received within a given timeout, clients SHOULD retry sending the packet with the same idempotency key. Clients SHOULD ensure there are multiple attempts to deliver the packet to the peer before the corresponding ILP Prepare expires.
+
+**NOTE:** Clients MUST ensure their peer supports [RFC 35: ILP over HTTP](.) draft 3 or later before retrying ILP Prepare requests to prevent duplicate side effects.
```
sappenin: Until there's broad consensus to deprecate Draft 2, I think this RFC should support both sync and async modes, or if there disagreement there, perhaps we should define this async behavior in a new specification.

0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -52,7 +63,58 @@ Content-Type: application/octet-stream
...
+
+### Idempotence Behavior
+
+Every Interledger packet corresponds to a transaction that may affect financial accounting balances. If a request fails, such as due to a network connection error, retrying ILP requests and responses with [idempotence](https://en.wikipedia.org/wiki/Idempotence) can prevent balance inconsistencies between peers.
+
+To enable this functionality, the two phases of an ILP packet lifecycle are separated into two different HTTP requests: one for the ILP Prepare, and one for the corresponding ILP Fulfill or ILP Reject.
+
+Responses with an HTTP `200 OK` status attest that the packet has been successfully accepted for processing, even if the packet may not necessarily be forwarded. If the connector responds with an HTTP `5xx` or `409 Conflict` status, the client SHOULD retry the request with the same idempotency key.
```
sappenin: To support both ILP-over-HTTP modes, what if we use `200 OK` to indicate that there's a Fulfill/Reject packet in the response, and use a `202 Accepted` response to indicate the async-mode?

0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -52,7 +63,58 @@ Content-Type: application/octet-stream
...
+
+**NOTE:** Clients MUST ensure their peer supports [RFC 35: ILP over HTTP](.) draft 3 or later before retrying ILP Prepare requests to prevent duplicate side effects.
 
-An endpoint MAY return standard HTTP errors, including but not limited to: a malformed or unauthenticated request, rate limiting, or an unresponsive upstream service. Connectors SHOULD either retry the request, if applicable, or relay an ILP Reject packet back to the original sender with an appropriate [Final or Temporary error code](./0027-interledger-protocol-4/0027-interledger-protocol-4#error-codes). Server errors (status codes 500-599) SHOULD be translated into ILP Reject packets with `T00: Temporary Error` codes.
+When a connector begins processing an incoming packet with an idempotency key it has not already tracked, it MUST persist that key. If a subsequent request of the same type (ILP Prepare, or ILP Fulfill/Reject) is encountered with the same idempotency key, the packet should be ignored, and the connector should respond with the same `200 OK` status. For safety, the connector MUST persist each idempotency key until the corresponding ILP Prepare expires.
```
sappenin: One potential problem with this advice is a case where the sending Connector retries near the expiry. Due to timing irregularities, it's possible the receiving Connector might have discarded the idempotency key, and treat the packet as a new Prepare.

Perhaps a simple remedy here in the spec would be to make this advice more overt (e.g., "For safety, the connector MUST persist each idempotency key until some amount of time after the corresponding ILP Prepare expires such that it doesn't accidentally process a duplicate Prepare packet."

0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -52,7 +63,58 @@ Content-Type: application/octet-stream
...
 
+Since ILP Prepare expirations are typically on the order of seconds and the transport layer will handle flow control, exponential backoff and jitter for retries are unnecessary.
```
sappenin: Which "transport" layer is this section referring to? The HTTP transport, or the STREAM tranport?

0035-ilp-over-http/0035-ilp-over-http.md
sappenin: Ah, well - I suppose my question makes sense if the spec support both Draft 2 and Draft3 at the same time. Makes less sense if only Draft3 is adopted.

0035-ilp-over-http/0035-ilp-over-http.md
kincaidoneil: STREAM/an end-to-end transport layer. If packets aren't getting through/timing out, STREAM *should* eventually back off in time.

0035-ilp-over-http/0035-ilp-over-http.md
kincaidoneil: Added this verbiage: https://github.com/interledger/rfcs/compare/003ad42...09916ea#diff-ccfda9b7df50a9cfd8618bb9c08e360aR126

0035-ilp-over-http/0035-ilp-over-http.md
kincaidoneil: Updated with 202 Accepted in the async mode: https://github.com/interledger/rfcs/compare/003ad42...09916ea#diff-ccfda9b7df50a9cfd8618bb9c08e360aR66-R68

0035-ilp-over-http/0035-ilp-over-http.md
kincaidoneil: Updated with David's suggestion [here](https://github.com/interledger/rfcs/pull/567#discussion_r359651531), feel free to re-open if that's insufficient.

kincaidoneil:
> When the sender has gotten a 200 from the receiver (but has not received a corresponding Fulfill or Reject), why is it the sender's responsibility to retry? The receiver already has the Prepare, and could keep retrying their fulfill/reject request until geting a 200 back from the original sender.

@sentientwaffle Yep! (The behavior you describe was how I intended it to be read, but I agree that "client" and "server" were unclear). I updated the language with "sender of ILP Prepare," "recipient of ILP Fulfill/Reject," etc. to try to better clarify this.

> Until there's broad consensus to deprecate Draft 2, I think this RFC should support both sync and async modes, or if there disagreement there, perhaps we should define this async behavior in a new specification.

@sappenin I reorganized and clarified the doc to better differentiate between the two modes and that they're both supported.

> Although it would break Coil's setup, because we horizontally scale our connector and having the fulfill go to a different shard than the one the prepare came from will make it stop working. Does the rust implementation have that issue too when sharded?

Good points @sharafian. Definitely agree it's worth measuring how prevalent this is, though even a very low failure rate could be costly in the time required for manual reconciliation. Re: horizontal scaling: the async mode does seem to complicate this. There's been a little discussion about adding crash recovery to Rust; if in-flight packets were written to e.g. a Redis store shared between instances, that might make it possible to support.

I think this is ready for re-review when/if anyone gets a chance!

0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -52,7 +59,72 @@ Content-Type: application/octet-stream
...
+
+In the asynchronous mode, if the request is semantically valid, the recipient MUST respond immediately that the ILP Prepare is accepted for processing, even if the packet will ultimately be rejected:
+
+```http
+HTTP/x.x 202 Accepted
+```
+
+#### Retry Behavior
+
+Two peers MAY retry ILP Prepare packets, but MUST both agree to support retries out-of-band to prevent duplicate side effects.
```
sharafian: In what situation are side effects possible?

0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -52,7 +59,72 @@ Content-Type: application/octet-stream
...
+
+If the recipient of the ILP Prepare responds with an `5xx` or `409 Conflict` status, or no HTTP response is received within a given timeout, the sender of the ILP Prepare SHOULD retry sending the packet with the same idempotency key.
```
sharafian: nit: an `5xx` -> a `5xx`

0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -52,7 +59,72 @@ Content-Type: application/octet-stream
...
+
+The sender MUST conclude retrying after receiving a response with a `2xx` status, `4xx` error, or receiving the corresponding ILP Fulfill/Reject packet.
```
sharafian: `4xx` error other than `409`. also what does the 409 conflict signify in this flow? It seems to have special meaning but is not fully explained

0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -52,7 +59,72 @@ Content-Type: application/octet-stream
...
+
+The sender SHOULD ensure there are multiple attempts to deliver the packet to the peer before the ILP Prepare expires.
```
sharafian: It seems like if a Prepare fails there's really nothing lost so it'd be best to fast-fail and let the sender retry if they so choose. That keeps packets faster, rathing than having everyone implement their own retry logic that might be waiting some timeout to retry the packet, slowing everyone down.

The only reason I see to not fast fail would be if there's a risk that a 5xx came back but the packet was actually passed on. But if you treated the 5xx as a reject and didn't honor a fulfill that came after it then that wouldn't really matter.

0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -52,7 +59,72 @@ Content-Type: application/octet-stream
...
+Idempotency-Key: 6ff99499-008e-4499-8644-048450627496
+
+< Body: Binary OER-Encoded ILP Fulfill or Reject Packet >
+```
+
+- **Path** &mdash; A connector MAY specify any HTTP path for their peer to send ILP packets to.
+- **Host Header** &mdash; The standard [HTTP Host Header](https://tools.ietf.org/html/rfc2616#section-14.23) indicating the domain of the HTTP server the Request is sent to.
+- **Content-Type Header** &mdash; MUST be set to `application/octet-stream`.
+- **Request Id Header** &mdash; Request ID from the corresponding ILP Prepare, which is a UUIDv4.
+- **Idempotency Key Header** &mdash; _Optional_. UUIDv4 to uniquely identify this packet.
```
sharafian: could clarify with "Uniquely identify this Fulfill/Reject packet"

0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -24,26 +25,32 @@ This is a minimal protocol built on HTTP. HTTP/2 is HIGHLY RECOMMENDED for perfo
...
 
 < Body: Binary OER-Encoded ILP Prepare Packet >
 ```
 
-**Field Details:**
+- **Path** &mdash; A connector MAY specify any HTTP path for their peer to send ILP packets to.
+- **Host Header** &mdash; The standard [HTTP Host Header](https://tools.ietf.org/html/rfc2616#section-14.23) indicating the domain of the HTTP server the Request is sent to.
+- **Content-Type / Accept Headers** &mdash; MUST be set to `application/octet-stream`.
+- **Request Id Header** &mdash; _Optional_. UUIDv4 to correlate the corresponding ILP Fulfill/Reject.
+- **Idempotency Key Header** &mdash; _Optional_. UUIDv4 to uniquely identify this packet.
```
sharafian: Maybe clarify with "uniquely identify this Prepare packet"

0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -52,7 +59,72 @@ Content-Type: application/octet-stream
...
+
+The sender of the ILP Fulfill/Reject MUST conclude retrying after receiving a response with a `2xx` status or `4xx` error.
+
+The sender SHOULD ensure there are multiple attempts to deliver the reply packet to the peer before the corresponding ILP Prepare expires.
+
+### Idempotence
+
+Every Interledger packet corresponds to a transaction that may affect financial accounting balances. If a request fails, such as due to a network connection error, retrying ILP requests and responses with [idempotence](https://en.wikipedia.org/wiki/Idempotence) can prevent balance inconsistencies between peers.
+
+When a connector begins processing an incoming packet with an idempotency key it has not already tracked, it MUST persist that key. If a subsequent request of the same type (ILP Prepare, or ILP Fulfill/Reject) is encountered with the same idempotency key, the packet should be ignored, and the connector should respond with the successful status. For safety, the connector MUST persist each idempotency key until some amount of time after the corresponding ILP Prepare expires so it doesn't accidentally process a duplicate ILP Prepare packet.
```
sharafian: if we want to make this nonbreaking we should clarify that idempotence is a feature of asynchronous mode only. If we want to make it nonbreaking and also add idempotence to synchronous mode then we should have a response code/header of some kind that makes it clear the idempotency key will be honored

kincaidoneil: Since ILP Prepare retries must be negotiated out of band, isn't this still non-breaking? If the peers don't retry ILP Prepares, the recipient of the ILP Prepare doesn't need to handle idempotent requests. Can I clarify that better?

0035-ilp-over-http/0035-ilp-over-http.md
kincaidoneil: > > The sender SHOULD ensure there are multiple attempts to deliver the packet to the peer before the ILP Prepare expires.

> It seems like if a Prepare fails there's really nothing lost so it'd be best to fast-fail and let the sender retry if they so choose.

If Bob's Prepare packet is temporarily interrupted in transit and doesn't reach Charlie, retrying at the link layer allows much quicker recovery. To Bob, this would appear as a network error, but they can't safely return a reject right then because it's possible Charlie received the packet and it'll ultimately get fulfilled (if Charlie's response was interrupted in transit, but Bob doesn't know). Currently:

1) All previous nodes in the path have to wait the entire timeout before the hold on the balance is released, which slows down the payment and prevents re-allocating that liquidity
2) Afterward, the original sender of the Prepare has to wait for the Reject to percolate all the way back before retrying (e.g. assuming the number of packets in flight is capped), which may increase the time to complete the payment

None of this is imperative, so it's an optional feature that two peers can choose to implement or enable.

0035-ilp-over-http/0035-ilp-over-http.md
kincaidoneil: Eh, maybe I should just remove retrying on `409`? (Stripe uses `409` for idempotent retries, but I'm not sure it matters)

0035-ilp-over-http/0035-ilp-over-http.md
kincaidoneil: https://github.com/interledger/rfcs/compare/09916ea8c1232f846610940066298f052f140c8f...8b43dafce2c84d073bc70afacf396f700d4ff8f5#diff-ccfda9b7df50a9cfd8618bb9c08e360aR48

0035-ilp-over-http/0035-ilp-over-http.md
kincaidoneil: https://github.com/interledger/rfcs/compare/09916ea8c1232f846610940066298f052f140c8f...8b43dafce2c84d073bc70afacf396f700d4ff8f5#diff-ccfda9b7df50a9cfd8618bb9c08e360aR99

0035-ilp-over-http/0035-ilp-over-http.md
kincaidoneil: Removed retrying on `409`: https://github.com/interledger/rfcs/compare/09916ea8c1232f846610940066298f052f140c8f...8b43dafce2c84d073bc70afacf396f700d4ff8f5#diff-ccfda9b7df50a9cfd8618bb9c08e360aR74

0035-ilp-over-http/0035-ilp-over-http.md
kincaidoneil: https://github.com/interledger/rfcs/compare/09916ea8c1232f846610940066298f052f140c8f...8b43dafce2c84d073bc70afacf396f700d4ff8f5#diff-ccfda9b7df50a9cfd8618bb9c08e360aR74

0035-ilp-over-http/0035-ilp-over-http.md
kincaidoneil: Added explanation of why both peers must support it/bad side effects, and why it's useful: https://github.com/interledger/rfcs/compare/09916ea8c1232f846610940066298f052f140c8f...8b43dafce2c84d073bc70afacf396f700d4ff8f5#diff-ccfda9b7df50a9cfd8618bb9c08e360aR72

0035-ilp-over-http/0035-ilp-over-http.md
kincaidoneil: Added an explanation here that hopefully better clarifies this: https://github.com/interledger/rfcs/compare/09916ea8c1232f846610940066298f052f140c8f...8b43dafce2c84d073bc70afacf396f700d4ff8f5#diff-ccfda9b7df50a9cfd8618bb9c08e360aR126-R132

adrianhopebailie:
This is a neat way to solve the issue, and as @matdehaast points out, this is the pattern we've seen used for Mojaloop.

I do wonder if the async pattern is a requirement for idempotency? It seems a pity to lose the one aspect of HTTP that was especially useful, request/response matching.

What if we simply added the idempotency key and expected a client to terminate the request and retry with the same key if it doesn't get a response?

If you use HTTP/2 this would only cost you a new stream which is very cheap.

0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -52,7 +59,78 @@ Content-Type: application/octet-stream
 < Body: Binary OER-Encoded ILP Fulfill or Reject Packet >
 ```
 
-All ILP Packets MUST be returned with the HTTP status code `200: OK`.
+If the request included a `Request-Id` header, the recipient handling the ILP Prepare SHOULD choose to handle the packet asynchronously and return the corresponding ILP Fulfill/Reject in a separate outgoing HTTP request.
```
sappenin: Thanks @kincaidoneil for pushing the PR forward - overall I like the proposal.

Are we sure we want to vary the behavior of the server based upon the presence of the `Request-Id` header? I'm not aware of any standard around doing such a thing, but this RFC seems like a good start: https://tools.ietf.org/html/rfc7240#section-4.1

That said, some alternatives for varying the response would be:

* Use a different URL prefix (e.g., `/ilp` vs `/ilp-async`)
* Use a different Media Type (`application/ilpv4-async` or `application/ilp-over-http-async`) or something like that.

The primary reason I bring this up is that it might be useful for an ILP-over-HTTP sync request/response that wants to use idempotence to _also_ use both the requestId and the `Idempotence Key` without triggering async behavior, (e.g., for logging or audit purposes).

kincaidoneil: I'd be in favor of `Prefer: respond-async`

kincaidoneil:
I made some changes to this proposal:

#### Added a `Callback-Url` header to the request

Previously, the idea was the callback URL for asynchronous replies would be statically configured for each peer. However, that complicates the implementation if the ILP Prepare sender operated multiple nodes behind a load balancer: they must cache which node handled the original ILP Prepare, in order to direct the ILP reply to the correct node. By including a callback URL in each individual HTTP request, this requires no static configuration, and enables the ILP reply to be sent directly to the node instance which sent the ILP Prepare and has its associated state available (they can still choose to use a load balancer).

#### Removed idempotent ILP Prepare retries (for now)

Retrying Prepares with idempotence requires the ILP Prepare receiver to operate a cache of idempotence keys, and is a breaking change for existing peering relationships (both peers must enable ILP Prepare retries for it to be safe). Note, however: this doesn't apply to idempotent retries on the ILP reply leg, which are still supported here.

I do still think we should add this feature, eventually. The downside of not retrying Prepares is if a request fails due to a transient error, the sender must wait the full timeout until they can retry it. For retail payments, the payment will take an additional 30 seconds (since holds prevent unsafe amounts in-flight). Also, for example, the `pay` STREAM implementation would terminate the payment, since it took too long -- which I believe is the right behavior.

0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -52,7 +62,60 @@ Content-Type: application/octet-stream
 < Body: Binary OER-Encoded ILP Fulfill or Reject Packet >
 ```
 
-All ILP Packets MUST be returned with the HTTP status code `200: OK`.
+If the request included a `Callback-Url` header and `Request-Id` header, the recipient handling the ILP Prepare SHOULD choose to handle the packet asynchronously and return the corresponding ILP Fulfill/Reject in a separate outgoing HTTP request.
+
+In the asynchronous mode, if the request is semantically valid, the recipient MUST respond immediately that the ILP Prepare is accepted for processing, even if the packet will ultimately be rejected:
```
sharafian, line 67: Is it a problem that until the response comes back the sender doesn't know whether asynchronous mode is being used? If it takes a couple seconds the connector isn't sure whether it's taking a long time to ACK or whether it's waiting for a fulfill/reject. I can't think of any concrete case where that's a problem but it seems slightly odd

sharafian (approved):

0035-ilp-over-http/0035-ilp-over-http.md
kincaidoneil: If the request is still pending but just taking a long time, no behavior changes since it's unsafe for any sender to retry (until we add idempotent Prepare retries).

Today, if a request is sent but then the connection is terminated, the sender immediately treats the packet as rejected. If the sender supports async, they must wait the full timeout even if the connection is terminated, since they may receive an async response. Note that this only happens when the reply is dropped, so it's the failure scenario which gets slightly worse (if the sender supports async but recipient doesn't).

0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -14,7 +15,7 @@ Scaling Interledger infrastructure to handle large volumes of ILP packets requir
 
 In an ILP Over HTTP connection, both peers run HTTP servers with accessible HTTPS endpoints. When peering, the peers exchange their respective URLs, authentication tokens or TLS certificates, ILP addresses, and settlement-related details.
 
-Each ILP Prepare packet is sent as the body of an HTTP request to the peer's server endpoint. ILP Fulfill or Reject packets are returned as the body of the HTTP response.
+Each ILP Prepare packet is sent as the body of an HTTP request to the peer's server endpoint. ILP Fulfill or Reject packets are returned as the body of the HTTP response in the synchronous mode, or sent in a separate HTTP request in the asynchronous mode.
```
matdehaast, line 18:
```suggestion
Each ILP Prepare packet is sent as the body of an HTTP request to the peer's server endpoint. ILP Fulfill or Reject packets are returned as the body of the HTTP response in synchronous mode, or sent in a separate HTTP request in asynchronous mode.
```

0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -52,7 +62,60 @@ Content-Type: application/octet-stream
...
+HTTP/x.x 202 Accepted
+```
+
+### Async ILP Fulfill/Reject Reply
+
+#### Request
+
+```http
+POST /ilp HTTP/x.x
+Host: example.com
```
matdehaast, line 79: Can we change these to reflect the `Callback-Url` used above for consistency.

matdehaast: It may also make sense to change the URL of the request and the reply to different domains so it's more obvious they are different parties

matdehaast (approved):
Just some general cleanup but otherwise good

0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -52,7 +62,60 @@ Content-Type: application/octet-stream
 < Body: Binary OER-Encoded ILP Fulfill or Reject Packet >
 ```
 
-All ILP Packets MUST be returned with the HTTP status code `200: OK`.
+If the request included a `Callback-Url` header and `Request-Id` header, the recipient handling the ILP Prepare SHOULD choose to handle the packet asynchronously and return the corresponding ILP Fulfill/Reject in a separate outgoing HTTP request.
```
sappenin, line 65: I'd like to use a more overt signal for "async-mode", such as what we discussed [in this comment](https://github.com/interledger/rfcs/pull/567#discussion_r376088183).

My rationale is mainly to allow this RFC to be extended later (after this proposal is committed) with other, non-callback-based styles of async communication (e.g., sender polling w\ batch responses). 

Any objection to including a 3rd `prefer:` header as the normatively required field to indicate sync vs async so  we don't have to assume `sync` in the absence of a `Callback-Url` header?

sappenin (changes requested):

0035-ilp-over-http/0035-ilp-over-http.md
kincaidoneil: I added the `Prefer: respond-async` header to specify asynchronous mode. I'd just note, sync is still assumed in the absence of a `Prefer` header

matdehaast (approved):
👍 

@kincaidoneil last question. If I receive a request that specifies `Prefer: respond-async` and I don't support it. Is there a normative response that I must reply with?

kincaidoneil:
@matdehaast For backwards compatibility, if the receiver doesn't support async, they just handle the request synchronously. The sender can identify which mode the receiver chose based on the status code of the response (200 vs 202) or if the body is non-empty.

0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -24,26 +25,37 @@ This is a minimal protocol built on HTTP. HTTP/2 is HIGHLY RECOMMENDED for perfo
...
+
+Asynchronous mode uses these additional headers:
 
-- **Path** - A connector MAY specify any HTTP path for their peer to send ILP packets to.
-- **Host Header** - The standard [HTTP Host Header](https://tools.ietf.org/html/rfc2616#section-14.23) indicating the domain of the HTTP server the Request is sent to.
-- **Content-Type / Accept Headers** - MUST be set to `application/octet-stream`.
-- **Body** - ILP Packet encoded using OER, as specified in [RFC 27: Interledger Protocol V4](./0027-interledger-protocol-4/0027-interledger-protocol-4.md).
+- **Prefer** &mdash; MUST be set to `respond-async`. If omitted, the reply behavior defaults to synchronous mode.
+- **Callback URL Header** &mdash; _Optional_. Callback URL of the origin connector to send an asynchronous HTTP request with the ILP Fulfill/Reject.
+- **Request Id Header** &mdash; _Optional_. UUIDv4 to uniquely identify this ILP Prepare, and correlate the corresponding ILP Fulfill/Reject.
```
sentientwaffle, line 54: These aren't `Optional` in async mode, right?

kincaidoneil: Thanks. Changed `Request-Id` so it's required, but keeping the `Callback-Url` as optional so peers can still use other mechanisms to set/exchange the callback URL.

sentientwaffle (approved):

0035-ilp-over-http/0035-ilp-over-http.md
sappenin: Sounds great!

sappenin (approved):

Issue #568, opened 2020-02-06
Add STREAM Receipts


sharafian:
This is something we've been talking about at Coil as a solution to make payment verification easier. Looking for feedback especially from those maintaining STREAM implementations or working on Web Monetization use cases

This proposal is meant to supercede the [Open Payments features of Sources](https://github.com/adrianhopebailie/open-payments/issues/19)

CC @adrianhopebailie @wilsonianb @kincaidoneil @matdehaast @sentientwaffle 

------

# STREAM Receipts

## Motivation

We've been discussing how to make payment verification on Interledger easy. This is immediately important for Web Monetization, where ILP payments must be verified to securely serve exclusive content. It also enables a host of other use cases including Codius and API micropayments.

Recently we've moved from a service that forwards ILP packets and signs JWTs that serve as proof of payment (https://github.com/sharafian/web-monetization-access) to a resource type in Open Payments (https://github.com/adrianhopebailie/open-payments/issues/19), and now we're exploring the idea of doing this receipt functionality at the STREAM layer.

## How would this work?

First, I want to define the different parties involved:

### Terminology

- **Receiver** - the party (probably a wallet) that is receiving funds over STREAM.
- **Verifier** - the party (for example a web monetized site's backend) that wishes to verify that ILP payments occurred to the receiver.
- **Sender** - the party that is sending funds over STREAM.

- **Receipt Nonce** - A random nonce used to identify a STREAM connection in receipts.
- **Receipt Seed** - An HMAC key known only to the verifier used to generate Receipt Secret.
- **Receipt Secret** - The key used to generate a Receipt's HMAC.
- **Receipt** - A binary object containing an integer amount of funds received on a stream, the Receipt Nonce for its STREAM connection, a Stream ID on the STREAM connection, a timestamp referring to the time that the stream was established, and an HMAC over all those fields.

### Flow

_This flow occurs for each STREAM connection with Receipts enabled._

First, the Verifier generates a random Receipt Nonce, and HMACs it with its Receipt Seed to create a Receipt Secret. This ensures the Receipt Secret will be unique to this STREAM connection.

The Verifier communicates the Receipt Secret and Receipt Nonce to the Receiver over some channel (see the SPSP section for an example). It's important that the Receipt Secret is shared between the Verifier and the Receiver, but not with the Sender. 

[...]

Issue #481, opened 2018-09-29
BTP GeneralizedTime's Leap-second should be "smeared"


sappenin:
In [IL-RFC-30](https://github.com/interledger/rfcs/blob/master/0030-notes-on-oer-encoding/0030-notes-on-oer-encoding.md#encoding), the spec mandates that the String value of a leap-second date should BTP-encode to a String that matches its ISO-8601 counterpart (i.e., keep the leap second in each String), like this:

`2016-12-31T23.59.60.852Z` -> `20161231235960.852Z` (leap second)

However, I suggest that the RFC should instead specify that leap-second dates should simply be invalid. As an example, it's unlikely that a system would produce a Timestamp like this: `2016-12-31T23.59.60.852Z`

My rationale is as follows:

**1. Object Construction in JS & Java**
  -  Both modern-Java (see [Instant.java](https://docs.oracle.com/javase/8/docs/api/java/time/Instant.html)) and Javascript (see [here](https://www.reddit.com/r/javascript/comments/6xu95x/where_do_leap_seconds_go/)) "smear" leap seconds into the broader day that includes a leap second. Thus, neither platform actually ever represents a leap-second, but instead smears a portion of the leap-second into the existing seconds of the day, so a day always has 86400 seconds. 
  - As an example, doing the following in JS produces a Date that does not contain the leap-second:
  ```
  new Date('Dec 31 2016 23:59:60 UTC')
  Sat, 31 Dec 2016 23:59:00 GMT
  ```

**2. Parsing**
  - Parsing a leap-second value does not work by default in Java, nor in Javascript. For example, see this [JSFiddle here](https://jsfiddle.net/m8zn3bxe/3/) where constructing a new `Date` with the String  `2016-12-31T23:59:60.852Z` produces an error. 
  - Likewise in Java, `Instant.parse("2016-12-31T23:59:60.852Z")` throws an exception. 
  - Also, it's likely that `btp-packet` is actually broken in this regard when handling leap-seconds per [this code here](https://github.com/interledgerjs/btp-packet/blob/master/index.js#L52). 

**3. UTC Standards**
  - This above "smearing" or "smoothing" is actually a UTC standard. See [UTC-SLS](https://www.cl.cam.ac.uk/~mgk25/time/utc-sls/).

adrianhopebailie:
Makes sense to me.

[...]

PR #570, opened 2020-02-27
Add STREAM receipts

wilsonianb:
https://github.com/interledger/rfcs/issues/568


0029-stream/0029-stream.md
```diff
@@ -436,6 +440,24 @@ In other words, if a sender resends data (e.g. because a packet was lost), it MU
...
+
+The `Receipt` MUST contain the following fields encoded using the [Octet Encoding Rules (OER)](http://www.oss.com/asn1/resources/books-whitepapers-pubs/Overview_of_OER.pdf):
+
+| Field | Type | Description |
+|---|---|---|
+| HMAC | UInt256 | HMAC-SHA256 over all other fields using a secret pre-shared between the verifying party and the receiver. |
+| Receipt Nonce | UInt128 | A random nonce pre-shared between the verifying party and the receiver used to identify the STREAM connection. |
+| Stream ID | UInt8 | Identifier of the stream this receipt refers to. |
+| Total Received | UInt64 | Total amount, denominated in the units of the receiver, that the receiver has received on this stream thus far. |
+| Stream Start Time | UInt64 | A UNIX timestamp referring to the time that the stream was established at the receiver. |
```
sharafian: It's worth considering whether we should use an InterledgerTimestamp here. It still is fixed length, and although it is a little trickier to parse it solves things like leap seconds. That said a unix timestamp is gonna be easier to use in a lot of different environments.

0029-stream/0029-stream.md
```diff
@@ -436,6 +440,24 @@ In other words, if a sender resends data (e.g. because a packet was lost), it MU
...
+| Field | Type | Description |
+|---|---|---|
+| Stream ID | VarUInt | Identifier of the stream this frame refers to. |
+| Receipt | 65-Byte OctetString | Proof provided by the receiver of the total amount received on this stream |
+
+The `Receipt` MUST contain the following fields encoded using the [Octet Encoding Rules (OER)](http://www.oss.com/asn1/resources/books-whitepapers-pubs/Overview_of_OER.pdf):
+
+| Field | Type | Description |
+|---|---|---|
+| HMAC | UInt256 | HMAC-SHA256 over all other fields using a secret pre-shared between the verifying party and the receiver. |
```
sharafian: we should explicitly name this pre-shared secret the Receipt Secret and be explicit that it's an additional optional parameter to create a STREAM connection. I could imagine someone reading this and thinking we're referring to the shared secret here

0029-stream/0029-stream.md
```diff
@@ -192,6 +193,8 @@ Client streams MUST be odd-numbered starting with 1 and server-initiated streams
 
 Money can be sent for a given stream by sending an ILP Prepare packet with a non-zero `amount` and a `StreamMoney` frame in the STREAM packet to indicate which stream the money is for. A single ILP Prepare can carry value destined for multiple streams and the `shares` field in each of the `StreamMoney` frames indicates what portion of the Prepare amount should go to each stream.
 
+The receiver SHOULD include `StreamReceipt` frames in the ILP Fulfill packet indicating the total amount of money received in each stream, unless a secret key with which to generate receipts was not pre-shared with the receiver. Receipts can be verified, in order to confirm payment, using the pre-shared secret.
```
sharafian: A Receipt Secret to generate receipts and a Receipt Nonce to include in those receipts. Also did we land on whether there's a required length for the Receipt Secret & Receipt Nonce? Might make sense to say they're both 32 bytes to prevent someone from DoSing an SPSP server with a massive Receipt Secret

0029-stream/0029-stream.md
wilsonianb: https://github.com/interledger/rfcs/pull/570/commits/bc04d01cfaddec6cd607b1cb0424390a4272c592 adds more details on Receipt Secret and Receipt Nonce

0029-stream/0029-stream.md
wilsonianb: Right now, it's 16 byte nonce and 32 byte secret (https://github.com/interledger/rfcs/pull/570/files#diff-0bbc536b39826341170954e2675f765cR72-R73).
Is it worth making the nonce 32 bytes?

0029-stream/0029-stream.md
```diff
@@ -192,6 +195,8 @@ Client streams MUST be odd-numbered starting with 1 and server-initiated streams
 
 Money can be sent for a given stream by sending an ILP Prepare packet with a non-zero `amount` and a `StreamMoney` frame in the STREAM packet to indicate which stream the money is for. A single ILP Prepare can carry value destined for multiple streams and the `shares` field in each of the `StreamMoney` frames indicates what portion of the Prepare amount should go to each stream.
 
+The receiver SHOULD include `StreamReceipt` frames in the ILP Fulfill packet indicating the total amount of money received in each stream, unless a Receipt Secret and Receipt Nonce were not pre-shared with the receiver. Receipts can be verified, in order to confirm payment, using the Receipt Secret.
```
sharafian: I think it's worth explaining a little bit how this works because I think the first thing someone reading this will get tripped up on is that the verifying party (who they may imagine to be the sender) can sign new receipts.

Something like: "To use STREAM receipts, the Receipt Secret is pre-shared between the verifier and the receiver. The sender presents the STREAM receipts to the verifier, which the verifier can trust because only verifier and receiver know the Receipt Secret."

0029-stream/0029-stream.md
sharafian: Well for the nonce we only need uniqueness, given it's public. So 16 is probably fine. 32 byte secret is good because it's often gonna be output of sha256. I think those values are good, but open to hear any other opinions

0029-stream/0029-stream.md
wilsonianb: included more details in https://github.com/interledger/rfcs/pull/570/commits/694f2d7ed1ec71a69e708dc1c15eb46dcc8f7768

sharafian:
https://github.com/interledger/rfcs/issues/481

This suggests that leap second support isn't really a valid reason to use InterledgerTime, making me lean towards just a Unix timestamp.

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -62,6 +63,17 @@ Host: example.com
...
+##### Request Headers
+
+The request MAY contain at least the following headers in order to pre-share STREAM Receipt details between the SPSP Server and a party verifying payments sent to the SPSP Server:
+
+| Header          | Description                                                |
+|:----------------|:-----------------------------------------------------------|
+| `Receipt-Nonce`  | A base64-encoded 16-byte random nonce used to identify the STREAM connection in STREAM Receipts. |
+| `Receipt-Secret` | A base64-encoded 32-byte key used to generate a STREAM Receipt's HMAC. |
+
+The SPSP Client MAY be provided with an SPSP Endpoint belonging to the receipt verifier, which would add the receipt headers and proxy the query to the SPSP Server.
```
kincaidoneil: How does the verifier know from the response whether the SPSP server supports receipts or not?

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -62,6 +63,17 @@ Host: example.com
...
 ```
 
+##### Request Headers
+
+The request MAY contain at least the following headers in order to pre-share STREAM Receipt details between the SPSP Server and a party verifying payments sent to the SPSP Server:
+
+| Header          | Description                                                |
+|:----------------|:-----------------------------------------------------------|
+| `Receipt-Nonce`  | A base64-encoded 16-byte random nonce used to identify the STREAM connection in STREAM Receipts. |
+| `Receipt-Secret` | A base64-encoded 32-byte key used to generate a STREAM Receipt's HMAC. |
```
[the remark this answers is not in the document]
sharafian: It's unlikely that the browser would be the verifier in this context, so I don't think these headers are needed in CORS policy. The expected flow is that the browser calls the verifier, verifier attaches headers and calls the wallet, and then response is returned to verifier who returns to browser. Then the browser passes receipts to verifier.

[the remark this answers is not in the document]
kincaidoneil: Right, thanks 👍

0029-stream/0029-stream.md
```diff
@@ -436,6 +444,24 @@ In other words, if a sender resends data (e.g. because a packet was lost), it MU
...
+
+The `Receipt` MUST contain the following fields encoded using the [Octet Encoding Rules (OER)](http://www.oss.com/asn1/resources/books-whitepapers-pubs/Overview_of_OER.pdf):
+
+| Field | Type | Description |
+|---|---|---|
+| HMAC | UInt256 | HMAC-SHA256 over all other fields using the 32 byte Receipt Secret, which is pre-shared between the verifying party and the receiver. |
+| Receipt Nonce | UInt128 | A random nonce pre-shared between the verifying party and the receiver used to identify the STREAM connection. |
+| Stream ID | UInt8 | Identifier of the stream this receipt refers to. |
+| Total Received | UInt64 | Total amount, denominated in the units of the receiver, that the receiver has received on this stream thus far. |
+| Stream Start Time | UInt64 | A UNIX timestamp referring to the time that the stream was established at the receiver. |
```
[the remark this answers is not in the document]
sharafian: This lets the verifier set a cache expiry on how long it remembers the highest amount seen for a Receipt Nonce + Stream Id. If we had no timestamp it would need to remember the highest receipt seen forever or else someone could present their receipt a year later and it would still be valid (and they would be credited the full total received instead of the diff of total received - last total received). So a verifier would basically have some reasonable cache expiry and not accept receipts from >cache expiry ago.

0029-stream/0029-stream.md
```diff
@@ -436,6 +444,24 @@ In other words, if a sender resends data (e.g. because a packet was lost), it MU
 
 Asset details exposed by this frame MUST NOT change during the lifetime of a Connection.
 
+#### 5.3.15. `StreamReceipt` Frame
+
+| Field | Type | Description |
+|---|---|---|
+| Stream ID | VarUInt | Identifier of the stream this frame refers to. |
+| Receipt | 65-Byte OctetString | Proof provided by the receiver of the total amount received on this stream |
```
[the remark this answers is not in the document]
sharafian: The receipt is kinda treated as a whole object that's passed around, for instance the receipt would be emitted in WM. So I think "grouping" like this can be helpful

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
sharafian: :+1: we should probably add some kind of discovery for this, either in a response header or in the response body for SPSP

0029-stream/0029-stream.md
```diff
@@ -436,6 +444,24 @@ In other words, if a sender resends data (e.g. because a packet was lost), it MU
...
+
+The `Receipt` MUST contain the following fields encoded using the [Octet Encoding Rules (OER)](http://www.oss.com/asn1/resources/books-whitepapers-pubs/Overview_of_OER.pdf):
+
+| Field | Type | Description |
+|---|---|---|
+| HMAC | UInt256 | HMAC-SHA256 over all other fields using the 32 byte Receipt Secret, which is pre-shared between the verifying party and the receiver. |
```
sentientwaffle: Nit: if the `HMAC` was the last field instead of the first, it would be easy to implement serialization with one fewer allocation.

wilsonianb: I think I put the hmac first with verification in mind, but it makes sense to me to optimize the serialization.

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
wilsonianb: Are we thinking the verifier would return a 4XX error if the SPSP server doesn't support receipts?
If so, maybe the SPSP server could just parrot the Receipt-Nonce header if it supports receipts.

0029-stream/0029-stream.md
wilsonianb: addressed in https://github.com/interledger/rfcs/pull/570/commits/8df93049769a21a0a8cd59b3d43902574b88e4ce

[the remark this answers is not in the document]
wilsonianb: Updated the description in https://github.com/interledger/rfcs/pull/570/commits/8df93049769a21a0a8cd59b3d43902574b88e4ce

0029-stream/0029-stream.md
```diff
@@ -436,6 +444,24 @@ In other words, if a sender resends data (e.g. because a packet was lost), it MU
...
+| Receipt Nonce | UInt128 | A random nonce pre-shared between the verifying party and the receiver used to identify the STREAM connection. |
+| Stream ID | UInt8 | Identifier of the stream this receipt refers to. |
```
sentientwaffle: FWIW: the ilp-stream spec doesn't guarantee that the stream ID will fit in a UInt8.

0029-stream/0029-stream.md
sentientwaffle: Why mandate the exact size of the receipt instead of leaving it open-ended for possible extension?

0029-stream/0029-stream.md
wilsonianb: Yeah, I based it off the recommended `default max stream ID of 20` in the spec, but we can make it larger if we want.

0029-stream/0029-stream.md
wilsonianb: :+1:
Should we also consider including a receipt version number?
Or could the receipt size be used to check compatibility, such as with receipt support discovery (https://github.com/interledger/rfcs/pull/570#discussion_r395327765)?

sharafian: Perhaps the receipt itself could have a version byte? Size might not be good enough if a future version is variable length. And if the receipt contains its own version info then STREAM doesn't need to emit extra information about receipt version

0029-stream/0029-stream.md
sentientwaffle: Ah, I missed that. I retract my objection.

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
sharafian: We could include `"receipts": true` in the SPSP response body. I feel like that's a little more clean than echoing the header. Another option would be some kind of SPSP features header if we wanted to make it extensible

0029-stream/0029-stream.md
wilsonianb: changed receipt to `VarOctetString` and added version number: https://github.com/interledger/rfcs/pull/570/commits/1633e86180f268e4808f9011610f91b6f325096b

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
wilsonianb: added optional `receipts` field to SPSP response body: https://github.com/interledger/rfcs/pull/570/commits/1633e86180f268e4808f9011610f91b6f325096b

0029-stream/0029-stream.md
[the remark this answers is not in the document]
kincaidoneil: Agree you wouldn't want to honor receipts indefinitely, but couldn't the verifier just cache the timestamp it generated the receipt nonce/responded to the SPSP query and use that to expire its cache? Why should it trust the recipient not to manipulate the timestamp?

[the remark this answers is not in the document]
wilsonianb: I like the idea of removing the timestamp from the receipt.
Verifiers can either cache the nonce and SPSP query time or encode the time in the nonce if they prefer to be more stateless.

kincaidoneil:
Since money can be sent over multiple streams, the verifier's accounting might be more complicated since they have to support tracking and totaling the amount received over each stream.

What if receipts attested the amount received over the whole connection instead? What are the arguments for/against a per-connection receipt vs a per-stream receipt?

wilsonianb:
https://github.com/wilsonianb/receipt-verifier/
:point_up_2: The current verifier implementation aggregates receipt amounts by a specified id (not necessarily the receipt nonce), so per-connection receipts wouldn't save it that much complexity.

I think it's still worth considering though. Would one type of accounting be easier to add to the Rust and Java STREAM implementations than the other?

wilsonianb:
On second thought now that I'm trying to update the verifier for receipts without the timestamp, per-connection receipts _would_ simplify things.
The verifier can create a cache entry with an expiry for the receipt nonce when the SPSP query happens. This entry can be used to store the latest receipt amount as well as determining when a submitted receipt is too old (the entry will have expired).

0029-stream/0029-stream.md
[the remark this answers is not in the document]
sharafian: :+1: that's a good observation, and saves us from having to specify a timestamp format which I'm a big fan of

sharafian:
> Since money can be sent over multiple streams, the verifier's accounting might be more complicated since they have to support tracking and totaling the amount received over each stream.

I was imagining that the key in the verifier's store would be the receipt nonce concatenated to stream ID. I'm not seeing right now how it's significantly simpler to just key with the receipt nonce.

> The verifier can create a cache entry with an expiry for the receipt nonce when the SPSP query happens. This entry can be used to store the latest receipt amount as well as determining when a submitted receipt is too old (the entry will have expired).

Gotcha, so it saves a lookup because we don't need a separate key tracking the expiration of the connection as a whole?

In general I'm not hugely opposed to dropping the STREAM id but I want to make sure it doesn't mess anything else up in the future. STREAM was designed with multiplexing as a 'first-class' feature, where the idea is everything is done on streams and you might have long-lived connections on which you do many streams (like a long-lived QUIC connection with several HTTP calls over it).

We always have the option of making a new version of the receipt format but I think it could save us from future work if the receipts conceptually make sense with the rest of STREAM by working with the Stream abstraction

wilsonianb:
>so it saves a lookup because we don't need a separate key tracking the expiration of the connection as a whole?

More or less yeah. This gives an idea of how the verifier might change from using a redis hash to a single value for per-connection receipts:
https://github.com/wilsonianb/receipt-verifier/commit/13f77b975c2510284e4807e5b084d34c0af26b3b

I don't have a strong opinion either way.

asn1/Stream.asn
```diff
@@ -174,4 +175,12 @@ StreamDataBlocked ::= SEQUENCE {
   maxOffset VarUInt
 }
 
+StreamReceipt ::= SEQUENCE {
+  -- Identifier of the stream
+  streamId VarUInt,
+
+  -- Receipt for verifying the total amount received on this stream
+  receipt VarBytes
```
sappenin: Is there a formal ASN.1 definition for `receipt`? I see it documented in the markdown for IL-RFC-29, but don't see anything formal in this file.

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -24,6 +24,7 @@ SPSP provides for exchanging basic server details needed by a client to set up a
 * **SPSP Server** - The application that handles incoming SPSP requests from the SPSP Client.
 * **SPSP Endpoint** - The specific HTTPS endpoint on the SPSP Server used for setting up a payment.
 * **STREAM Module** - Software included in the SPSP Client and Server that implements the [STREAM](../0029-stream/0029-stream.md) protocol.
+* **STREAM Receipt** - Proof provided by the payment recipient of the total amount received on a stream.
```
sappenin: Overall, these changes are looking good. One thing that I think would be really helpful (as a supplement to the changes proposed in this PR) is to _also_ create a new IL-RFC for **Stream Receipts**. That RFC should outline most of the information and flows that @sharafian [detailed in this post](https://github.com/interledger/rfcs/issues/568).

While the formalizations presented in this PR seem to support the minimal spec changes to enable Stream Receipts, I think having a formalization of how receipts are generated, used, verified, etc as per #568 is probably a good idea. 

Unless there's some particular reason not to be more specific than the changes here?

0029-stream/0029-stream.md
```diff
@@ -158,6 +159,8 @@ A server MUST communicate the following values to a client using an **authentica
 
 To avoid storing a 32 byte secret for each connection, a server MAY deterministically generate the shared secret for each connection from a single server secret and a nonce appended to the ILP Address given to a particular client, for example by using an HMAC.
 
+For each new connection, a server MAY be provided with a pre-shared 32 byte Receipt Secret to generate receipts and 16 byte Receipt Nonce to include in those receipts. To avoid storing these for each connection, a server MAY deterministically append them to the ILP Address given to a particular client. If doing so, the server MUST encrypt the Receipt Secret.
```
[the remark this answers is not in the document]
wilsonianb:
```suggestion
For each new connection, a server MAY be provided with a pre-shared 32 byte Receipt Secret (to generate receipts) and a 16 byte Receipt Nonce (to include in those receipts). To avoid storing this data for each connection, a Server MAY deterministically append this data to the ILP Address used for a Connection. If doing so, the server MUST encrypt the Receipt Secret.
```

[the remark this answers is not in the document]
sappenin: LGTM!

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -62,14 +63,26 @@ Host: example.com
 Accept: application/spsp4+json, application/spsp+json
 ```
 
+##### Request Headers
+
+The request MAY contain at least the following headers in order to pre-share STREAM Receipt details between the SPSP Server and a party verifying payments sent to the SPSP Server:
```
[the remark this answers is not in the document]
wilsonianb: It isn't explicitly stated anywhere, but a connection could technically have receipts without the headers if the receiver is the verifier.
I'll plan to explain that at least in the standalone STREAM Receipts RFC (https://github.com/interledger/rfcs/pull/570#discussion_r399577444) and then see how best to improve the wording here (and add links).

asn1/Stream.asn
sappenin: Resolving this - it's there now in Stream.asn.

0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
sappenin: This is addressed - thanks!

proposals/0000-stream-receipts.md
```diff
@@ -0,0 +1,93 @@
...
+
+If keeping a balance of how much the Sender has paid, the Verifier SHOULD store the Receipt amount under the Receipt Nonce + Stream ID. If there is already an amount stored for that Stream ID, the Verifier SHOULD only credit the Sender the difference in Receipt amounts.
+
+Alteratively, the Verifier MAY forego checking the Receipt amount if they are only concerned with whether or not the Sender paid at all, as opposed to how much.
+
+### Flow
+
+This flow occurs for each STREAM connection with Receipts enabled.
+
+1. Verifier MUST generate a random Receipt Nonce.
```
sharafian, line 51: this `MUST` contradicts with the suggestion above to include the timestamp in the receipt nonce. `MUST generate a unique Receipt Nonce` would be more accurate

proposals/0000-stream-receipts.md
```diff
@@ -0,0 +1,93 @@
...
+
+2. Verifier MUST communicate the Receipt Nonce and Receipt Secret to the Receiver. The Receipt Secret MUST NOT be communicated to the Sender.
+
+3. Once the STREAM connection is open and the Sender is sending, the Receiver MUST add an additional frame containing a Receipt to each ILP Fulfill packet.
+
+4. Sender MUST give the Receipt directly or indirectly to the Verifier.
+
+5. Verifier MUST verify the Receipt before crediting the Sender with the Receipt amount.
```
sharafian, line 59: `crediting` makes me think of wallet balances where the Sender will get that money back. Can we clarify, something like `accepting the Receipt amount as payment for services`

proposals/0000-stream-receipts.md
```diff
@@ -0,0 +1,93 @@
...
+
+### Stateless Verifier
+
+The Verifier MAY generate the Receipt Secret by producing an HMAC of the Receipt Nonce using a seed (known only to itself) as the HMAC key. This ensures the Receipt Secret will be unique to this STREAM connection and allows the Verifier to reproduce the Receipt Secret from any Receipts containing the Receipt Nonce.
+
+### Stateless Receiver
+
+Before a STREAM connection is established, the Receiver MAY use the Receipt Nonce and Receipt Secret as parameters to generate connection details (ILP Address and shared secret). The Receiver MAY encode the Receipt Nonce and Receipt Secret in the STREAM server ILP Address to allow the STREAM receiver to avoid persisting state. If doing so, the Receipt Secret MUST be encrypted. The Receipt Nonce MUST be unique each time STREAM parameters are generated, since it MUST be unique per STREAM connection.
```
sharafian, line 67: `since it MUST be unique per STREAM connection` the receipt nonce must be unique globally

asn1/Stream.asn
```diff
@@ -174,4 +194,12 @@ StreamDataBlocked ::= SEQUENCE {
   maxOffset VarUInt
 }
 
+StreamReceipt ::= SEQUENCE {
+  -- Identifier of the stream
+  streamId VarUInt,
+
+  -- Receipt for verifying the total amount received on this stream
+  receipt Receipt
```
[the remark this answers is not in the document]
wilsonianb: Given that the STREAM [spec](https://interledger.org/rfcs/0029-stream/#53-frames) and [implementation](https://github.com/interledgerjs/ilp-protocol-stream/blob/master/src/packet.ts#L491-L492) indicate that
>Each frame is encoded with its 1-byte type and length prefix.

the fact that the STREAM [asn.1 definition](https://interledger.org/rfcs/asn1/Stream.asn) doesn't specify the length prefix for the various frame `SEQUENCE`s, _seems_ to suggest that `SEQUENCE`s are length prefixed.
But I'm not finding an indication either way in the asn.1 docs.

0029-stream/0029-stream.md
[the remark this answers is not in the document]
sappenin: Did this get resolved?

asn1/Stream.asn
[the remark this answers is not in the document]
sappenin: I don't think the SEQUENCEs we use are length-prefixed because they're not `extensible`.

> A non-extensible sequence type is encoded as a presence bitmap followed by the concatenation of the encodings of the fields of the sequence type that are present in the value, in specification order. The presence bitmap is encoded as a bit string with a fixed size constraint (see 2.7, case 1), and has one bit for each field of the sequence type that has the keyword OPTIONAL or DEFAULT, in specification order.

(see [Overview_of_OER.pdf, Section 2.8](https://www.oss.com/asn1/resources/books-whitepapers-pubs/Overview_of_OER.pdf).

Though I'm curious: do we think the `Receipt` is going to change in the future and become extensible?

0029-stream/0029-stream.md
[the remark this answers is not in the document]
wilsonianb: yep, timestamp has been removed https://github.com/interledger/rfcs/pull/570/commits/ff2617a7d1b4ef592426b153d8f10a6b8f0ff180

asn1/Stream.asn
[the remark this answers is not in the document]
wilsonianb: Thanks for the link.
After discussing it with Ben, we'd like to follow the model of [ILP packets](https://github.com/interledger/rfcs/blob/master/asn1/InterledgerPacket.asn#L54-L59) by depending on the version byte to know what to expect for the remaining bytes (instead of making `Receipt` extensible and introducing the presence / extension presence bitmaps).
The entire receipt is length prefixed in the `StreamReceipt` frame in order for stream senders to be able to handle unsupported receipt versions.
https://github.com/interledger/rfcs/pull/570/commits/09172b7f714005bc4df682637829bf001b53a9bf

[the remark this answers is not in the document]
sappenin: > we'd like to follow the model of ILP packets by depending on the version byte to know what to expect for the remaining bytes

No objection here - I like this idea.

That said (and I'm no ASN expert, so there's that) -- the [current ASN](https://github.com/interledger/rfcs/commit/09172b7f714005bc4df682637829bf001b53a9bf#diff-b95395fdef50093ed11e454c0d9b050bR204) doesn't seem to have a strong connection between an actual `Receipt` and just random bytes because the type defined in the `StreamReceipt Frame` for the actual Receipt is just `VarBytes` (which is really just [OCTET STRING](https://github.com/interledger/rfcs/blob/master/asn1/GenericTypes.asn#L65)).

What if instead we more closely followed the concept of the InterledgerPacket here so it's obvious what's allowed in the `StreamReceipt` frame? 

Currently, InterlederPacket supports a bunch of different "types". In our case, we could mirror this style and just define new "types" in the future if we ever need a new Receipt version. Primarily, this would create a stronger connection between the `Receipt` field in the frame and the data allowed in it. ASN could look like this:

https://gist.github.com/sappenin/b5925e86a2cc6e8042650e6802800a7a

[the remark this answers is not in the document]
wilsonianb: Thanks for the gist.
I agree it's worth trying to associate `Receipt` with the `StreamReceiptFrame`.
A difference here from ILP and STREAM packets is that the receipt type/version is needed by the verifier but not the STREAM sender (who parses the frame). So I think including the type in the frame (as opposed to in the receipt) just adds a useless field for STREAM implementations to parse.
What do you think about this alternative which uses a CHOICE:
https://gist.github.com/wilsonianb/75ddc1b2a3780218016f1da0f28f9101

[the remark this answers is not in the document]
sappenin: I suppose I need to know more about what we think a future Receipt might look like. I see the comments above around a future `Receipt` having potentially a different length, but I don't quite have the context to know what that would mean. Can you elaborate there?

The reason I bring that up is that the `CHOICE` option seems unclear/complicated in terms of how to encode it (e.g., there's the tag itself, and then there appears to be different types of tags [in section 2.12](https://www.oss.com/asn1/resources/books-whitepapers-pubs/Overview_of_OER.pdf)).

Also, using `CHOICE` seems to be more geared towards payloads that should have one thing or another (e.g., `house` or `car`) -- it doesn't feel quite right for versioning (seems like extensibility is more correct there).

Wondering if anyone with more ASN.1 OER expertise should chime in here...  @adrianhopebailie  ?

> A difference here from ILP and STREAM packets is that the receipt type/version is needed by the verifier but not the STREAM sender (who parses the frame). 

Can you clarify this? In my mind the verifier could parse per-usual, but the STREAM sender wouldn't actually parse the `Receipt` if they don't need it. They can just treat this as a blob of data, no?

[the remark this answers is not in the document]
wilsonianb: I've updated the [gist](https://gist.github.com/wilsonianb/75ddc1b2a3780218016f1da0f28f9101) with a hypothetical receipts v2, but I hadn't looked at how `CHOICE` is handled in OER and agree that a tag is not desirable.

I don't think we have anything concrete in mind for what a future version might look like.
I'm mainly trying to support the possibility of future version(s) such that:
- the sender can parse a receipt blob of any length out of the `StreamReceipt` frame (without parsing the receipt itself, not even the version/type)
- the verifier can check a receipt's version/type to known if they support it.

>In my mind the verifier could parse per-usual, but the STREAM sender wouldn't actually parse the Receipt if they don't need it. They can just treat this as a blob of data, no?

That's all correct, but (using your [example gist](https://gist.github.com/sappenin/b5925e86a2cc6e8042650e6802800a7a)) the sender would need to parse the `ReceiptPacket`, where the type/version lives, in order to get at the `Receipt`.

How about this minor variation on your example
https://gist.github.com/wilsonianb/b164080a6c1b62e0e2be11d41efb8b13

The `Receipt` (formerly `ReceiptPacket`) is what's passed from the sender to the verifier.
The data in `Receipt` wouldn't be length prefixed <-- would that need to be represented differently in ASN.1?
The version number wouldn't be part of the hmac message, which should be fine.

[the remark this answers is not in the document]
wilsonianb: >The version number wouldn't be part of the hmac message, which should be fine.

Ben pointed out that could not be fine. A hypothetical subsequent version could swap the nonce and amount fields, and the receipt created as one version _could_ be submitted as the other version.

I've updated the hmac description in my gist:
https://gist.github.com/wilsonianb/b164080a6c1b62e0e2be11d41efb8b13
Does that seem workable, especially the non-length prefixed `Receipt` data field?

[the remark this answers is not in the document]
sappenin: @wilsonianb what's in your latest gist in the [comment](https://github.com/interledger/rfcs/pull/570#discussion_r409861591) directly above LGTM. 

> Ben pointed out that could not be fine. 

Great catch @sharafian !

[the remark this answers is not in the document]
wilsonianb: updated in https://github.com/interledger/rfcs/pull/570/commits/43039f71db1b815d54e42494aeea282ed18f5c4e

asn1/Stream.asn
```diff
@@ -67,14 +68,46 @@ FrameSet FRAME ::= {
...
+  {22 StreamDataBlocked} |
+  {23 StreamReceipt}
 }
 
 StreamFrame ::= SEQUENCE {
   type FRAME.&typeId ({FrameSet}),
   data FRAME.&Type ({FrameSet}{@type})
 }
 
+RECEIPT ::= CLASS {
```
sappenin, line 80: This is a minor nit, but can we move these definitions to the bottom of this file? As they are now, they sort of break the flow of the STREAM and frame definitions themselves, which makes the ASN harder to read.

We could even put these directly above the `StreamReceipt` frame definition on line 209? Or pull the receipt definitions themselves into a separate file? (I can't think of a use-case, but I suppose lines 80-109 can technically stand on their own independent of STREAM and be imported into this file).

wilsonianb: Moved to a separate file https://github.com/interledger/rfcs/pull/570/commits/7dd4111d08c640466d7d3aac82a0ec3f1658d0db
(force pushed because the commit wasn't originally appearing in the pull request, likely due to earlier github incident)

junautogroup (commented):
0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md-simple-payment-setup-protocol.mdhttps://github.com/junautogroup/rfcs/blob/f7f54131033cd8dbf7ff15953f58337a0a03fa9f/.github/stale.ymlasn1/GenericTypes.asn30asn1/GenericTypes.asn.github/stale.ymlThe `.github/stale.yml` configuration file is used for the Stale app in your repository to manage inactive issues. Here are the key configurations:

- `daysUntilStale`: 30 days of inactivity before an issue is marked as stale.
- `daysUntilClose`: 7 days of inactivity after being marked as stale before the issue is closed.
- `exemptLabels`: Issues with labels 'pinned' and 'security' are never considered stale.
- `staleLabel`: The label 'wontfix' is used to mark issues as stale.
- `markComment`: A comment is posted when an issue is marked as stale.
- `closeComment`: No comment is posted when a stale issue is closed (set to `false`).

junautogroup (commented):
1633e86180f268e4808f9011610f91b6f325096b

PR #572, opened 2020-04-24
Allocate ILRFC 39 to Stream Receipts

adrianhopebailie:


sappenin:
@wilsonianb I think this should be all fixed up now. If you agree, plz re-approve and I can merge.

PR #576, opened 2020-05-13
Fix broken link in 0026-payment-pointers

sharon-wang:
Replaced the broken link referencing "The Most Dangerous Code in the World".

Looks like I might need to include "Skip Version Check" in my commit message to pass CI, since 0026-payment-pointers has `draft: FINAL`.


sappenin:
@adrianhopebailie what's the path forward here? @sharon-wang can you try updating your commit message or just submit a new PR with a new message?

sharon-wang:
@sappenin I updated the commit message to include "Skip Version Check" and CI is now passing.

PR #578, opened 2020-05-27
docs(0009): recommend cors headers for spsp

sharafian:
It's an unwritten rule right now that SPSP servers include CORS headers to allow Web Monetization polyfills to function.

All ILP enabled wallets currently support this behavior already, but including it here will make it easier for future ILP enabled wallets to follow this recommendation.

### DIFF
--- a/0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
+++ b/0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
@@ -1,7 +1,7 @@
 ---
 title: The Simple Payment Setup Protocol (SPSP)
 type: working-draft
-draft: 11
+draft: 12
 ---
 # Simple Payment Setup Protocol (SPSP)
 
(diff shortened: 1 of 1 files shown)


0009-simple-payment-setup-protocol/0009-simple-payment-setup-protocol.md
```diff
@@ -63,6 +63,19 @@ Host: example.com
 Accept: application/spsp4+json, application/spsp+json
 ```
 
+##### Request Headers to Support Web Monetization Polyfills
+
+[Web Monetization](../0028-web-monetization/0028-web-monetization.md) polyfills may query SPSP from a non-privileged context they are implemented as a script rather than a browser extension. Sites may choose to use a script-based polyfill to enable Web Monetization for their visitors without requiring any browser extension or browser support.
```
sappenin, line 68:
```suggestion
[Web Monetization](../0028-web-monetization/0028-web-monetization.md) polyfills may query SPSP from a non-privileged context if they are implemented as a script rather than a browser extension. Sites may choose to use a script-based polyfill to enable Web Monetization for their visitors without requiring any browser extension or browser support.
```

PR #583, opened 2023-08-31
Make img src absolute path for importing

huijing:
The content of these spec files are being imported in other apps and relative URLs for assets are breaking.


huijing:
Resolved by #584

PR #589, opened 2024-02-28
Refer to IL-RFC-31 as now-deprecated

### LINKED ISSUE #588: RFC 1 refers to RFC 31 which no longer exists

michielbdejong:
Fixes #588


sabineschaller:
Hi @michielbdejong, nice to see you here again 😊 
I wasn't aware that this RFC was deprecated and I'm not sure it should be. We have scheduled an internal discussion to figure out whether it should be revived. I'll let you know. 
Note that all the deprecated specs are in the `/deprecated`, so no need to link an old commit in the docs.

PR #599, opened 2024-11-21
deprecate sync ilp over http

mkurapov:


0035-ilp-over-http/0035-ilp-over-http.md
```diff
@@ -117,7 +104,37 @@ The sender of the ILP Fulfill/Reject SHOULD ensure there are multiple attempts t
 
 An ILP Fulfill packet corresponds to a commitment which affects financial accounting balances. If an HTTP request carrying the ILP reply fails, such as due to a network connection error, retrying delivery of the ILP reply with [idempotence](https://en.wikipedia.org/wiki/Idempotence) can prevent balance inconsistencies between peers.
 
-If the sender of an ILP Prepare expects an asynchronous reply, they should only process the first ILP reply they receive corresponding to the in-flight ILP Prepare.
+The sender of the ILP Prepare should only process the first ILP reply they receive corresponding to the original ILP Prepare packet.
+
+### [Deprecated] Synchronous Mode
```
mkurapov, line 109: Is this section something we want to keep at all? Or should we just move the spec that has both versions to the deprecated folder, and make a new ILP-over-HTTP spec?

sabineschaller (commented):
I'm slightly leaning towards just getting rid of sync mode all together if we are sure nobody uses it.

